PR #29, opened 2023-10-17
Changes to get pypi project set up properly

saqadri:
Changes to get pypi project set up properly

* We don't need setup.py if we're using pyproject.toml
* Updated readme.md in /python folder
* Renamed `aiconfig_tools` -> `aiconfig`
* Added __init.py__ to every subdirectory
* Added dependencies in pyproject.toml


python/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2018 The Python Packaging Authority
```
suyoglastmileai, line 1: Should this be LastMIle?

Also I guess this is MIT licensed?

LastMileNewGrad (commented):
lgtm

LastMileNewGrad (approved):

python/LICENSE
saqadri: Will update, and yes it's the MIT license

PR #32, opened 2023-10-19
[Python] 1/n Serialize Setup

LastMileNewGrad:
[Python] 1/n Serialize Setup

Implemented Serialize in ConfigRuntime which calls ModelParser
`serialize`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/32).
* #33
* __->__ #32

### DIFF
--- a/python/src/aiconfig/Config.py
+++ b/python/src/aiconfig/Config.py
@@ -113,6 +113,25 @@ class AIConfigRuntime(AIConfig):
             data = resp.json()
             return cls.model_validate_json(data)
 
+    async def serialize(self, model_name: str, data: Dict, params: Optional[dict] = {},  prompt_name: Optional[str] = "") -> List[Prompt]:
+        """
+        Serializes the completion params into a Prompt object. Inverse of the 'resolve' function.
+
+        args:
+            model_name (str): The model name to create a Prompt object for
+            data (dict): The data to save as a Prompt Object
+            params (dict, optional): Optional parameters to save alongside the prompt
+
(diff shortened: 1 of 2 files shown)


python/src/aiconfig/Config.py
```diff
@@ -113,6 +113,25 @@ def load_from_workbook(cls, workbook_id: str) -> "AIConfigRuntime":
...
+            params (dict, optional): Optional parameters to save alongside the prompt
+
+        returns:
+            Prompt | List[Prompt]: A prompt or list of prompts representing the input data
+        """
+        model_parser = ModelParserRegistry.get_model_parser(model_name)
+        if not model_parser:
+            raise ValueError(f"Unable to serialize data: `{data}`\n Model Parser for model {model_name} does not exist.")
+        
+        prompts = model_parser.serialize("", data, self, params)
```
saqadri, line 132: What is the empty string for? If it's intentional please comment the variable name so it's easy to know

saqadri (approved):
Approving to unblock -- let's pass in prompt_name to the new serialize function so we don't pass empty string as the prompt name to model_parser.serialize

LastMileNewGrad:
Updated ModelParser and ConfigRuntime

> Approving to unblock -- let's pass in prompt_name to the new serialize function so we don't pass empty string as the prompt name to model_parser.serialize

Updated ModelParser and ConfigRuntime

PR #33, opened 2023-10-19
[Python] 4/n Serialize OpenAI Model Parser

LastMileNewGrad:
[Python] 4/n Serialize OpenAI Model Parser

Modified serialize return type to list of Prompts.

Implemented serialize for OpenAI ModelParser.

Tested with api wrapper with the diff on top.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/33).
* __->__ #33
* #38
* #39
* #37

### DIFF
--- a/python/src/aiconfig/default_parsers/openai.py
+++ b/python/src/aiconfig/default_parsers/openai.py
@@ -1,6 +1,6 @@
 from abc import abstractmethod
 import copy
-from typing import Dict, Optional, Union
+from typing import Dict, List, Optional, Union
 from typing import TYPE_CHECKING, Any, Dict, Optional
 from aiconfig import AIConfigSettings
 from aiconfig.AIConfigSettings import (
@@ -349,3 +349,47 @@ def refine_chat_completion_params(model_settings):
             completion_data[key] = model_settings[key]
 
     return completion_data
+
+
+def add_prompt_as_message(
+    prompt: Prompt, aiconfig: "AIConfigRuntime", messages: List, params=None
+):
+    """
+    Converts a given prompt to a message and adds it to the specified messages list.
+
+    Note:
+    - If the prompt contains valid input, it's treated as a user message.
+    - If the prompt has a custom role, function call, or name, these attributes are included in the message.
+    - If an AI model output exists, it is appended to the messages list.
+    """
+    if is_prompt_template(prompt):
+        resolved_prompt = resolve_prompt(prompt, params, aiconfig)
+        messages.append({"content": resolved_prompt, "role": "user"})
+    else:
+        # Assumes Prompt input will be in the format of ChatCompletionMessageParam (with content, role, function_name, and name attributes)
+        resolved_prompt = resolve_prompt(prompt.input.content, params, aiconfig)
+
+        prompt_input = prompt.input
+        role = prompt_input.role if hasattr(prompt_input, "role") else "user"
+        fn_call = prompt_input.function_call if hasattr(prompt_input, "function_call") else None
+        name = prompt_input.name if hasattr(prompt_input, "name") else None
+        messages.append(
+            {"content": resolved_prompt, "role": role, "function_call": fn_call, "name": name}
+        )
+
+    output = aiconfig.get_latest_output(prompt)
+    if output:
+        if output.output_type == "execute_result":
+            output_message = output.data
+            if output_message["role"] == "assistant":
+                messages.append(output_message)
+    return messages
+
+
+def is_prompt_template(prompt: Prompt):
+    """
+    Check if a prompt's input is a valid string.
+    """
+    return isinstance(prompt.input, str) or (
+        hasattr(prompt.input, "data") and isinstance(prompt.input.data, str)
+    )
--- a/python/src/aiconfig/default_parsers/openai.py
+++ b/python/src/aiconfig/default_parsers/openai.py
@@ -14,7 +14,7 @@ from aiconfig.AIConfigSettings import (
 )
 from aiconfig.default_parsers.parameterized_model_parser import ParameterizedModelParser
 from aiconfig.util.config_utils import get_api_key_from_environment
-from aiconfig.util.params import resolve_parameters, resolve_prompt, resolve_system_prompt
+from aiconfig.util.params import resolve_parameters, resolve_prompt, resolve_prompt_string, resolve_system_prompt
 
 import openai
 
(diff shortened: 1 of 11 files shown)


python/src/aiconfig/default_parsers/openai.py
```diff
@@ -74,40 +49,66 @@ def serialize(
...
-        # Serialize Prompt input
-        prompt_input = (
-            data["prompt"] if isinstance(data["prompt"], str) else {"data": data["prompt"]}
-        )
+        serialized_prompts = []
 
-        # Cosntruct model settings, does not include prompt
-        prompt_model_metadata = {key: value for key, value in data.items() if key != "prompt"}
+        # Combine conversation data with any extra keyword args
+        conversation_data = {**data} | kwargs
```
saqadri, line 55: Remove kwargs merging

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -74,40 +49,66 @@ def serialize(
...
+        global_model_settings = ai_config.get_global_settings(self.id())
 
-            override_settings = {key: prompt_model_metadata[key] for key in override_keys}
+        if global_model_settings:
+            # Identify the settings that differ from global settings
+            override_settings = {
+                key: copy.deepcopy(conversation_data[key])
+                for key in conversation_data
+                if key not in global_model_settings
+            }
```
saqadri, line 69: This isn't comparing the values of the keys, so if the global settings contain `temperature` but set to a 0.9, while input completion params contained `temperature` but set to 1, this will not consider that to be an override even though it should be.

saqadri: Also check everywhere else in the code where setting overrides are being calculated, I am guessing the same problem exists elsewhere too.

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -74,40 +49,66 @@ def serialize(
...
-            **kwargs
-        )
+            model_metadata = ModelMetadata(
+                **{"name": self.id(), "settings": copy.deepcopy(conversation_data)}
+            )
+        # Remove messages array from model metadata. Handled separately
+        model_metadata.settings.pop("messages", None)
+
+        system_prompt = ""
+        user_prompt = ""
```
saqadri, line 80: This is set to a string but then set to a Prompt object below, which is confusing. Maybe just set to None?

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -74,40 +49,66 @@ def serialize(
...
+                    model_metadata_with_system_prompt.settings["system_prompt"] = system_prompt
+
+                user_prompt = Prompt(
+                    name=prompt_name,
+                    input=user_prompt,
+                    metadata=PromptMetadata(model=model_metadata_with_system_prompt),
+                    parameters=parameters,
+                )
+                serialized_prompts.append(user_prompt)
+                user_prompt = ""
```
saqadri, line 109:
```suggestion
                user_prompt = None
```

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -74,40 +49,66 @@ def serialize(
...
+            model_metadata = ModelMetadata(
+                **{"name": self.id(), "settings": copy.deepcopy(conversation_data)}
+            )
+        # Remove messages array from model metadata. Handled separately
+        model_metadata.settings.pop("messages", None)
+
+        system_prompt = ""
+        user_prompt = ""
+
+        for message in conversation_data["messages"]:
```
saqadri, line 82: This doesn't handle `function` role

saqadri (changes requested):

LastMileNewGrad:
Added multiple changes and refactors, to match ts implementation; old comments are stale or fixed

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -74,40 +50,81 @@ def serialize(
...
-            model_metadata = {"name": model_name, "settings": prompt_model_metadata}
-
-        return Prompt(
-            name=prompt_name,
-            input=prompt_input,
-            metadata=PromptMetadata(model=PromptMetadata(**model_metadata)),
-            parameters=parameters,
-            **kwargs
-        )
+                user_prompt = Prompt(
```
saqadri, line 104: `prompt`

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -74,40 +50,81 @@ def serialize(
...
+                    outputs=[
+                        ExecuteResult(
+                            output_type="execute_result",
+                            execution_count=None,
+                            data=assistant_response,
+                            metadata={},
+                        )
+                    ]
+                    if assistant_response
+                    else [],
```
saqadri, line 123: `else None`

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -74,40 +50,81 @@ def serialize(
...
+                            "parameters": parameters,
+                            "remember_chat_context": True,
+                        }
+                    ),
+                    outputs=[
+                        ExecuteResult(
+                            output_type="execute_result",
+                            execution_count=None,
+                            data=assistant_response,
+                            metadata={},
```
saqadri, line 119: `metadata=None`

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -74,40 +50,81 @@ def serialize(
...
+                        )
+                    ]
+                    if assistant_response
+                    else [],
+                )
+                serialized_prompts.append(user_prompt)
+            i += 1
+        return serialized_prompts
```
saqadri, line 127: `prompts[prompts.length - 1].name = promptName;`

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -74,40 +50,81 @@ def serialize(
...
-        # check if config already has model settings in global metadata
-        model_name = prompt_model_metadata.get("model", self.id())
-        global_model_metadata = ai_config.metadata.models.get(model_name, {})
-
-        if global_model_metadata:
-            # Check if the model settings from the input data are the same as the global model settings
-            # Compute the difference between the global model settings and the model settings from the input data
-            # If there is a difference, then we need to add the different model settings as overrides on the prompt's metadata
-            override_keys = set(prompt_model_metadata.keys()) - set(global_model_metadata.keys())
+        serialized_prompts = []
```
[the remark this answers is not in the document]
LastMileNewGrad: updated

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -74,40 +50,81 @@ def serialize(
...
+        conversation_data = {**data}
+
+        if not "messages" in conversation_data:
+            raise ValueError("Data must have `messages` array to match openai api spec")
+
+        # Find first system prompt. Every prompt in the config will bet set to use this system prompt.
+        system_prompt = None
+        for message in conversation_data["messages"]:
+            if "role" in message and message["role"] == "system":
+                system_prompt = message["content"]
```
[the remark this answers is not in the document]
LastMileNewGrad: updated

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -74,40 +50,81 @@ def serialize(
...
+            if current_message_role == "user" or current_message_role == "function":
+                # Serialize User message as a prompt and save the assistant response as an output
+                assistant_response = None
+                if i + 1 < len(conversation_data["messages"]):
+                    next_message = conversation_data["messages"][i + 1]
+                    if next_message["role"] == "assistant":
+                        assistant_response = next_message
+                        i += 1
+                new_prompt_name = (
+                    f"{prompt_name}_{len(serialized_prompts) + len(ai_config.prompts) +1}"
```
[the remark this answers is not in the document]
LastMileNewGrad: changed to `len(prompt)` as it should be, not depending on the config

python/src/aiconfig/default_parsers/openai.py
LastMileNewGrad: Metadata is defined to be of type `Dict`, pydantic will enforce this and not allow None type unless explicitly defined with `None` or `Optional`.

Will make a new diff on top of the stack to refactor to allow None as a default value. Will also update other types as per @jonathanlastmileai 's previous suggestion regarding setting defaults as `None` and handling from there.

LastMileNewGrad:
> Add this to openai.ts in this diff as well.
> 
> ```
>  const input: PromptInput =
>           message.role === "user"
>             ? message.content ?? ""
>             : {
>                 ...message,
>               };
> ```

Updated ts 

- modified `serialize()` on the `Runtime`  object to require a prompt name, it is __not__ an optional attribute
- updated test/demo to reflect

python/src/aiconfig/default_parsers/openai.py
LastMileNewGrad: addressed in comment below

python/src/aiconfig/default_parsers/openai.py
saqadri: This hasn't been resolved. Rename variable to `prompt`

saqadri (approved):

PR #34, opened 2023-10-19
OpenAI Api Chat Completion wrapper

LastMileNewGrad:
OpenAI Api Chat Completion wrapper

Wrapper for openai.ChatCompletion.Create()

Saves the latest prompt and system prompt in each call to config and disk under `aiconfig.json`

TODO: make a test

### DIFF
--- a/python/src/aiconfig/ChatCompletion.py
+++ b/python/src/aiconfig/ChatCompletion.py
@@ -1,0 +1,48 @@
+"""
+wrapper around openai.ChatCompletion.create that will serialize prompts and save them to config 
+
+usage:
+
+    Normal Import:
+        ` import openai`
+
+    Modified:
+        ```
+        import openai 
+        
+        from aiconfig.ChatCompletion import create_and_save_to_config
+        openai.ChatCompletion.create = create_and_save_to_config
+        ```
+"""
+
+from aiconfig.AIConfigSettings import ExecuteResult
+from aiconfig.Config import AIConfigRuntime
+import openai
+import asyncio
+
+aiconfig = AIConfigRuntime.create("")
+
+openai_chat_completion_create = openai.ChatCompletion.create
+
+
+def create_and_save_to_config(*args, **kwargs):
+    response = openai_chat_completion_create(*args, **kwargs)
+
+    prompts = asyncio.run(aiconfig.serialize(kwargs.get("model"), kwargs))
+    for i, new_prompt in enumerate(prompts):
+        in_config = False
+        for config_prompt in aiconfig.prompts:
+            # check for duplicates (same input and settings.)
+            if (
+                config_prompt.input == new_prompt.input
+                and new_prompt.metadata == config_prompt.metadata
+            ):
+                in_config = True
+                break
+        if not in_config:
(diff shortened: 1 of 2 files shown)


LastMileNewGrad:
Testplan:

Created a test harness which runs 2 OpenAI ChatCompletions. Override the api call automatically creates aiconfig

https://github.com/lastmile-ai/aiconfig/assets/141073967/6bbb5566-192c-48c5-a5b1-284a25958d9a

python/src/aiconfig/ChatCompletion.py
```diff
@@ -0,0 +1,47 @@
...
+        ```
+"""
+
+from aiconfig.Config import AIConfigRuntime
+import openai
+import asyncio
+
+aiconfig = AIConfigRuntime.create("")
+
+original_create = openai.ChatCompletion.create
```
saqadri, line 24:
```suggestion
openai_chat_completion_create = openai.ChatCompletion.create
```

python/src/aiconfig/ChatCompletion.py
```diff
@@ -0,0 +1,47 @@
...
+                config_prompt.input == new_prompt.input
+                and new_prompt.metadata == config_prompt.metadata
+            ):
+                in_config = True
+                break
+        if not in_config:
+            new_prompt_name = "prompt{}".format(str(len(aiconfig.prompts)))
+            new_prompt.name = new_prompt_name
+            aiconfig.add_prompt(new_prompt.name, new_prompt)
+    aiconfig.save()
```
saqadri, line 45: Where is this being saved though? We never specified the aiconfig file path

python/src/aiconfig/ChatCompletion.py
```diff
@@ -0,0 +1,47 @@
...
+import openai
+import asyncio
+
+aiconfig = AIConfigRuntime.create("")
+
+original_create = openai.ChatCompletion.create
+
+
+def create_and_save_to_config(*args, **kwargs):
+    response = original_create(*args, **kwargs)
```
saqadri, line 28: use kwargs to include a few aiconfig-specific properties:
* aiconfig (pass in the AIConfigRuntime object to use to save the completion params and result)
* aiconfig_path (pass in the path to save a new AIConfigRuntime object to)
* Maybe also aiconfig_name and aiconfig_description

saqadri (approved):
Looks good! Just a few comments

python/src/aiconfig/ChatCompletion.py
LastMileNewGrad: This change would make the completion call more complicated since you now need to modify each API call

For instance:

Before: `openai.ChatCompletion.create(completion_params)`

After: `openai.ChatCompletion.create(completion_params, arg1, arg2, arg3, ...)`

Is this the intended change?

LastMileNewGrad: Incorporating changes in a new diff as discussed offlinem cc: @saqadri

python/src/aiconfig/ChatCompletion.py
LastMileNewGrad: updated

PR #35, opened 2023-10-20
[Python] Ease of Use Changes

LastMileNewGrad:
[Python] Ease of Use Changes

- Modified imports in __init__.py
- Updated import in test

This diff Allows imports to be smaller and easier to use.

`from aiconfig.AIConfigSettings import AIConfig` -> `from aiconfig import AIConfig`


saqadri (approved):

LastMileNewGrad:
Updated imports to include important classes

PR #36, opened 2023-10-20
Starting point of a Router Interface

LastMileNewGrad:
Starting point of a Router Interface

Starting point of a routing interface

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/36).
* __->__ #36
* #35

### DIFF
--- a/python/src/aiconfig/router.py
+++ b/python/src/aiconfig/router.py
@@ -1,0 +1,24 @@
+from abc import ABCMeta, abstractmethod
+from typing import List
+from aiconfig import Prompt
+from aiconfig.AIConfigSettings import Output
+from aiconfig.Config import AIConfigRuntime
+
+class Router(Metaclass=ABCMeta):
+
+    def __init__(self, config: AIConfigRuntime):
+        self.config = config
+
+    @abstractmethod
+    def find(input, **kwargs) -> List[Prompt]:
+        """
+        Search and select a list of suitable prompts
+        """
+        pass
+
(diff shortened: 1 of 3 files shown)


python/src/aiconfig/router.py
```diff
@@ -0,0 +1,24 @@
...
+
+    @abstractmethod
+    def find(input, **kwargs) -> List[Prompt]:
+        """
+        Search and select a list of suitable prompts
+        """
+        pass
+
+    @abstractmethod
+    def route(self, **kwargs) -> Output:
```
saqadri, line 20: This also needs the input to use to route with.
```suggestion
    def route(self, input, **kwargs) -> Output:
```

python/src/aiconfig/router.py
```diff
@@ -0,0 +1,24 @@
...
+from aiconfig.AIConfigSettings import Output
+from aiconfig.Config import AIConfigRuntime
+
+class Router(Metaclass=ABCMeta):
+
+    def __init__(self, config: AIConfigRuntime):
+        self.config = config
+
+    @abstractmethod
+    def find(input, **kwargs) -> List[Prompt]:
```
saqadri, line 13:
```suggestion
    def find(self, input, **kwargs) -> List[Prompt]:
```

saqadri (approved):
Just add some more comments to the class as well

PR #37, opened 2023-10-25
[Python] 1\n Added helpers for retrieving & merging model settings

LastMileNewGrad:
[Python] 1\n Added helpers for retrieving & merging model settings

This diff adds two helpers which ensure that retreived model settings properly contains relevant global settings.

- `extract_override_settings`
- `generate_model_metadata`

Added test for `extract_override_settings`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/37).
* #33
* #38
* #39
* __->__ #37

### DIFF
--- a/python/src/aiconfig/AIConfigSettings.py
+++ b/python/src/aiconfig/AIConfigSettings.py
@@ -1,6 +1,8 @@
+import copy
 from dataclasses import dataclass
 import json
 from typing import Any, Dict, Literal, Optional, Union, List
+from aiconfig.util.config_utils import extract_override_settings
 from pydantic import BaseModel
 
 # Pydantic doesn't handle circular type references very well, TODO: handle this better than defining as type Any
@@ -360,6 +362,30 @@ class AIConfig(BaseModel):
         # remove from prompt list
         self.prompts = [prompt for prompt in self.prompts if prompt.name != prompt_name]
 
+    def generate_model_metadata(
+        self, inference_settings: InferenceSettings, model_id: str
+    ) -> ModelMetadata:
+        """
+        Generate a model metadata object based on the provided inference settings
+
+        This function takes the inference settings and the model ID and generates a ModelMetadata object.
+
+        Args:
+            inference_settings (InferenceSettings): The inference settings.
+            model_id (str): The model id.
+
+        Returns:
+            ModelMetadata: The model metadata.
+        """
+
+        overriden_settings = extract_override_settings(self, inference_settings, model_id)
+
+        if not overriden_settings:
+            model_metadata = ModelMetadata(**{"name": model_id})
+        else:
+            model_metadata = ModelMetadata(**{"name": model_id, "settings": overriden_settings})
+        return model_metadata
+
     def update_model(
         self, model_metadata: Dict | ModelMetadata, prompt_name: Optional[str] = None
     ):
@@ -384,7 +410,7 @@ class AIConfig(BaseModel):
                 )
             prompt.metadata.model = model_metadata
         else:
-            self.metadata.models[model_metadata.name] = model_metadata
+            self.metadata.models[model_metadata.name] = model_metadata.settings
 
     def set_metadata(self, key: str, value: Any, prompt_name: Optional[str] = None):
         """
--- a/python/tests/test_programmatically_create_an_AIConfig.py
+++ b/python/tests/test_programmatically_create_an_AIConfig.py
@@ -1,3 +1,4 @@
+from aiconfig.util.config_utils import extract_override_settings
 import pytest
 from aiconfig.Config import AIConfigRuntime
 from aiconfig.AIConfigSettings import (
@@ -392,8 +393,9 @@ def test_get_prompt_nonexistent(ai_config_runtime: AIConfigRuntime):
 def test_update_model_ai_config(ai_config_runtime: AIConfigRuntime):
     model_metadata = {"name": "testmodel", "settings": {"topP": 0.9}}
     ai_config_runtime.update_model(model_metadata)
-    assert ai_config_runtime.metadata.models["testmodel"] == ModelMetadata(
-        **{"name": "testmodel", "settings": {"topP": 0.9}}
+    assert (
+        ai_config_runtime.metadata.models["testmodel"]
+        == ModelMetadata(**{"name": "testmodel", "settings": {"topP": 0.9}}).settings
     )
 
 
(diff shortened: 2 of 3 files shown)


python/src/aiconfig/AIConfigSettings.py
```diff
@@ -360,6 +361,60 @@ def delete_prompt(self, prompt_name: str):
...
+
+        Args:
+            inference_settings (InferenceSettings): The inference settings.
+            model_id (str): The model id.
+
+        Returns:
+            ModelMetadata: The model metadata.
+        """
+
+        overriden_settings = self.extract_override_settings(inference_settings, model_id)
```
jonathanlastmileai, line 410: just wondering why we need to distinguish the overridden settings from the rest

jonathanlastmileai (changes requested):
Code LG, just a question

python/src/aiconfig/AIConfigSettings.py
LastMileNewGrad: The idea is that aiconfig includes global model/inference settings and prompt-level settings.

As a result, any settings applied at the prompt level should act as overrides, and these changes should be reflected when updated the core data in the config

python/src/aiconfig/AIConfigSettings.py
```diff
@@ -360,6 +361,60 @@ def delete_prompt(self, prompt_name: str):
...
+            return override_settings
+        return inference_settings
+
+    def generate_model_metadata(
+        self, inference_settings: InferenceSettings, model_id: str
+    ) -> ModelMetadata:
+        """
+        Generate a model metadata object based on the provided inference settings
+
+        This function takes the inferene settings and the model ID and generates a ModelMetadata object.
```
saqadri, line 400:
```suggestion
        This function takes the inference settings and the model ID and generates a ModelMetadata object.
```

python/src/aiconfig/AIConfigSettings.py
```diff
@@ -384,7 +439,7 @@ def update_model(
                 )
             prompt.metadata.model = model_metadata
         else:
-            self.metadata.models[model_metadata.name] = model_metadata
+            self.metadata.models[model_metadata.name] = model_metadata.settings
```
saqadri, line 442: what if `.settings` doesn't exist?

python/tests/test_programmatically_create_an_AIConfig.py
```diff
@@ -465,17 +470,39 @@ def test_add_output_existing_prompt_no_overwrite(ai_config_runtime: AIConfigRunt
...
 
-    assert ai_config_runtime.get_latest_output("GreetingPrompt") == None
\ No newline at end of file
+    assert ai_config_runtime.get_latest_output("GreetingPrompt") == None
+
+
+def test_extract_override_settings(ai_config_runtime: AIConfigRuntime):
+    settings = {"topP": 0.9}
+    override = ai_config_runtime.extract_override_settings(settings, "testmodel")
+    assert override == {"topP": 0.9}
```
saqadri, line 500: What is this testing? What is the value that it's overriding?

python/tests/test_programmatically_create_an_AIConfig.py
```diff
@@ -465,17 +470,39 @@ def test_add_output_existing_prompt_no_overwrite(ai_config_runtime: AIConfigRunt
...
+
+    ai_config_runtime.add_model("testmodel", {"topP": 0.8})
+    override = ai_config_runtime.extract_override_settings(settings, "testmodel")
+    assert override == {"topP": 0.9}
+
+    ai_config_runtime.update_model(ModelMetadata(name="testmodel", settings={"topP": 0.9}))
+    override = ai_config_runtime.extract_override_settings(settings, "testmodel")
+    assert override == {}
```
saqadri, line 508: I'm confused, why is override empty?

python/src/aiconfig/AIConfigSettings.py
```diff
@@ -360,6 +361,60 @@ def delete_prompt(self, prompt_name: str):
         # remove from prompt list
         self.prompts = [prompt for prompt in self.prompts if prompt.name != prompt_name]
 
+    def extract_override_settings(self, inference_settings: InferenceSettings, model_id: str):
```
saqadri, line 364: Should this function just be somewhere in utils? It's a bit confusing why it's a core function in this class. Not a huge deal as long as it works properly.

saqadri (approved):
Approving to unblock, but please make sure this works properly with edge cases.

python/src/aiconfig/AIConfigSettings.py
jonathanlastmileai: That makes sense, but not sure why the overridden settings need to be distinguished explicitly.

python/src/aiconfig/AIConfigSettings.py
LastMileNewGrad: moved fn to util file

python/src/aiconfig/AIConfigSettings.py
LastMileNewGrad: the logic flow ensures that .settings always exists; if input is of type dict, it gets converted into a ModelMetadata object which always has a `.settings` (defaulted to an empty dictionary)

python/tests/test_programmatically_create_an_AIConfig.py
LastMileNewGrad: This test function takes in an empty `ai_config_runtime`. Test Cases in this test are: 

1.  
Config settings = {}
settings = {"topP": 0.9}
-> 
Expected Override = {"topP": 0.9}

2.
Config settings = {"topP": 0.8} 
settings = {"topP": 0.9}
-> 
Expected Override = {"topP": 0.9}

3. 
Config settings = {"topP": 0.9}
settings = {}
-> 
Expected Override = {"}

Updated test with comments for clarity.

python/tests/test_programmatically_create_an_AIConfig.py
LastMileNewGrad: described test in comment above

jonathanlastmileai (approved):
We whiteboarded this and came up with an alternative to recalculating which settings are overrides. Namely, can maintain state where globals and overrides are distinguished and make the update logic maintain those invariants.

This isn't blocking, just a strategy that was a bit more intuitive to me.

PR #39, opened 2023-10-25
[Python] 2\n helper fn's to Augment OpenAI Serialize()

LastMileNewGrad:
[Python] 2\n helper fn's to Augment OpenAI Serialize()

This diff introduces two new helper functions inside the openai model parser.

- `add_prompt_as_message`
- `has_valid_prompt_string`

First one  converts prompts into messages and appends them to a list of messages. Designed to augment serialize()

Second one validates if a prompt's input type.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/39).
* #33
* #38
* __->__ #39
* #37

### DIFF
--- a/python/src/aiconfig/default_parsers/openai.py
+++ b/python/src/aiconfig/default_parsers/openai.py
@@ -1,6 +1,6 @@
 from abc import abstractmethod
 import copy
-from typing import Dict, Optional, Union
+from typing import Dict, List, Optional, Union
 from typing import TYPE_CHECKING, Any, Dict, Optional
 from aiconfig import AIConfigSettings
 from aiconfig.AIConfigSettings import (
@@ -349,3 +349,47 @@ def refine_chat_completion_params(model_settings):
             completion_data[key] = model_settings[key]
 
     return completion_data
+
+
+def add_prompt_as_message(
+    prompt: Prompt, aiconfig: "AIConfigRuntime", messages: List, params=None
+):
+    """
+    Converts a given prompt to a message and adds it to the specified messages list.
+
+    Note:
+    - If the prompt contains valid input, it's treated as a user message.
+    - If the prompt has a custom role, function call, or name, these attributes are included in the message.
+    - If an AI model output exists, it is appended to the messages list.
+    """
+    if is_prompt_template(prompt):
+        resolved_prompt = resolve_prompt(prompt, params, aiconfig)
+        messages.append({"content": resolved_prompt, "role": "user"})
+    else:
+        # Assumes Prompt input will be in the format of ChatCompletionMessageParam (with content, role, function_name, and name attributes)
+        resolved_prompt = resolve_prompt(prompt.input.content, params, aiconfig)
+
+        prompt_input = prompt.input
+        role = prompt_input.role if hasattr(prompt_input, "role") else "user"
+        fn_call = prompt_input.function_call if hasattr(prompt_input, "function_call") else None
+        name = prompt_input.name if hasattr(prompt_input, "name") else None
+        messages.append(
+            {"content": resolved_prompt, "role": role, "function_call": fn_call, "name": name}
+        )
+
+    output = aiconfig.get_latest_output(prompt)
+    if output:
+        if output.output_type == "execute_result":
+            output_message = output.data
+            if output_message["role"] == "assistant":
+                messages.append(output_message)
+    return messages
+
+
+def is_prompt_template(prompt: Prompt):
+    """
+    Check if a prompt's input is a valid string.
+    """
+    return isinstance(prompt.input, str) or (
+        hasattr(prompt.input, "data") and isinstance(prompt.input.data, str)
+    )
--- a/python/src/aiconfig/AIConfigSettings.py
+++ b/python/src/aiconfig/AIConfigSettings.py
@@ -1,6 +1,8 @@
+import copy
 from dataclasses import dataclass
 import json
 from typing import Any, Dict, Literal, Optional, Union, List
+from aiconfig.util.config_utils import extract_override_settings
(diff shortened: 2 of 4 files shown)


python/src/aiconfig/default_parsers/openai.py
```diff
@@ -349,3 +349,45 @@ def refine_chat_completion_params(model_settings):
...
+    output = aiconfig.get_latest_output(prompt)
+    if output:
+        if output.output_type == "execute_result":
+            output_message = output.data
+            if output_message["role"] == "assistant":
+                messages.append(output_message)
+    return messages
+
+
+def has_valid_prompt_string(prompt: Prompt):
```
jonathanlastmileai, line 387: this feels a bit brittle. Can a prompt have a statically known structure?

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -349,3 +349,45 @@ def refine_chat_completion_params(model_settings):
             completion_data[key] = model_settings[key]
 
     return completion_data
+
+
+def add_prompt_as_message(
+    prompt: Prompt, aiconfig: "AIConfigRuntime", messages: list, params=None
```
jonathanlastmileai, line 355: `list` is a constructor; you want typing.List

jonathanlastmileai (changes requested):
Just a couple of questions here as well

python/src/aiconfig/default_parsers/openai.py
LastMileNewGrad: No, The type definition of the prompt input is either a string or arbitrary data. We would have dynamically "parse" it.

Any suggestions on how you would make this not as brittle?

python/src/aiconfig/default_parsers/openai.py
LastMileNewGrad: updated

python/src/aiconfig/default_parsers/openai.py
jonathanlastmileai: Need some more context - why can it be arbitrary data? What are the use cases?

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -349,3 +349,45 @@ def refine_chat_completion_params(model_settings):
...
+    Note:
+    - If the prompt contains valid input, it's treated as a user message.
+    - If the prompt has a custom role, function call, or name, these attributes are included in the message.
+    - If an AI model output exists, it is appended to the messages list.
+    """
+    if has_valid_prompt_string(prompt):
+        resolved_prompt = resolve_prompt(prompt, params, aiconfig)
+        messages.append({"content": resolved_prompt, "role": "user"})
+    else:
+        # prompt data can have function role
```
saqadri, line 369: Add comment in code that this assumes it is a ChatCompletionMessageParam object type (with content and role)

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -349,3 +349,45 @@ def refine_chat_completion_params(model_settings):
...
+        prompt_input = prompt.input
```
saqadri, line 370: Add
```
resolved_prompt = resolve_prompt(prompt.input.content, params, aiconfig)
```

python/src/aiconfig/default_parsers/openai.py
saqadri: Let's chat in person on this. For more context, this class is a parser for OpenAI models specifically. It serializes and deserializes prompts in a specific structure in aiconfig, and so this should be safe to do in this class because only prompts that were serialized by this class are processed by it.

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -349,3 +349,45 @@ def refine_chat_completion_params(model_settings):
...
+    output = aiconfig.get_latest_output(prompt)
+    if output:
+        if output.output_type == "execute_result":
+            output_message = output.data
+            if output_message["role"] == "assistant":
+                messages.append(output_message)
+    return messages
+
+
+def has_valid_prompt_string(prompt: Prompt):
```
saqadri, line 387: Rename to `is_prompt_template`

python/src/aiconfig/default_parsers/openai.py
LastMileNewGrad: updated comment

python/src/aiconfig/default_parsers/openai.py
LastMileNewGrad: updated

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -349,3 +349,47 @@ def refine_chat_completion_params(model_settings):
...
+    - If the prompt contains valid input, it's treated as a user message.
+    - If the prompt has a custom role, function call, or name, these attributes are included in the message.
+    - If an AI model output exists, it is appended to the messages list.
+    """
+    if is_prompt_template(prompt):
+        resolved_prompt = resolve_prompt(prompt, params, aiconfig)
+        messages.append({"content": resolved_prompt, "role": "user"})
+    else:
+        # Assumes Prompt input will be in the format of ChatCompletionMessageParam (with content, role, function_name, and name attributes)
+        resolved_prompt = resolve_prompt(prompt.input.content, params, aiconfig)
```
saqadri, line 370: what happens to `resolve_prompt` if content is null?

saqadri (approved):

jonathanlastmileai (approved):
Let's ship this and come back to the questions in a refactor if necessary

PR #42, opened 2023-10-27
ci/cd

LastMileNewGrad:
ci/cd

github workflow to run python unit tests

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/42).
* __->__ #42
* #41


LastMileNewGrad:
updated name per comment on other diff

python/requirements.txt
```diff
@@ -5,4 +5,5 @@ pydantic>=2.1
 pybars3
 google-generativeai
 openai
-python-dotenv
\ No newline at end of file
+python-dotenv
+huggingface_hub
```
saqadri, line 9: nit: don't think this needs to be part of this diff. But ok since I think you're adding it soon.

saqadri (approved):

PR #43, opened 2023-10-30
[Python] Tests File paths absolute -> relative

LastMileNewGrad:
[Python] Tests File paths absolute -> relative

Updated tests to allow relative file path handling for opening aiconfigs while testing.

- Created a util method that takes in a relative file path and converts it into an absolute file path
- Updated tests to use Util method

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/43).
* #46
* #42
* __->__ #43


python/tests/yarn.lock
```diff
@@ -0,0 +1,67 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
suyoglastmileai, line 1: You might have run `yarn add` in this folder which is why it made this `yarn.lock` in the `python/tests` folder (guessing it shouldn't be here?)

LastMileNewGrad: That was a remnant of my tests, removed

LastMileNewGrad:
Updated diff to be python only, TS on top

saqadri (approved):

PR #48, opened 2023-10-30
[Typescript][Long] Added HuggingFace Text Generation Parser.

LastMileNewGrad:
[Typescript] Added HuggingFace Text Generation Parser.

HF text gen model parser.
- Built off Ryan's previous diff on building Model Parser
- Doesn't require an api key but it can be specified.

TS HuggingFace api differs from the Python side in a couple of ways ie:
- different apis for streaming,
- Python lets you specify details to be included, is default in TS

- Modified aiconfig.registerModelParser() with generic Any, otherwise it defaults to JSONObject inside ModelParser and doesn't allow other types

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/48).
* __->__ #48
* #57

### DIFF
--- a/python/src/aiconfig/Config.py
+++ b/python/src/aiconfig/Config.py
@@ -1,6 +1,7 @@
 import json
 import sys
 import os
+from aiconfig.default_parsers.hf import HuggingFaceTextParser
 import requests
 from typing import ClassVar, Dict, List, Optional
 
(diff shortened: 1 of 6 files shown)


LastMileNewGrad:
Testplan:
Created a test harness for both py and ts and validated both HF model parsers work as expected and have the same outputs.

Minor differences in handling null, None, empty types

PR #49, opened 2023-10-31
AIConfig Schema updates

saqadri:
AIConfig Schema updates

A few schema updates to AIConfig based on feedback from bug bash and team brainstorming:

1. Add a `default_model` property to AIConfig metadata. If used, we can simplify the prompt to not have to specify any model information at all (see item #5 in Bug Bash: https://docs.google.com/document/d/1tMu-N53AzCKc_uPEeLARLgdnpVLM5HPQjlqOPs3Z0B4/edit)

2. Add `model_parsers` mapping property to AIConfig metadata. This allows us to specify a model parser ID in the config itself. This is based on Ryan's feedback for things like Hugging Face model parsers, which can support hundreds of different models. Now, the models that someone uses in the AIConfig can register the HF model parser in the config, saving us from having to manage a rapidly changing list of model IDs explicitly. (see https://github.com/lastmile-ai/aiconfig/pull/25)

3. Add `mime_type` property to Output type. This will allow us to support multi-modal outputs. Note that for PromptInput type, I have left it as-is intentionally. `PromptInput.data` is currently very flexible -- it can support multiple inputs and in theory can specify mime-types. We can iterate on multi-modal inputs separately.

### DIFF
--- a/typescript/lib/config.ts
+++ b/typescript/lib/config.ts
@@ -117,6 +117,25 @@ export class AIConfigRuntime implements AIConfig {
     );
 
     Object.assign(aiConfig, remainingProps);
+
+    // Update the ModelParserRegistry with any model parsers specified in the AIConfig
+    if (aiConfig.metadata.model_parsers) {
+      for (const [modelName, modelParserId] of Object.entries(
+        aiConfig.metadata.model_parsers
+      )) {
+        const modelParser = ModelParserRegistry.getModelParser(modelParserId);
+        if (!modelParser) {
+          throw new Error(
+            `E1001: Unable to load AIConfig: It specifies ${JSON.stringify(
+              aiConfig.metadata.model_parsers
+            )}, but ModelParser ${modelParserId} for model ${modelName} does not exist. Make sure you have registered the ModelParser using AIConfigRuntime.registerModelParser before loading the AIConfig.`
+          );
+        }
+
+        ModelParserRegistry.registerModelParser(modelParser, [modelName]);
+      }
+    }
+
     return aiConfig;
   }
 
@@ -139,7 +158,7 @@ export class AIConfigRuntime implements AIConfig {
       .then((response) => {
         if (response.status !== 200) {
           throw new Error(
-            `Failed to load workbook. Status code: ${response.status}`
+            `E1005: Failed to load workbook. Status code: ${response.status}`
           );
         }
 
@@ -275,7 +294,7 @@ export class AIConfigRuntime implements AIConfig {
     modelName: string,
     data: JSONObject,
     promptName: string,
-    params?: JSONObject,
+    params?: JSONObject
   ): Promise<Prompt | Prompt[]> {
     const modelParser = ModelParserRegistry.getModelParser(modelName);
     if (!modelParser) {
@@ -285,7 +304,7 @@ export class AIConfigRuntime implements AIConfig {
         )}: ModelParser for model ${modelName} does not exist`
       );
     }
-    
+
     const prompts = modelParser.serialize(promptName, data, this, params);
     return prompts;
   }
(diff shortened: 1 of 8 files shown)


typescript/lib/config.ts
```diff
@@ -117,6 +117,25 @@ export class AIConfigRuntime implements AIConfig {
...
+        const modelParser = ModelParserRegistry.getModelParser(modelParserId);
+        if (!modelParser) {
+          throw new Error(
+            `E1001: Unable to load AIConfig: It specifies ${JSON.stringify(
+              aiConfig.metadata.model_parsers
+            )}, but ModelParser ${modelParserId} for model ${modelName} does not exist. Make sure you have registered the ModelParser before loading the AIConfig.`
+          );
+        }
+
+        ModelParserRegistry.registerModelParser(modelParser, [modelName]);
```
LastMileNewGrad, line 135: Should we consider updating the `modelParser` ID to `modelName` before registering it? ie `modelParser.id=modelName`

This might not be necessary if ModelParser never needs an explicit reference to ModelName since it usually has a reference to the prompt data.

typescript/lib/config.ts
```diff
@@ -117,6 +117,25 @@ export class AIConfigRuntime implements AIConfig {
...
+    if (aiConfig.metadata.model_parsers) {
+      for (const [modelName, modelParserId] of Object.entries(
+        aiConfig.metadata.model_parsers
+      )) {
+        const modelParser = ModelParserRegistry.getModelParser(modelParserId);
+        if (!modelParser) {
+          throw new Error(
+            `E1001: Unable to load AIConfig: It specifies ${JSON.stringify(
+              aiConfig.metadata.model_parsers
+            )}, but ModelParser ${modelParserId} for model ${modelName} does not exist. Make sure you have registered the ModelParser before loading the AIConfig.`
```
LastMileNewGrad, line 131: Minor suggestion: consider updating error message `before loading the AIConfig with AIConfigRuntime.registerModelParser()`

typescript/lib/config.ts
```diff
@@ -506,6 +525,37 @@ export class AIConfigRuntime implements AIConfig {
...
   }
 
+  /**
+   * Sets the model to use for all prompts by default in the AIConfig.
+   * @param modelName The name of the model to default to.
+   */
+  public setDefaultModel(modelName: string | undefined) {
+    if (modelName == null) {
+      delete this.metadata.default_model;
+      return;
```
LastMileNewGrad, line 535: would it be beneficial to have a dedicated delete method?

LastMileNewGrad (approved):

typescript/types.ts
```diff
@@ -141,67 +165,23 @@ export type ExecuteResult = {
...
-  /**
-   * A mime-type keyed dictionary of data
-   */
-  data: {
-    /**
-     * mimetype output (e.g. text/plain), represented as either an array of strings or a string.
-     */
-    [k: string]: string | string[];
-  };
+  mime_type?: string;
```
LastMileNewGrad, line 175: Can you give an example of what this might look like?

typescript/lib/config.ts
saqadri: Resolved offline -- basically now we have a clearer definition of model parser ID vs. model ID. Both are needed.

typescript/lib/config.ts
saqadri: Discussed offline. Keeping this approach for now (consistent with `set env=` in CLI) -- we can add a delete method later if needed.

typescript/types.ts
saqadri: Discussed offline and went over mime types.

typescript/lib/config.ts
```diff
@@ -506,6 +525,37 @@ export class AIConfigRuntime implements AIConfig {
...
+
+    this.metadata.default_model = modelName;
+  }
+
+  /**
+   * Adds a model name : model parser ID mapping to the AIConfig metadata. This model parser will be used to parse Prompts in the AIConfig that use the given model.
+   * @param modelName The name of the model to set the parser for.
+   * @param modelParserId The ID of the model parser to use for the model. If undefined, the model parser for the model will be removed.
+   */
+  public setModelParser(modelName: string, modelParserId: string | undefined) {
```
LastMileNewGrad, line 546: Should this be synced to the ModelParserRegistry?

`config.set_model_parser("gpt-3.5-turbo", "HuggingFaceTextParser")` would update the the config but user would then have to manually run
`ModelParserRegistry.register_model_parser`

Could also make sense to make ModelParserRegistry a non static class and have it be an attribute of AIConfigRuntime

PR #51, opened 2023-10-31
[Typescript] Introduce the getGlobalSettings API Method

LastMileNewGrad:
[Typescript] Introduce the getGlobalSettings API Method

This diff introduces the `getGlobalSetting` API method, which retrieves global settings for a specified model.

One of the steps in enhancing ModelParser creation. More in diffs on top.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/51).
* #76
* #75
* #55
* #52
* #54
* __->__ #51
* #50

### DIFF
--- a/typescript/__tests__/testProgramaticallyCreateConfig.ts
+++ b/typescript/__tests__/testProgramaticallyCreateConfig.ts
@@ -1,0 +1,20 @@
+import { AIConfigRuntime } from "../lib/config";
+import { InferenceSettings } from "../types";
+
+describe("test Get Global Settings", () => {
+  // shared state
+  let aiConfigRuntime: AIConfigRuntime = AIConfigRuntime.create(
+    "Untitled AIConfig",
+    /*param description*/ undefined,
+    "latest",
+    { models: { testmodel: { topP: 0.9 } } }
+  );
+
+  test("Retrieving global setting from AIConfig with 1 model", () => {
+    const modelId: string = "testmodel";
+
+    const globalSettingsForTestModel =
+      aiConfigRuntime.getGlobalSettings(modelId);
+    expect(globalSettingsForTestModel).toEqual({ topP: 0.9 });
+  });
+});
--- a/python/src/aiconfig/util/config_utils.py
+++ b/python/src/aiconfig/util/config_utils.py
@@ -3,8 +3,9 @@ from typing import TYPE_CHECKING
 
 import copy
 
+
 if TYPE_CHECKING:
-    from aiconfig import AIConfigSettings
+    from ..AIConfigSettings import AIConfig
     from aiconfig.AIConfigSettings import InferenceSettings
     from aiconfig.Config import AIConfigRuntime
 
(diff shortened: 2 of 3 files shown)


typescript/__tests__/testProgramaticallyCreateConfig.ts
```diff
@@ -0,0 +1,20 @@
+import { AIConfigRuntime } from "../lib/config";
+import { InferenceSettings } from "../types";
+
+describe("test extract_override_settings", () => {
```
saqadri, line 4: Update description. Doesn't look correct for this test

typescript/__tests__/testProgramaticallyCreateConfig.ts
```diff
@@ -0,0 +1,20 @@
...
+  // shared state
+  let aiConfigRuntime: AIConfigRuntime = AIConfigRuntime.create(
+    "Untitled AIConfig",
+    undefined,
+    undefined,
```
saqadri, line 9: When adding "undefined" as a value, add /*param*/ as a comment on that line to describe which param is undefined.
`/*param*/ description`

Also can you double check if this is correct? Looks like you are setting schemaVersion to undefined? Just set it to "latest"

typescript/__tests__/testProgramaticallyCreateConfig.ts
```diff
@@ -0,0 +1,20 @@
...
+    { models: { testmodel: { topP: 0.9 } } }
+  );
+
+  it("Test Case 1: Retrieve global setting from AIConfig with 1 model", () => {
```
saqadri, line 13: nit: change from `it` to `test`. They are the same.

Also update the description:
`test("Retrieving global settings from AIConfig with 1 model")`.
Reson: in Jest the test names are supposed to read like a proper sentence. In this case: "test retrieving global settings from AIConfig with 1 model"

If you use `it` is can be `it("Retrieves global settings from AIConfig with 1 model")` so the sentence reads "it retrieves global settings from AIConfig with 1 model"

saqadri (approved):

typescript/__tests__/testProgramaticallyCreateConfig.ts
LastMileNewGrad: Updated to `test Get Global Settings` to accurately describe which method its testing

typescript/__tests__/testProgramaticallyCreateConfig.ts
LastMileNewGrad: updated and used `latest` as suggested

typescript/__tests__/testProgramaticallyCreateConfig.ts
LastMileNewGrad: that's a cool style guideline. Is there a javascript/typescript style guidelines like Python's PEP?

PR #52, opened 2023-10-31
[Typescript] Introduce Extract Override Settings method

LastMileNewGrad:
[Typescript] Introduce Extract Override Settings method

This diff introduces the `extractOverrideSettings` function. One of the abstractions which will make it easier to construct model sttings

- Abstracted existing logic to get overrides
- Added test. Same test & logic as the python side

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/52).
* #91
* #76
* #75
* #55
* __->__ #52

### DIFF
--- a/typescript/__tests__/testProgramaticallyCreateConfig.ts
+++ b/typescript/__tests__/testProgramaticallyCreateConfig.ts
@@ -1,5 +1,8 @@
 import { AIConfigRuntime } from "../lib/config";
-import { InferenceSettings } from "../types";
+import { JSONObject } from "../common";
+
+import { InferenceSettings, ModelMetadata } from "../types";
+import { extractOverrideSettings } from "../lib/utils";
 
 describe("test Get Global Settings", () => {
   // shared state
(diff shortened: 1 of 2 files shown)


typescript/__tests__/testProgramaticallyCreateConfig.ts
```diff
@@ -18,3 +20,69 @@ describe("test extract_override_settings", () => {
     expect(globalSettingsForTestModel).toEqual({ topP: 0.9 });
   });
 });
+
+describe("test extract_override_settings", () => {
```
saqadri, line 24: Update `describe and `test` as I explained in the diff below

typescript/lib/util/configUtils.ts
```diff
@@ -0,0 +1,52 @@
...
+ *    This function takes the inference settings and a model ID and returns a subset
+ *    of inference settings that have been overridden by model-specific settings. It
+ *    compares the provided settings with global settings, and returns only those that
+ *    differ or have no corresponding global setting.
+ * @param configRuntime The AIConfigRuntime that the prompt belongs to.
+ * @param inferenceSettings The model settings from the input data.
+ * @param modelId The model ID of the prompt.
+ * @returns The model settings from the input data.
+ */
+export function extractOverrideSettings(
```
saqadri, line 18: Just put this in `utils.ts`

typescript/lib/util/configUtils.ts
```diff
@@ -0,0 +1,52 @@
...
+  configRuntime: AIConfigRuntime,
+  inferenceSettings: InferenceSettings,
+  modelId: string
```
saqadri, line 21: `modelName` since that's what we call it everywhere

typescript/lib/util/configUtils.ts
```diff
@@ -0,0 +1,52 @@
...
+) {
+  let modelMetadata: ModelMetadata | string;
+  const modelName = modelId;
+  const globalModelSettings: InferenceSettings =
+    configRuntime.getGlobalSettings(modelName) ?? ({} as InferenceSettings);
```
saqadri, line 26: `const globalModelSettings = { ...(configRuntime.getGlobalSettings(modelName) ?? {} }`

Same for inferenceSettings below.

saqadri (approved):
Approving to unblock but some comments

typescript/__tests__/testProgramaticallyCreateConfig.ts
LastMileNewGrad: Updated. Please take another look

typescript/lib/util/configUtils.ts
LastMileNewGrad: moved to utils.ts

typescript/lib/util/configUtils.ts
LastMileNewGrad: updated

typescript/lib/util/configUtils.ts
LastMileNewGrad: updated.

typescript/__tests__/testProgramaticallyCreateConfig.ts
saqadri: Looks good, ship it

PR #55, opened 2023-10-31
[Typescript] Introduce GenerateModelMetadata API method.

LastMileNewGrad:
[Typescript] Introduce GenerateModelMetadata API method.

This diff introduces a new api method to config, `generateModelMetadata()` to make it much easier to implement `serialize()` when writing a modelParser

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/55).
* #91
* #76
* #75
* __->__ #55
* #52

### DIFF
--- a/typescript/lib/config.ts
+++ b/typescript/lib/config.ts
@@ -15,6 +15,7 @@ import _ from "lodash";
 import { getAPIKeyFromEnv } from "./utils";
 import { ParameterizedModelParser } from "./parameterizedModelParser";
 import { OpenAIChatModelParser, OpenAIModelParser } from "./parsers/openai";
+import { extractOverrideSettings } from "./utils";
 
 export type PromptWithOutputs = Prompt & { outputs?: Output[] };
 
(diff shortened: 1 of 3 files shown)


typescript/lib/config.ts
```diff
@@ -822,5 +823,29 @@ export class AIConfigRuntime implements AIConfig {
...
   }
 
+  /**
+   * Generates a ModelMetadata object based on the provided inference settings and model ID.
+   *
+   * @param inferenceSettings - The inference settings to be used for the model.
+   * @param modelId - The unique identifier for the model.
+   * @returns A ModelMetadata object that includes the model's name and optional settings.
+   */
+  public generateModelMetadata(
```
saqadri, line 833: Maybe `mergeModelMetadata`?

typescript/lib/config.ts
```diff
@@ -822,5 +823,29 @@ export class AIConfigRuntime implements AIConfig {
...
+    inferenceSettings: InferenceSettings,
+    modelId: string
```
saqadri, line 835: `modelName`

typescript/lib/config.ts
```diff
@@ -822,5 +823,29 @@ export class AIConfigRuntime implements AIConfig {
...
+  ) {
+    const overrideSettings = extractOverrideSettings(
+      this,
+      inferenceSettings,
+      modelId
+    );
+
+    if (!overrideSettings) {
```
saqadri, line 843: if (!overrideSettings || overrideSettings === {})

saqadri (approved):

typescript/lib/config.ts
LastMileNewGrad: updated

typescript/lib/config.ts
LastMileNewGrad: Since overrideSettings is an object, javascript won't let me compare to {} with `=== {}`. Updated to `Object.keys(overrideSettings).length === 0`

typescript/lib/config.ts
LastMileNewGrad: While this method is merging settings, its primary purpose is to generate the modelMetadata object. The act of merging is a part of the process but I think it doesn't define the overall purpose of the function. Could we maybe do something like `getOverridesAndCreateModelMetadata`?

Updated docstring to reflect merging & override extraction:
`Generates a ModelMetadata object from the inference settings by extracting the settings that override the global settings.`

saqadri: My issue is that even as a creator of the library I don't understand what this function does, so it won't be clear to any user of the library either. Maybe just `getModelMetadata` then.

LastMileNewGrad: Updated to `getModelMetadata`. Can comeback to this if we think of something better.

PR #56, opened 2023-10-31
Starting point for aiconfig CLI

saqadri:
Starting point for aiconfig CLI, using oclif


rholinshead:
Do we need all the github stuff as part of this?

saqadri:
Closing unless we need this. cc @suyoglastmileai @rholinshead

PR #61, opened 2023-10-31
AIConfig Schema updates

LastMileNewGrad:
AIConfig Schema updates

A few schema updates to AIConfig based on feedback from bug bash and team brainstorming:

1. Add a `default_model` property to AIConfig metadata. If used, we can simplify the prompt to not have to specify any model information at all (see item #5 in Bug Bash: https://docs.google.com/document/d/1tMu-N53AzCKc_uPEeLARLgdnpVLM5HPQjlqOPs3Z0B4/edit)

2. Add `model_parsers` mapping property to AIConfig metadata. This allows us to specify a model parser ID in the config itself. This is based on Ryan's feedback for things like Hugging Face model parsers, which can support hundreds of different models. Now, the models that someone uses in the AIConfig can register the HF model parser in the config, saving us from having to manage a rapidly changing list of model IDs explicitly. (see https://github.com/lastmile-ai/aiconfig/pull/25)

3. Add `mime_type` property to Output type. This will allow us to support multi-modal outputs. Note that for PromptInput type, I have left it as-is intentionally. `PromptInput.data` is currently very flexible -- it can support multiple inputs and in theory can specify mime-types. We can iterate on multi-modal inputs separately.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/61).
* #64
* #63
* #62
* __->__ #61

### DIFF
--- a/schema/aiconfig.schema.json
+++ b/schema/aiconfig.schema.json
@@ -4,6 +4,7 @@
   "additionalProperties": {},
   "properties": {
     "name": {
+      "description": "Friendly name descriptor for the AIConfig. Could default to the filename if not specified.",
       "type": "string"
     },
     "description": {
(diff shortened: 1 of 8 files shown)


LastMileNewGrad:
thanks sapling

PR #62, opened 2023-10-31
[Python][Easy][Schema Update] Refactor `InferenceResponse` to `ExecuteResult`

LastMileNewGrad:
[Python][Easy][Schema Update] Refactor `InferenceResponse` to `ExecuteResult`

Switch 'InferenceResponse' to 'ExecuteResult'

- Changed all instances of `InferenceResponse` to `ExecuteResult` in type hints and references aligning with the library's design.
- Removed any remaining implementations or references of `InferenceResponse`

To provide some additional context, InferenceResponse was a type made similar to ExecuteResult to handle outputs

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/62).
* #80
* #79
* #81
* #77
* #72
* #74
* #71
* #70
* #69
* #68
* #67
* #66
* #73
* #64
* #63
* __->__ #62

PR #66, opened 2023-11-01
[Python][Schema Update] Default Model & Model Parser CRUD API

LastMileNewGrad:
[Python][Schema Update] Default Model & Model Parser CRUD API

- Introduced the capability to set a default model and parser in the CRUD API.

Important to consideration:
- Unlike Typescript, Python does not have an undefined type.
- an alternate approach to mimic this could be by deleting the attribute from the object. However, attempts to access it subsequently (example: del a.B, variable = a.B) would raise an exception. You have to check hasattr(a, "B") to use this approach.
- To simplify the implementation and to align it with the Typescript version, chose to use None type to represent an undefined state. Ex:  `a.B = None`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/66).
* #80
* #79
* #81
* #77
* #72
* #74
* #71
* #70
* #69
* #68
* #67
* __->__ #66
* #73
* #64
* #63
* #62


python/src/aiconfig/AIConfigSettings.py
```diff
@@ -218,6 +218,29 @@ def get_model_name(self, prompt: Union[str, Prompt]):
...
+        """
+        Sets the model to use for all prompts by default in the AIConfig. Set to None to delete the default model.
+
+        Args:
+            model_name (str): The name of the default model.
+        """
+        self.metadata.default_model = model_name
+    
+    def set_model_parser(self, model_name: str, model_parser_id: Union[str, None]):
+        """
```
LastMileNewGrad, line 232: Should this be synced to the ModelParserRegistry? Just thought of this while implementing the python side.

ex:
config.set_model_parser("gpt-3.5-turbo", "HuggingFaceTextParser") would update the the config but user would then have to manually run
ModelParserRegistry.register_model_parser

Could also make sense to make ModelParserRegistry a non static class and have it be an attribute of AIConfigRuntime

saqadri: Hmm, good catch. Maybe yes -- add to this and to TS at the top of the stack?

saqadri (approved):

python/src/aiconfig/AIConfigSettings.py
LastMileNewGrad: Will update on top of stack

PR #75, opened 2023-11-01
[TypeScript] Fix forced Type when creating a model parser

LastMileNewGrad:
[TypeScript] Fix forced Type when creating a model parser

- Modified aiconfig.`registerModelParser()` by using a generic `Any` instead of defaulting to JSONObject within ModelParser. This change enables the acceptance of types that do not conform to JSONObject.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/75).
* #91
* #76
* __->__ #75
* #55
* #52

### DIFF
--- a/typescript/lib/config.ts
+++ b/typescript/lib/config.ts
@@ -15,6 +15,7 @@ import _ from "lodash";
 import { getAPIKeyFromEnv } from "./utils";
 import { ParameterizedModelParser } from "./parameterizedModelParser";
 import { OpenAIChatModelParser, OpenAIModelParser } from "./parsers/openai";
+import { extractOverrideSettings } from "./utils";
 
 export type PromptWithOutputs = Prompt & { outputs?: Output[] };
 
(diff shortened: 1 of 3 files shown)


typescript/lib/config.ts
```diff
@@ -243,7 +243,10 @@ export class AIConfigRuntime implements AIConfig {
    * @param modelParser The model parser to add to the registry.
    * @param ids Optional list of model IDs to register the model parser for. If unspecified, the model parser will be registered for modelParser.id.
    */
-  public static registerModelParser(modelParser: ModelParser, ids?: string[]) {
+  public static registerModelParser(
+    modelParser: ModelParser<any, any>,
```
saqadri, line 247: This isn't great. Let's revert this. Not sure why this is needed

saqadri: Spoke offline. Will keep this and revisit in the future

saqadri (approved):

PR #76, opened 2023-11-01
[Typescript] Added HuggingFace Text Generation Parser.

LastMileNewGrad:
[Typescript] Added HuggingFace Text Generation Parser.

HF text gen model parser.
- Built off Ryan's previous diff on building Model Parser
- Doesn't require an api key but it can be specified.

TS HuggingFace api differs from the Python side in a couple of ways ie:
- different apis for streaming,
- - utilized both inside the TS impl
- Python lets you specify details to be included, no option in TS but default is most of those details

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/76).
* #91
* __->__ #76
* #75
* #55
* #52

### DIFF
--- a/typescript/demo/test-hf.ts
+++ b/typescript/demo/test-hf.ts
@@ -1,0 +1,66 @@
+import OpenAI from "openai";
+import * as path from "path";
+import { AIConfigRuntime } from "../lib/config";
+import { HuggingFaceTextGenerationModelParser } from "../lib/parsers/hf";
+import { Prompt } from "../types";
+
+async function run() {
+  const config = AIConfigRuntime.load(
+    path.join(__dirname, "/mistral-config.json")
+  );
+
+  // register HF MP
+  const mistralModelParser = new HuggingFaceTextGenerationModelParser(
+    "mistralai/Mistral-7B-v0.1",
+    true
+  );
+  AIConfigRuntime.registerModelParser(mistralModelParser, [
+    "mistralai/Mistral-7B-v0.1",
+  ]);
+
+  console.log("Deserialize Prompt1: ");
+  console.log(await config.resolve("prompt1"));
+  console.log("\nRun Prompt1: ");
+  console.log(await config.run("prompt1"));
+
+  console.log("Latest output: ", config.getOutputText("prompt1"));
+
+  console.log("serialize prompt2: ");
+  const prompts = (await config.serialize(
+    "mistralai/Mistral-7B-v0.1",
+    { inputs: "Hello, world!" },
+    "prompt2"
+  )) as Prompt[];
+
+  const prompt2 = prompts[0];
+
+  console.log("Prompt2: ", prompt2);
+  console.log("adding prompt2, ", config.addPrompt(prompt2, prompt2.name));
+
+  console.log("Deserialize Prompt2: ");
+  console.log(await config.resolve("prompt2"));
+  console.log("\nRun Prompt2: ");
+  console.log(await config.run("prompt2"));
+
+  //stream
+  const inferenceOptions = {
+    callbacks: {
+      streamCallback: (data: any, _accumulatedData: any, _index: any) => {
+        process.stdout.write(data);
+      },
+    },
+  };
+
+  console.log("starting to stream: \n");
+  await config.run("prompt1", undefined, inferenceOptions);
+
+  config.save("config.json");
+}
+
+run();
+
+const hfMP = new HuggingFaceTextGenerationModelParser(
+  "mistralai/Mistral-7B-v0.1",
+  true
+);
+AIConfigRuntime.registerModelParser(hfMP, ["mistralai/Mistral-7B-v0.1"]);
--- a/typescript/lib/parsers/hf.ts
+++ b/typescript/lib/parsers/hf.ts
@@ -1,0 +1,220 @@
+import {
+  HfInference,
+  TextGenerationArgs,
+  TextGenerationOutput,
+  TextGenerationStreamOutput,
+} from "@huggingface/inference";
+
+import {
+  Prompt,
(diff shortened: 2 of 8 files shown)


LastMileNewGrad:
testplan:

```typescript
{
  "name": "exploring nyc through chatgpt config",
  "description": "",
  "schema_version": "latest",
  "metadata": {
    "parameters": {},
    "models": {
      "mistralai/Mistral-7B-v0.1": {
        "model": "mistralai/Mistral-7B-v0.1",
        "top_p": 0.9,
        "temperature": 0.9,
        "stream": true
      }
    }
  },
  "prompts": [
    {
      "name": "prompt1",
      "input": "Hi! Tell me 10 cool things to do in NYC.",
      "metadata": {
        "model": {
          "name": "mistralai/Mistral-7B-v0.1"
        },
        "remember_chat_context": true
      }
    },
    {
      "name": "prompt2",
      "input": "Hello, world!",
      "metadata": {
        "model": {
          "name": "mistralai/Mistral-7B-v0.1",
          "settings": {}
        },
        "parameters": {}
      }
    }
  ]
}```

File: typescript/demo/test-hf.ts [hunk 1]
```diff
@@ -0,0 +1,74 @@
...
+      "../../python/tests/aiconfigs/basic_chatgpt_query_config.json"
+    )
+  );
+
+  // register HF MP
+  const hfMP = new HuggingFaceTextGenerationModelParser(
+    "mistralai/Mistral-7B-v0.1",
+    true
+  );
+  AIConfigRuntime.registerModelParser(hfMP, ["mistralai/Mistral-7B-v0.1"]);
```
saqadri, line 20: You shouldn't need to register `["mistralai/Mistral-7B-v0.1"])` again, right?

saqadri: In fact this is where using the `model_parsers` would be better

typescript/lib/parsers/hf.ts
```diff
@@ -0,0 +1,227 @@
...
+
+export class HuggingFaceTextGenerationModelParser extends ParameterizedModelParser<TextGenerationArgs> {
+  private hfClient: HfInference;
+
+  public constructor(modelId: string, use_api_token?: Boolean) {
+    super();
+    let token = use_api_token
+      ? getAPIKeyFromEnv("HUGGING_FACE_API_TOKEN")
+      : undefined;
+    this.hfClient = new HfInference(token);
```
saqadri, line 31: Should we instantiate this in `run`? So that the API Key isn't needed for serialize and deserialize?

typescript/lib/parsers/hf.ts
```diff
@@ -0,0 +1,227 @@
...
+  }
+}
+
+/**
+ * Handles and constructs the output for a stream response.
+ * @param response
+ * @param options
+ * @returns
+ */
+async function construct_stream_output(
```
saqadri, line 183: `constructStreamOutput`

typescript/lib/parsers/hf.ts
```diff
@@ -0,0 +1,227 @@
...
+  response: AsyncGenerator<TextGenerationStreamOutput>,
+  options: InferenceOptions
+): Promise<Output> {
+  let accumulated_message = "";
```
saqadri, line 187: `accumulatedMessage`

typescript/lib/parsers/hf.ts
```diff
@@ -0,0 +1,227 @@
...
+  let output = {} as ExecuteResult;
+
+  (async () => {
```
saqadri, line 190: why?

saqadri: You can just remove this (async() => {}) wrapper

typescript/lib/parsers/hf.ts
```diff
@@ -0,0 +1,227 @@
...
+        execution_count: index,
+        metadata: metadata,
+      } as ExecuteResult;
+    }
+  })();
+
+  return output;
+}
+
+function constructRegularOutput(response: TextGenerationOutput): Output {
```
saqadri, line 215: `constructOutput`

saqadri (approved):
Nice!

typescript/lib/parsers/hf.ts
LastMileNewGrad: updated

typescript/lib/parsers/hf.ts
LastMileNewGrad: Yes that would make sense. Can instantiate on first run by checking if client isn't defined, and if not define client.

typescript/demo/test-hf.ts
LastMileNewGrad: Yes with the new schema update this isn't necessary! Will update on a new diff with changes in the comment below this as well, as this would entail a couple of small changes:
- Move construction of the HF Client to run to address api key
- Does not require  modelName; id will always return `HuggingFaceTextGeneration` parser 
- test harness doesn't need to instantiate a modelParser

typescript/demo/test-hf.ts
```diff
@@ -0,0 +1,74 @@
...
+async function run() {
+  const config = AIConfigRuntime.load(
+    path.join(
+      __dirname,
+      "../../python/tests/aiconfigs/basic_chatgpt_query_config.json"
+    )
+  );
+
+  // register HF MP
+  const hfMP = new HuggingFaceTextGenerationModelParser(
```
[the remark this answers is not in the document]
LastMileNewGrad: Updated but per comment below this isn't needed anymore.

LastMileNewGrad:
rebase

PR #83, opened 2023-11-03
[Python] Update base imports

LastMileNewGrad:
[Python] Update base imports

Added  a couple more imports to base init.py to allow for cleaner imports at application layer

### DIFF
--- a/python/src/aiconfig/__init__.py
+++ b/python/src/aiconfig/__init__.py
@@ -1,14 +1,23 @@
+# Core Data Classes
 from .AIConfigSettings import (
     AIConfig,
     ConfigMetadata,
     ExecuteResult,
+    JSONObject,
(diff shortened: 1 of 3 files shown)


python/src/aiconfig/__init__.py
```diff
@@ -1,14 +1,23 @@
+# Core Data Classes
 from .AIConfigSettings import (
```
saqadri, line 2: Can you also rename this to Schema? Or AIConfigSchema?

saqadri (approved):

python/src/aiconfig/__init__.py
LastMileNewGrad: will update in a new diff ontop to schema.py as discussed offline

PR #89, opened 2023-11-06
[Python] Updates after 1.0.4 publish

saqadri:
[Python] Updates after 1.0.4 publish


python/src/python_aiconfig.egg-info/PKG-INFO
```diff
@@ -1,6 +1,6 @@
 Metadata-Version: 2.1
```
LastMileNewGrad, line 1: Haven't gotten a chance to understand the process of publishing, you aren't using the `.egg` to publish right?

LastMileNewGrad (approved):

PR #90, opened 2023-11-06
[Python][Easy] Refactor AIConfigsettings.py to schema.py

LastMileNewGrad:
[Python][Easy] Refactor AIConfigsettings.py to schema.py

as discussed offline cc: @saqadri,

rename the file `AIConfigSettings.py` to `schema.py`. This is in attempt to align with TS impl where the type definitions are under `types.ts`

Not an exact 1:1 but the type definitions in python will now live in `schema.py`


saqadri (approved):
Looks much cleaner! Maybe as a follow up we can update types.ts to schema.ts, but not super important at the moment.

PR #98, opened 2023-11-07
[Python] Refactor save() to use file from load()

LastMileNewGrad:
[Python] Refactor save() to use file from load()

## WHAT

The Config.save() method has been updated to automatically save the aiconfig back to the original source file unless an alternative file path is provided

- Added attribute `file_path` to AIConfigRuntime.
- on `config.load()` `file_path` attribute is set to the file
- Modified save() to use file path if available else default to `aiconfig.json`

## WHY
The flow from loading and saving a config should be as easy as possible. This way the user can easily update their config without having the extra step of specifying the same file path that was used to load.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/98).
* #99
* __->__ #98


saqadri (approved):
Approving to unblock, some comments in diff above in stack.

PR #99, opened 2023-11-07
[Typescript] Refactor save() to use file from load()

LastMileNewGrad:
[Typescript] Refactor save() to use file from load()

## WHAT

The Config.save() method has been updated to automatically save the aiconfig back to the original source file unless an alternative file path is provided

- Added attribute `file_path` to AIConfigRuntime.
- on `config.load()` `file_path` attribute is set to the file
- Modified save() to use file path if available else default to `aiconfig.json`

## WHY
The flow from loading and saving a config should be as easy as possible. This way the user can easily update their config without having the extra step of specifying the same file path that was used to load.

# Test Plan:
Local testing
- Load Config()
- Save config() with no params specified.
- Config saved under original file path

### DIFF
--- a/test.py
+++ b/test.py
@@ -1,5 +1,0 @@
-from aiconfig import AIConfigRuntime
-
-x = AIConfigRuntime.load("python/demo/GPT3-WhatAreTransformers-Stream-Enabled.json")
-
-x.save()
--- a/typescript/lib/config.ts
+++ b/typescript/lib/config.ts
@@ -69,6 +69,8 @@ export class AIConfigRuntime implements AIConfig {
   metadata: AIConfig["metadata"];
   prompts: PromptWithOutputs[];
 
+  filePath?: string;
+
   public constructor(
     name: string,
     description?: string,
@@ -92,7 +94,11 @@ export class AIConfigRuntime implements AIConfig {
   public static load(aiConfigFilePath: string) {
     const aiConfigString = fs.readFileSync(aiConfigFilePath, "utf8");
     const aiConfigObj = JSON.parse(aiConfigString);
-    return this.loadJSON(aiConfigObj);
+
+    const config = this.loadJSON(aiConfigObj);
+    config.filePath = aiConfigFilePath;
+
+    return config;
   }
 
   /**
@@ -201,9 +207,10 @@ export class AIConfigRuntime implements AIConfig {
    * @param filePath The path to the file to save to.
    * @param saveOptions Options that determine how to save the AIConfig to the file.
    */
-  public save(filePath: string, saveOptions?: SaveOptions) {
+  public save(filePath?: string, saveOptions?: SaveOptions) {
     try {
       let aiConfigObj: AIConfigRuntime = _.cloneDeep(this);
+
       if (!saveOptions?.serializeOutputs) {
         const prompts = [];
         for (const prompt of aiConfigObj.prompts) {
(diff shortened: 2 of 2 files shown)


test.py
```diff
@@ -0,0 +1,5 @@
+from aiconfig import AIConfigRuntime
```
saqadri, line 1: Was this file meant to be committed? Looks like a local test

typescript/lib/config.ts
```diff
@@ -69,6 +69,8 @@ export class AIConfigRuntime implements AIConfig {
   metadata: AIConfig["metadata"];
   prompts: PromptWithOutputs[];
 
+  file_path?: string;
```
saqadri, line 72:
```suggestion
  filePath?: string;
```

typescript/lib/config.ts
```diff
@@ -92,7 +94,11 @@ export class AIConfigRuntime implements AIConfig {
   public static load(aiConfigFilePath: string) {
     const aiConfigString = fs.readFileSync(aiConfigFilePath, "utf8");
     const aiConfigObj = JSON.parse(aiConfigString);
-    return this.loadJSON(aiConfigObj);
+
+    const config = this.loadJSON(aiConfigObj);
+    config.file_path = aiConfigFilePath;
```
saqadri, line 99:
```suggestion
    config.filePath = aiConfigFilePath;
```

typescript/lib/config.ts
```diff
@@ -201,19 +207,28 @@ export class AIConfigRuntime implements AIConfig {
...
+      // Remove the file_path property from the to-be-saved AIConfig
+      aiConfigObj.file_path = undefined;
 
       // TODO: saqadri - make sure that the object satisfies the AIConfig schema
       const aiConfigString = JSON.stringify(aiConfigObj, null, 2);
+  
+      if (!filePath) {
+        filePath = this.file_path ?? "aiconfig.json";
+      }
+
```
saqadri, line 230: Remove newline (there are 2)

typescript/lib/config.ts
```diff
@@ -201,19 +207,28 @@ export class AIConfigRuntime implements AIConfig {
...
         aiConfigObj.prompts = prompts;
       }
+      // Remove the file_path property from the to-be-saved AIConfig
+      aiConfigObj.file_path = undefined;
 
       // TODO: saqadri - make sure that the object satisfies the AIConfig schema
       const aiConfigString = JSON.stringify(aiConfigObj, null, 2);
+  
+      if (!filePath) {
+        filePath = this.file_path ?? "aiconfig.json";
```
saqadri, line 228:
```suggestion
        filePath = this.filePath ?? "aiconfig.json";
```

typescript/lib/config.ts
```diff
@@ -201,19 +207,28 @@ export class AIConfigRuntime implements AIConfig {
...
+
       if (!saveOptions?.serializeOutputs) {
         const prompts = [];
         for (const prompt of aiConfigObj.prompts) {
           prompts.push(_.omit(prompt, "outputs"));
         }
         aiConfigObj.prompts = prompts;
       }
+      // Remove the file_path property from the to-be-saved AIConfig
+      aiConfigObj.file_path = undefined;
```
saqadri, line 222:
```suggestion
      aiConfigObj.filePath = undefined;
```

saqadri: Just confirm this works

saqadri (approved):
Approving to unblock.

test.py
LastMileNewGrad: Not intended to be committed, thanks for catching. Removed.

typescript/lib/config.ts
LastMileNewGrad: Thanks!

typescript/lib/config.ts
LastMileNewGrad: Added testplan video

LastMileNewGrad:
Testplan:
1. Load and save existing config. file_path is not a field in the saved json as expected.
2. Create a config with code and save to file. Defaults to aiconfig.json.

https://github.com/lastmile-ai/aiconfig/assets/141073967/66b31568-12a2-4b9b-9b07-d4895dc63a5b

LastMileNewGrad:
Rebase

PR #111, opened 2023-11-08
[Typescript][Callbacks][1/n] CallbackManager impl

LastMileNewGrad:
[Typescript][Callbacks][1/n] CallbackManager impl

### DIFF
--- a/typescript/lib/callback.ts
+++ b/typescript/lib/callback.ts
@@ -1,0 +1,147 @@
+import { writeFileSync, appendFileSync } from "fs";
+import { EventEmitter } from "events";
+
+// TypeScript interfaces for type definitions
+interface Result<T> {
+  Ok?: T;
+  Err?: string;
+}
+
+// Record class equivalent in TypeScript
+class Record {
+  readonly model_config: object;
+
+  constructor(modelConfig: object) {
+    this.model_config = Object.freeze(modelConfig);
+  }
+}
+
+// CallbackEvent equivalent in TypeScript
+class CallbackEvent {
+  name: string;
+  data: any;
+  ts_ns: number;
+
+  constructor(name: string, data: any, ts_ns: number = Date.now()) {
+    this.name = name;
+    this.data = data;
+    this.ts_ns = ts_ns;
+  }
+}
+
+// CallbackResult equivalent in TypeScript
+class CallbackResult extends Record {
+  result: any;
+
+  constructor(result: any) {
+    super({ strict: true });
+    this.result = result;
+  }
+}
+
+// Callback type equivalent in TypeScript
+type Callback = (
+  event: CallbackEvent,
+  run_id: string
+) => Promise<CallbackResult>;
+
+async function runThunkSafe<T>(
+  thunk: () => Promise<T>,
+  timeout: number
+): Promise<Result<T>> {
+  return new Promise((resolve) => {
+    const timer = setTimeout(() => {
+      resolve({ Err: "Timeout" });
+    }, timeout);
+
+    thunk()
+      .then((response) => {
+        clearTimeout(timer);
+        resolve({ Ok: response });
+      })
+      .catch((error) => {
+        clearTimeout(timer);
+        resolve({ Err: error.message });
+      });
+  });
+}
+
+// CallbackManager equivalent in TypeScript
+class CallbackManager {
+  callbacks: ReadonlyArray<Callback>;
+  results: Result<any>[];
+  run_id: string | undefined;
+
+  constructor(callbacks: ReadonlyArray<Callback>) {
+    this.callbacks = callbacks;
+    this.results = [];
+    this.run_id = undefined;
+  }
+
+  async runCallbacks(event: CallbackEvent): Promise<void> {
+    for (const callback of this.callbacks) {
+      const result = await runThunkSafe(
+        () => callback(event, this.run_id!),
+        1000
+      );
+      this.results.push(result);
+    }
+  }
+
+  resetRunState(run_id?: string): void {
+    this.results = [];
+    this.run_id = run_id;
+  }
+
+  static default(): CallbackManager {
+    // Placeholder for the actual default callback, which would write to a JSON file
+    const defaultCallback: Callback = async (
+      event: CallbackEvent,
+      run_id: string
+    ) => {
+      const dataEvent = event; // Assuming event has a method to dump itself to a model
+      const dataWrite = { run_id, ...dataEvent };
+      appendFileSync(
+        "/var/logs/callbacks.json",
+        JSON.stringify(dataWrite, null, 2) + "\n"
+      );
+      return new CallbackResult(0); // Placeholder for actual result
+    };
+    return new CallbackManager([defaultCallback]);
+  }
+}
+
+// Serialization function
+function safeSerializeJson(obj: any): string {
+  return JSON.stringify(
+    obj,
+    (key, value) => {
+      if (typeof value === "object" && value !== null) {
+        return value;
+      }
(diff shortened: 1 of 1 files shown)


typescript/lib/callback.ts
```diff
@@ -0,0 +1,97 @@
...
+    this.data = data;
+    this.ts_ns = ts_ns;
+  }
+}
+
+// Type Aliases
+type Callback = (event: CallbackEvent) => Promise<any>;
+// type Result = Ok<any> | Err<any>;
+
+async function executeCoroutineWithTimeout(
```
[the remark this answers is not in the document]
rholinshead: Might be nice to pass callback and event into this instead, then can reject with info about the event for handling below

typescript/lib/callback.ts
```diff
@@ -0,0 +1,97 @@
...
+    this.callbacks = callbacks;
+    this.results = [];
+    this.timeout = timeout;
+  }
+
+  async runCallbacks(event: CallbackEvent): Promise<void> {
+    const tasks = this.callbacks.map((callback) =>
+      executeCoroutineWithTimeout(callback(event), this.timeout)
+    );
+    this.results = await Promise.all(tasks);
```
rholinshead, line 63: If we rejected in `withTimeout` and still want to swallow errors, we could do something like:
```
// Use .allSettled if we want to avoid propagating errors and just log something
const results = await Promise.allSettled(promises);
results.forEach(result => {
  if (result.status === "rejected") {
    console.error('Error running callback for event: ', result.event, result.error);
   }
});
```

But fwiw, I think we should keep as Promise.all and just update `withTimeout` to reject so that it throws the error instead of swallowing it

typescript/lib/callback.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  static createDefaultManager(): CallbackManager {
```
rholinshead, line 66: nit, can we call this something like `createManagerWithLogging(filePath: string = "aiconfig.log")` to make it clear that it will log? 'default' terminology makes it seem like all CM should be using it and logging to the file.

Also, supporting filePath as a param with default is ideal

typescript/lib/callback.ts
```diff
@@ -0,0 +1,97 @@
+import { createLogger, format, transports } from "winston";
+
+// Constants
+const DEFAULT_TIMEOUT: number = 5; // Default timeout for callback execution in seconds
```
rholinshead, line 4:
```suggestion
const DEFAULT_TIMEOUT_SECONDS: number = 5; // Default timeout for callback execution in seconds
```

typescript/lib/callback.ts
```diff
@@ -0,0 +1,97 @@
...
+
+export class CallbackEvent {
+  name: string;
+  file: string;
+  data: any;
```
rholinshead, line 9: Not a big fan of using `any` in ts in general (also, surprised the linter didn't complain), and in this case I think it does result in extra friction on people using callbacks since they need to dig through code to see what possible data values will be for different event types.

Not blocking on this, but it seems that there's a pretty limited set of events that are defined and could just do something like

```
export interface CallbackEvent<D=unknown> {
  name: string,
  data: D,
  ts_ns?: number,
}

// In ModelParser

export interface ModelParserEvent<D=unknown> extends CallbackEvent<D> {
  parser: string; // "openai", etc.
};

export type SerializeStartEventData<SD=JSONObject> = {
  promptName: string,
  data: SD,
  params: JSONObject,
};

export interface SerializeStartEvent<SD=JSONObject> extends ModelParserEvent<
  SerializeStartEventData<SD>
> {
  name: "on_serialize_start",
};

export interface SerializeEndEvent extends ModelParserEvent<Prompt[]> {
  name: "on_serialize_end"
};

export interface DeserializeStartEvent extends ModelParserEvent<
  {prompt: Prompt, params?: JSONObject}
> {name: "on_deserialize_start"};

export interface DeserializeEndEvent<R = JSONObject> extends ModelParserEvent<{result: R}> {
   name: "on_deserialize_end"
};

export interface RunStartEvent extends ModelParserEvent<{
  prompt: Prompt,
  options?: InferenceOptions,
  params?: JSONObject,
}> {
  name: "on_run_start",
};

export interface RunEndEvent extends ModelParserEvent<{
  result: Output | Output[],
}> {
  name: "on_run_end",
};
```

Then we could leverage stronger typing of data while still supporting arbitrary events,

rholinshead: Also just want to reiterate that this isn't a blocker here, just something that can be looked into later to improve the typing a bit

typescript/lib/callback.ts
```diff
@@ -0,0 +1,97 @@
...
+  ts_ns: number;
```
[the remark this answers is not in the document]
LastMileNewGrad: ts_ns means timestamp in nanoseconds. Updated with comments

typescript/lib/callback.ts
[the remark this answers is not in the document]
LastMileNewGrad: Updated to `withTimeout`. agree that it was a confusing with python terminology. Updated input param to `promise`, does `runPromise` make more sense here?

typescript/lib/callback.ts
```diff
@@ -0,0 +1,97 @@
...
+type Callback = (event: CallbackEvent) => Promise<any>;
+// type Result = Ok<any> | Err<any>;
+
+async function executeCoroutineWithTimeout(
+  coroutine: Promise<any>,
+  timeout: number
+): Promise<any> {
+  return new Promise((resolve) => {
+    setTimeout(() => {
+      resolve(new Error("Timeout"));
```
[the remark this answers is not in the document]
LastMileNewGrad: Updated to reject()

typescript/lib/callback.ts
```diff
@@ -0,0 +1,97 @@
...
+
+// Type Aliases
+type Callback = (event: CallbackEvent) => Promise<any>;
+// type Result = Ok<any> | Err<any>;
+
+async function executeCoroutineWithTimeout(
+  coroutine: Promise<any>,
+  timeout: number
+): Promise<any> {
+  return new Promise((resolve) => {
```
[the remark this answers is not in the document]
LastMileNewGrad: Added, though with promise.race(), resolve isn't needed

typescript/lib/callback.ts
```diff
@@ -0,0 +1,97 @@
...
+    setTimeout(() => {
+      resolve(new Error("Timeout"));
+    }, timeout * 1000);
+
+    coroutine
+      .then((result) => {
+        resolve(result);
+      })
+      .catch((error) => {
+        resolve(error);
```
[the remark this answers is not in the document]
LastMileNewGrad: updated, using reject()

typescript/lib/callback.ts
LastMileNewGrad: Updated withTimeout as advised, throws reject instead of swallowing it

typescript/lib/callback.ts
LastMileNewGrad: updated to `createManagerWithLogging`, added a param for file path

typescript/lib/callback.ts
LastMileNewGrad: updated

typescript/lib/callback.ts
LastMileNewGrad: adding to linear todo to revisit

typescript/lib/callback.ts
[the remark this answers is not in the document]
rholinshead: `promise` is fine

rholinshead (approved):

PR #112, opened 2023-11-08
[Typescript][Callbacks][2/n] Add CallbackManager to AIConfigRuntime

LastMileNewGrad:
[Typescript][Callbacks][2/n] Add CallbackManager to AIConfigRuntime

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/112).
* #198
* #116
* #113
* __->__ #112
* #111
* #115

### DIFF
--- a/typescript/lib/config.ts
+++ b/typescript/lib/config.ts
@@ -17,6 +17,7 @@ import { ParameterizedModelParser } from "./parameterizedModelParser";
 import { OpenAIChatModelParser, OpenAIModelParser } from "./parsers/openai";
 import { extractOverrideSettings } from "./utils";
 import { HuggingFaceTextGenerationParser } from "./parsers/hf";
+import { CallbackManager } from "./callback";
 
 export type PromptWithOutputs = Prompt & { outputs?: Output[] };
 
@@ -73,6 +74,7 @@ export class AIConfigRuntime implements AIConfig {
   prompts: PromptWithOutputs[];
 
   filePath?: string;
+  callbackManager: CallbackManager = CallbackManager.createManagerWithLogging();
 
   public constructor(
     name: string,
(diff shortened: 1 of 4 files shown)


typescript/lib/config.ts
```diff
@@ -70,6 +71,7 @@ export class AIConfigRuntime implements AIConfig {
   prompts: PromptWithOutputs[];
 
   filePath?: string;
+  callbackManager: CallbackManager = new CallbackManager([]);
```
rholinshead, line 74: It seems strange to make it required and construct it by default vs just having it optional and optionally passed into the constructor

typescript/lib/config.ts
```diff
@@ -409,6 +411,10 @@ export class AIConfigRuntime implements AIConfig {
     return result;
   }
 
+  public setCallbackManager(callbackManager: CallbackManager) {
```
rholinshead, line 414: In which scenarios would someone construct the config and then pass the CM afterwards, vs just supporting it passed through in the constructor?

Also, seems like this should make some sort of noise if there's already a CM that would be replaced by this call?

LastMileNewGrad: The idea here is that you can modify, delete the callback manager with one command. If it might should it throw an error or do something else?

typescript/lib/config.ts
LastMileNewGrad: updated to default logger

typescript/lib/config.ts
rholinshead: Maybe just check and do a `console.warn('Replaced existing CallbackManager for config')`?

rholinshead (approved):

PR #113, opened 2023-11-08
[Typescript][Callbacks][3/n] Add Callback events to AIConfigRuntime

LastMileNewGrad:
[Typescript][Callbacks][3/n] Add Callback events to AIConfigRuntime

### DIFF
--- a/typescript/lib/config.ts
+++ b/typescript/lib/config.ts
@@ -16,6 +16,7 @@ import { getAPIKeyFromEnv } from "./utils";
 import { ParameterizedModelParser } from "./parameterizedModelParser";
 import { OpenAIChatModelParser, OpenAIModelParser } from "./parsers/openai";
 import { extractOverrideSettings } from "./utils";
+import { CallbackManager } from "./callback";
 
 export type PromptWithOutputs = Prompt & { outputs?: Output[] };
 
@@ -70,6 +71,7 @@ export class AIConfigRuntime implements AIConfig {
   prompts: PromptWithOutputs[];
 
   filePath?: string;
+  callbackManager?: CallbackManager;
 
   public constructor(
     name: string,
@@ -218,12 +220,13 @@ export class AIConfigRuntime implements AIConfig {
         }
         aiConfigObj.prompts = prompts;
       }
-      // Remove the filePath property from the to-be-saved AIConfig
+      // Remove the filePath, callbackManager property from the to-be-saved AIConfig
       aiConfigObj.filePath = undefined;
+      aiConfigObj.callbackManager = undefined;
 
       // TODO: saqadri - make sure that the object satisfies the AIConfig schema
       const aiConfigString = JSON.stringify(aiConfigObj, null, 2);
-  
+
       if (!filePath) {
         filePath = this.filePath ?? "aiconfig.json";
       }
@@ -406,6 +409,10 @@ export class AIConfigRuntime implements AIConfig {
       params
     );
     return result;
+  }
+
+  public setCallbackManager(callbackManager: CallbackManager) {
+    this.callbackManager = callbackManager;
   }
 
   //#endregion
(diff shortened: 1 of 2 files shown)


typescript/lib/config.ts
```diff
@@ -71,7 +71,7 @@ export class AIConfigRuntime implements AIConfig {
   prompts: PromptWithOutputs[];
 
   filePath?: string;
-  callbackManager: CallbackManager = new CallbackManager([]);
+  callbackManager: CallbackManager = CallbackManager.createDefaultManager();
```
rholinshead, line 74: Hmm, I still think having this optional and default undefined is best. It seems a bit aggressive to default logging to a specific file -- there could be issues with file permissions, adding log files by default to git commits, etc. that would make this a bad default experience; we should just make it easy to add the logging instead (e.g. they can just call CallbackManager.createWithLogging()` and pass through the constructor if they want a CM that logs to file

typescript/lib/config.ts
```diff
@@ -288,6 +288,8 @@ export class AIConfigRuntime implements AIConfig {
    * that can be used to call the GPT-4 API. It will combine all the parameters, references and metadata specified in the AIConfig."
    */
   public async resolve(promptName: string, params: JSONObject = {}) {
+    const startEvent = new CallbackEvent("on_resolve_start", __filename, {promptName, params});
```
rholinshead, line 291: Something to look into as a potential followup: instead of requiring fileName passed through, we could probably do a bit of a hack in the `runCallbacks` implementation to grab the exact place it was called using a temp error stack -- https://stackoverflow.com/questions/28631260/how-to-get-filename-and-line-number-of-where-a-function-is-called-in-node

typescript/lib/config.ts
```diff
@@ -302,6 +304,8 @@ export class AIConfigRuntime implements AIConfig {
     }
 
     const resolvedPrompt = modelParser.deserialize(prompt, this, params);
+    const endEvent = new CallbackEvent("on_resolve_end", __filename, {'result': resolvedPrompt});
```
rholinshead, line 307: nit - don't need the `' '` around result. Also, could probably just do `{resolvedPrompt}` as the key/val

typescript/lib/config.ts
```diff
@@ -328,6 +334,8 @@ export class AIConfigRuntime implements AIConfig {
     }
 
     const prompts = modelParser.serialize(promptName, data, this, params);
+    const endEvent = new CallbackEvent("on_serialize_end", __filename, {'result': prompts});
```
rholinshead, line 337: Same here r.e. `'result'` => `{prompts}`

typescript/lib/config.ts
```diff
@@ -368,6 +378,8 @@ export class AIConfigRuntime implements AIConfig {
       prompt.outputs = [result];
     }
 
+    const endEvent = new CallbackEvent("on_run_end", __filename, {'result': prompt.outputs});
```
rholinshead, line 381: `'result'` => `outputs: prompt.outputs`

rholinshead (approved):
Just small things / nits

typescript/lib/config.ts
rholinshead: Could also do `className: this.name or this.constructor.name` instead of __fileName here. Looking over next PR, it seems mainly beneficial for distinguishing between parser classes

typescript/lib/config.ts
LastMileNewGrad: I can see why default logging could seem a bit aggressive. @saqadri wdyt?

LastMileNewGrad: adding a todo to linear task to revisit specifics on whether or not to include defaults

typescript/lib/config.ts
LastMileNewGrad: that's pretty cool, will add to linear todo as well

typescript/lib/config.ts
LastMileNewGrad: previous

typescript/lib/config.ts
LastMileNewGrad: see previous

PR #115, opened 2023-11-09
[Typescript] refactor save() for omitted fields

LastMileNewGrad:
[Typescript] refactor save() for omitted fields

## What
Improved the modularity of the field exclusion logic when saving config to disk.

## Why
The current saving mechanism is too rigid

### DIFF
--- a/typescript/lib/config.ts
+++ b/typescript/lib/config.ts
@@ -208,8 +208,11 @@ export class AIConfigRuntime implements AIConfig {
    * @param saveOptions Options that determine how to save the AIConfig to the file.
    */
   public save(filePath?: string, saveOptions?: SaveOptions) {
+    const keysToOmit = ['filePath'] as const;
+  
     try {
-      let aiConfigObj: AIConfigRuntime = _.cloneDeep(this);
+      // Create a Deep Copy and omit fields that should not be saved
+      let aiConfigObj: Omit<AIConfigRuntime, typeof keysToOmit[number]> = _.omit(_.cloneDeep(this), keysToOmit)
 
       if (!saveOptions?.serializeOutputs) {
         const prompts = [];
@@ -218,8 +221,6 @@ export class AIConfigRuntime implements AIConfig {
         }
         aiConfigObj.prompts = prompts;
       }
-      // Remove the filePath property from the to-be-saved AIConfig
-      aiConfigObj.filePath = undefined;
 
       // TODO: saqadri - make sure that the object satisfies the AIConfig schema
       const aiConfigString = JSON.stringify(aiConfigObj, null, 2);


LastMileNewGrad:
Testplan:
- Loaded a config.
- Verified Config has attribute `filePath`
- Save() config. Showcase saved config does not have attribute `filePath`


https://github.com/lastmile-ai/aiconfig/assets/141073967/d00dca5c-ad9f-4a45-b452-f82b4f0d2917

typescript/lib/config.ts
```diff
@@ -208,8 +208,11 @@ export class AIConfigRuntime implements AIConfig {
    * @param saveOptions Options that determine how to save the AIConfig to the file.
    */
   public save(filePath?: string, saveOptions?: SaveOptions) {
+    const keysToOmit = ['filePath'] as const;
```
rholinshead, line 211: Hmm, do we want the serialized AIConfig to have dynamic top-level fields that we want to persist or will it always match the keys defined in the interface (excluding the open `[k: string]: any;`)?

If we want it to match the interface instead of the runtime interface,  rather than omitting specific fields here, it may be easier to just copy the exact fields we want persisted from the serialized schema. That way the runtime can have any number of added fields without having to worry about coming here and omitting them.

Something like:
```
const aiConfigObj: AIConfig = {
    name: this.name;
    description: this.description,
    schema_version: this.schema_version;
    metadata: {
      ...this.metadata, // could technically cloneDeep this but doesn't look like this function is modifying anything anyway so doesn't seem necessary
    },
    prompts: [
      ...this.prompts.map(prompt => {
        if (saveOptions.serializeOutputs === true) {
          return prompt;
        } else {
        return {
          ...prompt,
          outputs: [],
         }
       }
     })
    ],
}
```

Otherwise, if we want to support persisting dynamic/generic values on the config other than filePath and callbackManager then this should be ok to start. Just a couple suggestions for improvements in that case:

1. We might also want to support an `omitKeys` array in SaveOptions for users to exclude keys they've defined on the config as well. Probably not necessary for v0, but might be worth adding an Issue for.
2. We could have an `_internal` key on the config type which contains `filePath`, `callbackManager`, etc. for the internal implementation that we know we don't want to persist. Then this can be done simply by omitting `_internal` without tracking any keys to omit here

rholinshead (approved):
Accepting to unblock, just some suggestions

typescript/lib/config.ts
```diff
@@ -208,8 +208,11 @@ export class AIConfigRuntime implements AIConfig {
...
+  
     try {
-      let aiConfigObj: AIConfigRuntime = _.cloneDeep(this);
+      // Create a Deep Copy and omit fields that should not be saved
+      let aiConfigObj: Omit<AIConfigRuntime, typeof keysToOmit[number]> = _.omit(_.cloneDeep(this), keysToOmit)
```
rholinshead, line 215: Can use `const` here since it's never reassigned

rholinshead:
Also side note, we should have a test for `save` that mocks fs and ensures `writeFileSync` is called with the correct contents (especially wrt removing outputs); that will also make it easier to ensure the default behaviour correctly removes any new fields that are added to the class that shouldn't be serialized

typescript/lib/config.ts
LastMileNewGrad: updated

typescript/lib/config.ts
LastMileNewGrad: Adding to Linear Task

PR #116, opened 2023-11-09
[Typescript][Callbacks][4/n] Add Callback events to OpenAI Model Parser

LastMileNewGrad:
[Typescript][Callbacks][4/n] Add Callback events to OpenAI Model Parser

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/116).
* #198
* __->__ #116
* #113
* #112
* #111
* #115

### DIFF
--- a/typescript/lib/parsers/openai.ts
+++ b/typescript/lib/parsers/openai.ts
@@ -19,6 +19,7 @@ import {
 } from "openai/resources";
 import _ from "lodash";
 import { InferenceOptions } from "../modelParser";
+import { CallbackEvent } from "../callback";
 
 export class OpenAIModelParser extends ParameterizedModelParser<CompletionCreateParams> {
   private openai: OpenAI | null = null;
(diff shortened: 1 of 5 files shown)


typescript/lib/parsers/openai.ts
```diff
@@ -381,6 +388,10 @@ export class OpenAIChatModelParser extends ParameterizedModelParser<Chat.ChatCom
     // Rename the last prompt to the requested prompt name
     prompts[prompts.length - 1].name = promptName;
 
+    const endEvent = new CallbackEvent("on_serialize_end", __filename, {
+      result: prompts,
```
rholinshead, line 392: Same here r.e. `result` just being `prompts` key instead -- it doesn't seem like we are using a common `result` key for anything so using the more specific key makes more sense

rholinshead (approved):
LGTM, just a comment on using the more specific property names instead of generic `result`

typescript/lib/parsers/openai.ts
LastMileNewGrad: Revisiting the 'result' and the output data as a todo, but will probably get rid of the `result` as suggested

LastMileNewGrad:
https://linear.app/lastmile-ai/issue/LAS-454/callback-handler-improvements

PR #118, opened 2023-11-09
[python][callbacks] default callback

LastMileNewGrad:
[python][callbacks] default callback

### DIFF
--- a/python/src/aiconfig/Config.py
+++ b/python/src/aiconfig/Config.py
@@ -63,7 +63,7 @@ class AIConfigRuntime(AIConfig):
 
         This method creates a new AI configuration with the provided parameters and sets it as the current AI configuration.
         """
-        return cls(
+        new_aiconfig_object =  cls(
             **{
                 "name": name,
                 "description": description,
@@ -72,6 +72,10 @@ class AIConfigRuntime(AIConfig):
                 "prompts": prompts,
             }
         )
+
+        default_callback_manager = CallbackManager([])
+
+        return new_aiconfig_object
 
     @classmethod
     def load(cls, json_config_filepath) -> "AIConfigRuntime":
--- a/python/src/aiconfig/Config.py
+++ b/python/src/aiconfig/Config.py
@@ -1,7 +1,7 @@
 import json
 import sys
 import os
-from aiconfig.callback import CallbackManager
+from aiconfig.callback import CallbackEvent, CallbackManager
 import requests
 from typing import ClassVar, Dict, List, Optional
 
@@ -137,6 +137,9 @@ class AIConfigRuntime(AIConfig):
         returns:
             Prompt | List[Prompt]: A prompt or list of prompts representing the input data
         """
+        event = CallbackEvent("on_serialize_start", {"input": locals()})
+        await self.callback_manager.run_callbacks(event)
+
         model_parser = ModelParserRegistry.get_model_parser(model_name)
         if not model_parser:
             raise ValueError(
@@ -144,6 +147,9 @@ class AIConfigRuntime(AIConfig):
             )
 
         prompts = await model_parser.serialize(prompt_name, data, self, params)
+
+        event = CallbackEvent("on_serialize_complete", {"prompts": prompts})
+        await self.callback_manager.run_callbacks(event)
         return prompts
 
     async def resolve(
@@ -162,6 +168,8 @@ class AIConfigRuntime(AIConfig):
         Returns:
             str: The resolved prompt.
         """
+        event = CallbackEvent("on_resolve_start", {"input": locals()})
+        await self.callback_manager.run_callbacks(event)
         if prompt_name not in self.prompt_index:
             raise IndexError(
                 "Prompt not found in config, available prompts are:\n {}".format(
@@ -174,7 +182,9 @@ class AIConfigRuntime(AIConfig):
         model_provider = AIConfigRuntime.get_model_parser(model_name)
 
         response = await model_provider.deserialize(prompt_data, self, params)
-
+        
+        event = CallbackEvent("on_resolve_complete", {"output": response})
+        await self.callback_manager.run_callbacks(event)
         return response
 
     async def run(
@@ -194,6 +204,8 @@ class AIConfigRuntime(AIConfig):
         Returns:
             object: The response object returned by the AI-model's API.
         """
+        event = CallbackEvent("on_run_start", {"input": locals()})
+        await self.callback_manager.run_callbacks(event)
         if prompt_name not in self.prompt_index:
             raise IndexError(
                 "Prompt not found in config, available prompts are:\n {}".format(
@@ -205,7 +217,10 @@ class AIConfigRuntime(AIConfig):
         model_name = self.get_model_name(prompt_data)
         model_provider = AIConfigRuntime.get_model_parser(model_name)
 
-        response = await model_provider.run(prompt_data, self, options, params, **kwargs)
+        response = await model_provider.run(prompt_data, self, options, params, callback_manager = self.callback_manager, **kwargs)
+
+        event = CallbackEvent("on_run_complete", {"prompt_name": prompt_name, "params": params, "options": options, "kwargs": kwargs, "response": response})
+        await self.callback_manager.run_callbacks(event)
         return response
 
     #
--- a/python/src/aiconfig/Config.py
+++ b/python/src/aiconfig/Config.py
@@ -1,6 +1,7 @@
 import json
 import sys
 import os
+from aiconfig.callback import CallbackManager
 import requests
 from typing import ClassVar, Dict, List, Optional
 
@@ -40,6 +41,8 @@ class AIConfigRuntime(AIConfig):
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
         self.file_path = None
+        # Blank CallbackManager does nothing
+        self.callback_manager = CallbackManager([])
 
     @classmethod
     def create(
@@ -219,9 +222,11 @@ class AIConfigRuntime(AIConfig):
             json_config_filepath (str, optional): The file path to the JSON configuration file.
                 Defaults to "aiconfig.json".
         """
+        # AIConfig json should only contain the core data fields. These are auxiliary fields that should not be persisted
         exclude_options = {
             "prompt_index": True, 
-            "file_path": True
+            "file_path": True,
+            "callback_manager": True,
         }
 
         if not include_outputs:
@@ -287,3 +292,6 @@ class AIConfigRuntime(AIConfig):
                 )
             )
         return ModelParserRegistry.get_model_parser(model_id)
+
+    def set_callback_manager(self, callback_manager: CallbackManager):
+        self.callback_manager = callback_manager
--- a/python/src/aiconfig/Config.py
+++ b/python/src/aiconfig/Config.py
@@ -140,7 +140,7 @@ class AIConfigRuntime(AIConfig):
                 f"Unable to serialize data: `{data}`\n Model Parser for model {model_name} does not exist."
             )
 
-        prompts = model_parser.serialize(prompt_name, data, self, params)
+        prompts = await model_parser.serialize(prompt_name, data, self, params)
         return prompts
 
     async def resolve(
--- a/python/src/aiconfig/callback.py
+++ b/python/src/aiconfig/callback.py
@@ -1,0 +1,121 @@
+import asyncio
+import json
+import time
+from result import Err, Ok
+from pydantic import BaseModel, ConfigDict
+from typing import (
+    Coroutine,
+    Final,
+    List,
+    Sequence,
+    TextIO,
+    TypeAlias,
+    Union,
+    Any,
+    Awaitable,
+    Callable,
+    Optional,
+)
+
+
+class CallbackEvent:
+    """
+    Data class that represents an event that triggers a callback.
+    Backed by a pydantic model for easy serialization.
+    """
+
+    def __init__(self, name: str, data: Any, ts_ns: int = time.time_ns()):
+        self.name = name
+        # Anything available at the time the event happens.
+        # It is passed to the callback.
+        self.data = data
+        self.ts_ns = ts_ns
+
+
+# Type declaration for Asyn function
+Callback = Callable[[CallbackEvent], Awaitable]
+Result: TypeAlias = Union[Ok[Any], Err[Any]]
+
+
+async def execute_coroutine_with_timeout(
+    coroutine: Coroutine[Any, Any, Any], timeout: int
+) -> Result:
+    try:
+        task = asyncio.create_task(coroutine)
+        res = await asyncio.wait_for(task, timeout=timeout)
+        return Ok(res)
+    except Exception as e:
+        return Err(str(e))
+
+
+class Record(BaseModel):
+    model_config = ConfigDict(strict=True, frozen=True)
+
+
+class pydanticBackedCallBackEvent(Record):
+    """
+    Data class that represents an event that triggers a callback. A copy of CallbackEvent
+    but backed by a pydantic model for easy serialization.
+    """
+    name: str
+    data: Any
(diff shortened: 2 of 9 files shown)


python/src/aiconfig/callback.py
```diff
@@ -0,0 +1,139 @@
...
+        self.data = data
+        self.ts_ns = ts_ns
+
+
+# Type Aliases
+Callback = Callable[[CallbackEvent], Awaitable]
+Result: TypeAlias = Union[Ok[Any], Err[Any]]
+
+
+async def execute_coroutine_with_timeout(
```
jonathanlastmileai, line 45: nit: let's keep names consistent? at some point we'll either have a shared lib or unify the repos, and that will make the dedup easier

cc @saqadri

python/src/aiconfig/callback.py
```diff
@@ -0,0 +1,139 @@
...
+    coroutine: Coroutine[Any, Any, Any], timeout: int
```
jonathanlastmileai, line 46: try to type this parametrically if you can

python/src/aiconfig/callback.py
```diff
@@ -0,0 +1,139 @@
...
+) -> Result:
```
jonathanlastmileai, line 47: this also needs to be parametrized. Pyright catches this for me: do you have strict config?

python/src/aiconfig/callback.py
```diff
@@ -0,0 +1,139 @@
...
+    
+    Returns:
+        An Ok result if the coroutine executes successfully or an Err on Exception
+    """
+    try:
+        task = asyncio.create_task(coroutine)
+        res = await asyncio.wait_for(task, timeout=timeout)
+        return Ok(res)
+    except Exception as e:
+        return Err(str(e))
```
jonathanlastmileai, line 63: This feels like we're officially committing to Result. Everybody good? @saqadri , @rholinshead , @suyoglastmileai

python/src/aiconfig/callback.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+class Record(BaseModel):
+    model_config = ConfigDict(strict=True, frozen=True)
```
jonathanlastmileai, line 67: Maybe let's try to have some kind of annotation for stuff that's duplicated, or at least preserve the exact directory structure so its easier to find? see above

python/src/aiconfig/callback.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+class pydanticBackedCallBackEvent(Record):
```
jonathanlastmileai, line 70: nit: why rename?

python/src/aiconfig/callback.py
```diff
@@ -0,0 +1,139 @@
...
+        event = pydanticBackedCallBackEvent(**event.__dict__)
+        tasks = []
+        for callback in self.callbacks:
+            task = execute_coroutine_with_timeout(callback(event), self.timeout)
+            tasks.append(task)
+
+        self.results = await asyncio.gather(*tasks)
+
+    def clear_results(self) -> None:
+        self.results = []
```
jonathanlastmileai, line 99: Nit don't think we need this. can reinstantiate

python/src/aiconfig/callback.py
```diff
@@ -0,0 +1,139 @@
...
+    name: str
+    data: Any
+    ts_ns: int = time.time_ns()
+
+
+class CallbackManager:
+    """
+    Manages a sequence of callbacks to be executed in response to Events
+    """
+    def __init__(self, callbacks: Sequence[Callback], timeout: int = DEFAULT_TIMEOUT) -> None:
```
jonathanlastmileai, line 84: One thing to consider about default arguments: if the user accidentally passes down None, it will overwrite this. None. is often used as a sentinel for "use whatever the default is" so that would result in unexpected behavior.

I usually set to the real value inside the function for this reason.

python/src/aiconfig/callback.py
```diff
@@ -0,0 +1,139 @@
...
+        self.callbacks: Final[Sequence[Callback]] = callbacks
+        self.results: List[Result] = []
+        self.timeout = timeout
+
+    async def run_callbacks(self, event: CallbackEvent) -> None:
+        event = pydanticBackedCallBackEvent(**event.__dict__)
```
jonathanlastmileai, line 90: Arent you going to have a lot of keys in here that will break the constructor?

python/src/aiconfig/callback.py
```diff
@@ -0,0 +1,139 @@
...
+            tasks.append(task)
+
+        self.results = await asyncio.gather(*tasks)
+
+    def clear_results(self) -> None:
+        self.results = []
+
+    @classmethod
+    def create_default_manager(cls) -> "CallbackManager":
+        callback = create_json_logging_callback("callbacks.json")
```
jonathanlastmileai, line 103: nit: keep names consistent? see above

python/src/aiconfig/callback.py
```diff
@@ -0,0 +1,139 @@
...
+        return CallbackManager([callback])
+
+
+def serialize_object_to_json(obj: Any, **kwargs):  # type: ignore [fixme]
```
jonathanlastmileai, line 107: same

python/src/aiconfig/callback.py
```diff
@@ -0,0 +1,139 @@
...
+    def handle_non_serializable_objects(o: Any):
+        return f"<<non-serializable: {type(o).__qualname__}>>"
+
+    return json.dumps(obj, default=handle_non_serializable_objects, **kwargs)
+
+
+def create_json_logging_callback(file: Union[str, TextIO]) -> Callback:
+    """
+    Higher-order function that returns a callback that writes events to a file.
```
jonathanlastmileai, line 116: nit: you probably dont need to mention it's a HOP. It's clear what it returns

python/src/aiconfig/callback.py
```diff
@@ -0,0 +1,139 @@
...
+    Args:
+        file: Either a file path or a file object to write to.
+
+    Returns:
+        A callback that writes events to the specified file.
+    """
+
+    def write_event_to_file(event: CallbackEvent, file: TextIO) -> int:
+        data_event = event.model_dump()
+        data_write = dict(**data_event)
```
jonathanlastmileai, line 127: is this necessary?

python/src/aiconfig/callback.py
```diff
@@ -0,0 +1,139 @@
...
+        return file.write("\n" + serialize_object_to_json(data_write, indent=2))
+
+    async def _callback(event: CallbackEvent) -> Any:
+        if isinstance(file, str):
+            with open(file, "a+") as f:
```
jonathanlastmileai, line 132: Remember to check for I/O errors here. This is something I think I haven't done yet, but dont want it to slip through the cracks here too.


We can have indirection like this (see SR core.py)

```
def text_file_write(path: str, contents: str):
    try:
        with open(path, "w") as f:
            return f.write(contents)
    except Exception as e:
        # TODO [P1] deal with this
        logger.critical("exn=" + str(e))
        return ""
        
        
```

jonathanlastmileai (commented):
LG

python/src/aiconfig/Config.py
```diff
@@ -40,6 +41,8 @@ class AIConfigRuntime(AIConfig):
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
         self.file_path = None
+        # Blank CallbackManager does nothing
+        self.callback_manager = CallbackManager([])
```
jonathanlastmileai, line 45: why dont we make it do something 🗡️

python/src/aiconfig/Config.py
```diff
@@ -287,3 +292,6 @@ def get_model_parser(model_id: str) -> ModelParser:
                 )
             )
         return ModelParserRegistry.get_model_parser(model_id)
+
+    def set_callback_manager(self, callback_manager: CallbackManager):
+        self.callback_manager = callback_manager
```
jonathanlastmileai, line 316: I dont think we need getters and setters - it's extra boilerplate. Thoughts?

python/src/aiconfig/Config.py
```diff
@@ -194,6 +204,8 @@ async def run(
         Returns:
             object: The response object returned by the AI-model's API.
         """
+        event = CallbackEvent("on_run_start", {"input": locals()})
```
jonathanlastmileai, line 211: > locals()

Sledgehammer approach... I dont know, feels like a code smell :D

jonathanlastmileai (commented):
LG

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -50,6 +56,8 @@ async def serialize(
         Returns:
             str: Serialized representation of the prompt and inference settings.
         """
+        event = CallbackEvent("on_serialize_start", {"input": locals()})
```
jonathanlastmileai, line 59: locals?

jonathanlastmileai (commented):
LG

python/src/aiconfig/Config.py
```diff
@@ -73,6 +73,10 @@ def create(
             }
         )
 
+        default_callback_manager = CallbackManager([])
```
jonathanlastmileai, line 76: Should default write to logs?

jonathanlastmileai (commented):
LG

python/src/aiconfig/callback.py
```diff
@@ -0,0 +1,139 @@
...
+        self.name = name
+        # Anything available at the time the event happens.
+        # It is passed to the callback.
+        self.data = data
+        self.ts_ns = ts_ns
+
+
+# Type Aliases
+Callback = Callable[[CallbackEvent], Awaitable]
+Result: TypeAlias = Union[Ok[Any], Err[Any]]
```
jonathanlastmileai, line 42: This is defined already in the Result library. Also, let's try to parametrize with generics where possible and avoid any

python/src/aiconfig/default_parsers/openai.py
LastMileNewGrad: `locals()`  gets the function params. Updating this to the relevant input params, discussed offline w/ @saqadri

python/src/aiconfig/callback.py
LastMileNewGrad: could you share your Pyright setup? I've been rolling with the stand pylint checker and haven't had it complain

python/src/aiconfig/callback.py
LastMileNewGrad: Will be updating this to be a default logger and as per your other comment about the default

python/src/aiconfig/callback.py
LastMileNewGrad: The reason behind this is that constructing classes with Pydantic Base Models is a slight inconvenience. There isn't really a need for pydantic here besides easy JSON serialization. That way the user can just create the regular `CallbackEvent` class and it gets converted into the Pydantic `BaseModel` version here for the benefits.  Will update the docstring/comment

Could this end up being a bad thing?

python/src/aiconfig/callback.py
LastMileNewGrad: I'm not sure I follow on how it could break, all keys will be under `data` in 

```
CallbackEvent:
     data: Any
 ```
 
The reason behind the second class def is that constructing classes with Pydantic Base Models is a slight inconvenience. There isn't really a need for pydantic here besides easy JSON serialization. That way the user can just create the regular CallbackEvent class and it gets converted into the Pydantic BaseModel version here for the benefits. Will update the docstring/comment

Could this end up being a bad thing?

File: python/src/aiconfig/callback.py (hunk 20 above)
Comment (LastMileNewGrad) [reply python/src/aiconfig/callback.py:41]: What's the reason to avoid Any?

File: python/src/aiconfig/callback.py (hunk 2 above)
Comment (LastMileNewGrad) [reply python/src/aiconfig/callback.py:45]: By parametrically do you mean with generics?

Comment (jonathanlastmileai) [reply python/src/aiconfig/callback.py:45]: yes

File: python/src/aiconfig/callback.py (hunk 3 above)
Comment (jonathanlastmileai) [reply python/src/aiconfig/callback.py:46]: https://github.com/lastmile-ai/semantic-retrieval/blob/main/pyrightconfig.json

Strict with a couple of specific checks I disabled, I think because they were impossible to reconcile with numpy.

It's probably better to start as strict as we want to get, early, because going the opposite direction will be harder.

File: python/src/aiconfig/callback.py (hunk 14 above)
Comment (jonathanlastmileai) [reply python/src/aiconfig/callback.py]: this one is about handling I/O errors

File: python/src/aiconfig/default_parsers/openai.py (hunk 18 above)
Comment (jonathanlastmileai) [reply python/src/aiconfig/default_parsers/openai.py]: Curious for @saqadri 's thinking. My initial reaction was "there's a ton of stuff in there we may not want or need to log."

File: python/src/aiconfig/callback.py (hunk 20 above)
Comment (jonathanlastmileai) [reply python/src/aiconfig/callback.py:41]:
```
# static type error (good, we want the error)
def f(x: str) -> int:
    return x

def g(y: int): -> int
  return y + 1

g(f("3"))

# no static type error (bad, we dont wan't to have a runtime error)
def f(x: str) -> Any:
    return 1

# TypeError: cannot add "3" to 1.
g(f("3"))
```

Similar idea with generic type arguments.

File: python/src/aiconfig/callback.py (hunk 3 above)
Comment (jonathanlastmileai) [reply python/src/aiconfig/callback.py:46]: @LastMileNewGrad , this is not blocking

File: python/src/aiconfig/callback.py (hunk 11 above)
Comment (jonathanlastmileai) [reply python/src/aiconfig/callback.py]: Consistency with SR names is not blocking

File: python/src/aiconfig/callback.py (hunk 12 above)
Comment (jonathanlastmileai) [reply python/src/aiconfig/callback.py]: not blocking

File: python/src/aiconfig/callback.py (hunk 13 above)
Comment (jonathanlastmileai) [reply python/src/aiconfig/callback.py]: not blocking

File: python/src/aiconfig/callback.py (hunk 14 above)
Comment (jonathanlastmileai) [reply python/src/aiconfig/callback.py]: This is a quick fix: can use code like my example

File: python/src/aiconfig/callback.py (hunk 4 above)
Comment (jonathanlastmileai) [reply python/src/aiconfig/callback.py:62]: Not blocking, this does not have to be an irreversible commitment.

File: python/src/aiconfig/callback.py (hunk 6 above)
Comment (jonathanlastmileai) [reply python/src/aiconfig/callback.py:69]: remember to not leak implementation detail in names. Also classes are always uppercase

File: python/src/aiconfig/callback.py (hunk 9 above)
Comment (jonathanlastmileai) [reply python/src/aiconfig/callback.py:90]: I was able to just have CallbackEvent be a Record (which is a type of BaseModel) without issues. No need to convert here. Just instantiate at callsite, e.g.

```
...run_callbacks(
            CallbackEvent(
                name="openai_embeddings_fragments_embedded",
                data=dict(
                    n_batches=len(batches),
                    n_embeddings=len(out),
                ),
            )
```


are you having serialization issues?

File: python/src/aiconfig/Config.py [hunk 21]
```diff
@@ -137,15 +137,22 @@ async def serialize(self, model_name: str, data: Dict,  prompt_name: str, params
         returns:
             Prompt | List[Prompt]: A prompt or list of prompts representing the input data
         """
+        event = CallbackEvent("on_serialize_start", __name__, {model_name: model_name, "data": data, "prompt_name": prompt_name, "params": params})
```
jonathanlastmileai, line 140: Should think of a more robust way to get info in there aside from __name__ or __file__, e.g. using traceback lib. Not blocking

jonathanlastmileai (approved):
As long as it's tested, LG

jonathanlastmileai (approved):

jonathanlastmileai (approved):

jonathanlastmileai (approved):

jonathanlastmileai (approved):

PR #121, opened 2023-11-09
[Python] remove old code from  run_with_dependencies

LastMileNewGrad:
[Python] remove old code from  run_with_dependencies

- align run_with_dependencies impl with new changes
- updated stream calculation logic in openai model parser
- - updated test


saqadri (approved):
Approving to unblock, but the stream setting is very complicated -- may be better to switch it to if/else

LastMileNewGrad:
- Tested the Getting started guide. Docs will be updated on top of this diff, works as intended.

- Investigating a problem with running prompts with dependencies and streaming  being out of order after multiple executions, need to take a closer look.

PR #147, opened 2023-11-10
[Typescript] Changes to make the getting started guide start working

LastMileNewGrad:
[Typescript] Changes to make the getting started guide start working

The schema update makes every field on a prompt besides Name and Input optional. Changes in this diff reflect this update.


saqadri (approved):
Thanks for the fixes!

PR #190, opened 2023-11-13
[Python] update AIConfig & OpenAI model parser compatibility with sdk api version 1.0.0

LastMileNewGrad:
[Python] update AIConfig & OpenAI model parser compatibility with sdk api version 1.0.0

title

1. Modified ChatCompletion api to match the new API structure
- `openai.ChatCompletions.create()` -> `openai.chat.completions.create()`
2. Refactored tests
- New method requires env API Key to be set even when attempting to mock
- - set env before mocking method. Diff looks huge due to spacing changes but there's no other changes made there
- Modified mock ChatCompletion request to mimic new typing (see next bullet)
3. OpenAI API response is now backed with Pydantic Typing.
- AIConfig uses dicts for non core types (core types defined in schema.py). Now, On response from Completion API, Updated to convert response to dict and exclude any values that are None typed
- - ex: `response = response.model_dump(exclude_none=True)`
4. update requirements.txt & pyproject.toml to 1.0.0
5. Updated introspection api wrapper.
- Note that with the new change (see below), user will have to modify the client object's chat completion method. Still only ~1 line change.
- Updated demo notebook to reflect changes in diff above

Note: openai reccomends instatiating the client rather than using the base openai one. Wasn't able to figure out how to mock calls that way for test, sticking with the base for now

Reccomended:
```python
client = Client()

client.chat.completions.create()
```

Base:

```python
openai.chat.completions.create()
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/190).
* #194
* __->__ #190

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -23,7 +23,7 @@ dependencies = [
   "pydantic>=2.1",
   "pybars3",
   "google-generativeai",
-  "openai<1.0.0",
+  "openai>=1.0.0",
   "python-dotenv",
   "huggingface_hub",
   ]
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -4,7 +4,7 @@ pytest
 pydantic>=2.1
 pybars3
 google-generativeai
-openai<1.0.0
+openai>=1.0.0
 python-dotenv
 huggingface_hub
 result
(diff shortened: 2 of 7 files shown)


saqadri (approved):
Looks good, nice work. Ship it!

PR #198, opened 2023-11-14
[Typescript] test for save()

Ankush-lastmile:
[Typescript] test for save()

### DIFF
--- a/typescript/__tests__/testSave.ts
+++ b/typescript/__tests__/testSave.ts
@@ -1,0 +1,29 @@
+// FileWriter.test.ts
+import * as fs from "fs";
+import * as path from "path";
+import { tmpdir } from "os";
+import { AIConfigRuntime } from "../lib/config";
+
+describe("AIConfigRuntime save()", () => {
+  it("saves the config and checks if the config json doesn't have key filePath", () => {
+    const tmpDir = tmpdir(); // Get the system's temporary directory
+    const newFileName = "modified_config.json";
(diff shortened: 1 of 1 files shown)


typescript/__tests__/testSave.ts
```diff
@@ -0,0 +1,29 @@
...
+      "samples",
+      "basic_chatgpt_query_config.json"
+    );
+
+    const aiconfig = AIConfigRuntime.load(filePath);
+    // Modify and save the configuration
+    aiconfig.save(tempFilePath);
+
+    // Open and parse the saved JSON file
+    const savedConfig = JSON.parse(fs.readFileSync(tempFilePath, "utf8"));
```
rholinshead, line 24: Instead of actually writing to a temp file, I think you could just
```
const mockWriteFileSync = jest.fn();

jest.mock("fs", () => ({
  ...jest.requireActual("fs"),
  writeFileSync: mockWriteFileSync,
}));
```

Then in the test, do
```
aiconfig.save('testPath');
expect(mockFileSync).toHaveBeenCalledWith(['testPath', expectedConfigJSON]);
```

rholinshead (approved):
Just a suggestion to mock instead of writing to temp file, not a blocker

Ankush-lastmile:
rebase

Ankush-lastmile:
> Just a suggestion to mock instead of writing to temp file, not a blocker

Never got around to this. Created #598

PR #213, opened 2023-11-15
[extensions] 1/n refactor HuggingFace TextGeneration MP impl

Ankush-lastmile:
[extensions] 1/n refactor HuggingFace TextGeneration MP impl

modelparser class name
- HuggingFaceTextGenerationModelParser
- model.id() -> `HuggingFaceTextGenerationModelParser`

updated root imports
- init.py
- index.ts

Copied over implementations from core library.

Looks like the files got duplicate when the folder was renamed `Extensions` -> `extensions`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/213).
* #218
* #210
* __->__ #213

### DIFF
--- a/Extensions/HuggingFace/python/src/aiconfig_extension_huggingface_textgeneration/__init__.py
+++ b/Extensions/HuggingFace/python/src/aiconfig_extension_huggingface_textgeneration/__init__.py
@@ -1,0 +1,5 @@
+"""
+Hugging Face Extension for AIConfig
+"""
+
+from .hf import HuggingFaceTextParser
--- a/Extensions/HuggingFace/typescript/hf.ts
+++ b/Extensions/HuggingFace/typescript/hf.ts
@@ -13,10 +13,10 @@ import {
   ExecuteResult,
   AIConfigRuntime,
   InferenceOptions,
+  ParameterizedModelParser,
 } from "aiconfig";
-import { CompletionCreateParams } from "openai/resources";
+
 import _ from "lodash";
-import * as aiconfig from "aiconfig";
 import { JSONObject } from "aiconfig/dist/common";
 
 export function getAPIKeyFromEnv(apiKeyName: string) {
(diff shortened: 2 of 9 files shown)


Extensions/HuggingFace/typescript/hf.ts
```diff
@@ -0,0 +1,246 @@
...
+
+    if (stream) {
+      const response = await this.hfClient.textGenerationStream(
+        textGenerationArgs
+      );
+      const output = await ConstructStreamOutput(
+        response,
+        options as InferenceOptions
+      );
+      return output;
```
rholinshead, line 159: Need to add outputs to prompt

Extensions/HuggingFace/python/src/aiconfig_extension_huggingface_textgeneration/__init__.py
rholinshead, line 1: Hmm, this is adding back the `Extensions` folder -- can you just add this stuff to `extensions` instead?

rholinshead (changes requested):
Mainly requesting changes so we don't re-introduce `Extensions` folder instead of just updating the files in `extensions` folder.

But some comments on the ts implementation -- I think you can just copy/paste the current core hf.ts parser here and it will have them all resolved

Extensions/HuggingFace/python/src/aiconfig_extension_huggingface_textgeneration/__init__.py
LastMileNewGrad: I'm not sure what happened here since I remember deleting the `Extensions` Folder.  I will remake this diff stack to to make it explicit and easy to understand.

rholinshead: Git (or maybe it's an OS thing) struggles with moves/renames that change only in casing; for move, normally `sl mv` will work but for this one in particular I had to use a temp folder first (https://github.com/lastmile-ai/aiconfig/pull/199)

rossdanlm:
Will follow up later, closing for now

PR #214, opened 2023-11-15
[TS] Add LlamaModelParser Extension

rholinshead:
[TS][WIP] Add LlamaModelParser Extension

# [TS] Add LlamaModelParser Extension

Implemented a `LlamaModelParser` extension, which supports local execution of LLaMA models in GGUF format. All GGUF LLaMA models from https://huggingface.co/TheBloke?search_models=gguf should be supported. Under the hood, this package leverages [node-llama-cpp](https://withcatai.github.io/node-llama-cpp/) for prompt execution.

By default, LlamaModelParser will look for .gguf models within `./models` folder relative to the running process. To use a different directory, specify the model directory when constructing the parser:

```
const parser = new LlamaModelParser(path.resolve('path', 'to', 'models'));
```

The directory must contain models meeting the following criteria:

- Must be valid GGUF format models with `.gguf` extension
- Name must contain the model name registered for use in the parser; e.g. `AIConfigRuntime.registerModelParser(llamaModelParser, ["llama-2-7b-chat"])` must have a `.gguf` file with name containing `llama-2-7b-chat`

We recommend downloading the models provided by TheBloke on HuggingFace: https://huggingface.co/TheBloke?search_models=gguf. For example, to download a few different models:

```
curl -L https://huggingface.co/TheBloke/Llama-2-7B-Chat-GGUF/resolve/main/llama-2-7b-chat.Q4_K_M.gguf --output ./models/llama-2-7b-chat.Q4_K_M.gguf
curl -L https://huggingface.co/TheBloke/Llama-2-13B-chat-GGUF/resolve/main/llama-2-13b-chat.Q4_K_M.gguf --output ./models/llama-2-13b-chat.Q4_K_M.gguf
curl -L https://huggingface.co/TheBloke/CodeUp-Llama-2-13B-Chat-HF-GGUF/resolve/main/codeup-llama-2-13b-chat-hf.Q4_K_M.gguf --output ./models/codeup-llama-2-13b-chat-hf.Q4_K_M.gguf
```

## Testing:
Added a cookbook example for these models and ensured correctness of responses.

TODO:
Before landing: publish the package and update the cookbook to use it instead of the local package
Asap: add callback handling

## Dev notes:
I initially tried to use `llama-node` under the hood for the completions, but ran into several issues:
- ESM usage in commonjs (resolved with `inclusion`, which is also needed for the same reason here)
- Issue with internals of LLM constructor (was saying `llm is not a constructor`)
Transitioned to using `node-llama-cpp` which seems more up-to-date / at least has no other issues besides the esm imports


cookbooks/llama/typescript/package.json
```diff
@@ -0,0 +1,15 @@
+{
+    "name": "aiconfig-extension-llama",
+    "version": "1.0.0",
+    "main": "main.ts",
+    "license": "MIT",
+    "dependencies": {
+        "aiconfig": "^1.0.2",
+        "aiconfig-extension-llama": "file:/Users/ryanholinshead/Projects/aiconfig/extensions/llama/typescript/aiconfig-extension-llama-v1.0.0.tgz"
```
saqadri, line 8: nit: note to update after publishing it

extensions/llama/typescript/llama.ts
```diff
@@ -0,0 +1,343 @@
...
+      completionParams,
+      completionParams.model
+    );
+
+    const prompt: Prompt = {
+      name: promptName,
+      input,
+      metadata: {
+        model: modelMetadata,
+        parameters: params ?? {},
```
saqadri, line 122: nit: can parameters be null? Do we need `?? {}`?

extensions/llama/typescript/llama.ts
```diff
@@ -0,0 +1,343 @@
...
+    return new LlamaChatSession({ context, conversationHistory, systemPrompt });
+  }
+
+  public serialize(
+    promptName: string,
+    data: LlamaCompletionParams,
+    aiConfig: AIConfigRuntime,
+    params?: JSONObject | undefined
+  ): Prompt | Prompt[] {
+    const { prompt: input, conversationHistory, ...completionParams } = data;
```
saqadri, line 108: Looks like `conversationHistory` isn't serialized -- should that get saved each as separate Prompt objects? I believe that's what the openai model parser does.

extensions/llama/typescript/llama.ts
```diff
@@ -0,0 +1,343 @@
+import {
```
saqadri, line 1: The functions are missing callback events.

Would recommend following the pattern in openai.ts:
```
    const startEvent = {
      name: "on_serialize_start",
      file: __filename,
      data: {
        promptName,
        data,
        params,
      },
    } as CallbackEvent;
    aiConfig.callbackManager.runCallbacks(startEvent);
```

This more broadly shows a gap where every model parser creator will have to do the right think with callback events. Perhaps we modify the interface where we have abstract implementations for `run`, `deserialize` and `serialize` which wrap a `runImpl` etc around the callback events, so it's consistent across all model parsers.

cc @LastMileNewGrad @jonathanlastmileai (same for python side)

saqadri (approved):
Approving to unblock -- a few comments

extensions/llama/typescript/llama.ts
rholinshead: Just used hf as an example, but looks like it can be removed

extensions/llama/typescript/llama.ts
rholinshead: Hmm, I'm not sure -- this is where the serialize/deserialize is confusing. The Llama model would never have `conversationHistory` as a param to its completion request and is only something constructed by the parser at run time based on existing prompts. I can try to add that using openai as an example, but seems a bit strange imo

extensions/llama/typescript/llama.ts
rholinshead: @saqadri this is why I mentioned exporting them -- they aren't exposed in the package for me to actually add them :)

rholinshead: I think we can do that for these types of callbacks since they all seem to be before/after events around the method implementations anyway

extensions/llama/typescript/llama.ts
rholinshead: Done

extensions/llama/typescript/llama.ts
rholinshead: Done

rholinshead:
Updated to use published version

Going to land to make sure py side stays unblocked, but can do followups:
- Update to use callbacks after aiconfig npm is updated
- Update `id` from `LLaMA` to `llama` if agreed upon

PR #219, opened 2023-11-15
[AIC-PY][1/n] llama extension+cookbook

jonathanlastmileai:
[AIC-PY][1/n] llama extension+cookbook

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/219).
* #238
* __->__ #219

### DIFF
--- a/cookbooks/llama/python/llama.py
+++ b/cookbooks/llama/python/llama.py
@@ -1,0 +1,1 @@
+/Users/jonathan/Projects/aiconfig/extensions/llama/python/llama.py
--- a/extensions/llama/python/llama.py
+++ b/extensions/llama/python/llama.py
@@ -1,0 +1,68 @@
+from typing import Any
+from aiconfig import Output, Prompt
+from aiconfig.Config import AIConfigRuntime
+from aiconfig.default_parsers.parameterized_model_parser import ParameterizedModelParser
+from aiconfig.model_parser import InferenceOptions
+from aiconfig.schema import ExecuteResult
+
+from llama_cpp import CreateCompletionResponse, Llama
+
+from aiconfig.util.params import resolve_prompt
+
+
+SUPPORTED_MODELS = ["llama-2-7b-chat"]
+
+
+class LlamaModelParser(ParameterizedModelParser):
+    def __init__(self, model_path: str) -> None:
+        super().__init__()
+        self.model_path = model_path
+
+    async def serialize(
+        self,
+        prompt_name: str,
+        data: Any,
+        ai_config: AIConfigRuntime,
+        parameters: dict | None = None,
+        **kwargs,
+    ) -> Prompt:
+        parameters = parameters or {}
+        out = Prompt(name="TODO", input="TODO")
+        return out
+
+    async def deserialize(
+        self, prompt: Prompt, aiconfig: AIConfigRuntime, params: dict | None = None
+    ) -> dict:
+        out = {}
+        return out
+
+    def id(self) -> str:
+        return "LLaMA"
+
+    async def run_inference(
+        self,
+        prompt: Prompt,
+        aiconfig: AIConfigRuntime,
+        options: InferenceOptions | None,
+        parameters: dict,
+    ) -> ExecuteResult:
+        resolved = resolve_prompt(prompt, parameters, aiconfig)
+
+        llm = Llama(self.model_path)
+        response = llm(resolved)
+
+        try:
+            texts = [r["choices"][0]["text"] for r in response]
+        except TypeError:
+            texts = [response["choices"][0]["text"]]
+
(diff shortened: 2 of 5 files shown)


cookbooks/llama/python/llama.py
```diff
@@ -0,0 +1 @@
+/Users/jonathan/Projects/aiconfig/extensions/llama/python/llama.py
```
saqadri, line 1: I'm guessing this will be replaces with a requirements.txt that imports the llama extension once published?

extensions/llama/python/llama.py
```diff
@@ -0,0 +1,68 @@
...
+        return out
+
+    async def deserialize(
+        self, prompt: Prompt, aiconfig: AIConfigRuntime, params: dict | None = None
+    ) -> dict:
+        out = {}
+        return out
+
+    def id(self) -> str:
+        return "LLaMA"
```
saqadri, line 40: nit: keep lowercase

saqadri (approved):

extensions/llama/python/llama.py
rholinshead: @jonathanlastmileai  I can make this change upstream in ts if you agree on lowercase?

extensions/llama/python/llama.py
```diff
@@ -0,0 +1,68 @@
...
+from aiconfig.default_parsers.parameterized_model_parser import ParameterizedModelParser
+from aiconfig.model_parser import InferenceOptions
+from aiconfig.schema import ExecuteResult
+
+from llama_cpp import CreateCompletionResponse, Llama
+
+from aiconfig.util.params import resolve_prompt
+
+
+SUPPORTED_MODELS = ["llama-2-7b-chat"]
```
rholinshead, line 13: nit, can remove since many can be supported, right?

extensions/llama/python/llama.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class LlamaModelParser(ParameterizedModelParser):
+    def __init__(self, model_path: str) -> None:
```
rholinshead, line 17: Any reason to do file path vs folder to reuse a model parser for all local llama models?

extensions/llama/python/llama.py
```diff
@@ -0,0 +1,68 @@
...
+    async def serialize(
+        self,
+        prompt_name: str,
+        data: Any,
+        ai_config: AIConfigRuntime,
+        parameters: dict | None = None,
+        **kwargs,
+    ) -> Prompt:
+        parameters = parameters or {}
+        out = Prompt(name="TODO", input="TODO")
```
rholinshead, line 30: Implementation missing?

extensions/llama/python/llama.py
```diff
@@ -0,0 +1,68 @@
...
+    async def run_inference(
+        self,
+        prompt: Prompt,
+        aiconfig: AIConfigRuntime,
+        options: InferenceOptions | None,
+        parameters: dict,
+    ) -> ExecuteResult:
+        resolved = resolve_prompt(prompt, parameters, aiconfig)
+
+        llm = Llama(self.model_path)
```
rholinshead, line 51: Does this support streaming?

extensions/llama/python/llama.py
rholinshead: Just wanted to double-check here r.e. ModelParser ids; I had used LLaMA from the perspective that, if this were a model provided in LastMile, its name would probably be that. It also looks like ChatGPT and HuggingFaceTextGenerationParser are other examples of capitalized ids for other 'generalized' model parsers that support underlying request models in lower case (e.g. gpt-4, llama-2-7b-chat)

jonathanlastmileai: no opinion

extensions/llama/python/llama.py
jonathanlastmileai: not exactly sure im following - you need to load a specific model right?

extensions/llama/python/llama.py
jonathanlastmileai: yep, took a while just to get the hello world of this model parser :(

extensions/llama/python/llama.py
jonathanlastmileai: next diff!

extensions/llama/python/llama.py
rholinshead: Yes, but you could find the model based on name in the folder, so that we can register a single LlamaModelParser instance for any # of supported model names (associated with model binaries in the specified directroy) vs having one parser instance per model name (associated with model binary at the specified file path)

PR #237, opened 2023-11-16
Add Callback Events to LlamaModelParser

rholinshead:
# Add Callback Events to LlamaModelParser

With the Callbacks exposed for use in the latest `aiconfig` npm package, this PR updates the `LlamaModelParser` extension in ts to emit all the default ModelParser callback events.

Also, fixes the `run` implementation to save the outputs to the prompt for now

Updated & ran the `ask-llama.ts` cookbook to test out callback handlers being called with the events

### DIFF
--- a/cookbooks/llama/typescript/ask-llama.ts
+++ b/cookbooks/llama/typescript/ask-llama.ts
@@ -1,5 +1,6 @@
 import { AIConfigRuntime } from "aiconfig";
 import { LlamaModelParser } from "aiconfig-extension-llama";
+import { CallbackEvent, CallbackManager } from "aiconfig/dist/lib/callback";
 
 async function main() {
   const llamaModelParser = new LlamaModelParser();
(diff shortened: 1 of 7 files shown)


extensions/llama/typescript/llama.ts
```diff
@@ -343,15 +396,27 @@ export class LlamaModelParser extends ParameterizedModelParser<LlamaCompletionPa
...
+    const outputs = [
       {
         output_type: "execute_result",
         data: response,
         execution_count: 0,
         metadata: {},
       } as ExecuteResult,
     ];
+
+    prompt.outputs = outputs;
```
rholinshead, line 408: Tacked on this fix to set the outputs in the prompt as well

cookbooks/llama/typescript/ask-llama.ts
```diff
@@ -1,5 +1,6 @@
 import { AIConfigRuntime } from "aiconfig";
```
saqadri, line 1: Not for this diff, but @rholinshead can you add a video to the cookbook readme showing ask-llama in action? Can be pretty basic, but visual would be useful

saqadri (approved):
🚀 🦙

PR #224, opened 2023-11-15
[AIC-PY] cli-mate demo

jonathanlastmileai:
[AIC-PY] cli-mate demo

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/224).
* __->__ #224
* #240

PR #223, opened 2023-11-15
[AIC-py] py linter

jonathanlastmileai:
[AIC-py] py linter

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/223).
* #240
* #238
* __->__ #223

PR #238, opened 2023-11-16
[AIC-PY][2/n] llama extension+cookbook: streaming

jonathanlastmileai:
[AIC-PY][2/n] llama extension+cookbook: streaming

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/238).
* #224
* #240
* __->__ #238
* #223

### DIFF
--- a/cookbooks/llama/python/ask_llama.py
+++ b/cookbooks/llama/python/ask_llama.py
@@ -1,7 +1,9 @@
-from aiconfig import AIConfigRuntime
-from aiconfig.model_parser import InferenceOptions
 import asyncio
-from llama import LlamaModelParser, SUPPORTED_MODELS as LLAMA_MODELS
+
+from aiconfig.model_parser import InferenceOptions
+from llama import LlamaModelParser
+
+from aiconfig import AIConfigRuntime
 
 
 async def main():
@@ -9,17 +11,34 @@ async def main():
         model_path="models/llama-2-7b-chat.Q4_K_M.gguf"
     )
 
-    for lm in LLAMA_MODELS:
+    for lm in [
+        "llama-2-7b-chat",
+        "llama-2-13b-chat",
+        "codeup-llama-2-13b-chat-hf",
+    ]:
         AIConfigRuntime.register_model_parser(llama_model_parser, lm)
 
     config = AIConfigRuntime.load("cookbooks/llama/llama-aiconfig.json")
 
-    inference_options = InferenceOptions()
+    print("Getting response without streaming...")
+    response7b = await config.run("prompt7b", {})
+    texts = config.get_output_text("prompt7b", response7b)
+    print(f"\n\n\nFinal output:\n{texts[0]}")
 
-    response7b = await config.run("prompt7b", inference_options=inference_options)
+    print("\n\n\n\nGetting streaming response...")
+
+    def stream_callback(data, accumulated_message, index):
+        print(data, end="", flush=True)
+
+    inference_options = InferenceOptions(
+        stream=True,
+        stream_callback=stream_callback,
+    )
+
+    response7b = await config.run("prompt7b", {}, options=inference_options)
 
     texts = config.get_output_text("prompt7b", response7b)
-    print(f"Output:\n{texts[0]}")
+    print(f"\n\n\nFinal output:\n{texts[0]}")
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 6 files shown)


cookbooks/llama/python/ask_llama.py
```diff
@@ -1,25 +1,44 @@
...
+        stream=True,
+        stream_callback=stream_callback,
+    )
 
-    response7b = await config.run("prompt7b", inference_options=inference_options)
+    response7b = await config.run("prompt7b", {}, options=inference_options)
 
     texts = config.get_output_text("prompt7b", response7b)
-    print(f"Output:\n{texts[0]}")
+    print(f"\n\n\nFinal output:\n{texts[0]}")
```
rholinshead, line 41: This will just repeat the stuff printed via streaming, right? Can probably remove to reduce confusion

extensions/llama/python/llama.py
```diff
@@ -48,15 +45,30 @@ async def run_inference(
...
 
-        llm = Llama(self.model_path)
-        response = llm(resolved)
+        # print(f"{prompt=}, {resolved=}, {parameters=}, {options=},{aiconfig=}")
 
-        try:
-            texts = [r["choices"][0]["text"] for r in response]
-        except TypeError:
-            texts = [response["choices"][0]["text"]]
+        stream = True
```
rholinshead, line 50: Should be from options

extensions/llama/python/llama.py
```diff
@@ -48,15 +45,30 @@ async def run_inference(
...
+            return ExecuteResult(output_type="execute_result", data=[acc], metadata={})
+        else:
+            response = llm(resolved)
+            try:
+                texts = [r["choices"][0]["text"] for r in response]
+            except TypeError:
+                texts = [response["choices"][0]["text"]]
 
-        return ExecuteResult(output_type="execute_result", data=texts, metadata={})
+            return ExecuteResult(output_type="execute_result", data=texts, metadata={})
```
rholinshead, line 71: `texts` is an array of strings here, right? If so, should update `get_output_text` to join the data to a single string?

rholinshead (approved):
Accepting to unblock, just small things

extensions/llama/python/llama.py
rholinshead, line 71: Oh, should also put this into the prompt outputs before returning

extensions/llama/python/llama.py
```diff
@@ -48,15 +45,30 @@ async def run_inference(
...
+        acc = ""
+        if stream:
+            for res in llm(resolved, stream=True):
+                raw_data = res["choices"][0]
+                data = raw_data["text"]
+                index = int(raw_data["index"])
+                acc += data
+                if options:
+                    options.stream_callback(data, acc, index)
+            print(flush=True)
```
rholinshead, line 62:
```suggestion
```

extensions/llama/python/llama.py
```diff
@@ -48,15 +45,30 @@ async def run_inference(
     ) -> ExecuteResult:
         resolved = resolve_prompt(prompt, parameters, aiconfig)
```
rholinshead, line 46: Should add the logging for start/end events in the methods as well

extensions/llama/python/llama.py
jonathanlastmileai: Good question - what is it supposed to return if there's more than one string? cc @LastMileNewGrad

jonathanlastmileai: sorry not following :D

extensions/llama/python/llama.py
jonathanlastmileai: This is just for a newline

extensions/llama/python/llama.py
jonathanlastmileai: via callbacks?

extensions/llama/python/llama.py
rholinshead: i.e. 
```
outputs = [ExecuteResult(output_type="execute_result", data=texts, metadata={})]
prompt.outputs = outputs
return outputs
```

extensions/llama/python/llama.py
rholinshead: Yep

PR #239, opened 2023-11-16
[Docs] Simplify docs structure further

saqadri:
[Docs] Simplify docs structure further

Have a single page for extensibility for now, and also get rid of "SDK Overview" (content from that should just be in Usage Guide)


saqadri:
fyi @rholinshead I moved the model parser extensibility page to be extensibility.md

PR #240, opened 2023-11-16
[AIC-PY][3/n] llama extension+cookbook: remember

jonathanlastmileai:
[AIC-PY][3/n] llama extension+cookbook: remember

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/240).
* #224
* __->__ #240

### DIFF
--- a/extensions/llama/python/llama.py
+++ b/extensions/llama/python/llama.py
@@ -14,6 +14,7 @@ class LlamaModelParser(ParameterizedModelParser):
     def __init__(self, model_path: str) -> None:
         super().__init__()
         self.model_path = model_path
+        self.qa = []
 
     async def serialize(
         self,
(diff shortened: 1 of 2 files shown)


extensions/llama/python/llama.py
```diff
@@ -24,35 +25,57 @@ async def serialize(
...
+        return {"model_input": model_input}
 
     def id(self) -> str:
         return "LLaMA"
 
+    def history(self) -> str:
+        def _fmt(q, a):
+            return f"Q: {q}\nA: {a}"
+
+        out = "\n".join(_fmt(q, a) for q, a in self.qa)
```
rholinshead, line 55: Conversation history should be derived from previous prompts of the same model

jonathanlastmileai: Not sure what you mean exactly

rholinshead (changes requested):
For chat history, we should:
- in deserialize, grab input and response from previous prompts of the same model to construct the history from, instead of maintaining history in model parser state
- in serialize, need to support adding prev prompts if given a list of messages for the conversation history

extensions/llama/python/llama.py
rholinshead: The aiconfig runtime will contain a prompt for each message in the history already; we should use that instead of maintaining it in state in the model parser. See the ts implementation for example, or the openai model parser

jonathanlastmileai: It seems outputs is empty, so I only have the history of questions available. cc @LastMileNewGrad

rholinshead: Lol, added a comment on the streaming PR about that -- for now model parsers need to set prompt.outputs within `run_inference` implementation

jonathanlastmileai: uhh so what's the advantage really? :D It's trivial to just do it like this

rholinshead: The advantage is you can take in a config constructed from a conversation, add a new prompt and run that single new prompt -- and it continues the conversation. This current implementation wouldn't do that since it would require manually running all prompts (even though the history is in the config to use)

rholinshead:
I think this got accidentally merged with the cli-mate cookbook, but please fix forward per comments @jonathanlastmileai

jonathanlastmileai:
> I think this got accidentally merged with the cli-mate cookbook, but please fix forward per comments @jonathanlastmileai

ah strange. re: saving history?

rholinshead:
> > I think this got accidentally merged with the cli-mate cookbook, but please fix forward per comments @jonathanlastmileai
> 
> ah strange. re: saving history?

Yes. I think this implementation should leverage the existing previous prompts of the same model for the chat history. It also still needs the prompt.outputs set in `run_inference`

PR #262, opened 2023-11-17
[python][easy] hf update update method from refactor

LastMileNewGrad:
[python][easy] hf update update method from refactor

## What
generate_model_metadata -> get_model_metadata

## Why
This wasn't updating during previous refactor


LastMileNewGrad:
cc: @rossdanlm thanks for catching that

saqadri (approved):

PR #281, opened 2023-11-18
Close Open Handle in `withTimeout`

rholinshead:
# Close Open Handle in `withTimeout`

When running `yarn test` there's an informative warning:
> Jest did not exit one second after the test run has completed.

'This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.

Re-running with `--detectOpenHandles` surfaced:
```
Jest has detected the following 4 open handles potentially keeping Jest from exiting:

  ●  Timeout

      23 | ): Promise<any> {
      24 |   const timeoutPromise = new Promise((_, reject) => {
    > 25 |     setTimeout(() => {
         |     ^
      26 |       reject(new Error("Timeout"));
      27 |     }, timeout * 1000);
      28 |   });

      at lib/callback.ts:25:5
      at lib/callback.ts:24:26
```

The problem is the timeout is still active ones the Promise is resolved. We can close the open handle by clearing the timeout if the input Promise resolves before the timeout promise.

## Testing:
Re-running `yarn test` has no warning about open handles & callback tests work

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/281).
* #283
* #282
* __->__ #281
* #280
* #279

PR #280, opened 2023-11-18
Add Callbacks to Core HF ModelParser

rholinshead:
# Add Callbacks to Core HF ModelParser

This core ModelParser was missing all the callbacks, so add them here.

## Testing:
- Added jest tests in https://github.com/lastmile-ai/aiconfig/pull/283
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/280).
* #283
* #282
* #281
* __->__ #280
* #279

PR #279, opened 2023-11-18
[ts] Add outputs to Prompt in HF Parser

rholinshead:
# [ts] Add outputs to Prompt in HF Parser

When using the HF parser as an example for Llama, I noticed it didn't correctly set the prompt output during `run`. This PR correctly sets it, similar to the openai parser

## Testing:
- Added jest tests in https://github.com/lastmile-ai/aiconfig/pull/283

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/279).
* #283
* #282
* #281
* #280
* __->__ #279

PR #282, opened 2023-11-18
Fix eslint tsconfigRoot

rholinshead:
# Fix eslint tsconfigRoot

Open a `.test.ts` file on main and you'll see this error at the top:
```
Parsing error: Cannot read file '/users/ryanholinshead/projects/aiconfig/tsconfig.json'
```

Fix this by setting `tsconfigRootDir: __dirname` in `eslintrc`. Note that we need to convert this to an actual object in `js` since `__dirname` is an environment variable and not accessible in raw json. The `.js` and `.json` files should be equivalent otherwise

## Testing:
No more red squigglies in test files & tests still work

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/282).
* #283
* __->__ #282
* #281
* #280
* #279


typescript/.eslintrc.js
```diff
@@ -0,0 +1,17 @@
+export default {
+  root: true,
+  extends: ["eslint:recommended", "plugin:@typescript-eslint/recommended"],
+  parser: "@typescript-eslint/parser",
+  parserOptions: { project: "./tsconfig.json", tsconfigRootDir: __dirname },
```
LastMileNewGrad, line 5: I am unfamiliar with the nuances of the eslint, is the tsconfigRootDir required as a parserOption? I noticed it wasn't set in the LLama extension

LastMileNewGrad (approved):
lgtm

typescript/.eslintrc.js
rholinshead: It's not required, but seems needed in this case to have it found in the __tests__. It's not set in the llama extension since there's no __tests__ dir where the issue would be seen (or any directory excluded from compilation in tsconfig)

PR #283, opened 2023-11-18
Fix HF Model Parser OutputText & Add Tests

rholinshead:
Fix HF Model Parser OutputText & Add Tests


rholinshead:
cc @LastMileNewGrad for visibility in this stack here, since I know you were porting the HF parser over as an extension example

typescript/lib/parsers/hf.ts
```diff
@@ -170,7 +240,8 @@ export class HuggingFaceTextGenerationParser extends ParameterizedModelParser<Te
     }
 
     if (output.output_type === "execute_result") {
-      return output.data as string;
+      return (output.data as TextGenerationOutput | TextGenerationStreamOutput)
+        .generated_text as string;
```
LastMileNewGrad, line 244: I also noticed this bug, however I think this should be fixed when constructing the result, (from my wip diff)[https://github.com/lastmile-ai/aiconfig/pull/213/files#diff-a31886f1d5938735e6e6d480f8a8dadb925d6ae6163f953db5cf7c56cb82201cR95-R109]

- Stream output to return accumulated message, (not last token)[https://github.com/lastmile-ai/aiconfig/pull/213/files#diff-a31886f1d5938735e6e6d480f8a8dadb925d6ae6163f953db5cf7c56cb82201cR95]
- regular output to set data to the string .generated_text (link)[https://github.com/lastmile-ai/aiconfig/pull/213/files#diff-a31886f1d5938735e6e6d480f8a8dadb925d6ae6163f953db5cf7c56cb82201cR115]

LastMileNewGrad (approved):
lgtm, had one comment but not blocking

typescript/lib/parsers/hf.ts
rholinshead: I think it's correct for the output data to match the response structure, instead of pulling the raw text out of it, fwiw. In any case, this is just fixing the existing code to not type the object as string so going to land as-is and any updates to the result in your PR can be propagated back to the core class as well

PR #285, opened 2023-11-18
[RFC][WIP] readme: explain 2 format modes and use cases more clearly

jonathanlastmileai:
[RFC][WIP] readme: explain 2 format modes and use cases more clearly

I took a swing at updating the AIConfig docs to reconcile artifact mode with
source-controllable mode.

The premise of this change is that ipynb is unsource-controllable, which means
AIConfig format cannot be simultaneously analogous to ipynb and config.

The answer @sarmad and I discussed is that there are 2 separate
modes or use cases each backed by the format.

### DIFF
--- a/README.md
+++ b/README.md
@@ -543,3 +543,15 @@ There are 2 areas where we are going beyond what notebooks offer:
 
 1. `aiconfig` is more **source-control friendly** than `ipynb`. `ipynb` stores binary data (images, etc.) by encoding it in the file, while `aiconfig` recommends using file URI references instead.
 2. `aiconfig` can be imported and **connected to application code** using the AIConfig SDK.
+
+### How is AIConfig _simultaneously_ like Jupyter notebooks and source-controllable configuration?
+
+AIConfig has two broad classes of use case that use the storage format slightly differently:
+(a) Production dev cycle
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,7 +11,15 @@
 
 ## Overview
 
-AIConfig saves prompts, models and model parameters as source control friendly configs. This allows you to iterate on prompts and model parameters _separately from your application code_.
+AIConfig is an SDK that eases the development of LLM apps in two broad ways, each roughly corresponding to a class of use cases:
```
saqadri, line 14: This isn't the right spot for this nuance. The overview should convey the basic premise.

This nuance either belongs in an FAQ to start, or in a dedicated page of its own on the docs website.

saqadri (changes requested):

README.md
jonathanlastmileai: Fair enough, but I think the previous version of this is not doing justice to the library - we don't only separate code from config in a source controllable way; we also make it easier to do quick iteration in notebook/playground mode

saqadri: That's a good point. We can update the overview to show more of that value. It just has to be done in a succinct way

PR #286, opened 2023-11-18
Update getting-started.md

activescott:
When using `pip install ...` I received `pip Permission denied: 'LICENSE'` on my jupyter stack which is pretty vanilla (https://github.com/activescott/my-jupyter). I fixed it by using `pip install --user ...`. 

My understanding is that if not using virtualenv then using `--user` is at least perfectly fine and often preferred. A couple other references:
- https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-user
- https://stackoverflow.com/questions/33004708/osx-el-capitan-sudo-pip-install-oserror-errno-1-operation-not-permitted

Or maybe this should just be a note about possible errors and an alternative way of installing. Hope this helps!

### DIFF
--- a/aiconfig-docs/docs/getting-started.md
+++ b/aiconfig-docs/docs/getting-started.md
@@ -44,7 +44,7 @@ $ yarn add aiconfig
 <TabItem value="pip">
 
 ```bash
-$ pip install python-aiconfig
+$ pip install --user python-aiconfig
 ```
 
 </TabItem>


saqadri (approved):
Thanks for catching this! Approving this so you can merge, also cc @rossdanlm if we need to update these instructions with pip3 here as well

saqadri:
Also, separately @activescott would love your thoughts on the library and project

activescott:
@saqadri First impression upon an hour or so in is good. I plan to try it out on a small project this coming week. I'll keep sending issues/or PRs as I find things. Always happy to chat though too.

rossdanlm:
> Thanks for catching this! Approving this so you can merge, also cc @rossdanlm if we need to update these instructions with pip3 here as well

I'd prefer to keep explicit `pip3` to make it easier for intro-level devs (https://github.com/lastmile-ai/aiconfig/pull/276) but thanks for the change, looks great!

saqadri:
Merging this and @rossdanlm feel free to update on top of stack

PR #288, opened 2023-11-19
[dall-e parser][1/n] Initial AIConfig Python parser for Dall-E 3 (non-streaming)

rossdanlm:
[dall-e parser][1/n] Initial AIConfig Python parser for Dall-E 3 (non-streaming)

This takes in inputs from the AIConfig and saves them as outputs in the AIConfig prompt.

Right now this only supports

- non-streaming (will add streaming next diff)
- Dall-E 3 (will make the parser class more generic to support both Dall-E 3 and Dall-E 2 in future diff)
- Python

Results
Using color as yellow (the metadata text says green with yellow hue)
Screenshot 2023-11-19 at 18 27 31

Using color as red
Screenshot 2023-11-19 at 18 27 39

The input and output AIConfigs
Screenshot 2023-11-19 at 18 28 22

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/288).
* #295
* #294
* #290
* __->__ #288

### DIFF
--- a/python/src/aiconfig/default_parsers/dalle.py
+++ b/python/src/aiconfig/default_parsers/dalle.py
@@ -1,0 +1,204 @@
+import copy
+from typing import TYPE_CHECKING, Any, Dict, Optional
+
+# Type hint imports
+from aiconfig import (
+    ExecuteResult,
+    Output,
+    Prompt,
+    PromptMetadata,
+)
+
+# ModelParser Utils
+from aiconfig import ParameterizedModelParser
+from aiconfig import resolve_prompt
+from aiconfig import InferenceOptions
+from aiconfig.util.config_utils import get_api_key_from_environment
+import openai
+from openai import OpenAI
+
+# Dall-E API imports
+from openai.types import ImagesResponse, Image
+
+# Circuluar Dependency Type Hints
+if TYPE_CHECKING:
+    from aiconfig.Config import AIConfigRuntime
+
+
+# Step 1: define Helpers
+
+# TODO: Centralize this into utils function with the HuggingFaceTextParser class
+def refine_image_completion_params(model_settings):
+    """
+    Refines the completion params for the Dall-E request API. Removes any unsupported params.
+    The supported keys were found by looking at the OpenAI Dall-E API: https://platform.openai.com/docs/api-reference/images/create?lang=python`
+    """
+    supported_keys = {
+        "model",
+        "n",
+        "quality",
+        "response_format",
+        "size",
+        "style"
+    }
+
+    completion_data = {}
+    for key in model_settings:
+        if key.lower() in supported_keys:
+            completion_data[key.lower()] = model_settings[key]
+
+    return completion_data
+
+
+def construct_regular_output(image_data: Image, execution_count: int) -> Output:
+    output = ExecuteResult(
+        **{
+            "output_type": "execute_result",
+            "data": image_data.b64_json or image_data.url,
+            "execution_count": execution_count,
+            "metadata": {"revised_prompt": image_data.revised_prompt},
+        }
+    )
+    return output
+
+
+class DallE3ImageGenerationParser(ParameterizedModelParser):
+    """
+    A model parser for Dall-E 3 text-->image generation models.
+    """
+    
+    def __init__(self):
+        """
+        Usage:
+
+        1. Create a new model parser object
+                parser = DallE3ImageGenerationParser()
+        2. Add the model parser to the registry.
+                config.register_model_parser(parser)
+
+        The model parser will require an API token to be set in the environment variable OPENAI_API_KEY.
+        """
+        super().__init__()
+        self.client = None
+
+    def id(self) -> str:
+        """
+        Returns an identifier for the model (e.g. dall-e-2, dall-e-3, etc.).
+        """
+        return "dall-e-3"
+
+
+    # TODO: Test this sometime from getting a response from the API and saving it to a file
+    def serialize(
+        self,
+        prompt_name: str,
+        data: Any,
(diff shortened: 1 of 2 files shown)


python/src/aiconfig/default_parsers/dalle.py
```diff
@@ -0,0 +1,197 @@
...
+        self.client = OpenAI(api_key=get_api_key_from_environment("OPENAI_API_KEY"))
+
+    def id(self) -> str:
+        """
+        Returns an identifier for the model (e.g. dall-e-2, dall-e-3, etc.).
+        """
+        return "dall-e-3"
+
+
+    # TODO: Test this sometime from getting a response from the API and saving it to a file
```
rossdanlm, line 90: @LastMileNewGrad I was planning just to test by calling `DallE3ImageGenerationParser.serialize()` but do you have any other suggestions? I could also directly move the serialize logic from `run_inference()` into here too (I think that's more intuitive)

python/src/aiconfig/default_parsers/dalle.py
```diff
@@ -0,0 +1,197 @@
...
+            not "stream" in completion_data or completion_data.get("stream") != False
+        )
+
+        print("Calling image generation. This can take several seconds, please hold on...")
+        response : ImagesResponse = self.client.images.generate(**completion_data)
+
+        outputs = []
+        if not stream:
+            # ImageResponse object also contains a "created" field for timestamp, should I store that somewhere?
+            # Ex: response=ImagesResponse(created=1700347843, data=[...])
```
rossdanlm, line 173: @LastMileNewGrad what do you think?

python/src/aiconfig/default_parsers/dalle.py
```diff
@@ -0,0 +1,197 @@
...
+        The model parser will require an API token to be set in the environment variable OPENAI_API_KEY.
+        """
+        super().__init__()
+        self.client = OpenAI(api_key=get_api_key_from_environment("OPENAI_API_KEY"))
+
+    def id(self) -> str:
+        """
+        Returns an identifier for the model (e.g. dall-e-2, dall-e-3, etc.).
+        """
+        return "dall-e-3"
```
rossdanlm, line 87: Will generalize this to also work for `dall-e-2` model later in diff stack

python/src/aiconfig/default_parsers/dalle.py
```diff
@@ -0,0 +1,204 @@
...
+
+        outputs = []
+        if not stream:
+            # ImageResponse object also contains a "created" field for timestamp, should I store that somewhere?
+            # Ex: response=ImagesResponse(created=1700347843, data=[...])
+            for execution_count, image_data in enumerate(response.data):
+                output = construct_regular_output(image_data, execution_count)
+                outputs.append(output)
+        else:
+            # TODO (rossdamlm): Enable streaming next diff
```
LastMileNewGrad, line 185: does the dall-e api support streaming? I checked the OpenAI Api source code and I'm not sure I see streaming as an option for image generation

python/src/aiconfig/default_parsers/dalle.py
LastMileNewGrad: ya you should be able to test by calling serialize, only the run step requires mocking. There should be an example of this test for the openai model parser.

>could also directly move the serialize logic from run_inference() into here too 

Could you clarify what this entails?

python/src/aiconfig/default_parsers/dalle.py
LastMileNewGrad: One option is you can dump all the relevant data in metadata of the output

python/src/aiconfig/default_parsers/dalle.py
rossdanlm: I think you're right. Added https://github.com/lastmile-ai/aiconfig/pull/290 to ignore streaming options

python/src/aiconfig/default_parsers/dalle.py
rossdanlm: Yea it's just that it seems a bit annoying to add this for every image output (since created field is the same for all of them), but I think I'll do it just to keep all the details. Will do this in future diff to keep commits small

python/src/aiconfig/default_parsers/dalle.py
rossdanlm: > Could you clarify what this entails?

In run_inference() we generally have:

1. Building completion params with `deserialize()`
2. Make API request
3. Parse API response into Prompt Output (this step instead of manually parsing, we could call `serialize()` like we do for `deserialize()` in step 1)

LastMileNewGrad: I see where you're getting at, I think that could be a pretty near api design improvement.

Currently the api for `serialize()` is designed to take completion params and return a `Prompt` Object.

The implicit "serialize" in the run step takes the completion and constructs an Output for a Prompt Object. How would you approach merging the two?

LastMileNewGrad:
cc: @saqadri should the dall-e model parser be a core parser or an extension? 

Given that it only relies on the OpenAI library, thereby not introducing any additional dependencies, I see no problem making it a core parser

python/src/aiconfig/default_parsers/dalle.py
rossdanlm: Added tests for serialize: https://github.com/lastmile-ai/aiconfig/pull/294

LastMileNewGrad:
discussed offline, any openai parser can be a core parser

LastMileNewGrad (approved):
lgtm

PR #290, opened 2023-11-20
[dall-e parser][2/n] Do not read streaming options

rossdanlm:
[dall-e parser][2/n] Do not read streaming options

I don't think that the image generation, edit, or variations support streaming for the Dall-E 2/3 APIs. I just removed this logic because we could have streaming enabled for other prompts (ex: text-based ones)

Even for image variations (dall-e 2), the examples use a BytesIO stream, but only to get the final value of it. If you look at the API we need to pass in a full file: https://platform.openai.com/docs/guides/images/using-in-memory-image-data

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/290).
* #297
* #294
* __->__ #290

### DIFF
--- a/python/src/aiconfig/default_parsers/dalle.py
+++ b/python/src/aiconfig/default_parsers/dalle.py
@@ -50,7 +50,7 @@ def refine_image_completion_params(model_settings):
     return completion_data
 
 
-def construct_regular_output(image_data: Image, execution_count: int) -> Output:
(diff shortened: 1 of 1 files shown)


LastMileNewGrad (approved):
lgtm

PR #298, opened 2023-11-20
[dall-e parser][5/n] Add mime type for urls (png) and base64 (plaintext)

rossdanlm:
[dall-e parser][5/n] Add mime type for urls (png) and base64 (plaintext)

Didn't know about this until now, wasn't adding it earlier.

The common MIME types are listed here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types

| Before | After |
|---|---|
| Screenshot 2023-11-20 at 18 11 17 | Screenshot 2023-11-20 at 18 15 55 |

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/298).
* __->__ #298
* #297

### DIFF
--- a/python/src/aiconfig/default_parsers/dalle.py
+++ b/python/src/aiconfig/default_parsers/dalle.py
@@ -57,12 +57,13 @@ def construct_output(image_data: Image, execution_count: int) -> Output:
             "data": image_data.b64_json or image_data.url,
             "execution_count": execution_count,
             "metadata": {"revised_prompt": image_data.revised_prompt},
+            "mime_type": "image/png",
         }
     )
     return output
 
 
-class DallEImageGenerationParser(ParameterizedModelParser):
+class DalleImageGenerationParser(ParameterizedModelParser):
     """
     A model parser for Dall-E 2 and Dall-E 3 text-->image generation models.
     """
@@ -72,7 +73,7 @@ class DallEImageGenerationParser(ParameterizedModelParser):
         Usage:
 
         1. Create a new model parser object
-                parser = DallEImageGenerationParser("dall-e-3")
+                parser = DalleImageGenerationParser("dall-e-3")
         2. Add the model parser to the registry.
                 config.register_model_parser(parser)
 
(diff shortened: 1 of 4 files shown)


python/src/aiconfig/default_parsers/dalle.py
```diff
@@ -51,45 +51,64 @@ def refine_image_completion_params(model_settings):
 
 
 def construct_output(image_data: Image, execution_count: int) -> Output:
+    mime_type : Optional[str] = None
+    data : Optional[str] = None
+    
+    if image_data.b64_json:
+        mime_type = "text/plain"
+        data = image_data.b64_json
```
saqadri, line 59: I don't think this is correct. Even if the data is base-64 encoded, the MIME type is for what the data represents. So I believe it should always be `image/png`. cc @rholinshead as well

python/src/aiconfig/default_parsers/dalle.py
```diff
@@ -51,45 +51,64 @@ def refine_image_completion_params(model_settings):
...
             "execution_count": execution_count,
             "metadata": {"revised_prompt": image_data.revised_prompt},
+            "mime_type": mime_type,
         }
     )
     return output
 
 
-class DallE3ImageGenerationParser(ParameterizedModelParser):
+class DallEImageGenerationParser(ParameterizedModelParser):
```
saqadri, line 76: nit
```suggestion
class DalleImageGenerationParser(ParameterizedModelParser):
```

saqadri (approved):
A few comments, but approving to unblock.

python/src/aiconfig/default_parsers/dalle.py
rholinshead: Yep, that's correct. One thing we should consider is if we want to just enforce serialized data as uris instead of supporting binary data dumped into the config. The latter will be a huge pain code review and likely for performance down-the-line (e.g. saving/loading a big json blob with many images is not going to be super fast)

PR #299, opened 2023-11-20
Apply Schema to Existing Config JSON Files

rholinshead:
# Apply Schema to Existing Config JSON Files

As it turns out, we already have a valid schema representation for the AIConfig in the repo! We can immediately start linting against it in VS Code by adding the appropriate `$schema` property pointed to that file: 
```
"$schema": "https://github.com/lastmile-ai/aiconfig/raw/main/schema/aiconfig.schema.json",
```

A couple discussion points:
- Do we want to version this schema file under a v1 folder to point to that, so we can add new versions without 'breaking' existing configs (i.e. having VS Code lint warnings show up for them)?
- Looks like quite a few configs were already not matching the schema (and linted accordingly), mainly missing metadata of having `null` instead of undefined keys for some values (programmatic creation)
- Do we want to include this schema value in the serialized JSON so that the files immediately lint in VS Code when opening after serializing to disk?

### DIFF
--- a/cookbooks/Basic-Prompt-Routing/assistant_aiconfig.json
+++ b/cookbooks/Basic-Prompt-Routing/assistant_aiconfig.json
@@ -1,4 +1,5 @@
 {
+  "$schema": "https://github.com/lastmile-ai/aiconfig/raw/main/schema/aiconfig.schema.json",
   "name": "assistant_config",
   "schema_version": "latest",
   "metadata": {
--- a/cookbooks/Chain-of-Verification/cove_config.json
+++ b/cookbooks/Chain-of-Verification/cove_config.json
@@ -1,4 +1,5 @@
 {
+  "$schema": "https://github.com/lastmile-ai/aiconfig/raw/main/schema/aiconfig.schema.json",
   "name": "Chain of Verification (CoVe) ",
   "schema_version": "latest",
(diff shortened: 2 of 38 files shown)


rholinshead:
Need to upstream changes to the type to make prompt metadata optional first

Ankush-lastmile:
>Do we want to include this schema value in the serialized JSON so that the files immediately lint in VS Code when opening after serializing to disk?

Does vscode automatically do that without any extra changes/additions?

rholinshead:
> >Do we want to include this schema value in the serialized JSON so that the files immediately lint in VS Code when opening after serializing to disk?
> 
> 
> 
> Does vscode automatically do that without any extra changes/additions?

If the $schema value is present, VS code will immediately lint the file with that schema type. But we need to explicitly set that $schema value inside the config json files

Ankush-lastmile:
>Do we want to version this schema file under a v1 folder

I would say yes, we define the code-schema as v1 In the code already. I don't see how it could hurt in any way

[code ref](https://github.com/lastmile-ai/aiconfig/blob/main/typescript/types.ts#L217)

rholinshead:
Going to close this out for now since we'll need to update the ts interface and JSON schema to make prompt metadata optional first

PR #294, opened 2023-11-20
[dall-e parser][3/n] Add basic tests for load and serialize

rossdanlm:
[dall-e parser][3/n] Add basic tests for load and serialize

Pretty simple, just adding in the tests which I already did manually to test the previous commits

Specifically I added tests for `load()` and `serialize()`. Found an error on serialize through this test (had to define function as `async`!

Should now be able to run tests by entering:
```
cd ~/Projects/aiconfig/python && pytest --disable-warnings -vv
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/294).
* #301
* #298
* #297
* __->__ #294

PR #301, opened 2023-11-21
[dall-e parser][6/n] Add test file for running image generation configs

rossdanlm:
[dall-e parser][6/n] Add test file for running image generation configs

Tests that the code runs and response is well formatted to prompt

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/301).
* __->__ #301
* #298
* #297
* #294

### DIFF
--- a/python/src/aiconfig/Config.py
+++ b/python/src/aiconfig/Config.py
@@ -3,7 +3,7 @@ import sys
 import os
 from aiconfig.callback import CallbackEvent, CallbackManager
 from .default_parsers.hf import HuggingFaceTextGenerationParser
-from .default_parsers.dalle import DallE3ImageGenerationParser
+from .default_parsers.dalle import DallEImageGenerationParser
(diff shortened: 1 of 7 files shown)


rossdanlm:
This is broken now because the mock API key isn't allowing it to be run: 

Screenshot 2023-11-20 at 20 29 11

cc @LastMileNewGrad any quick fixes? Did you run into this issue when adding `test_get_output_text()` to the `test_openai_util.py` file?

Ankush-lastmile:
Yep this is the issue I was talking about previously. When Instantiating a client when using the openai library, the pytest fixture to mock the completion call isn't working as intended.

Looking at your error message, `incorrect api key`, seems like the completion api request went through. This means the mock method to mock the completion/api did not work as intended

PR #310, opened 2023-11-21
[aic-py] autoformat with black

jonathanlastmileai:
[aic-py] autoformat with black

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/310).
* __->__ #310
* #309
* #304


.lint_glob_excludes
```diff
@@ -0,0 +1 @@
+python/build/**
```
Anpala94, line 1: what happens if you don't include this?

Ankush-lastmile, line 1: what happens if you don't include this?

jonathanlastmileai: I think build is from pypi builds? So pylint will complain about it. but it's just copies of what's in src. same with `dist`. I'm actually not sure why there's both dist and build.

CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @saqadri @rholinshead @suyoglastmileai @Ankush-lastmile @jonathanlastmileai @rossdanlm
```
Ankush-lastmile, line 1: nit: would prefer if any small changes were also reflected or mentioned in commit description somewhere

Ankush-lastmile (approved):
lgtm, though looks like you have some tests failing

CODEOWNERS
jonathanlastmileai: Right, good find. This actually merged with #304 but this was stacked before the merge.

PR #314, opened 2023-11-22
[extensions][ts] 1/n Update to Model Parser impl

Ankush-lastmile:
[extensions][ts] 1/n Update to Model Parser impl

## What

1. Copied over ts implementation from core library

2. two smaller changes

- model.id()
- - `HuggingFaceTextGenerationParser` -> `HuggingFaceTextGenerationModelParserExtension`
- modelparser class name
- - `HuggingFaceTextGenerationParser` -> `HuggingFaceTextGenerationModelParserExtension`

## Why

Starting diff for stack to create an extension that serves as an example on how to write a model parser extension including comprehensive docs, docstrings, etc.

Consequently, the model id has to differ as the library already has HuggingFace model parser

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/314).
* #315
* __->__ #314
* #312

### DIFF
--- a/extensions/HuggingFace/typescript/hf.ts
+++ b/extensions/HuggingFace/typescript/hf.ts
@@ -9,12 +9,13 @@ import {
   Prompt,
   Output,
   PromptInput,
+  ParameterizedModelParser,
   ModelMetadata,
   ExecuteResult,
   AIConfigRuntime,
   InferenceOptions,
+  CallbackEvent
 } from "aiconfig";
-import { CompletionCreateParams } from "openai/resources";
 import _ from "lodash";
 import * as aiconfig from "aiconfig";
 import { JSONObject } from "aiconfig/dist/common";
@@ -32,9 +33,9 @@ export function getAPIKeyFromEnv(apiKeyName: string) {
  * Set the environment variable HUGGING_FACE_API_TOKEN to use your HuggingFace API token.
  * A HuggingFace API token is not required to use this model parser.
  */
-export class HuggingFaceTextGenerationModelParser extends aiconfig.ParameterizedModelParser<TextGenerationArgs> {
+export class HuggingFaceTextGenerationModelParserExtension extends ParameterizedModelParser<TextGenerationArgs> {
   private hfClient: HfInference | undefined;
-  _id = "HuggingFaceTextGenerationModelParser";
+  _id = "HuggingFaceTextGenerationModelParserExtension";
 
   public constructor() {
     super();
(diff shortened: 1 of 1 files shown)


rossdanlm:
nit: rename `HuggingFaceTextGenerationModelParser` to `HuggingFaceTextGenerationModelParserExtension`

extensions/HuggingFace/typescript/hf.ts
```diff
@@ -148,20 +205,33 @@ export class HuggingFaceTextGenerationModelParser extends aiconfig.Parameterized
...
       );
-      return output;
     } else {
       const response = await this.hfClient.textGeneration(textGenerationArgs);
-      const output = constructOutput(response);
-      return output;
+      output = constructOutput(response);
     }
+
+    prompt.outputs = [output];
```
rossdanlm, line 223: Why do we do this here and not before?

extensions/HuggingFace/typescript/hf.ts
```diff
@@ -178,7 +248,8 @@ export class HuggingFaceTextGenerationModelParser extends aiconfig.Parameterized
     }
 
     if (output.output_type === "execute_result") {
-      return output.data as string;
+      return (output.data as TextGenerationOutput | TextGenerationStreamOutput)
+        .generated_text as string;
```
rossdanlm, line 252: This makes sense to me. I feel we should change the core lib too? Although tbh most code will hit the L242 anyways and never reach here

rossdanlm (approved):
Generally lgtm, left comments for questions

extensions/HuggingFace/typescript/hf.ts
Ankush-lastmile: I'm assuming you were referring to similar discussion as here: https://github.com/lastmile-ai/aiconfig/pull/283#discussion_r1399248692

I'll update the python to be similar, sounds like we want to keep the response structure as much as possible

extensions/HuggingFace/typescript/hf.ts
Ankush-lastmile: Most of these changes are from [PR#283](https://github.com/lastmile-ai/aiconfig/pull/283)

Ankush-lastmile:
>nit: rename HuggingFaceTextGenerationModelParser to HuggingFaceTextGenerationModelParserExtension

Great suggestion, Thanks!

renamed `HuggingFaceTextGenerationModelParser` to `HuggingFaceTextGenerationModelParserExtension`

updated pr description.

PR #317, opened 2023-11-22
[AIC-py] promptfoo integration prototype

jonathanlastmileai:
[AIC-py] promptfoo integration prototype

Make a custom provider using AIConfig and configure it with promptfoo yaml.

Run (TS connector):
`npx promptfoo@latest eval -c typescript/scripts/eval/promptfoo/examples/travel/travel_promptfooconfig.yaml`

Run (Py connector):
`npx promptfoo@latest eval -c python/src/aiconfig/eval/promptfoo/examples/travel/travel_promptfooconfig.yaml`

image

### DIFF
--- a/python/src/aiconfig/eval/promptfoo/README.md
+++ b/python/src/aiconfig/eval/promptfoo/README.md
@@ -1,0 +1,30 @@
+# Promptfoo integration
+
+Use case: I'm a SWE who wants to run my AIConfig against a set of test cases specified in a config file. Each test case has the input and a success condition of my choosing.
+
+## Philosophy / design
+
+Prompfoo has a pretty nice interface (both input and outputs) for addressing the use case. Tests are specified in a yaml file and the test suite can be run with a simple command. The same config file makes it easy to connect your test suite to an AI config with a small amount of code.
+
+## How-to guide
+
+1. Write your test cases in a Promptfoo config file. See examples/travel/travel_promtfooconfig.yaml as an example.
+2. Define an AIConfig test suite settings file. It should have the prompt name and path to your aiconfig. See examples/travel/travel_aiconfig_test_suite_settings.json for example.
+3. Set your provider to point to run_aiconfig.py with your settings file as the argument. For e.g. see examples/travel/travel_promtfooconfig.yaml. Like this:
+
+```
+providers:
+  - exec:python ../../run_aiconfig.py ./travel_aiconfig_test_suite_settings.json
+```
+
+4. export your provider API key if needed so it's available to subprocess environments:
+   `export OPENAI_API_KEY=...`
+
+5. Make sure your shell environment (including subshells) contains a python3 executable called `python` on its path. One way to do this is to set up an anaconda/miniconda environment.
+
+6. Run `npx promptfoo@latest eval -c path/to/promptfooconfig.yaml`
+You should see one passing test and one failing test.
+
+### Debugging / Troubleshooting
+Help, something went wrong! 
+-> Try adding --verbose flag at the end of the `npx` command.
--- a/typescript/scripts/eval/promptfoo/run_aiconfig.ts
+++ b/typescript/scripts/eval/promptfoo/run_aiconfig.ts
@@ -1,0 +1,31 @@
+import { AIConfigRuntime } from "../../../lib/config";
+import fs from "fs";
+import path from "path";
+
+async function main() {
+  let settings_path = process.argv[2];
+  let question = process.argv[3];
+  let settings = _load_settings(settings_path);
+  let prompt_name = settings["prompt_name"];
+  let aiconfig_path = settings["aiconfig_path"];
+
+  let fullAIConfigPath = path.join(__dirname, aiconfig_path);
+  let runtime = AIConfigRuntime.load(fullAIConfigPath);
+  let params = {
+    the_query: question,
+  };
+  let result = await runtime.run(prompt_name, params);
+  let final_output: string;
+
+  let r0 = Array.isArray(result) ? result[0] : result;
+  final_output = runtime.getOutputText(prompt_name, r0);
+  console.log(final_output);
+}
+
+function _load_settings(settings_path: string) {
+  let rawdata = fs.readFileSync(settings_path, "utf-8");
+  let settings = JSON.parse(rawdata);
+  return settings;
+}
+
+main();
--- a/python/src/aiconfig/eval/promptfoo/examples/travel/travel_aiconfig_test_suite_settings.json
+++ b/python/src/aiconfig/eval/promptfoo/examples/travel/travel_aiconfig_test_suite_settings.json
@@ -1,0 +1,4 @@
+{
+  "prompt_name": "get_activities",
+  "aiconfig_path": "travel_parametrized_for_testing.aiconfig.json"
+}
--- a/python/src/aiconfig/eval/promptfoo/examples/travel/travel_parametrized_for_testing.aiconfig.json
+++ b/python/src/aiconfig/eval/promptfoo/examples/travel/travel_parametrized_for_testing.aiconfig.json
@@ -1,0 +1,36 @@
+{
+  "name": "NYC Trip Planner",
+  "description": "Intrepid explorer with ChatGPT and AIConfig",
+  "schema_version": "latest",
+  "metadata": {
+    "models": {
+      "gpt-3.5-turbo": {
+        "model": "gpt-3.5-turbo",
+        "top_p": 1,
+        "temperature": 1
+      },
+      "gpt-4": {
+        "model": "gpt-4",
+        "max_tokens": 3000,
+        "system_prompt": "You are an expert travel coordinator with exquisite taste."
+      }
+    },
+    "default_model": "gpt-3.5-turbo"
+  },
+  "prompts": [
+    {
+      "name": "get_activities",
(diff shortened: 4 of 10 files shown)


saqadri:
I don't fully understand this yet -- is this the full integration with promptfoo? I don't see an interface to set this up for any aiconfig. Assuming this is just to show it running with aiconfig for the travel example

JLessinger:
> I don't fully understand this yet -- is this the full integration with promptfoo? I don't see an interface to set this up for any aiconfig. Assuming this is just to show it running with aiconfig for the travel example 

 Can you clarify what kind of interface would be needed on top of promptfoo's yaml format? Are we trying to allow people to evaluate an AIConfig with less boilerplate? Is it more about easily doing many AIConfigs in a batch?

saqadri:
Well what is the process for using promptfoo for evaluating aiconfigs? The code doesn't make it clear.

Is there anything needed to be generated, like the prompt.txt, from an aiconfig to be able to use it?

jonathanlastmileai:
> Well what is the process for using promptfoo for evaluating aiconfigs? The code doesn't make it clear.
> 
> Is there anything needed to be generated, like the prompt.txt, from an aiconfig to be able to use it?

I see, ok. Let me update with a readme and let's go from there for next iteration.

cookbooks/Getting-Started/eval/promptfoo/prompt.txt
```diff
@@ -0,0 +1 @@
+{{question}}
```
saqadri, line 1: This seems like something that could or should be autogenerated

cookbooks/Getting-Started/eval/promptfoo/travel.py
```diff
@@ -0,0 +1,30 @@
...
+
+import openai
+from dotenv import load_dotenv
+
+from aiconfig import AIConfigRuntime
+from aiconfig.model_parser import InferenceOptions
+import sys
+
+
+async def main():
```
saqadri, line 12: This script seems like something that should be an interface or function in the core library which can set this up.

Inputs to it: the aiconfig, and prompt to run. Rest is boilerplate .

Or even more general, could take in an AIConfigRuntime object and a run function

cookbooks/Getting-Started/eval/promptfoo/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+## How-to guide
+
+1. Make a file `prompt.txt` that just has the contents
+
+   `"{{question}}"`
+   
+    (literally, don't put any actual question there).
+2. Write your test cases in a Promptfoo config file. See promtfooconfig.yaml as an example. Don't worry about "providers" for now; we'll get to that.
+3. Write a "connector" script to tell Promptfoo how to test your AIConfig. For example, `travel.aiconfig.json` has the corresponding `travel.py` script.
```
saqadri, line 17: This is where I think we can provide a simple interface that takes an aiconfig (either file or object), prompt to run, parameter overrides, and whether to run with dependencies or not (I.e. same signature as run function).

Then people don't need to create this script themselves for the vast majority of use cases.

cookbooks/Getting-Started/eval/promptfoo/travel.py
jonathanlastmileai: Yeah makes sense, almost all the boilerplate here (probably?) can be automated. There doesn't seem to be a great way to run it as a library though - the most sensible procedure I can think of is to have an AIConfig testing util script. Something like:

```
python aiconfig_test.py setup path/to/aiconfig prompt_name path/to/new/config.yaml
python aiconfig_test.py test path/to/new/config.yaml
```

jonathanlastmileai: Actually you don't even need the `test` subcommand. It's just `npx promptfoo@latest eval`.

cookbooks/Getting-Started/eval/promptfoo/README.md
jonathanlastmileai: OK, let me take a swing at this.

cookbooks/Getting-Started/eval/promptfoo/travel.py
jonathanlastmileai: > Or even more general, could take in an AIConfigRuntime object and a run function

At that point, why not have the user write their own script? I think inverting control at that level of complexity can make debugging harder more than it makes authoring easier.

cookbooks/Getting-Started/eval/promptfoo/README.md
jonathanlastmileai: > parameter overrides, and whether to run with dependencies or not (I.e. same signature as run function).


I'm a little confused about this part - AIConfigRuntime.run takes the prompt name, inputs (aka params), and inference options that I don't think are relevant for testing, right? What's the run with dependencies option? cc @Ankush-lastmile

Ankush-lastmile: run_with_dependencies is a kwarg you pass into run() to invoke run_with_dependencies.

I think the idea here is that the interface is built on top of config.run(), and all the attributes for run are passed through from the interface

jonathanlastmileai: Right, I see. I think for testing we need run_with_dependencies==False exactly when we want to unit test a config. i.e. pass a test suite specific inputs to one prompt without running its dependencies. 

Let's note that this is a special case of running an arbitrary sub-DAG, which we probably want to support eventually.  So we will want to generalize run_with_dependencies to something like run_with_leaves() where leaves sets the input params for the sub-DAG we want to run.

jonathanlastmileai: We have an indirection layer now (settings.json) where we can add arbitrary config as needed like run_with_dependencies=False. Can that be a follow up?

saqadri:
Is this ready for review?

jonathanlastmileai:
> Is this ready for review?

Yes

typescript/scripts/eval/promptfoo/run_aiconfig.ts
```diff
@@ -0,0 +1,35 @@
...
+
+function _load_settings(settings_path: string) {
+  let rawdata = fs.readFileSync(settings_path, "utf-8");
+  let settings = JSON.parse(rawdata);
+  return settings;
+}
+
+if (process.env.npm_lifecycle_event != null) {
+  main();
+}
```
suyoglastmileai, line 35: What is this w/ process.env.npm_lifecycle_event and why is it needed? (I've never seen this used before btw, why can't I just call `main()`?).

python/src/aiconfig/eval/promptfoo/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Philosophy / design
+
+Prompfoo has a pretty nice interface (both input and outputs) for addressing the use case. Tests are specified in a yaml file and the test suite can be run with a simple command. The same config file makes it easy to connect your test suite to an AI config with a small amount of code.
+
+## How-to guide
+
+1. Write your test cases in a Promptfoo config file. See examples/travel/travel_promtfooconfig.yaml as an example.
+2. Define an AIConfig test suite settings file. It should have the prompt name and path to your aiconfig. See examples/travel/travel_aiconfig_test_suite_settings.json for example.
+3. Set your provider to point to run_aiconfig.py with your settings file as the argument. For e.g. see examples/travel/travel_promtfooconfig.yaml. Like this:
```
suyoglastmileai, line 13:
```suggestion
1. Write your test cases in a Promptfoo config file. See examples/travel/travel_promptfooconfig.yaml as an example.
2. Define an AIConfig test suite settings file. It should have the prompt name and path to your aiconfig. See examples/travel/travel_aiconfig_test_suite_settings.json for example.
3. Set your provider to point to run_aiconfig.py with your settings file as the argument. For e.g. see examples/travel/travel_promptfooconfig.yaml. Like this:
```

package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "devDependencies": {
+    "@types/node": "^20.10.0"
+  },
+  "dependencies": {
+    "dotenv": "^16.3.1",
+    "fs": "^0.0.1-security",
+    "openai": "^4.20.1"
+  }
+}
```
suyoglastmileai, line 10: These should be inside of the typescript folder's package.json right? (and then you don't need package-lock.json or yarn.lock in the root folder

typescript/scripts/eval/promptfoo/run_aiconfig.ts
```diff
@@ -0,0 +1,33 @@
+import { config as load_dotenv } from "dotenv";
```
suyoglastmileai, line 1: I don't think dotenv is a dependency inside of `typescript/package.json` so this will have an error when it hits main now.

To test this you would need to `cd ~/Projects/aiconfig && rm -rf node_modules`, then try running this again since your previous commit version had installed dotenv.

Fix would be add dotenv as dependency to `typescript/package.json`, run `yarn` in the typescript directory & add `package.json` and `yarn.lock` to this diff.

jonathanlastmileai: Removed import.

suyoglastmileai (approved):

PR #318, opened 2023-11-22
[py] streaming respect stream value

Ankush-lastmile:
[py] streaming refactor wip

## What

refactor stream check to respect stream options

## Why

streaming options wasn't respecting the stream attribute

### DIFF
--- a/python/src/aiconfig/default_parsers/openai.py
+++ b/python/src/aiconfig/default_parsers/openai.py
@@ -235,7 +235,7 @@ class OpenAIInference(ParameterizedModelParser):
         # const stream = options?.stream ?? completionParams.stream ?? true;
         stream = True  # Default value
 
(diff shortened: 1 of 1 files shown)


python/src/aiconfig/default_parsers/openai.py
```diff
@@ -235,7 +235,7 @@ async def run_inference(
         # const stream = options?.stream ?? completionParams.stream ?? true;
         stream = True  # Default value
 
-        if options is not None and options.stream:
+        if options is not None and options.stream is not None:
```
Ankush-lastmile, line 238: without this change, this condition never gets hit if its false

jonathanlastmileai (approved):
Trusting you've tested this reasonably thoroughly :D

PR #320, opened 2023-11-22
Make prompt.metadata Optional in Schema and TS Prompt Type

rholinshead:
Make prompt.metadata Optional in Schema and TS Prompt Type

# Make prompt.metadata Optional in Schema and TS Prompt Type
This ensure the typescript implementation and the schema definition match existing python implementation and desired functionality -- the metadata isn't always needed and no point in having it be an empty object

## Testing
```
ryanholinshead@Ryans-MacBook-Pro typescript % yarn test
yarn run v1.22.19
$ jest --runInBand
 PASS  __tests__/testProgramaticallyCreateConfig.ts
  test Get Global Settings
    ✓ Retrieving global setting from AIConfig with 1 model (1 ms)
  ExtractOverrideSettings function
    ✓ Should return initial settings when no global settings are defined
    ✓ Should return an override when initial settings differ from global settings
    ✓ Should return empty override when global settings match initial settings (10 ms)
    ✓ Should return empty override when Global settings defined and initial settings are empty (2 ms)

 PASS  __tests__/config.test.ts
  Loading an AIConfig
    ✓ loading a basic chatgpt query config (7 ms)
    ✓ loading a prompt chain (4 ms)
    ✓ deserialize and re-serialize a prompt chain (3 ms)
    ✓ serialize a prompt chain with different settings (1 ms)

 PASS  __tests__/parsers/hf/hf.test.ts
  HuggingFaceTextGeneration ModelParser
    ✓ uses HuggingFace API token from environment variable if it exists (6 ms)
    ✓ serializing params to config prompt (1 ms)
    ✓ serialize callbacks (1 ms)
    ✓ deserializing config prompt to params (1 ms)
    ✓ deserialize callbacks
    ✓ run prompt, non-streaming (3 ms)
    ✓ run prompt, streaming (3 ms)
    ✓ run callbacks (1 ms)

Test Suites: 3 passed, 3 total
Tests:       17 passed, 17 total
Snapshots:   0 total
Time:        1.118 s
Ran all test suites.
✨  Done in 2.08s.
```

### DIFF
--- a/typescript/lib/config.ts
+++ b/typescript/lib/config.ts
@@ -546,7 +546,10 @@ export class AIConfigRuntime implements AIConfig {
         );
       }
 
-      prompt.metadata.model = modelMetadata;
+      prompt.metadata = {
+        ...prompt.metadata,
+        model: modelMetadata,
+      };
     } else {
       if (!this.metadata.models) {
         this.metadata.models = {};
@@ -570,7 +573,10 @@ export class AIConfigRuntime implements AIConfig {
         );
       }
 
-      prompt.metadata.model = modelMetadata;
+      prompt.metadata = {
+        ...prompt.metadata,
+        model: modelMetadata,
+      };
     } else {
       if (!this.metadata.models) {
         this.metadata.models = {};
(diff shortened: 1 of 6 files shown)


typescript/lib/config.ts
```diff
@@ -701,7 +709,10 @@ export class AIConfigRuntime implements AIConfig {
         );
       }
 
-      prompt.metadata[key] = value;
+      prompt.metadata = {
+        ...prompt.metadata,
```
Ankush-lastmile, line 713: does the spread operator automatically ignore optionals if it doesn't exist?

saqadri (approved):

typescript/lib/config.ts
rholinshead: Yep!
Screenshot 2023-11-22 at 6 21 58 PM

PR #332, opened 2023-11-28
[hotfix][AIC-py] accept openai response: either dict or pydantic model

jonathanlastmileai:
[hotfix][AIC-py] accept openai response: either dict or pydantic model


saqadri (approved):
Approving to unblock -- did openai completion api change or did we have a regression here?

jonathanlastmileai:
> Approving to unblock -- did openai completion api change or did we have a regression here?

It seems we tried to update with the OpenAI API following this: https://github.com/openai/openai-python/discussions/742

> Response objects are now [pydantic models](https://docs.pydantic.dev/latest/usage/models/)


but the official openAI docs show it as a dict, not a pydantic model. I confirmed from running the cookbook it's a dict. So I'm not sure exactly what's going on, i.e. why that discussion says its a Pydantic model. @Ankush-lastmile any idea?

Ankush-lastmile:
>official openAI docs show it as a dict
I'm not sure if that has been updated but the api uses pydantic.

The issue here isn't the model parser, but the wrapper

https://github.com/lastmile-ai/aiconfig/blob/main/python/src/aiconfig/ChatCompletion.py#L83

jonathanlastmileai:
> > official openAI docs show it as a dict
> 
> I'm not sure if that has been updated but the api uses pydantic.
> 
> I think The issue here isn't the model parser, but the wrapper
> 
> https://github.com/lastmile-ai/aiconfig/blob/main/python/src/aiconfig/ChatCompletion.py#L83
> 
> This should be updated to create a new object when dumping the model rather than converting the existing object. The object gets passed forward since its a stream, but it shouldn't be converted for the user

When I run  `response = openai.chat.completions.create(**completion_data)` directly, it does indeed return a BaseModel. Do we know why it doesn't when run from openai.py?

Ankush-lastmile:
>Do we know why it doesn't when run from openai.py?

From [the wrapper](https://github.com/lastmile-ai/aiconfig/blob/main/python/src/aiconfig/ChatCompletion.py#L83), notice when streaming,  this chunk object is getting converted into a dict with `model_dump`, then gets passed forward with the [yield statement](https://github.com/lastmile-ai/aiconfig/blob/main/python/src/aiconfig/ChatCompletion.py#L100). But the user might still be expecting that BaseModel, in openai.py its expecting that basemodel

PR #333, opened 2023-11-28
[cookbooks] wrapper notebook fix last cell

Ankush-lastmile:
[cookbooks] wrapper notebook fix last cell

## What

update aiconfig run cell to use original chat completion api

## Why

openai wrapper is used to create configs. It must be reset to the original api, `openai.chat.completions.create()` in order for the OpenAI model parser to work.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/333).
* #334
* __->__ #333


Ankush-lastmile:
closing, found a fix for the wrapper itself

PR #334, opened 2023-11-28
wrapper don't mutate stream chunks

Ankush-lastmile:
wrapper don't mutate stream chunks

updated notebook to expect basemodel as wellz

## Why

the stream chunk gets forwarded to the user (not pictured in this diff) through the yield statement several lines down


jonathanlastmileai:
Can you show an example of exactly what was broken and how this fixes it?

jonathanlastmileai (approved):

saqadri (approved):

PR #341, opened 2023-11-28
[ez][2/n][AIC-py] rm unused imports

jonathanlastmileai:
[ez][2/n][AIC-py] rm unused imports

### DIFF
--- a/python/src/aiconfig/util/config_utils.py
+++ b/python/src/aiconfig/util/config_utils.py
@@ -3,7 +3,7 @@ import os
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
-    from aiconfig.Config import AIConfigRuntime
+    pass
 
     from aiconfig.schema import InferenceSettings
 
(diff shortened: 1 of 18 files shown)


python/tests/test_run_config.py
```diff
@@ -3,7 +3,7 @@
 from aiconfig.Config import AIConfigRuntime
 from mock import patch
 
-from .conftest import mock_openai_chat_completion, set_temporary_env_vars
+from .conftest import mock_openai_chat_completion
```
rholinshead, line 6: Just confirming it's not needed in the function def below? Not sure if the pytest decorator does magic like calling with this as the param

python/src/aiconfig/util/config_utils.py
```diff
@@ -3,7 +3,7 @@
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
-    from aiconfig.Config import AIConfigRuntime
+    pass
```
rholinshead, line 6: Should this just be removed and move the InferenceSettings import up?

rholinshead (approved):

python/tests/test_run_config.py
jonathanlastmileai: No magic AFAIK. test passes

PR #343, opened 2023-11-28
[eng week] Set run output type to be list[Output] instead of Output

rossdanlm:
[eng week] Set run output type to be list[Output] instead of Output

Makes it more scalable. The Prompt schema already refers to `outputs` as `List[Output]` anyways

### DIFF
--- a/cookbooks/HuggingFace/hf.py
+++ b/cookbooks/HuggingFace/hf.py
@@ -1,5 +1,5 @@
 import copy
-from typing import TYPE_CHECKING, Any, Dict, Optional
+from typing import TYPE_CHECKING, Any, Dict, List, Optional
 
 # HuggingFace API imports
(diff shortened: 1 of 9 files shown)


cookbooks/HuggingFace/hf.py
```diff
@@ -1,5 +1,5 @@
 import copy
-from typing import TYPE_CHECKING, Any, Dict, Optional
+from typing import TYPE_CHECKING, Any, Dict, List, Optional
```
saqadri, line 2: Can you also do this for the typescript side?

rossdanlm:
Note for adding TS support, Ryan did this so can reference his syntax: https://github.com/lastmile-ai/aiconfig/pull/473

Ankush-lastmile:
#617

PR #344, opened 2023-11-28
[hf parsers][1/n] Transformer for text generation

rossdanlm:
[hf parsers][1/n] Transformer for text generation

This is super basic, just getting started.

1) Have not supported streaming
2) Have not supported using any of the completion params in the generator call
3) Hardcoded to using the 'gpt2' model, will make it more generic in the future
4) Have not added automated test yet for serialize and loading the AIConfig

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/344).
* #380
* #379
* #351
* #363
* #348
* #345
* __->__ #344

### DIFF
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -14,3 +14,10 @@ prompt_toolkit
 mock
 pytest-asyncio
 lastmile_utils==0.0.3
+transformers
+
+# TODO (rossdanlm): remove from core lib after moving to extension: https://github.com/lastmile-ai/aiconfig/pull/344
+transformers
+torch
+torchvision
+torchaudio
--- a/python/src/aiconfig/Config.py
+++ b/python/src/aiconfig/Config.py
@@ -10,6 +10,7 @@ from aiconfig.model_parser import InferenceOptions, ModelParser
 
 from .default_parsers.dalle import DalleImageGenerationParser
 from .default_parsers.hf import HuggingFaceTextGenerationParser
+from .default_parsers.hugging_face_transformers.text_generation import HuggingFaceTextGenerationTransformer
 from .registry import (
     ModelParserRegistry,
     update_model_parser_registry_with_config_runtime,
@@ -35,6 +36,7 @@ for model in gpt_models:
 ModelParserRegistry.register_model_parser(PaLMChatParser())
 ModelParserRegistry.register_model_parser(PaLMTextParser())
 ModelParserRegistry.register_model_parser(HuggingFaceTextGenerationParser())
+ModelParserRegistry.register_model_parser(HuggingFaceTextGenerationTransformer('gpt2'))
 dalle_image_generation_models = [
     "dall-e-2",
     "dall-e-3",
--- a/python/src/aiconfig/default_parsers/hugging_face_transformers/text_generation.py
+++ b/python/src/aiconfig/default_parsers/hugging_face_transformers/text_generation.py
@@ -1,0 +1,208 @@
+import copy
+from typing import TYPE_CHECKING, Any, Dict, List, Optional
+from transformers import pipeline
+
+from aiconfig.util.params import resolve_prompt
+from aiconfig.default_parsers.parameterized_model_parser import ParameterizedModelParser
+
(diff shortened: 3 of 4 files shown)


rossdanlm:
Testing with using `num_return_sequences = 4` in the generator call. See docs from https://huggingface.co/docs/transformers/main_classes/text_generation for all params that are supported

Screenshot 2023-11-28 at 18 14 52

python/src/aiconfig/Config.py
```diff
@@ -35,6 +36,7 @@
 ModelParserRegistry.register_model_parser(PaLMChatParser())
 ModelParserRegistry.register_model_parser(PaLMTextParser())
 ModelParserRegistry.register_model_parser(HuggingFaceTextGenerationParser())
+ModelParserRegistry.register_model_parser(HuggingFaceTextGenerationTransformer('gpt2'))
```
Ankush-lastmile, line 39: can we make this an extension? Its easy to develop here locally but this shouldn't be part of core

python/requirements.txt
```diff
@@ -13,4 +13,10 @@ nest_asyncio
...
 pytest-asyncio
-lastmile_utils==0.0.3
\ No newline at end of file
+lastmile_utils==0.0.3
+transformers
+
+#Pytorch
+torch
+torchvision
+torchaudio
```
Ankush-lastmile, line 22: just a note that the pyproject.toml isn't dynamically linked to the requirements.txt (this can be changed). If you introduce dependencies make sure to update the pyproject accordingly

Ankush-lastmile, line 22: Another thing we need to keep in mind for extensibility is that some models use different versions of pytorch. Not a blocker right now but something to keep in mind since pip sucks at managing version conflicts

python/src/aiconfig/default_parsers/hugging_face_transformers/text_generation.py
```diff
@@ -0,0 +1,208 @@
...
+        "seed",
+        "stop_sequences",
+        "temperature",
+        "top_k",
+        "top_p",
+        "truncate",
+        "typical_p",
+        "watermark",
+        "decoder_input_details",
+    }
```
Ankush-lastmile, line 40: what happened to the other 40 keys? I remember you mentioned there was a lot more

python/src/aiconfig/default_parsers/hugging_face_transformers/text_generation.py
```diff
@@ -0,0 +1,208 @@
...
+        prompt.outputs = outputs
+        return prompt.outputs
+
+    def get_output_text(
+        self,
+        prompt: Prompt,
+        aiconfig: "AIConfigRuntime",
+        output: Optional[Output] = None,
+    ) -> str:
+        if not output:
```
Ankush-lastmile, line 198: nit: `output is not None` is better than checking for a Falsey object. Realizing this as I read this code

Ankush-lastmile:
lgtm but move this to a new extension folder before merging

python/src/aiconfig/default_parsers/hugging_face_transformers/text_generation.py
Ankush-lastmile: nvm see it on the diff onto

rossdanlm:
> lgtm but move this to a new extension folder before merging

cc @Ankush-lastmile I'm going to do this at the end of this stack to avoid merge conflicts

python/requirements.txt
rossdanlm: Fixed in https://github.com/lastmile-ai/aiconfig/pull/358

rossdanlm: Hmmm, yea that seems tough, not sure best solution for now, good to discuss more later

rossdanlm:
> > lgtm but move this to a new extension folder before merging
> 
> cc @Ankush-lastmile I'm going to do this at the end of this stack to avoid merge conflicts

Adding this to https://github.com/lastmile-ai/aiconfig/pull/344

PR #348, opened 2023-11-29
[hf parsers][3/n] Add streaming for text generation transformer

rossdanlm:
[hf parsers][3/n] Add streaming for text generation transformer

Streaming now enabled using the `TextIteratorStreamer` object: https://huggingface.co/docs/transformers/v4.35.2/en/internal/generation_utils#transformers.TextIteratorStreamer

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/348).
* #380
* #379
* #351
* #363
* __->__ #348
* #345
* #344

PR #351, opened 2023-11-29
[hf parsers][4/n] Do not use hardcoded value for model

rossdanlm:
[hf parsers][4/n] Do not use hardcoded value for model

TSIA, we're now able to pass in a model name during the constructor through the AIConfig

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/351).
* #380
* #379
* __->__ #351
* #363
* #348
* #345
* #344

### DIFF
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -14,3 +14,10 @@ prompt_toolkit
 mock
 pytest-asyncio
 lastmile_utils==0.0.3
+transformers
+
+# TODO (rossdanlm): remove from core lib after moving to extension: https://github.com/lastmile-ai/aiconfig/pull/344
+transformers
+torch
+torchvision
+torchaudio
--- a/python/src/aiconfig/default_parsers/hugging_face_transformers/text_generation.py
+++ b/python/src/aiconfig/default_parsers/hugging_face_transformers/text_generation.py
@@ -1,11 +1,12 @@
 import copy
 from typing import TYPE_CHECKING, Any, Dict, List, Optional
-from transformers import pipeline
+from transformers import AutoTokenizer, pipeline, TextIteratorStreamer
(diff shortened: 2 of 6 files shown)


python/src/aiconfig/default_parsers/hugging_face_transformers/text_generation.py
```diff
@@ -0,0 +1,282 @@
...
+        streamer (TextIteratorStreamer): Streams the output. See https://huggingface.co/docs/transformers/v4.35.2/en/internal/generation_utils#transformers.TextIteratorStreamer
+        options (InferenceOptions): The inference options. Used to determine the stream callback.
+
+    """
+    accumulated_message = ""
+    output = ExecuteResult(
+            **{
+                "output_type": "execute_result",
+                "data": accumulated_message,
+                "execution_count": 0, #Multiple outputs are not supported for streaming
```
Ankush-lastmile, line 113: Although you will have trouble streaming the different outputs, the multiple outputs are still generated into the response object. Bit a of a tricky situation but i'll leave it upto you if you want to keep it this way

python/requirements.txt
```diff
@@ -13,4 +13,10 @@ nest_asyncio
 prompt_toolkit
 mock
 pytest-asyncio
-lastmile_utils==0.0.3
\ No newline at end of file
+lastmile_utils==0.0.3
+transformers
```
Ankush-lastmile, line 17: lets pin transformers. I noticed their api docs state that the api will change over time, don't want this to fail unexpectedly.

Ankush-lastmile:
lgtm but lets move this to an extension

PR #352, opened 2023-11-29
1/n align deserialize() to library design

Ankush-lastmile:
1/n align deserialize() to library design

## What

Exclude Previous Run Outputs when performing config.resolve(), aka, model parser `.deserialize()`

## Why

When resolving (deserializing) a prompt, if the prompt as executed before, the output from the previous execution is currently included in the deserialized state. While this was the expected behbaiour in the past, it has become apparent that including previous outputs is counterintuitive and not user-friendly

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/352).
* #371
* #353
* __->__ #352

### DIFF
--- a/python/src/aiconfig/default_parsers/openai.py
+++ b/python/src/aiconfig/default_parsers/openai.py
@@ -209,7 +209,7 @@ class OpenAIInference(ParameterizedModelParser):
                 )
 
         # Add in the latest prompt
-        add_prompt_as_message(prompt, aiconfig, completion_params["messages"], params)
+        add_prompt_as_message(prompt, aiconfig, completion_params["messages"], params, is_last_prompt=True)
         await aiconfig.callback_manager.run_callbacks(
             CallbackEvent(
                 "on_deserialize_complete", __name__, {"output": completion_params}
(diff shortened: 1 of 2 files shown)


jonathanlastmileai (changes requested):
I think this makes sense high-level, but it would be helpful to see some examples / a lot more context. What's an example use case of deserialize, why did it seem before like previous outputs should be kept, and why does it now seem otherwise?

Ankush-lastmile:
This was discussed offline w/ @saqadri 

> What's an example use case of deserialize

Consider this scenario:

1. First Run

- User Prompt: What is a good condiment for a burger? Respond with just the answer.
- GPT Response: Mustard

2. Subsequent Run of the Same Prompt
- Expected Behavior: The GPT should replicate its previous answer.
- Current Behavior: The GPT adds a new response while retaining the old one.

User Prompt: What is a good condiment for a burger? Respond with just the answer.
GPT previous Response included in request: Mustard
______
GPT Response: Ketchup


The desired outcome is for the GPT to consistently return only the initial response (Mustard) for repeated prompts, rather than accumulating additional responses

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -452,7 +452,7 @@ def refine_chat_completion_params(model_settings):
 
 
 def add_prompt_as_message(
-    prompt: Prompt, aiconfig: "AIConfigRuntime", messages: List, params=None
+    prompt: Prompt, aiconfig: "AIConfigRuntime", messages: List, params=None, is_last_prompt: bool = False
```
jonathanlastmileai, line 455: when should it be false?

Ankush-lastmile:
Updated instead in #491

PR #360, opened 2023-11-29
[WIP] [do not land] Add attachments field to PromptInput

rossdanlm:
Add attachments field to PromptInput

Making it so that we can pass in non-text input into AIConfig. Most of this format is taken from the PromptOutput

### DIFF
--- a/python/src/aiconfig/schema.py
+++ b/python/src/aiconfig/schema.py
@@ -59,9 +59,23 @@ class PromptMetadata(BaseModel):
     class Config:
         extra = "allow"
 
+class Attachment(BaseModel):
+    '''
+    Attachment used to pass data in PromptInput for non-text inputs (ex: image, audio)
+    '''
+    # The data representing the attachment
+    data: Any
+    # The MIME type of the result. If not specified, the MIME type will be assumed to be plain text.
+    mime_type: Optional[str] = None
+    # Output metadata
+    metadata: Dict[str, Any]
 
 class PromptInput(BaseModel):
-    # Any additional inputs to the model
+    # Attachements can be used to pass in non-text inputs (ex: image, audio)
+    attachments: List[Attachment] = []
+
+    # Freeform data for the overall prompt input (ex: document answering question 
+    # requires both images (attachments) and question (data))
     data: Optional[Any] = None
 
     class Config:
--- a/typescript/lib/parsers/openai.ts
+++ b/typescript/lib/parsers/openai.ts
@@ -32,6 +32,7 @@ export class OpenAIModelParser extends ParameterizedModelParser<CompletionCreate
 
   public serialize(
     promptName: string,
+    // TODO (rossdanlm): Generalize this openai serializer to work with multi-modal inputs, not just text completion
     data: CompletionCreateParams,
     aiConfig: AIConfigRuntime,
     params?: JSONObject
@@ -42,6 +43,7 @@ export class OpenAIModelParser extends ParameterizedModelParser<CompletionCreate
       input = data.prompt;
     } else {
       input = {
+        attachments: [],
         data: data.prompt,
       };
     }
@@ -274,7 +276,7 @@ export class OpenAIChatModelParser extends ParameterizedModelParser<Chat.ChatCom
     } else if (typeof prompt.input?.data === "string") {
       return prompt.input?.data;
     } else {
-      const message = prompt.input as Chat.ChatCompletionMessageParam;
+      const message = prompt.input?.data as Chat.ChatCompletionMessageParam;
       return message.content ?? "";
     }
   }
@@ -361,8 +363,11 @@ export class OpenAIChatModelParser extends ParameterizedModelParser<Chat.ChatCom
           }
         }
 
-        const input: PromptInput =
-          message.role === "user" ? message.content ?? "" : { ...message };
+        const input: PromptInput = {
+          attachments: [],
+          data:
+            message.role === "user" ? message.content ?? "" : { ...message },
+        };
 
         const prompt: Prompt = {
           name: `${promptName}_${prompts.length + 1}`,
--- a/typescript/types.ts
+++ b/typescript/types.ts
@@ -72,8 +72,31 @@ export type SchemaVersion =
   | "v1"
   | "latest";
 
+export type Attachment = {
+  /**
+   * The data representing the attachment
+   */
+  data: any;
+
+  /**
+   * The MIME type of the result. If not specified, the MIME type will be
+   * assumed to be plain text
+   */
+  mime_type: string | undefined;
+
+  /**
+   * Attachment metadata
+   */
+  metadata?: JSONValue;
+};
+
 export type PromptInput =
   | {
+      /**
+       * Attachements can be used to pass in non-text inputs (ex: image, audio)
+       */
+      attachments: Attachment[];
+
       /**
        * Input to the model. This can represent a single input, or multiple inputs.
        * The structure of the data object is up to the ModelParser. For example,
--- a/typescript/__tests__/config.test.ts
+++ b/typescript/__tests__/config.test.ts
@@ -1,6 +1,7 @@
 import { AIConfigRuntime } from "../lib/config";
 import path from "path";
 import { Prompt } from "../types";
+import { PromptInput } from "aiconfig";
 
 const originalProcessEnv = { ...process.env };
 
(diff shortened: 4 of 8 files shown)


python/src/aiconfig/schema.py
```diff
@@ -59,9 +59,25 @@ class PromptMetadata(BaseModel):
     class Config:
         extra = "allow"
 
+class Attachment(BaseModel):
+    '''
+    Attachment used to pass data in PromptInput for non-text inputs (ex: image, audio)
+    '''
+    # index of the attachment
+    input_count: Union[int, None] = None
```
suyoglastmileai, line 67: Is this needed if attachments is a list below?

rossdanlm: Yea you're probably right, will remove

python/src/aiconfig/schema.py
```diff
@@ -59,9 +59,23 @@ class PromptMetadata(BaseModel):
...
 
+class Attachment(BaseModel):
+    '''
+    Attachment used to pass data in PromptInput for non-text inputs (ex: image, audio)
+    '''
+    # The data representing the attachment
+    data: Any
+    # The MIME type of the result. If not specified, the MIME type will be assumed to be plain text.
+    mime_type: Optional[str] = None
+    # Output metadata
```
rholinshead, line 70:
```suggestion
    # Attachment metadata
```
Assuming this was not intended as metadata specific to an output

rholinshead (commented):
I think this looks good. Only alternative I can think of is just having non-text data supported as an Input instead of an Attachment in an input, and having Prompt support inputs array. The main place that would be beneficial would be if there are multiple text inputs since we already support multiple non-text inputs via attachments array here (would need to have multiple text inputs keyed within data I think)

rholinshead:
Can you also update the ts side and the schema to reflect this change as well so we keep them all consistent?

rossdanlm:
> I think this looks good. Only alternative I can think of is just having non-text data supported as an Input instead of an Attachment in an input, and having Prompt support inputs array. The main place that would be beneficial would be if there are multiple text inputs since we already support multiple non-text inputs via attachments array here (would need to have multiple text inputs keyed within data I think)

In that case it's probably better to make the inputs an array instead of keyed, since ordering might be important. But yea I agree, they can just be as part of data within `Any` format and when parsing, we can check if it's a list type and process in order

rossdanlm:
Can't get `yarn test` to work, mainly due to two errors:

1) `TypeError: mockedHfInference.mockClear is not a function`
2) `ENOENT: no such file or directory, open '/Users/rossdancraig/Projects/aiconfig/typescript/dist/__tests__/samples/basic_chatgpt_query_config.json'`

Timeboxing this for now, can investigate more later

python/src/aiconfig/schema.py
```diff
@@ -59,9 +59,25 @@ class PromptMetadata(BaseModel):
...
+    metadata: Dict[str, Any]
 
 class PromptInput(BaseModel):
-    # Any additional inputs to the model
+    # Attachements can be used to pass in non-text inputs (ex: image, audio)
+    attachments: List[Attachment] = []
```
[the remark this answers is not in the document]
rholinshead: Ah, sorry, missed this. I think having it optional is probably best, in case other prompts have data but no attachments

typescript/lib/parsers/openai.ts
```diff
@@ -274,7 +276,7 @@ export class OpenAIChatModelParser extends ParameterizedModelParser<Chat.ChatCom
     } else if (typeof prompt.input?.data === "string") {
       return prompt.input?.data;
     } else {
-      const message = prompt.input as Chat.ChatCompletionMessageParam;
+      const message = prompt.input?.data as Chat.ChatCompletionMessageParam;
```
rholinshead, line 279: Oh, nice fix

typescript/types.ts
```diff
@@ -72,8 +72,31 @@ export type SchemaVersion =
...
+  /**
+   * The data representing the attachment
+   */
+  data: any;
+
+  /**
+   * The MIME type of the result. If not specified, the MIME type will be
+   * assumed to be plain text
+   */
+  mime_type: string | undefined;
```
rholinshead, line 85:
```suggestion
  mime_type?: string
```

typescript/types.ts
```diff
@@ -72,8 +72,31 @@ export type SchemaVersion =
...
+  /**
+   * Attachment metadata
+   */
+  metadata?: JSONValue;
+};
+
 export type PromptInput =
   | {
+      /**
+       * Attachements can be used to pass in non-text inputs (ex: image, audio)
```
rholinshead, line 96:
```suggestion
       * Attachments can be used to pass in non-text inputs (ex: image, audio)
```

typescript/types.ts
```diff
@@ -72,8 +72,31 @@ export type SchemaVersion =
...
+       */
+      attachments: Attachment[];
```
rholinshead, line 98:
```suggestion
      attachments?: Attachment[];
```

rholinshead (approved):
Accepting to unblock, but strongly think it should be optional. That will also remove the need to change the test files, which may help with your yarn test. Fwiw I just pulled main, cd'd to typescript dir and ran `yarn && yarn test` without issue

python/src/aiconfig/schema.py
```diff
@@ -59,9 +59,23 @@ class PromptMetadata(BaseModel):
...
+class Attachment(BaseModel):
+    '''
+    Attachment used to pass data in PromptInput for non-text inputs (ex: image, audio)
+    '''
+    # The data representing the attachment
+    data: Any
+    # The MIME type of the result. If not specified, the MIME type will be assumed to be plain text.
+    mime_type: Optional[str] = None
+    # Output metadata
+    metadata: Dict[str, Any]
```
suyoglastmileai, line 71: Something that I'm thinking of when getting GPT-4V working - should attachments have optional names? 

GPT-4V just takes a list of images, but other models could have specific named attachments (the one I'm thinking of is Stable Diffusion with Control net could take one style of control image, a second one, and a prompt - in a model parser might be easier to reference by name rather than by index)

rholinshead: Ah, was just thinking something similar when working on the prompt schema. Mainly, parameters right now are really only supporting the prompt.input === "string" scenario, but for attachments we'll need to either support the scenario for input.data === "string" or, better yet, support reference by key so we could potentially also include references to attachments by key as well?

suyoglastmileai: I did this for GPT-4V support btw https://github.com/lastmile-ai/lastmile/blob/743d4713bff9663179ec0e2d8508b511fce804d3/src/helpers/client/workbookCellTypes.ts#L45 ; still using an array, but have an optional name so that you can still filter by name (O(n) because array, but at least it's not just indexed).

python/src/aiconfig/schema.py
```diff
@@ -59,9 +59,23 @@ class PromptMetadata(BaseModel):
...
         extra = "allow"
 
+class Attachment(BaseModel):
+    '''
+    Attachment used to pass data in PromptInput for non-text inputs (ex: image, audio)
+    '''
+    # The data representing the attachment
+    data: Any
+    # The MIME type of the result. If not specified, the MIME type will be assumed to be plain text.
+    mime_type: Optional[str] = None
```
rholinshead, line 69: Left the same comment in the GPT4-V PR as well: do we need two "mime type" specific fields here -- one for the serialized data representation and one for what the data actually is? For example, if we have an image attachment the mime_type could be img/png but the serialized data itself could be uri, base64 string, etc.

We need a way to understand both:
- what the data is for (in order to show the UI for attaching image input / showing after attachment)
- how the data is saved (in order to render the img to show it)

typescript/types.ts
```diff
@@ -72,8 +72,31 @@ export type SchemaVersion =
...
+  /**
+   * The MIME type of the result. If not specified, the MIME type will be
+   * assumed to be plain text
+   */
+  mime_type: string | undefined;
+
+  /**
+   * Attachment metadata
+   */
+  metadata?: JSONValue;
```
rholinshead, line 90: Should be object to match elsewhere

```suggestion
  metadata?: {
    [k: string]: any;
  };
```

typescript/types.ts
```diff
@@ -72,8 +72,31 @@ export type SchemaVersion =
   | "v1"
   | "latest";
 
+export type Attachment = {
+  /**
+   * The data representing the attachment
+   */
+  data: any;
```
rholinshead, line 79:
```suggestion
  data: JSONValue;
```

rholinshead (changes requested):
Just putting back into queue with latest comments

typescript/lib/parsers/openai.ts
rholinshead: Oh wait, not a fix. The intention is to support raw dump of completion params in the input instead of under .data

rossdanlm:
No longer needed since Ryan landed https://github.com/lastmile-ai/aiconfig/pull/473

PR #363, opened 2023-11-29
Add more docstrings to ModelParserRegistry

rossdanlm:
Add more docstrings to ModelParserRegistry

Making this easier to understand what's going on

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/363).
* #380
* #379
* #351
* __->__ #363
* #348
* #345
* #344

### DIFF
--- a/python/src/aiconfig/registry.py
+++ b/python/src/aiconfig/registry.py
@@ -1,5 +1,5 @@
 import typing
-from typing import Dict, List
+from typing import Dict, List, Optional
 
 from aiconfig.schema import Prompt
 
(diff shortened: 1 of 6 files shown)


python/src/aiconfig/registry.py
```diff
@@ -10,21 +10,37 @@
...
+    This then gets updated into the desired format 1) by calling the 
+    `update_model_parser_registry_with_config_runtime()` command
+    
+    Each model_name is only allowed to map to a single model_parser for an AIConfigRuntime.
+    """
     _parsers: Dict[str, ModelParser] = {}
 
     @staticmethod
-    def register_model_parser(model_parser: ModelParser, ids: List[str] = None):
+    def register_model_parser(model_parser: ModelParser, model_names: List[str] = None):
```
saqadri, line 33: Wouldn't this be a breaking change on the AIConfig API? I think we should revert back to `ids`, not worth a breaking change

saqadri (changes requested):

python/src/aiconfig/registry.py
rossdanlm: Yea it would. I feel it's worth updating since it's more clear though. I can just update the docstrings for now though

PR #364, opened 2023-11-29
[AIC-py] Improve openai wrapper + cookbook

jonathanlastmileai:
[AIC-py] Improve openai wrapper + cookbook


python/src/aiconfig/ChatCompletion.py
```diff
@@ -199,3 +203,135 @@ async def run_and_await_serialize():
...
+            output_aiconfig_ref=output_aiconfig_ref, openai_api=api
+        ),
+    )
+    return cast(ModuleType, new_module)
+
+
+def get_completion_create_wrapped_openai_client(
+    output_aiconfig_ref: str | AIConfigRuntime,
+) -> openai.OpenAI:
+    api = openai.Client()
```
Ankush-lastmile, line 328: When the user is instantiating a client, we can expect them to construct it with their own settings and options. For example, [instantiating an azure client](https://learn.microsoft.com/en-us/azure/ai-services/openai/quickstart?tabs=command-line%2Cpython-new&pivots=programming-language-python#:~:text=client%20%3D%20AzureOpenAI(%0A%20%20%20%20api_key%3Dos.getenv(%22AZURE_OPENAI_KEY%22)%2C%20%20%0A%20%20%20%20api_version%3D%222023%2D10%2D01%2Dpreview%22%2C%0A%20%20%20%20azure_endpoint%20%3D%20os.getenv(%22AZURE_OPENAI_ENDPOINT%22)%0A%20%20%20%20))

This should probably still work if you can pass in the client or something similar

python/src/aiconfig/ChatCompletion.py
```diff
@@ -199,3 +203,135 @@ async def run_and_await_serialize():
     else:
         serialized_prompts = asyncio.run(run_and_await_serialize())
     return serialized_prompts
+
+
+# TODO type this
+def create_and_save_to_config_v2(
```
Ankush-lastmile, line 209: if only v2 is used, can we get rid of the v1? can do on top

cookbooks/OpenAI-ChatCompletion-AIConfigWrapper/openai_wrapper.ipynb
```diff
@@ -677,7 +941,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.11.6"
+   "version": "3.10.13"
```
Ankush-lastmile, line 944: We should have a cell showcasing the wrapper with streaming as well. The rest of the notebook lgtm

python/tests/test_chatcompletion_mock_wrapper.py
```diff
@@ -0,0 +1,25 @@
...
+    completions = MockRealCompletions()
+
+
+class MockRealOpenAI:
+    """
+    We have to mock the real openAI to test the actual mock because the real openai needs an API key.
+    """
+
+    some_real_field: str = "the_real_value"
+    chat = MockRealChat()
```
Ankush-lastmile, line 18: What's the use for this? I'm assuming tests be added ontop

Ankush-lastmile (approved):
lgtm for the most part, couple comments ptal

python/tests/test_chatcompletion_mock_wrapper.py
jonathanlastmileai: I had to mock the real object so I could test the real mock. lol

Otherwise the test can't run because it doesnt have the API key.

python/src/aiconfig/ChatCompletion.py
Ankush-lastmile: only place its used was in the example workbook, and the default export in root __init__.py. Its no longer used now with your v2. It should be safe to remove

cookbooks/OpenAI-ChatCompletion-AIConfigWrapper/openai_wrapper.ipynb
jonathanlastmileai: I was a little unclear why that's an important thing to demo here. Isn't it just about creating an aiconfig from existing code?

Ankush-lastmile: Yes that's correct. Its important because it showcases to the user that this wrapper works with complex code that is required for streaming.

PR #353, opened 2023-11-29
2/n [ts] tests

Ankush-lastmile:
2/n [ts] tests

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/353).
* #371
* __->__ #353
* #352

PR #371, opened 2023-11-30
3/n [py] tests (top of stack)

Ankush-lastmile:
3/n [py] tests (top of stack)

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/371).
* __->__ #371
* #353
* #352

### DIFF
--- a/python/src/aiconfig/default_parsers/openai.py
+++ b/python/src/aiconfig/default_parsers/openai.py
@@ -209,7 +209,7 @@ class OpenAIInference(ParameterizedModelParser):
                 )
 
         # Add in the latest prompt
-        add_prompt_as_message(prompt, aiconfig, completion_params["messages"], params)
+        add_prompt_as_message(prompt, aiconfig, completion_params["messages"], params, is_last_prompt=True)
         await aiconfig.callback_manager.run_callbacks(
             CallbackEvent(
                 "on_deserialize_complete", __name__, {"output": completion_params}
@@ -452,7 +452,7 @@ def refine_chat_completion_params(model_settings):
 
 
 def add_prompt_as_message(
-    prompt: Prompt, aiconfig: "AIConfigRuntime", messages: List, params=None
+    prompt: Prompt, aiconfig: "AIConfigRuntime", messages: List, params=None, is_last_prompt: bool = False
 ):
     """
     Converts a given prompt to a message and adds it to the specified messages list.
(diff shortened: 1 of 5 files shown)


python/tests/test_resolve.py
```diff
@@ -25,3 +29,39 @@ async def test_resolve_default_model_config_with_openai_parser():
...
         "top_p": 1,
     }
+
+
+@pytest.mark.asyncio
+async def test_resolve_after_run(set_temporary_env_vars):
+    """
+    After running a prompt, resolving the same prompt should return the same result as if it was never run assuming no parameters or dependencies are at play.
+
+    This test is specific to the openai model parser
```
jonathanlastmileai, line 39: hmm, why?

python/tests/test_resolve.py
```diff
@@ -25,3 +29,39 @@ async def test_resolve_default_model_config_with_openai_parser():
         "temperature": 1,
         "top_p": 1,
     }
+
+
+@pytest.mark.asyncio
+async def test_resolve_after_run(set_temporary_env_vars):
+    """
+    After running a prompt, resolving the same prompt should return the same result as if it was never run assuming no parameters or dependencies are at play.
```
jonathanlastmileai, line 37: Nice test.

python/tests/test_resolve.py
```diff
@@ -25,3 +29,39 @@ async def test_resolve_default_model_config_with_openai_parser():
...
+
+    This test is specific to the openai model parser
+    """
+    with patch.object(
+        openai.chat.completions, "create", side_effect=mock_openai_chat_completion
+    ):
```
jonathanlastmileai, line 43: I think this is OK for now, but please consider passing this down explicitly (either manually or via a DI framework). 

Zen of python:

> Explicit is better than implicit.

jonathanlastmileai (changes requested):
Back to you for now just because this is stacked on 1/n

python/tests/test_resolve.py
Ankush-lastmile: can you give me an example of what that looks like? I don't see how it would make sense to pass down the client or your inject all the way from config.run() to the end api call

saqadri (changes requested):
This isn't a good approach. Explained why in the comments. Can be much simpler. This won't work well in many situations because now `deserialize` is broken

python/src/aiconfig/default_parsers/openai.py
saqadri, line 1: This also needs to be updated for the other model parsers. 
If you go with Option 3 described below, you can call `delete_output` in the `run` function in Config.py, which would cover all model parsers.

Ankush-lastmile:
https://github.com/lastmile-ai/aiconfig/issues/490

PR #375, opened 2023-11-30
[ts] 1/n PaLM Model Parser Add Dependencies

Ankush-lastmile:
[ts] 1/n PaLM Model Parser Add Dependencies

Start of Diff stack for PaLM Text Model Parser

## What

Added Node Dependencies for the Palm API, namely:
- `@google-ai/generativelanguage": "^1.1.0",`
- `"google-auth-library": "^9.3.0"`

## Why

To use PaLM

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/375).
* #377
* #396
* #397
* #394
* #376
* __->__ #375
* #395


package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "dependencies": {
+    "@google-ai/generativelanguage": "^1.1.0",
+    "google-auth-library": "^9.3.0"
```
rossdanlm, line 4: In general what's the best practice for pinning dependencies?

rossdanlm (approved):
lgtm

package.json
Ankush-lastmile: we haven't pinned any dependencies in AIConfig besides ones that may have broken with latest. We could probably pin dependencies now that AIConfig is a bit more mature than when we were first building

PR #376, opened 2023-11-30
[ts] 2/n mp stubs

Ankush-lastmile:
[ts] 2/n mp stubs

## What

Stubbed out methods for PaLM Text Model Parser. Implement on top

## Why

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/376).
* #377
* #396
* #397
* #394
* __->__ #376
* #375
* #395


typescript/lib/parsers/palm.ts
```diff
@@ -0,0 +1,50 @@
...
+ *  Refines the completion params for the PALM text generation api. Removes any unsupported params.
+ *  The supported keys were found by looking at the PaLM text generation api. `INSERT TYPE HERE`
+ */
+function refineTextGenerationParams() {
+  throw new Error("Method not implemented.");
+}
+
+function constructOutputs() {
+  throw new Error("Method not implemented.");
+}
```
rossdanlm, line 50: nit: for future let's not add stub for generic helper functions until we use them. I get having stubs for extending the `ParameterizedModelParser` and skeleton code, but the helper here makes future diffs harder to understand where/why they were created

typescript/lib/parsers/palm.ts
```diff
@@ -0,0 +1,50 @@
...
+import { GoogleAuth } from "google-auth-library";
+import { getAPIKeyFromEnv } from "../utils";
+
+/**
+ * Model Parser for PaLM Text Generation. PaLM API Currently does not support streaming
+ * models/chat-bison-001
+ */
+
+export class PaLMTextParser extends ParameterizedModelParser {
+  _id = "models/text-bison-001";
```
rossdanlm, line 16: Nit: in future, add TODOs like `"Do not hardcode model"` so it's clear this will be addressed (or not addressed, but at least intent to address) later

Ankush-lastmile: not a TODO, PaLM api only supports two models, we'll have one parser for each

typescript/lib/parsers/palm.ts
```diff
@@ -0,0 +1,50 @@
...
+import { ParameterizedModelParser } from "../parameterizedModelParser";
+import { GoogleAuth } from "google-auth-library";
+import { getAPIKeyFromEnv } from "../utils";
+
+/**
+ * Model Parser for PaLM Text Generation. PaLM API Currently does not support streaming
+ * models/chat-bison-001
+ */
+
+export class PaLMTextParser extends ParameterizedModelParser {
```
rossdanlm, line 15: nit: `PaLMTextGenerationParser` is more descriptive, but not a big deal

Ankush-lastmile: naming convention comes from the models/api being referred to as PaLM Text and PaLM Chat

typescript/lib/parsers/palm.ts
```diff
@@ -0,0 +1,50 @@
...
+  _id = "models/text-bison-001";
+
+  public serialize(promptName: string, data: JSONObject, aiConfig: AIConfigRuntime, params?: JSONObject | undefined): Prompt | Prompt[] {
```
rossdanlm, line 18: Didn't we want only to return array type instead of singular type `Prompt[]`? We didn't fix this during Eng Week? Let's create a TODO + link to the Github Issues. Actually yea, let's do all TODOs linked to GH issues. I'll do this too

rossdanlm (approved):
lgtm but pls read comments

typescript/lib/parsers/palm.ts
Ankush-lastmile: Yes thanks for catching that, looks like I had an unpushed change, updated. Also created #416

PR #384, opened 2023-12-01
Add pyproject.toml to HuggingFaceTransformers extension

rossdanlm:
Add pyproject.toml to HuggingFaceTransformers extension

Need this so we can publish


extensions/HuggingFaceTransformers/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2023 LastMile AI
```
Ankush-lastmile, line 1: do we need to state MIT License? https://github.com/lastmile-ai/aiconfig/blob/main/LICENSE#L1

Ankush-lastmile (approved):

PR #406, opened 2023-12-05
[ez] Add prompt name to error messages

rossdanlm:
[ez] Add prompt name to error messages

Pretty simple. Should probably unify all the TS error messages, but just wanted to get this done first for now quickly. Added a TODO

### DIFF
--- a/python/src/aiconfig/Config.py
+++ b/python/src/aiconfig/Config.py
@@ -207,9 +207,7 @@ class AIConfigRuntime(AIConfig):
 
         if prompt_name not in self.prompt_index:
             raise IndexError(
-                "Prompt not found in config, available prompts are:\n {}".format(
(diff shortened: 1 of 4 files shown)


python/src/aiconfig/Config.py
```diff
@@ -207,7 +207,8 @@ async def resolve(
 
         if prompt_name not in self.prompt_index:
             raise IndexError(
-                "Prompt not found in config, available prompts are:\n {}".format(
+                "Prompt '{}' not found in config, available prompts are:\n {}".format(
```
jonathanlastmileai, line 210: nit: prefer f-strings

```suggestion
                f"Prompt '{promp_name}' not found in config, available prompts are:\n {list(self.prompt_index.keys())}"
```

jonathanlastmileai (approved):

PR #408, opened 2023-12-05
[rfc] batch execution

Ankush-lastmile:
[rfc] batch execution mvp

Request For Comments: standlone batch execution interface

### DIFF
--- a/python/src/aiconfig/batch_execution/mvp.py
+++ b/python/src/aiconfig/batch_execution/mvp.py
@@ -1,0 +1,85 @@
+from typing import Dict
+from aiconfig import AIConfigRuntime
+
+
+# Decide how to the config gets run
+run_options = {
+    "run_all": True,
+    "run last": False
+}
+
+# Decide params for a run id
+params = {
+    "1": {
+        'descriptor': "bouncy"
+    }
+}
+
+# Decide how many times to run the config.
+n = 1
+
+def input_interface(aiconfig: AIConfigRuntime, params: Dict, n: int, run_options: Dict):
+    """
+    User expected to pass in the following items:
+
+    1. aiconfig: AIConfigRuntime
+        It would make sense to pass in the loaded AIConfigRuntime rather than a file ref to an aiconfig. 
+        Reason being that there could be some intermediate setup (callbacks, registering model parsers, etc) that needs to be done before the AIConfig is ready to be used.
+
+    2. params: dict.
+        This is a nested dictionary of parameters that the user would like to pass into the AIConfig.
+        The user should be able to provide parameters for each run. They also need to specify which params will be used for each run. Similar to json/csv in AI Workflows
+
+        The structure of the dictionary is as follows:
+        {
+            run_id: {
+                param_name: param_value
+                param_name2: param_value
+            },
+            run_id+1: {
+                param_name: param_value
+                param_name2: param_value
+            },
+        }
+
+    3. n: int
+        The number of runs that the user would like to execute on the same aiconfig.
+        Example:
+        config: {
+            prompt:
+                {
+                    name: "generate haiku",
+                    input: "Write me a haiku?"
+                },
+                }
+        }
+        This config could be run 10 times to generate 10 different haikus. Dependent on a high temp and model producing different results.
+
+
+    Total number of runs = len(params) * n
+    """
+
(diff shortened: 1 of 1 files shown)


python/src/aiconfig/batch_execution/mvp.py
```diff
@@ -0,0 +1,85 @@
+from typing import Dict
```
jonathanlastmileai, line 1: nit: I learned recently learned this is deprecated (https://docs.python.org/3/library/typing.html#deprecated-aliases)

python/src/aiconfig/batch_execution/mvp.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+# Decide how to the config gets run
+run_options = {
+    "run_all": True,
+    "run last": False
+}
+
+# Decide params for a run id
+params = {
```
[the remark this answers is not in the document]
jonathanlastmileai: "parameters" is overloaded, but that might be inevitable. Both sklearn and axsweep pretty much use the same word.

python/src/aiconfig/batch_execution/mvp.py
```diff
@@ -0,0 +1,85 @@
...
+        """
+        Run all prompts in an aiconfig
+        """
+        try:
+            for prompt in aiconfig.prompts:
+                await aiconfig.run(prompt.name, params= run_params)
+
+    outputs = {}
+    # perform an execution for every run
+    for nth_execution in range(n):
```
jonathanlastmileai, line 72: Let's decouple repetitions for evaluating stochastic behavior. I think this is a bit of a can of worms and should be considered with focus, as a separate issue. Analogous to https://github.com/lastmile-ai/semantic-retrieval/issues/204

python/src/aiconfig/batch_execution/mvp.py
```diff
@@ -0,0 +1,85 @@
...
+        # Perform an execution for every set of params
+        for run_id, run_params in params.items():
+            if run_options["run_all"]:
+                # run the aiconfig with the params
+                run_all(aiconfig, run_params)
+
+                # Save config with outputs
+                aiconfig.save(json_config_filepath= f'{nth_execution}_execution_{run_id}_params',include_outputs=True)
```
jonathanlastmileai, line 80: A few things here:
1. We should structure the results like the user specified them, or at least think through how the results are going to get joined back up with the grid spec. Here, it seems we're expecting the user to remember their mapping from run id -> params and do the join themselves. Maybe that's the right answer, but it needs to be considered. My inclination would be to do that for the user to help them avoid mistaking one grid mapping from another. e.g. if I run, then change my grid spec file, then rerun, any subsequent join I do will be wrong. If we do this internally and take responsibility for the user, we need to carefully consider whether we're going to normalize like this and join back up by run_id, or just store everything as a (denormalized) view or artifact. 
2. Whatever the answer to 1, let's store any metadata in a structured way, not embedded in a file path.
3. Let's decouple serialization from inference unless we have a reason to couple them. The user might want to persist the results, but not necessarily, and it can easily be done as a subsequent step (like aiconfig.save is its own step to begin with). 
4. Whether or not we persist or just return K AIConfig instances, let's really think about whether that's convenient or even necessary. What data is the user really interested in? Will they want to open each materialized config one by one in notebook mode? If they do, will they be able to conveniently find the right config file by run_id? That seems to be what this implementation intends, but I don't yet see this as a use case. I think the clearer use case is a tabular (promptfoo-style) output.

jonathanlastmileai:
Let's scope this down to a simple loop: maybe running a fixed aiconfig on a list of inputs and getting one string output for each. 

It could generalize this to supporting more than just "the_query": https://github.com/lastmile-ai/aiconfig/blob/9669b9e5614d5b1b0603ddf3ec61556cd79bfc14/python/src/aiconfig/eval/lib.py#L196

Here's the loop that uses the above: https://github.com/lastmile-ai/aiconfig/blob/9669b9e5614d5b1b0603ddf3ec61556cd79bfc14/python/src/aiconfig/eval/lib.py#L159-L160


It actually now occurs to me that these two nominal types can be shared; not specific to eval: https://github.com/lastmile-ai/aiconfig/blob/9669b9e5614d5b1b0603ddf3ec61556cd79bfc14/python/src/aiconfig/eval/lib.py#L59-L60

Ankush-lastmile:
closed in favor of #469

PR #410, opened 2023-12-06
[hf parser] Pass in model name to text generator

rossdanlm:
[hf parser] Pass in model name to text generator

I forgot to do this before, which would default to None model (I forget which model they used). This explains why the mistral detokenizer wasn't working properly even though ChaptGPT was.

For now just storing these different generators in memory through a dict where the key is model name. No matter what model, they all follow the same text generation API so it's ok

### DIFF
--- a/extensions/HuggingFaceTransformers/python/src/aiconfig_extension_hugging_face_transformers/text_generation.py
+++ b/extensions/HuggingFaceTransformers/python/src/aiconfig_extension_hugging_face_transformers/text_generation.py
@@ -1,6 +1,6 @@
 import copy
 from typing import TYPE_CHECKING, Any, Dict, List, Optional
-from transformers import AutoTokenizer, pipeline, TextIteratorStreamer
+from transformers import AutoTokenizer, Pipeline, pipeline, TextIteratorStreamer
 import threading
 
 from aiconfig.default_parsers.parameterized_model_parser import ParameterizedModelParser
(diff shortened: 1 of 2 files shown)


rossdanlm:
| gpt2 | mistral |
|---|---|
| https://drive.google.com/file/d/1AhJ6mtVtpzyT2GTMPd2rHWqvOrRU9z6q/view?usp=sharing | https://drive.google.com/file/d/1kA84Qrx5idPi2ypuSeNDK7B76xpeNeY8/view?usp=sharing |

extensions/HuggingFaceTransformers/python/src/aiconfig_extension_hugging_face_transformers/text_generation.py
```diff
@@ -224,27 +224,31 @@ async def run_inference(
         """
         completion_data = await self.deserialize(prompt, aiconfig, options, parameters)
         completion_data["text_inputs"] = completion_data.pop("prompt", None)
+        
+        model_name : str = aiconfig.get_model_name(prompt)
+        if isinstance(model_name, str) and model_name not in self.generators:
+            self.generators[model_name] = pipeline('text-generation', model=model_name)
+        generator = self.generators[model_name]
```
Ankush-lastmile, line 231: I'm not sure how I feel about this.  Two things here. Not blocking but something to think about and if this could be done differently

1. This change means the pipeline gets loaded at execution/run time
2.  If the user has many models, all of them get tied down to this one model parser, as opposed to one per model

Ankush-lastmile (approved):
lgtm. Left a comment but not blocking

extensions/HuggingFaceTransformers/python/src/aiconfig_extension_hugging_face_transformers/text_generation.py
rossdanlm: Create a GH "issue" for us to keep track, good design flagging: https://github.com/lastmile-ai/aiconfig/issues/415

PR #432, opened 2023-12-08
[AIC-py][eval] Rejigger data structures so each metric can have a unique ID

jonathanlastmileai:
[AIC-py][eval] Rejigger data structures so each metric can have a unique ID

Each metric should have a unique ID for extra clarity (debugging etc).
The ID should be deterministic and unique for each metric, including distinguishing
between parameter values (like which substring is being checked for)

Test:

pytest, npx, notebook

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/432).
* #435
* __->__ #432


python/tests/test_eval.py
```diff
@@ -172,3 +174,138 @@ async def test_run_test_suite_with_inputs(data: st.DataObject):
...
+    return (
+        _extract_id(metric_fn=metric, metric_input=input_pair[0]),
+        _extract_id(metric_fn=metric, metric_input=input_pair[1]),
+    )
+
+
+def _extract_id_pair_for_metric_pair(
+    metric_pair: tuple[SampleEvaluationFunction[str], SampleEvaluationFunction[str]],
+    metric_input: str,
+) -> tuple[str, str]:
```
rossdanlm, line 197: Can you give example comment docstrings on how input pair vs. metric pair is different?

python/tests/test_eval.py
```diff
@@ -172,3 +174,138 @@ async def test_run_test_suite_with_inputs(data: st.DataObject):
...
+    return (
+        _extract_id(metric_fn=metric_pair[0], metric_input=metric_input),
+        _extract_id(metric_fn=metric_pair[1], metric_input=metric_input),
+    )
+
+
+@st.composite
+def _sample_string_pairs(draw: st.DrawFn) -> tuple[str, str]:
+    return draw(
```
rossdanlm, line 206: what does draw do?

python/tests/test_eval.py
```diff
@@ -172,3 +174,138 @@ async def test_run_test_suite_with_inputs(data: st.DataObject):
...
+        st.tuples(st.text(min_size=1), st.text(min_size=1)).filter(
+            lambda t: t[0] != t[1]
```
rossdanlm, line 208: What does this function do? What is the st.text fields?

python/tests/test_eval.py
jonathanlastmileai: It samples randomly from the data type you specified. Docs: https://hypothesis.readthedocs.io/en/latest/data.html#composite-strategies

python/src/aiconfig/eval/lib.py
```diff
@@ -154,7 +160,13 @@ async def _run(input_datum: str) -> Result[TextOutput, str]:
...
         )
 
-    res_outputs = await cu.result_reduce_list_all_ok_async(list(map(_run, all_inputs)))
+    # TODO: fix the race condition and then use gather
+    # https://github.com/lastmile-ai/aiconfig/issues/434
+    res_outputs_: list[Result[TextOutput, str]] = []
+    for input_datum in all_inputs:
+        res_outputs_.append(await _run(input_datum))
+    res_outputs = cu.result_reduce_list_all_ok(res_outputs_)
+    # res_outputs = await cu.result_reduce_list_all_ok_async(list(map(_run, all_inputs)))
```
rossdanlm, line 169: delete?

rossdanlm (approved):
Nice!

PR #438, opened 2023-12-08
[extensions] LLama-Guard

Ankush-lastmile:
[extensions] LLama-Guard

Model Parser Extension for LLama-Guard

Loads LLama-Guard, a 7b model into memory

### DIFF
--- a/extensions/LLama-Guard/python/LICENSE
+++ b/extensions/LLama-Guard/python/LICENSE
@@ -1,0 +1,19 @@
+Copyright (c) 2023 LastMile AI
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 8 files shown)


Ankush-lastmile:
- removed unused import from transformers -> tensorflow
- updated to 0.0.2

Could probably update readme/pyproject.toml with better links

Ankush-lastmile:
uncommented pyright config, forgot that I had commented it out

saqadri (approved):
Going to merge this and we can review as a fast follow. It works as validated in https://colab.research.google.com/drive/1CfF0Bzzkd5VETmhsniksSpekpS-LKYtX#scrollTo=qgMP6Q7Jjc-8

PR #452, opened 2023-12-11
2/n add default stream = True

Ankush-lastmile:
2/n add default stream = True

## What

default stream to true


Ankush-lastmile:
on top of #451

rossdanlm (approved):
lgtm!

Issue #401, opened 2023-12-04
Batch inference 


saqadri:
Similar to AI Workflows, we should enable local batch inference.

Let's discuss in this issue what the API should look like and what data format the results are created as.

Request from @2timesjay.

jonathanlastmileai:
Spoke offline with @Ankush-lastmile. I think it would be incredibly helpful here to be very concretely use-case driven. In the case of eval, I found it extremely helpful to start with the use case and user story (https://docs.google.com/document/d/17tjQgLDmAqyq26XJx4GqVYS73K73GykvhywEiFCozps/edit#bookmark=id.5uogvi8dxrv4).

Proposal for a use case to zoom in on: 

I'm building a haiku-generating AIConfig and I have some example haiku types and some idea of how to evaluate the output quality. (This is an eval use case, but batch inference is a key implementation component). I want to explore not only the different outputs I get for different Haiku types, but how the output quality changes depending on the AIConfig itself. e.g. Llama vs. GPT4.

Ankush-lastmile:
Created a draft pr #404 rfc .

From my understanding of Jacob's conversation, the primary use case for batch execution can be viewed as an exploratory use case, wherein the user is privy to both inputs and outputs. This suggests that batch execution is closely tied to an evaluation or an eval interface.


+1 the specific use case from @2timesjay would be great to work with.


Considering a suitable interface that provides users with access to batch execution for AIConfig, here's my proposal:

### Input Interface: 
Users can provide 4 inputs
1. The AIConfig
2. Parameters
3. The number of executions for the same AIConfig
4. Execution options, essentially a run-config that specifies how to execute the AIConfig. This can be options such as the following:
- `run all` - run all prompts
-  `run last` - run the last prompt

### Output Interface
A key question here is determining who or what will consume the output? 

Proposal:
1. We can persist the outputs for each execution using the config.save(save_with_outputs=True) method. This separates execution and evaluation as each "execution" is saved as an AIConfig artifact.

Another alternative is to return the data in memory that includes the outputs corresponding to each execution. This would probably conform to a dictionary mapping. Something like this

```json
{
     run_1: { 
          prompt: "prompt1"
          output: "output"
     }
}
```

jonathanlastmileai:
@Ankush-lastmile , I'm glad we're use-case driven and totally agree with this:

[...]

PR #469, opened 2023-12-12
[py] Batch Execution Interface

Ankush-lastmile:
[py] Batch Execution Interface

## What

Starting point to create a batch execution interface for AIConfig. This diff creates the interface in python. We can implement the typescript equivalent on top.

Methods Implemented:
- `config.py` (AIConfigRuntime)
- - run_batch() calls `run_batch` of the model parser
- `model_parser.py` (abstract model parser class)
- - run_batch()

## Usage
User will need to construct a list of params for batch, pass it into the `run_batch` method. Very similar to the `run()` method.

### Output type
The resulting object of a Batch Execute will contain the Execute Result, and the resolved completion params with paramters. It would look smoething like this:

```
[
  (
    ExecuteResult(), aiconfig.resolve()
   ),
  (
    ExecuteResult(), aiconfig.resolve()
   )
]
```

Check out the implementation for more details, an example is provided in the docstring

Discussed interface design with @saqadri. We can iterate on top of this but this should get us started.

### Note:

This diff does not implement concurrent request batching. One things would be required on top of this diff for a robust batch exec to be implemented in the model parsers.
1. Exponential Backoff inside model parser implementations at request time. For ex: [OpenAI Backoff impl](https://platform.openai.com/docs/guides/rate-limits/error-mitigation?context=tier-free#:~:text=Example%203%3A-,Manual%20backoff%20implementation,-If%20you%20don%27t)

## Test Plan

Created a config which has a parameterized dependency chain. Batch executed on a list of params

Draft PR containing test plan: [#472](https://github.com/lastmile-ai/aiconfig/pull/472)

https://github.com/lastmile-ai/aiconfig/assets/141073967/e071361f-637f-442b-86c5-edec68b5eba7

This is for issue [#401](https://github.com/lastmile-ai/aiconfig/issues/401)

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/469).
* #472
* __->__ #469


python/src/aiconfig/model_parser.py
```diff
@@ -78,6 +79,40 @@ async def run(
...
+    ) -> list["AIConfigRuntime"]:
+        """
+        Run inference on multiple sets of parameters, one set at a time concurrently.
+        Default implementation for the run_batch method. Model Parsers may choose to override this method if they need to implement custom batch execution logic.
+        For each dictionary of parameters in `params_list``, the `run` method is invoked. All iterations are separate as we use a deep copy of `aiconfig` in each iteration.
+
+        Args:
+            prompt (Prompt): The prompt for running the inference
+            aiconfig (AIConfigRuntime): The AIConfig object containing all necessary configurations (prompts and parameters) for running the inference.
+            params_list (list[dict[str, Any]]): A List of dictionaries, where each dictionary is a set of parameters that directly influence the behaviour of inference.
```
saqadri, line 98: parameters_list

python/src/aiconfig/model_parser.py
```diff
@@ -78,6 +79,40 @@ async def run(
...
+        """
+        # Intiailize an empty list to hold the results
+        inference_results = []
+
+        # Create a deep copy of the aiconfig object to prevent mutations that could affect other iterations
+        aiconfig_deep_copy = copy.deepcopy(aiconfig)
+
+        for params in parameters_list:
+            await self.run(prompt, aiconfig_deep_copy, options, params, **kwargs)
+            inference_results.append(aiconfig_deep_copy)
```
saqadri, line 113: I think we need to persist the parameters that the prompt was run with into the aiconfig, right?

python/src/aiconfig/model_parser.py
```diff
@@ -78,6 +79,40 @@ async def run(
...
+            list[AIConfigRuntime]: A list of AIConfigRuntime objects. Each object contains the state of the AIConfigRuntime after each run using the corresponding parameter set from params_list.
+        """
+        # Intiailize an empty list to hold the results
+        inference_results = []
+
+        # Create a deep copy of the aiconfig object to prevent mutations that could affect other iterations
+        aiconfig_deep_copy = copy.deepcopy(aiconfig)
+
+        for params in parameters_list:
+            await self.run(prompt, aiconfig_deep_copy, options, params, **kwargs)
```
saqadri, line 112: Why is this done serially? Shouldn't we run this in parallel across all of them?

python/src/aiconfig/model_parser.py
```diff
@@ -78,6 +79,40 @@ async def run(
...
+        Returns:
+            list[AIConfigRuntime]: A list of AIConfigRuntime objects. Each object contains the state of the AIConfigRuntime after each run using the corresponding parameter set from params_list.
+        """
+        # Intiailize an empty list to hold the results
+        inference_results = []
+
+        # Create a deep copy of the aiconfig object to prevent mutations that could affect other iterations
+        aiconfig_deep_copy = copy.deepcopy(aiconfig)
+
+        for params in parameters_list:
```
saqadri, line 111: We also need exponential back off... can come in a diff above but it's critical if run() fails.

python/src/aiconfig/Config.py
```diff
@@ -274,6 +274,74 @@ async def run(
...
+            self,
+            parameters_list,
+            options,
+            callback_manager=self.callback_manager,
+            **kwargs,
+        )
+
+        # Extract the prompts from the result
+        prompts = []
+        for aiconfig in batch_results:
```
saqadri, line 337: Same problem as other function that we have no idea which parameters were applied to get this particular aiconfig.

Alternatively you could do prompt.resolve with the parameters  and return the fully resolved prompt back.

saqadri (changes requested):
Some comments

Ankush-lastmile:
Added async execution. Although the current model parsers all use synchronous api calls. Something to keep in mind and update forward

Ankush-lastmile:
>we have no idea which parameters were applied to get this particular aiconfig.

Updated output of `batch_execute()` to be a list of Execution Results, along with the corresponding completion params.

python/src/aiconfig/Config.py
```diff
@@ -274,6 +265,77 @@ async def run(
         await self.callback_manager.run_callbacks(event)
         return response
 
+    async def run_batch(
+        self,
+        prompt_name: str,
+        parameters_list: list[dict[str, Any]],
+        options: Optional[InferenceOptions] = None,
+        **kwargs,
+    ) -> list[Tuple["ExecuteResult", JSONObject | Any]]:
```
saqadri, line 274: Just return the corresponding params for that run in the Tuple as well.
So Tuple["ExecuteResult", JSONObject | Any, Dict]

saqadri (approved):
Looks good, just one comment, and then ship it

Ankush-lastmile:
> So Tuple["ExecuteResult", JSONObject | Any, Dict]

Updated to Tuple["ExecuteResult", JSONObject | Any, Dict]

looks like some formatting changes made it into the commit .

saqadri (approved):
Ship it!

saqadri (approved):
Ship it!

saqadri (approved):
Ship it!

saqadri:
lol no idea why it pasted my comment 3 times

PR #473, opened 2023-12-12
Add attachments field to PromptInput

rholinshead:
# Add attachments field to PromptInput

"Commandeering" https://github.com/lastmile-ai/aiconfig/pull/360 here (not sure if it's possible to update the existing PR) and addressing the feedback + reflecting changes in the JSON schema itself.

This adds the `attachments` field to to PromptInput objects to support attaching non-text data of different MIME types to prompt inputs. Note that this was technically already supported by the generic input data, but we should instead use attachments as the de facto way to do this.

### Typescript:
- `yarn test` for typescript tests
```
ryanholinshead@Ryans-MacBook-Pro typescript % yarn test
yarn run v1.22.19
$ jest --runInBand
 PASS  __tests__/config.test.ts
  Loading an AIConfig
    ✓ loading a basic chatgpt query config (8 ms)
    ✓ loading a prompt chain (4 ms)
    ✓ deserialize and re-serialize a prompt chain (2 ms)
    ✓ serialize a prompt chain with different settings (1 ms)

 PASS  __tests__/parsers/palm-text/palm.test.ts
  PaLM Text ModelParser
    ✓ serializing params to config prompt (5 ms)
    ✓ deserializing params to config (2 ms)
    ✓ run prompt, non-streaming (77 ms)

 PASS  __tests__/parsers/hf/hf.test.ts
  HuggingFaceTextGeneration ModelParser
    ✓ uses HuggingFace API token from environment variable if it exists (6 ms)
    ✓ serializing params to config prompt (1 ms)
    ✓ serialize callbacks (1 ms)
    ✓ deserializing config prompt to params (1 ms)
    ✓ deserialize callbacks (2 ms)
    ✓ run prompt, non-streaming (4 ms)
    ✓ run prompt, streaming (1 ms)
    ✓ run callbacks (3 ms)

 PASS  __tests__/testProgramaticallyCreateConfig.ts
  test Get Global Settings
    ✓ Retrieving global setting from AIConfig with 1 model
  ExtractOverrideSettings function
    ✓ Should return initial settings when no global settings are defined
    ✓ Should return an override when initial settings differ from global settings (1 ms)
    ✓ Should return empty override when global settings match initial settings
    ✓ Should return empty override when Global settings defined and initial settings are empty

Test Suites: 4 passed, 4 total
Tests:       20 passed, 20 total
Snapshots:   0 total
Time:        1.995 s, estimated 2 s
Ran all test suites.
```
- Ran demo scripts successfully (`npx ts-node demo/demo.ts`, `npx ts-node demo/function-call-stream.ts`)

### Python:
- pytest, no failures
```
========================================================================= 85 passed, 34 warnings in 7.94s ==========================================================================
(aiconfig) ryanholinshead@Ryans-MacBook-Pro python % 
```

### DIFF
--- a/typescript/demo/demo.ts
+++ b/typescript/demo/demo.ts
@@ -70,20 +70,18 @@ async function openAIWithAIConfig() {
   //     console.log(aiConfig.getOutputText("demoPrompt", output));
   //   }
 
-  const inferenceOptions: InferenceOptions = {
-    callbacks: {
-      streamCallback: (data: any, _accumulatedData: any, _index: any) => {
-        process.stdout.write(data?.content || "\n");
-      },
-    },
-  };
-
   const streamCallback = (data: any, _accumulatedData: any, _index: any) => {
     process.stdout.write(data?.content || "\n");
   };
 
+  const inferenceOptions: InferenceOptions = {
+    callbacks: {
+      streamCallback,
+    },
+  };
+
   // Streaming:
-  let result2 = await aiConfig.run(
+  await aiConfig.run(
     "demoPrompt",
     /*params*/ {
       name: "Streaming Demo",
--- a/python/src/aiconfig/schema.py
+++ b/python/src/aiconfig/schema.py
@@ -60,8 +60,25 @@ class PromptMetadata(BaseModel):
         extra = "allow"
 
 
+class Attachment(BaseModel):
+    """
+    Attachment used to pass data in PromptInput for non-text inputs (ex: image, audio)
+    """
+
+    # The data representing the attachment
+    data: Any
+    # The MIME type of the result. If not specified, the MIME type will be assumed to be text/plain
+    mime_type: Optional[str] = None
+    # Output metadata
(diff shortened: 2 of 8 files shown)


typescript/demo/demo.ts
```diff
@@ -70,20 +70,18 @@ async function openAIWithAIConfig() {
   //     console.log(aiConfig.getOutputText("demoPrompt", output));
   //   }
 
+  const streamCallback = (data: any, _accumulatedData: any, _index: any) => {
```
rholinshead, line 73: Minor cleanup:
- use defined `streamCallback` here, since it was just an unused variable before
- remove unnecessary `let result2` below

Technically should be a separate diff but I had to test this script out and easier to just include these small things here

saqadri (approved):
Looks awesome!

rholinshead:
Going to land this and I'll look into publishing tomorrow morning (thanks Rossdan for the detailed publish instructions)

PR #474, opened 2023-12-12
Remove package-lock.json

rholinshead:
# Remove package-lock.json

Just realized this file snuck in from https://github.com/lastmile-ai/aiconfig/pull/473, which had it from https://github.com/lastmile-ai/aiconfig/pull/360

We should be using `yarn` only, so only `yarn.lock` should ever be produced/committed.


rholinshead:
cc @rossdanlm for visibility, in case you're also running `npm` commands -- should only use `yarn` since they aren't entirely compatible

saqadri (approved):

PR #475, opened 2023-12-13
{WIP} Exponential Backoff w/openai model parser

Ankush-lastmile:
{WIP} Exponential Backoff w/openai model parser

wip exponential backoff with the openai model parser.

- Create a wrapper around sending the api request
- use as decorator to completion request

### DIFF
--- a/python/src/aiconfig/util/parser_utils.py
+++ b/python/src/aiconfig/util/parser_utils.py
@@ -1,0 +1,48 @@
+# imports
+import random
+import time
+ 
+import openai
+# define a retry decorator
+def retry_with_exponential_backoff(
+    func,
+    initial_delay: float = 1,
+    exponential_base: float = 2,
+    jitter: bool = True,
+    max_retries: int = 10,
+    errors: tuple = (openai.RateLimitError,),
+):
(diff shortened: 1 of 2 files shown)


python/src/aiconfig/util/parser_utils.py
```diff
@@ -0,0 +1,48 @@
+# imports
+import random
+import time
+ 
+import openai
+# define a retry decorator
+def retry_with_exponential_backoff(
```
jonathanlastmileai, line 7: Please unify with https://github.com/lastmile-ai/lastmile-utils/blob/7e5a977cdb3facbe2c76b0608cf3074144c5af36/python/src/lastmile_utils/lib/core/utils.py#L186 in lastmile-utils so we can use it anywhere. 

The signatures can be combined to get the best of both worlds:
- initial_delay, exponential_base, jitter, max_retries, errors
- logger
- strict pyright type check
- [proposal] thunk only (to force user to decouple the function inputs from the error handling)

jonathanlastmileai (changes requested):

Ankush-lastmile:
closing this for now. Keeping track with #596

Issue #463, opened 2023-12-12
[model integration] add support for gemini models 12/13


tanya-rai:

saqadri:
It's shipped, but needs support for vision model

Ankush-lastmile:
Currently Gemini Model Parser doesn't implement the `serialize()` method. This is a high priority.

Next steps is incorporating MultiModality for Gemini-pro-vision

Ankush-lastmile:
Keeping this issue open. Couple of [non priority items remain](https://github.com/lastmile-ai/aiconfig/blob/main/extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py#L22) with Gemini, such as Multimodal.

PR #478, opened 2023-12-13
[extensions][py] Gemini

Ankush-lastmile:
[extensions][py] Gemini

[extensions][py] Gemini

## What

Gemini Model Parser Extension

Supports Gemini Models.

Gemini api seems to only allow candidate of 1. This means number of outputs is always 1.

Gemini models are:
Gemini Pro
Gemini pro Visio (similar to gpt -4v)

This model parser supports Gemini Pro T2T functionality

## TestPlan:

Cookbook: https://github.com/lastmile-ai/aiconfig/tree/main/cookbooks/Gemini

## Issue

#463

### DIFF
--- a/extensions/Gemini/python/README.md
+++ b/extensions/Gemini/python/README.md
@@ -1,0 +1,8 @@
+# Using Gemini with AIConfig
+
+We are prioritizing the Python library Usage
+
+1. Install `pip install aiconfig-extension-gemini`
+2. Import the extension `import aiconfig_extension_gemini`
+
+And you're all good to go! This extension automatically updates the model parser registry with the model "gemini-pro". Check out the Gemini Cookbook in the aiconfig repo for a full example.
--- a/extensions/Gemini/python/pyproject.toml
+++ b/extensions/Gemini/python/pyproject.toml
@@ -1,0 +1,45 @@
+[build-system]
+requires = ["setuptools", "wheel"]
+
+[project]
+name = "aiconfig_extension_gemini"
+version = "0.0.4"
+authors = [
+  { name="LastMile AI" },
+  { name="Ankush Pala", email="ankush@lastmileai.dev" },
+]
+description = "An extension for using Gemini with aiconfig"
+readme = "README.md"
+requires-python = ">=3.10"
+classifiers = [
+    "Programming Language :: Python :: 3",
+    "License :: OSI Approved :: MIT License",
+    "Operating System :: OS Independent",
+]
+dynamic = ["dependencies"]
+
+[tool.setuptools.dynamic]
+dependencies = {file = ["requirements.txt"]}
+
+[project.urls]
+Homepage = "https://github.com/lastmile-ai/aiconfig/"
+Issues = "https://github.com/lastmile-ai/aiconfig/issues"
+
+# Black formatting
+[tool.black]
+line-length = 99
+include = '\.pyi?$'
+exclude = '''
+/(
+      .eggs         # exclude a few common directories in the
+    | .git          # root of the project
+    | .hg
+    | .mypy_cache
+    | .tox
+    | venv
+    | _build
+    | buck-out
+    | build
+    | dist
+  )/
+'''
--- a/extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
+++ b/extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
@@ -1,0 +1,295 @@
+# Define a Model Parser for LLama-Guard
+from typing import TYPE_CHECKING, Dict, List, Optional, Any
+
+import google.generativeai as genai
+
+
+from aiconfig.default_parsers.parameterized_model_parser import ParameterizedModelParser
+from aiconfig.model_parser import InferenceOptions
+from aiconfig.schema import ExecuteResult, Output, Prompt
+from aiconfig.util.params import resolve_prompt
+from aiconfig import CallbackEvent, get_api_key_from_environment, AIConfigRuntime
+
+
+# Circuluar Dependency Type Hints
+if TYPE_CHECKING:
+    from google.generativeai.types import AsyncGenerateContentResponse
+
+def construct_regular_outputs(response: "AsyncGenerateContentResponse") -> list[Output]:
+    """
+    Construct regular output per response result, without streaming enabled
+    """
+    output_list = []
+    for i, candidate in enumerate(response.candidates):
+        output = ExecuteResult(
+            **{
+                "output_type": "execute_result",
+                "data": candidate.content.parts[0].text,
+                "execution_count": i,
+                "metadata": response.prompt_feedback.__dict__,
+            }
+        )
+
+        output_list.append(output)
+
+    return output_list
+
+
+async def construct_stream_outputs(
+    response: "AsyncGenerateContentResponse", options: InferenceOptions
+) -> list[Output]:
+    """
+    Construct Outputs while also streaming the response with stream callback
+
+    Gemini api seems to not support more than one candidate, so we'll just use the first one
+    """
+    output_list = []
+
+    acc = ""
+
+    async for chunk in response:
+        # deconstruct chunk
+
+        data = chunk.parts[0].text
+        acc += data
+        if options is not None and options.stream_callback is not None:
+            options.stream_callback(data, acc, 0)
+
+    output = ExecuteResult(
+        **{
+            "output_type": "execute_result",
+            "data": response.candidates[0].content.parts[0].text,
+            "execution_count": 0,  # Hard coded for now. If api supports multiple candidates this can be updated.
+            "metadata": response.prompt_feedback.__dict__,
+        }
+    )
+
+    output_list.append(output)
+
+    return output_list
+
+
+class GeminiModelParser(ParameterizedModelParser):
+    def __init__(self, id: str = "gemini-pro"):
+        super().__init__()
+        self.model = genai.GenerativeModel(id)
+
+    def id(self) -> str:
+        """
+        Returns an identifier for the model (e.g. llama-2, gpt-4, etc.).
+        """
+        return "GeminiModelParser"
+
+    async def serialize(
+        self,
+        prompt_name: str,
+        data: Dict,
+        ai_config: "AIConfigRuntime",
+        parameters: Optional[Dict] = None,
+        **kwargs,
+    ) -> List[Prompt]:
+        """
+        Defines how a prompt and model inference settings get serialized in the .aiconfig.
+
+        Args:
+            prompt (str): The prompt to be serialized.
+            inference_settings (dict): Model-specific inference settings to be serialized.
+
+        Returns:
+            str: Serialized representation of the prompt and inference settings.
+        """
+
+        event = CallbackEvent(
+            "on_serialize_start",
+            __name__,
+            {
+                "prompt_name": prompt_name,
+                "data": data,
+                "parameters": parameters,
+                "kwargs": kwargs,
+            },
+        )
+        await ai_config.callback_manager.run_callbacks(event)
+
+        raise Exception("Not Implemented")
+
+        event = CallbackEvent("on_serialize_complete", __name__, {"result": prompts})
+        await ai_config.callback_manager.run_callbacks(event)
+
+    async def deserialize(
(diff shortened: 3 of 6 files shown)


extensions/Gemini/python/README.md
```diff
@@ -0,0 +1,15 @@
+# Using LLama-Guarde for model parsing with AIConfig
```
saqadri, line 1: Update

saqadri:
Since PaLM is already part of the the base aiconfig, should we add Gemini directly to aiconfig core? Just a thought, not tied to the idea

Ankush-lastmile:
> Since PaLM is already part of the the base aiconfig, should we add Gemini directly to aiconfig core? Just a thought, not tied to the idea

We'll do this in the future, gemini uses the same library as PaLM

Ankush-lastmile:
One more type error on colab, removed the type in attempt to get the google-generative ai library working on colab

extensions/Gemini/python/README.md
```diff
@@ -0,0 +1,8 @@
+# Using Gemini with AIConfig
+
+We are prioritizing the Python library Usage
+
+1. Install `pip install aiconfig-extensions-gemini`
```
rholinshead, line 5:
```suggestion
1. Install `pip install aiconfig-extension-gemini`
```

extensions/Gemini/python/README.md
```diff
@@ -0,0 +1,8 @@
...
+2. Import the extension `import aiconfig_extension_gemini`
+
+And you're all good to go! This extension automatically updates the model parser registry with the model "Gemini-pro". Check out the Gemini Cookbook in the aiconfig repo for a full example.
```
rholinshead, line 8: We should make sure we're consistent with the model name between this and workbooks, for when we download as aiconfig, right? cc @saqadri 

Also, should this be `gemini-pro` (lowercase)?

extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
```diff
@@ -0,0 +1,295 @@
...
+    Construct regular output per response result, without streaming enabled
+    """
+    output_list = []
+    for i, candidate in enumerate(response.candidates):
+        output = ExecuteResult(
+            **{
+                "output_type": "execute_result",
+                "data": candidate.content.parts[0].text,
+                "execution_count": i,
+                "metadata": response.prompt_feedback.__dict__,
```
rholinshead, line 29: Where is prompt_feedback defined for Gemini? I've been looking at https://cloud.google.com/vertex-ai/docs/generative-ai/model-reference/gemini#gemini-pro for the req/res but does `genai` sdk add this data?

extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
```diff
@@ -0,0 +1,295 @@
...
+            {
+                "prompt_name": prompt_name,
+                "data": data,
+                "parameters": parameters,
+                "kwargs": kwargs,
+            },
+        )
+        await ai_config.callback_manager.run_callbacks(event)
+
+        raise Exception("Not Implemented")
```
rholinshead, line 114:
```suggestion

```

Should be removed and implemented

extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
```diff
@@ -0,0 +1,295 @@
...
+
+        # TODO: check and handle api key here
+        completion_data = await self.deserialize(prompt, aiconfig, parameters)
+
+        stream = True  # Default value
+        if options is not None and options.stream is not None:
+            stream = options.stream
+        elif "stream" in completion_data:
+            stream = completion_data["stream"]
+        completion_data["stream"] = stream
```
rholinshead, line 194: Is this actually a supported param?

extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
```diff
@@ -0,0 +1,295 @@
...
+
+        if not output:
+            return ""
+
+        if output.output_type == "execute_result":
+            output_data = output.data
+            if isinstance(output_data, str):
+                return output_data
+            else:
+                raise Exception("Not Implemented")
```
rholinshead, line 227: Shouldn't happen, right?

extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
```diff
@@ -0,0 +1,295 @@
...
+                    previous_prompt_template = resolve_prompt(
+                        previous_prompt, params, aiconfig
+                    )
+                    previous_prompt_output = aiconfig.get_latest_output(previous_prompt)
+                    previous_prompt_output_text = self.get_output_text(
+                        previous_prompt, aiconfig, previous_prompt_output
+                    )
+
+                    messages.append(
+                        {"role": "user", "parts": [{"text": previous_prompt_template}]}
```
rholinshead, line 266: Just double-checking lowercase role works instead of USER and MODEL ?

extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
```diff
@@ -0,0 +1,295 @@
...
+                            "role": "model",
+                            "parts": [{"text": previous_prompt_output_text}],
+                        }
+                    )
+
+        return messages
+
+
+def refine_chat_completion_params(model_settings):
+    # completion parameters to be used for Palm's chat completion api
```
rholinshead, line 279:
```suggestion
    # completion parameters to be used for Gemini's chat completion api
```

extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
```diff
@@ -0,0 +1,295 @@
...
+    # messages handled seperately
+    supported_keys = {
+        "contents",
+        "generation_config",
+        "safety_settings",
+        "stream",
```
rholinshead, line 285: 'tools' can be added later on as well. For generation_config, do we want that to be sub-keyed under model_settings, e.g.
```
model.settings = { generation_config: {topK: ..., etc.} }
```

Or serialized top-level in the settings:
```
model.settings = {topK: ...}
```

The web app is doing the latter I think (cc @saqadri ).

Do we need to match camelCase of https://cloud.google.com/vertex-ai/docs/generative-ai/model-reference/gemini#gemini-pro or does the api here support this snake case?

rholinshead:
For the test plan, can you link this local package and test in the cookbook?

extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
Ankush-lastmile: The Python sdk Docs were pretty lacking when building this, it was pretty inconsistent throughout their api docs and some of their guides didn't work. Also note that I mainily used their public api docs (not through vertex). Link: https://ai.google.dev/docs

The response object had a couple of attributes but it didn't make sense how to use all of them. Printing the object didn't give me any information either so I inspected it by doing `print(response.__dict__)` which shows me the object fields. Accessing fields like 

Screenshot 2023-12-14 at 5 22 55 PM

extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
Ankush-lastmile: Yep, there were 4 main attributes: 

Screenshot 2023-12-14 at 5 26 46 PM

extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
Ankush-lastmile: Not expected, but I hadn't thoroughly understood the expected result types. I put this in as a just in case

extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
Ankush-lastmile: Not sure if it matters, the palm equivalent was `author: 0`, `author: 1` but the api/model only checked that it was alternating. Fairly certain that the Gemini is doing something similar.

the lowercase user and model worked for this

extensions/Gemini/python/pyproject.toml
```diff
@@ -0,0 +1,45 @@
...
+[tool.setuptools.dynamic]
+dependencies = {file = ["requirements.txt"]}
+
+[project.urls]
+Homepage = "https://github.com/lastmile-ai/aiconfig/"
+Issues = "https://github.com/lastmile-ai/aiconfig/issues"
+
+# Black formatting
+[tool.black]
+line-length = 99
```
jonathanlastmileai, line 30: I don't think we need this section in these toml files if we're using vscode settings?

extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
jonathanlastmileai: Seems like a `ValueError`

extensions/Gemini/python/src/aiconfig_extension_gemini/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .Gemini import GeminiModelParser
+
+__ALL__ = ["GeminiModelParser"]
```
jonathanlastmileai, line 3: I think this is supposed to be lowercase
```suggestion
__all__ = ["GeminiModelParser"]
```

extensions/Gemini/python/README.md
Ankush-lastmile: This could also be updated in the adapter logic if its not consistent. There are already a couple of differences in aiconfig vs workbooks right now.

Ankush-lastmile:
Addressed Nits

extensions/Gemini/python/README.md
saqadri: It should be lowercased here too

extensions/Gemini/python/pyproject.toml
```diff
@@ -0,0 +1,45 @@
+[build-system]
+requires = ["setuptools", "wheel"]
+
+[project]
+name = "aiconfig_extension_gemini"
```
saqadri, line 5: It should have dashes instead of underscores, right? That's how the other extensions are, let's be consistent on names

extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
saqadri: @Ankush-lastmile make sure this is tracked and you don't close the GH issue until this is done.

extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
saqadri: Agreed let's update the exception at least

extensions/Gemini/python/pyproject.toml
Ankush-lastmile: can you confirm this is the case? What's the difference otherwise?

Ankush-lastmile:
- Updated `get_output_text` to throw ValueError instead of base Exception
- updated init.py `__ALL__` -> `__all__`

rebase to main again (ontop of fix that un-breaks current tests)

extensions/Gemini/python/pyproject.toml
Ankush-lastmile: the pyproject.toml can not contain dashes, underscores are required here. This will break the package build system. The import statement also uses underscores
ex:
```python
import aiconfig_extension_gemini
```

Where as the pip install needs dashes
```
pip install aiconfig-extension-gemini
```

extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
Ankush-lastmile: Left a comment on #463. This is my next high-pri focus

extensions/Gemini/python/pyproject.toml
saqadri: Ah, understood. Thanks for the clarification

saqadri (approved):
Approving to unblock.

PR #479, opened 2023-12-13
Bump pypi package version to 1.1.6 to support Prompt Input Attachments

rholinshead:
Bump pypi package version to 1.1.6 to support Prompt Input Attachments

With the Attachments added to the config schema, bumping the package version to 1.1.6 prior to publishing to pypi.

Tested successfully on the Getting Started and Function Calling cookbooks (moved python function calling one to python dir)

### DIFF
--- a/cookbooks/Getting-Started/getting_started.ipynb
+++ b/cookbooks/Getting-Started/getting_started.ipynb
@@ -1,6 +1,7 @@
 {
   "cells": [
     {
+      "attachments": {},
       "cell_type": "markdown",
       "metadata": {
         "id": "kNmnUZnUFzt3"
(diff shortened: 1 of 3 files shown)


cookbooks/Getting-Started/getting_started.ipynb
```diff
@@ -29,6 +31,7 @@
       ]
     },
     {
+      "attachments": {},
```
saqadri, line 34: Why is this updated? Maybe just the notebook editor?

saqadri (approved):

Ankush-lastmile:
>Why is this updated? Maybe just the notebook editor?

Yep looks like it

Screenshot 2023-12-13 at 1 30 45 PM

cookbooks/Getting-Started/getting_started.ipynb
rholinshead: Ya, I changed the comment on setting .env variable and saved and it auto-added attachments.

Same for when I executed the function calling one it set language_info to some weird metadata (will remove with vim)

PR #486, opened 2023-12-14
Prioritize python example code tab in docs

rholinshead:
# Prioritize python example code tab in docs

Make sure all example code / tabs prioritizes showing python since that's the most common language for AIConfig usage. The order is actually determined by the order in the `aiConfigLanguages` array in tab constants, but this PR also updates the tabs defined in the md code to reflect the intended ordering.

## Testing:
- Went through each page and each tab toggle to ensure correctness
e.g.

https://github.com/lastmile-ai/aiconfig/assets/5060851/b852005d-2620-414e-a226-7a885cc95622

### DIFF
--- a/aiconfig-docs/core/tabConstants.js
+++ b/aiconfig-docs/core/tabConstants.js
@@ -29,8 +29,8 @@ const pythonPackageManagers = [
 const defaultPythonPackageManager = "pip";
 
 const aiConfigLanguages = [
-  { label: "TypeScript", value: "node" },
   { label: "Python", value: "python" },
+  { label: "TypeScript", value: "node" },
 ];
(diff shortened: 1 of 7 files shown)


aiconfig-docs/core/tabConstants.js
```diff
@@ -29,8 +29,8 @@ const pythonPackageManagers = [
 const defaultPythonPackageManager = "pip";
 
 const aiConfigLanguages = [
```
rholinshead, line 31: Note the ordering here is actually what defines the tab ordering. I've just updated the tab ordering in code for clarity

saqadri (approved):

PR #491, opened 2023-12-14
[py][ts] Clear outputs before running

Ankush-lastmile:
[py][ts] Clear outputs before running

## What

Remove the outputs of the prompt when run() is called

## Why

previous outputs stay cached. This introduces unintended side effects

## Testplan

Run a prompt twice, expect the same output without any changes. This is the result of the output no longer being cached on a new run

TS:
Screenshot 2023-12-14 at 3 06 17 PM
Py:
Added print statements to verify outputs were getting removed
Screenshot 2023-12-14 at 3 07 47 PM

### DIFF
--- a/python/src/aiconfig/Config.py
+++ b/python/src/aiconfig/Config.py
@@ -261,6 +261,9 @@ class AIConfigRuntime(AIConfig):
         model_name = self.get_model_name(prompt_data)
         model_provider = AIConfigRuntime.get_model_parser(model_name)
 
+        # Clear previous run outputs if they exist
+        self.delete_output(prompt_name)
+
         response = await model_provider.run(
             prompt_data,
             self,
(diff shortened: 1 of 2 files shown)


saqadri:
Add a test plan for how this change was validated+

python/src/aiconfig/Config.py
```diff
@@ -261,6 +261,9 @@ async def run(
         model_name = self.get_model_name(prompt_data)
         model_provider = AIConfigRuntime.get_model_parser(model_name)
 
+        # Clear previous run outputs if they exist
+        self.delete_output(prompt_name)
```
saqadri, line 265: Does this remove the output from the config?

saqadri (approved):

python/src/aiconfig/Config.py
Ankush-lastmile: Yes `delete_output` deletes the outputs and returns them. Typescript does the same with `deleteOutput`
py: https://github.com/lastmile-ai/aiconfig/blob/main/python/src/aiconfig/schema.py#L530

PR #505, opened 2023-12-15
[AIC-py][eval] robust function calls

jonathanlastmileai:
[AIC-py][eval] robust function calls

main thing: call `run_thunk_safe()` for eval functions
- async, catch exceptions, timeout
- make a few important downstream changes, e.g. ID for logging/debugging

### DIFF
--- a/python/src/aiconfig/eval/common.py
+++ b/python/src/aiconfig/eval/common.py
@@ -26,7 +26,7 @@ MetricValue = int | float | str | bool | CustomMetricValue
 
 class EvaluationFunction(Protocol, Generic[T_OutputDatum]):
     @abstractmethod
-    def __call__(self, output_datum: T_OutputDatum) -> MetricValue:
+    async def __call__(self, output_datum: T_OutputDatum) -> MetricValue:
         pass
 
 
@@ -70,9 +70,15 @@ class EvaluationMetricMetadata(cu.Record, Generic[T_OutputDatum]):
     # e.g. {"substring": "hello", "case_sensitive": False}
     extra_metadata: dict[str, Any] = {}
 
+    def __repr__(self) -> str:
+        fields = self.__dict__
+        fields["id"] = self.id
+        s_json = json.dumps(fields, indent=2)
+        return f"EvaluationMetricMetadata({s_json})"
+
 
 class SampleMetricValue(cu.Record, Generic[T_OutputDatum]):
-    value: MetricValue
+    value: MetricValue | None
     metric_metadata: EvaluationMetricMetadata[T_OutputDatum]
 
     @root_validator(pre=True)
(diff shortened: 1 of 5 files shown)


python/src/aiconfig/eval/common.py
```diff
@@ -54,25 +54,25 @@ class EvaluationMetricMetadata(cu.Record, Generic[T_OutputDatum]):
...
     description: str
     best_value: MetricValue | None = None
     worst_value: MetricValue | None = None
     # e.g. {"substring": "hello", "case_sensitive": False}
     extra_metadata: dict[str, Any] = {}
+    id: str = "UNSET"
+
+    @model_validator(mode="before")
+    def set_id(cls, values: dict[str, Any]) -> dict[str, Any]:
+        values["id"] = cu.hash_id(f"{values['name']}{values['description']}{values['best_value']}{values['worst_value']}".encode("utf-8"))
```
rholinshead, line 67: Won't this overwrite the `id` that the user has requested to be set?

python/src/aiconfig/eval/common.py
```diff
@@ -54,25 +54,25 @@ class EvaluationMetricMetadata(cu.Record, Generic[T_OutputDatum]):
...
+        return values
+
+    def _serialize_extra_metadata(self) -> str:
+        return json.dumps(self.extra_metadata, sort_keys=True)
 
 
 class SampleMetricValue(cu.Record, Generic[T_OutputDatum]):
-    value: MetricValue
+    value: MetricValue | None
```
rholinshead, line 75: What's the intention of a None value here?

python/src/aiconfig/eval/common.py
jonathanlastmileai: Yes and the intention is to set it automatically. I will figure out a different approach that prevents the user from setting it manually.

python/src/aiconfig/eval/common.py
jonathanlastmileai: Placeholder for errors

python/src/aiconfig/eval/lib.py
```diff
@@ -86,23 +86,34 @@ def __str__(self) -> str:
...
+        def _ok_with_log(res_: Result[MetricValue, str]) -> MetricValue | None:
+            match res_:
+                case Ok(res):
+                    return res
+                case Err(e):
+                    LOGGER.error(f"Error evaluating {eval_params=}: {e}")
+                    return None
+
+        # TODO: figure out the right timeout
+        res_ = await cu.run_thunk_safe(_calculate(), timeout=1)
```
saqadri, line 112: is this 1 second? Would be good to add a quick comment, and ideally have timeout be passed in to the function

saqadri (approved):

PR #506, opened 2023-12-15
Initial .gitattributes

andrew-lastmile:


.gitattributes
```diff
@@ -0,0 +1,2 @@
+# reclassify ipynb notebooks as Python
+*.ipynb linguist-language=Python
```
Ankush-lastmile, line 2: are there any docs that tell us how this works?

Ankush-lastmile (approved):
lgtm

.gitattributes
andrew-lastmile: https://github.com/github-linguist/linguist/blob/master/docs/overrides.md

PR #513, opened 2023-12-16
[AIC-py][eval] add structured model-based evals

jonathanlastmileai:
[AIC-py][eval] add structured model-based evals

- add `make_structured_llm_metric()` which takes some configs and creates a metric
that queries an LLM with the text to evaluate and some evaluation criteria,
and returns a structured eval result
- wrap with `make_openai_structured_llm_metric()`
- make example `gpt3_5_text_ratings()`
- add tests that mock the openai call

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/513).
* #539
* #523
* #522
* __->__ #513


python/src/aiconfig/eval/common.py
```diff
@@ -1,14 +1,21 @@
 import json
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
-from typing import Any, Generic, Protocol, TypeVar
+from typing import Any, Generic, NewType, Protocol, Type, TypeVar
 
 import lastmile_utils.lib.core.api as cu
-from pydantic import root_validator
+import result
+from aiconfig.eval import common
```
rholinshead, line 8: Isn't this import referring to this module/file?

python/src/aiconfig/eval/common.py
```diff
@@ -127,3 +145,21 @@ def check_value_range(cls, values: dict[str, Any]) -> dict[str, Any]:
...
+
+class TextRatingsData(cu.Record):
+    conciseness_rating: int
+    conciseness_confidence: float
+    conciseness_reasoning: str
+
+
+def get_llm_structured_response(
+    input_text: str,
+    chat_completion_create: CompletionTextToSerializedJSON,
```
rholinshead, line 158: nit, should this be `text_completion_create` instead? Chat seems pretty specific

python/src/aiconfig/eval/metrics.py
```diff
@@ -1,11 +1,16 @@
...
 from functools import total_ordering
-from typing import Any, Generic
+from typing import Any, Generic, Type
 
+import lastmile_utils.lib.core.api as cu
 import nltk
 import pandas as pd
-from aiconfig.eval.common import CustomMetricValue, EvaluationFunction, EvaluationMetricMetadata, T_OutputDatum
+from aiconfig.eval import common
+from aiconfig.eval.common import CustomMetricValue, EvaluationFunction, EvaluationMetricMetadata, T_BaseModel, T_OutputDatum, TextRatingsData
```
rholinshead, line 9: Is the first import needed (i.e. second one is redundant right now)?

python/src/aiconfig/eval/metrics.py
```diff
@@ -147,3 +152,128 @@ async def _get_overall_positive_sentiment(output_datum: str) -> TextOverallPosit
...
+            extra_metadata=dict(
+                basemodel_type_name=pydantic_basemodel_type.__name__,
+                eval_llm_name=eval_llm_name,
+                field_descriptions_json=json.dumps(field_descriptions, sort_keys=True),
+            ),
+        ),
+    )
+
+
+def _make_openai_structured_llm_metric_helper(
```
rholinshead, line 196: Seems strange to have openai-specific methods in this more general file -- can this be moved to eval/openai.py?

python/tests/test_eval.py
```diff
@@ -231,8 +235,102 @@ async def test_exception_metric(caplog):
...
+            _mock_response(
+                function_arguments_return,
+            )
+        )
+
+    return _mock_openai_chat_completion_create
+
+
+@pytest.mark.asyncio
+async def test_openai_structured_eval():
```
rholinshead, line 288: Should this move to an openai eval test?

python/src/aiconfig/eval/common.py
jonathanlastmileai: Yes, in this case I'm using sort of partly-qualified identifiers like `common.SerializedJSON`. This is often better to disambiguate, e.g. it's the standard way to use pandas/numpy/pytorch/etc. Unfortunately vscode's auto-import fix does it the other way, so this is done inconsistently.

python/src/aiconfig/eval/metrics.py
jonathanlastmileai: Yeah, it's inconsistent (see above), and I don't know if vscode can fix this. I'll try and get cli-mate to do it as a follow-up.

python/src/aiconfig/eval/metrics.py
jonathanlastmileai: Yeah, I can move this but it'll require shuffling some other stuff (like the definition of metric) to avoid circular imports. Even if it's just type imports which *could* be made to work cyclically, it tends to be cleaner to make the modules totally non-cyclic.

python/tests/test_eval.py
jonathanlastmileai: Yeah I'm tracking small follow ups that can be bundled here: https://docs.google.com/document/d/15gx4vVmKVgI5bDHjSDd8xCQmTGcC8RrtLU0iCa32AIQ/edit#heading=h.3brc4h36z6i0

python/src/aiconfig/eval/common.py
jonathanlastmileai: I'll just tackle a bunch of these easy things together if that's ok (https://docs.google.com/document/d/15gx4vVmKVgI5bDHjSDd8xCQmTGcC8RrtLU0iCa32AIQ/edit)

rholinshead (approved):
I think this LGTM with the followups to come later

python/src/aiconfig/eval/openai.py
```diff
@@ -0,0 +1,87 @@
+from abc import abstractmethod
```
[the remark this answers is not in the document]
jonathanlastmileai: Yeah so Sarmad asked this too, and the answer is we should probably be using AIConfig for model-graded eval, but it was slightly easier to prototype this and get something working e2e using openai directly. This is actually a problem with our function calling docs - I couldn't quite figure it out (cc @Ankush-lastmile, @saqadri , @saqadri  ).


Also, due to that race condition, we can't do parallel batch inference yet.

python/src/aiconfig/eval/openai.py
```diff
@@ -0,0 +1,87 @@
...
+from typing import Protocol
+
+import lastmile_utils.lib.core.api as cu
+import openai
+import openai.types.chat as openai_types
+from aiconfig.eval import common
+from result import Err, Ok, Result
+
+
+@dataclass
```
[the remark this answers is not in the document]
jonathanlastmileai: Good question. According to the zen of python, "There should be one-- and preferably only one --obvious way to do it." Therefore, the Python community created at least two complicatedly overlapping and non-obvious solutions to essentially the same problem, which is the fact that Python doesn't have an immutable record type.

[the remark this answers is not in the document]
jonathanlastmileai: tl;dr I've been using immutable strict BaseModel (which is core_utils.Record) because it does auto-strict serialization and deserialization. However, that can be a blocker/PITA sometimes, which is when I go back to good old dataclass which essentially is just record with auto-constructor.

python/src/aiconfig/eval/metrics.py
```diff
@@ -147,3 +152,128 @@ async def _get_overall_positive_sentiment(output_datum: str) -> TextOverallPosit
...
+
+gpt3_5_text_ratings = make_openai_structured_llm_metric(
+    eval_llm_name="gpt-3.5-turbo-0613",
+    pydantic_basemodel_type=TextRatingsData,
+    metric_name="text_ratings",
+    metric_description="Text ratings",
+    field_descriptions=dict(
+        conciseness_rating="1 to 5 rating of conciseness",
+        conciseness_confidence="0 to 1.0 rating of confidence in conciseness rating",
+        conciseness_reasoning="reasoning behind the conciseness rating",
```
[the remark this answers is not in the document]
jonathanlastmileai: Not exactly following - this is basically the structured prompt engineering you do when creating your model-graded metric.

python/src/aiconfig/eval/metrics.py
```diff
@@ -147,3 +152,128 @@ async def _get_overall_positive_sentiment(output_datum: str) -> TextOverallPosit
...
+
+    # User interface: unwrap and raise
+    match res_metric:
+        case Ok(metric):
+            return metric
+        case Err(e):
+            raise ValueError(f"Error making metric: {e}")
+
+
+gpt3_5_text_ratings = make_openai_structured_llm_metric(
```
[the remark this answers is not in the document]
jonathanlastmileai: It's an example, but it can also really be used directly off the shelf (like brevity or substring_match). I'm tracking this problem of library structure/clarity in the follow-ups doc: "straighten up modules / init / public/private / underscore / api"

python/src/aiconfig/eval/common.py
```diff
@@ -127,3 +145,21 @@ def check_value_range(cls, values: dict[str, Any]) -> dict[str, Any]:
...
+    basemodel_type: Type[common.T_BaseModel],
+) -> Result[common.T_BaseModel, str]:
+    return result.do(
+        cu.safe_model_validate_json(response_ok, basemodel_type)
```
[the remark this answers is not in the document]
jonathanlastmileai: Yeah I can just codemod this globally

python/src/aiconfig/eval/common.py
jonathanlastmileai: @rholinshead , after looking again, this is supposed to be an interface to generalize `openai.chat.completions.create` but _not_ necessarily to be more general than chat. The idea here was to allow mocking the openai function. So unless we think there's a reason otherwise, I think it makes sense to keep it as is.

python/src/aiconfig/eval/metrics.py
jonathanlastmileai: We're probably going to end up generalizing this by moving to AIConfig anyway, so I'm going to leave this as-is for now so I don't do work that we know will be undone

PR #511, opened 2023-12-15
[editor] Input Attachments Starting Point

rholinshead:
[editor] Input Attachments Starting Point

# [editor] Input Attachments Starting Point

Initial implementation for prompt input attachments rendering. This supports rendering existing attachments and adds an upload field to add new attachments or update existing attachments.

Some follow-ups (ordered by priority):
- Implement upload endpoint to actually store new attachments
- UI improvements
- Implement attachment metadata editing

https://github.com/lastmile-ai/aiconfig/assets/5060851/21d17037-45fa-446b-a204-e8f437842603

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/511).
* #515
* #514
* __->__ #511
* #510

PR #515, opened 2023-12-17
[editor] Initial Output Rendering

rholinshead:
[editor] Initial Output Rendering

# [editor] Initial Output Rendering

Get prompt outputs to render. The main consideration here is that we can't really render directly based on the raw AIConfig outputs since those can be any arbitrary data defined by the model parsers. Instead, we add Client-specific types (mainly `ClientExecuteResult`) for the AIConfig which are constructed on load of the config, and which we will be able to properly handle rendering. For non-text outputs, we'll probably have an 'attachments' type to render those output attachments based on mimetype.

Side note -- maybe we just have a single type which can support both text & attachments? We should revisit if we have such models. This should do for now as a starting point.

Screenshot 2023-12-17 at 5 36 56 PM

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/515).
* __->__ #515
* #514
* #511
* #510

### DIFF
--- a/cli/aiconfig-editor/pages/api/aiconfig/load.ts
+++ b/cli/aiconfig-editor/pages/api/aiconfig/load.ts
@@ -1,9 +1,14 @@
 import type { NextApiRequest, NextApiResponse } from "next";
-import { AIConfig, AIConfigRuntime } from "aiconfig";
+import { AIConfigRuntime } from "aiconfig";
 import { ErrorResponse } from "@/src/shared/serverTypes";
+import {
+  ClientAIConfig,
+  ClientExecuteResult,
+  ClientPromptOutput,
+} from "@/src/shared/types";
 
 type Data = {
-  aiconfig: AIConfig;
+  aiconfig: ClientAIConfig;
 };
 
 type RequestBody = {
@@ -24,10 +29,34 @@ export default function handler(
     return res.status(500).json({ error: "No path provided" });
   }
 
-  // TODO: Needed to add node-fetch to the dependencies of aiconfig-editor for this to work
-  // This should be fixed up in library though?
   // TODO: load should probably be async?
   const aiconfig = AIConfigRuntime.load(body.path);
 
-  res.status(200).json({ aiconfig });
+  // Refine outputs for client-side rendering. We only care about displaying (and deleting)
+  // outputs directly from the editor
+  const clientAIConfig: ClientAIConfig = {
+    ...aiconfig,
+    prompts: aiconfig.prompts.map((prompt) => ({
+      ...prompt,
+      outputs: prompt.outputs?.map((output) => {
+        if (output.output_type === "execute_result") {
+          const text = aiconfig.getOutputText(prompt, output);
+          if (text) {
+            // TODO: Once AIConfig output types are updated to be more structured, revisit and
+            // ideally remove Client-specific types
+            return {
+              ...output,
+              renderData: {
+                type: "text",
+                text,
+              },
+            } as ClientExecuteResult;
+          }
+        }
+        return output as ClientPromptOutput;
+      }),
+    })),
+  };
+
+  res.status(200).json({ aiconfig: clientAIConfig });
 }
--- a/cli/aiconfig-editor/src/shared/types.ts
+++ b/cli/aiconfig-editor/src/shared/types.ts
@@ -1,3 +1,5 @@
+import { AIConfig, Prompt, Error, ExecuteResult } from "aiconfig";
+
 export type EditorFile = {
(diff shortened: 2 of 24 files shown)


rholinshead:
Once AIConfig has common output shapes we can revisit and update (ideally can remove the 'Client'- types and use the actual AIConfig defined types everywhere)

cli/aiconfig-editor/pages/api/aiconfig/load.ts
```diff
@@ -24,10 +29,32 @@ export default function handler(
...
+    ...aiconfig,
+    prompts: aiconfig.prompts.map((prompt) => ({
+      ...prompt,
+      outputs: prompt.outputs?.map((output) => {
+        if (output.output_type === "execute_result") {
+          const text = aiconfig.getOutputText(prompt, output);
+          if (text) {
+            return {
+              ...output,
+              client: {
```
saqadri, line 47: The naming is a bit confusing here. I'm guessing this means this is a client-only field, but perhaps a better name may be `renderData` or something that describes what it will be used for.

cli/aiconfig-editor/pages/api/aiconfig/load.ts
```diff
@@ -24,10 +29,32 @@ export default function handler(
...
 
-  res.status(200).json({ aiconfig });
+  // Refine outputs for client-side rendering. We only care about displaying (and deleting)
+  // outputs directly from the editor
+  const clientAIConfig: ClientAIConfig = {
+    ...aiconfig,
+    prompts: aiconfig.prompts.map((prompt) => ({
+      ...prompt,
+      outputs: prompt.outputs?.map((output) => {
+        if (output.output_type === "execute_result") {
```
saqadri, line 42: I'm guessing this is where we'll also add handling for error output types?

cli/aiconfig-editor/src/shared/types.ts
```diff
@@ -1,7 +1,43 @@
...
   isDirectory: boolean;
   disabled?: boolean;
 };
+
+export type TextExecuteResultOutput = {
+  type: "text";
+  text: string;
+};
+
+export type ClientExecuteResultOutput = TextExecuteResultOutput; // TODO: Add non-text types
```
saqadri, line 16: Not blocking for this diff, but this is a bit tricky. I'm wondering if we can avoid this type altogether by having a standardized format with which to store outputs. 

The other thing I'm concerned about at the moment is handling things like function calling and other more complex flows. We don't have to support everything now, but just design it such that they could be supported.

Ideally we could just use the main AIConfig type instead of needing this kind of intermediate type.

cli/aiconfig-editor/travel.aiconfig.json
```diff
@@ -1,8 +1,8 @@
 {
   "name": "NYC Trip Planner",
-  "description": "Intrepid explorer with ChatGPT and AIConfig",
   "schema_version": "latest",
   "metadata": {
+    "parameters": {},
```
saqadri, line 5: nit: remove

saqadri (approved):
Approving to unblock, but a bit concerned about how generalizable this approach will be

cli/aiconfig-editor/pages/api/aiconfig/load.ts
rholinshead: Error types should already be matching the Error type and we can get the error value/trace from that without special handling

cli/aiconfig-editor/src/shared/types.ts
rholinshead: Ya, left a comment above. @rossdanlm is looking into implementing a generalized Output type that will allow us to get the text / mime-type data directly, and once that's done we can come back and ideally remove all of this.

Note that this is only being done to support outputs right now -- function calling properties are already supported in model settings. I'm not sure what the expectation will be for those within the editor itself though since it needs a python runtime (i.e. running code outside of AIConfig) to call the functions, right?

saqadri: Yeah, mostly just thinking about how it gets rendered for an aiconfig that contains `functions`. The output in that case would contain the function call to execute, so just thinking about how it will get rendered by the editor

rholinshead: Ah, I see. Would it make sense to also treat function calls as just text to show the literal function call? They would probably be used to then propagate the results into the config via parameters in most cases, right? It's tough to say what the best UI would be for that output otherwise

PR #518, opened 2023-12-18
[py] pin openai dependency to < 1.5

Ankush-lastmile:
[py] pin openai to a range of versions

## What

pin openai dependency to 1.0.0 < x < 1.5

## Why

OpenAI updated the api to 1.5. https://github.com/openai/openai-python/releases/tag/v1.5.0

Looks like the new version updates the Chat Completion type with a new field `logprobs`

### DIFF
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -6,7 +6,7 @@ pytest
 pydantic>=2.1
 pybars3
 google-generativeai
-openai>=1.0.0
+openai >= 1.0.0, < 1.5
 python-dotenv
 huggingface_hub
 result


rholinshead:
Accepting to unblock release. We should probably fast-follow with the required updates to use 1.5 so that we have the latest changes; doesn't look like anything major other than supporting the logprops param

rholinshead (approved):

Ankush-lastmile:
> We should probably fast-follow with the required updates to use 1.5 so that we have the latest changes; doesn't look like anything major other than supporting the logprops param

created #519

PR #522, opened 2023-12-18
[AIC-py][eval] refactor nltk score metrics for mocking

jonathanlastmileai:
[AIC-py][eval] refactor nltk score metrics for mocking

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/522).
* #539
* #523
* __->__ #522
* #513


python/src/aiconfig/eval/metrics.py
```diff
@@ -103,55 +73,58 @@ def __lt__(self, other: Any) -> bool:
...
-    sid = SentimentIntensityAnalyzer()
-    mapping: dict[str, float] = sid.polarity_scores(output_datum)  # type: ignore
+class GetPolarityScores(Protocol):
+    @abstractmethod
+    def __call__(self, text: str) -> dict[str, float]:
+        pass
+
+
+def _get_nltk_polarity_scores(text: str, model: str) -> dict[str, float]:
+    nltk.download(model, quiet=True)  # type: ignore
```
rholinshead, line 83: Does this get cached, I assume?

rholinshead (approved):
I think this LGTM, just a question about caching the downloaded data. A lot of this is a bit hard to follow (at least for someone with little python & eval experience) so good documentation will be really important

python/src/aiconfig/eval/metrics.py
jonathanlastmileai: maybe under the hood, but when my flight wifi cut out it was hanging, so Im not entirely sure. That's what motivated this PR. This is really there just as an example, dont think we need to manually cache. I just wanted it not to run for real during tests.

PR #523, opened 2023-12-18
[AIC-py][eval] fix the types pt 1

jonathanlastmileai:
[AIC-py][eval] fix the types pt 1

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/523).
* #539
* __->__ #523
* #522
* #513


python/src/aiconfig/eval/metrics.py
```diff
@@ -98,52 +87,251 @@ def __lt__(self, other: Any) -> bool:
...
+# 1. functions that return metrics intended to be called directly
+
+
+def make_openai_structured_llm_metric(
+    eval_llm_name: str,
+    pydantic_basemodel_type: Type[T_BaseModel],
+    metric_name: str,
+    metric_description: str,
+    field_descriptions: dict[str, str] = {},
+    openai_chat_completion_create: OpenAIChatCompletionCreate | None = None,
```
saqadri, line 252: What is this particular one for? Is this to enable [openai API introspection](https://aiconfig.lastmileai.dev/docs/overview/create-an-aiconfig/#openai-api-python-wrapper)? Curious why we wouldn't just go with the AIConfigRuntime interface instead

saqadri (approved):
I don't fully understand all of this, but approving to unblock. Would love to get an in-person overview of this from you sometime this week

python/src/aiconfig/eval/metrics.py
jonathanlastmileai: You're right, ideally we should use aiconfig runtime here. Directly using OpenAI function calling was actually slightly easier / clearer to get going, so I did that in the interest of unblocking this. We should make it easier to get started with function calling using AIConfig. cc @Ankush-lastmile , @rossdanlm

jonathanlastmileai:
> I don't fully understand all of this, but approving to unblock. Would love to get an in-person overview of this from you sometime this week

Any time

python/src/aiconfig/eval/metrics.py
jonathanlastmileai: Also, `evaluate` uses asyncio.gather to run multiple blocking I/O calls concurrently, whereas AIConfig does not currently do that correctly (issue 434). We have to either fix that or pass batches back to the provider APIs to enable efficient inference and therefore eval.

rossdanlm: I also was asking why we are specifically using OpenAI params in https://github.com/lastmile-ai/aiconfig/pull/513#discussion_r1431914899

jonathanlastmileai: hmm, as opposed to?

PR #531, opened 2023-12-19
2/n Gemini Serialize()

Ankush-lastmile:
2/n Gemini Serialize()

## What

Implemented the serialize() method of the Gemini Model parser

The gemini api is a bit trickier than previous api's we've seen before. Using the OpenAI api as a comparison point, the Gemini api is a little bit more complex.

The request takes in a contents field, which can have various types. This is because the api supports
- one shot prompts
- one shot prompting but with multiple "parts"
- multi-turn conversation with multiple parts
- multi-modal one shot with multiple parts (out of scope of this diff)

Note: `part` refers to some form of data. Either a string, a proto, a blob, or mime typed data. Gemini api has done a good job in supporting a wide array of use cases.

For simplicity, I've limited the serialize() functionality to string objects or list/dict objects with strings. If the object contains something else, an error is thrown

Notes:
- Function Calling is not supported by the public release beta. This is only available in a beta not available to the general public. Incorporating this into serialize/deserialize() may require an extra step of parsing the input. Specifically, function calling takes in a param called `tools`.

## Why

Was not implemented previously, planned as a fast follow.

## Testplan

Serialize Test plan 1
serialize test plan 2

## Issue

#463

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/531).
* #558
* #534
* __->__ #531
* #532

### DIFF
--- a/extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
+++ b/extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
@@ -2,13 +2,14 @@
 from typing import TYPE_CHECKING, Dict, List, Optional, Any
 
 import google.generativeai as genai
-
+import copy
 
 from aiconfig.default_parsers.parameterized_model_parser import ParameterizedModelParser
 from aiconfig.model_parser import InferenceOptions
 from aiconfig.schema import ExecuteResult, Output, Prompt
-from aiconfig.util.params import resolve_prompt
-from aiconfig import CallbackEvent, get_api_key_from_environment, AIConfigRuntime
+from aiconfig.util.params import resolve_prompt, resolve_prompt_string
+from aiconfig import CallbackEvent, get_api_key_from_environment, AIConfigRuntime, PromptMetadata, PromptInput
+from google.generativeai.types import content_types
 
 
 # Circuluar Dependency Type Hints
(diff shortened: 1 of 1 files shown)


extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
```diff
@@ -1,5 +1,6 @@
 # Define a Model Parser for LLama-Guard
-from typing import TYPE_CHECKING, Dict, List, Optional, Any
+from typing import TYPE_CHECKING, Dict, Iterable, List, Optional, Any
```
rossdanlm, line 2: This is the police! You are not allowed to use unused modules: `Iterable`

extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
```diff
@@ -16,6 +17,21 @@
...
+DOCSTRING = """
+Model Parser for Gemini text to text models. Doesn't support image generation yet. 
+Function calling is not available on the public release Gemini api. Function Calling is also not supported in this model parser.
+See this link for more information on api versions: https://ai.google.dev/docs/api_versions
+
+@ankush-lastmile 
+TODO: This model Parser does not support multimodal
+TODO: This model Parser does not support function calling (not available)
+TODO: This model Parser does not support serializing all different types of the Gemini API (ie protos)
+TODO: This model Parser does not support strongly structuring the input data containing the configmetadata for the Gemini API
```
rossdanlm, line 30: The `GenerationConfig` params must be specified explicitly underneath `settings` as it's own nested dict (see "gH#1232 test plan for details")

extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
```diff
@@ -90,14 +106,50 @@ async def serialize(
         **kwargs,
     ) -> List[Prompt]:
         """
-        Defines how a prompt and model inference settings get serialized in the .aiconfig.
+        Defines how a prompt or a multi turn prompt chain along with  model inference settings get serialized into prompts for AIConfig.
```
rossdanlm, line 109: there's an extra space :o

rossdanlm (approved):
Great work! pls test!

Ankush-lastmile:
discussed and resolved other comments offline cc: @rossdanlm

PR #539, opened 2023-12-19
[AIC-py][eval] bundle of small easy fixes

jonathanlastmileai:
[AIC-py][eval] bundle of small easy fixes

- freeze dataclasses
- wire up eval fn timeout further (there is a little more to go)
- move mock aiconfig runtime to separate file
-

see https://docs.google.com/document/d/15gx4vVmKVgI5bDHjSDd8xCQmTGcC8RrtLU0iCa32AIQ/edit

### DIFF
--- a/python/src/aiconfig/eval/lib.py
+++ b/python/src/aiconfig/eval/lib.py
@@ -1,6 +1,7 @@
 import asyncio
 import logging
 from dataclasses import dataclass
+from functools import partial
 from typing import Any, Generic, NewType, Sequence, Tuple, TypeVar
 
 import lastmile_utils.lib.core.api as cu
@@ -90,7 +91,7 @@ MetricList = list[Metric[T_OutputDatum, Any]]
 
 
 async def _evaluate_for_sample(
-    eval_params: SampleEvaluationParams[T_InputDatum, T_OutputDatum, T_MetricValue]
+    eval_params: SampleEvaluationParams[T_InputDatum, T_OutputDatum, T_MetricValue], timeout: int
 ) -> SampleEvaluationResult[T_InputDatum, T_OutputDatum, T_MetricValue]:
     sample, metric = (
         eval_params.output_sample,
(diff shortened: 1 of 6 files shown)


python/src/aiconfig/eval/lib.py
```diff
@@ -90,7 +91,7 @@ def __str__(self) -> str:
 
 
 async def _evaluate_for_sample(
-    eval_params: SampleEvaluationParams[T_InputDatum, T_OutputDatum, T_MetricValue]
+    eval_params: SampleEvaluationParams[T_InputDatum, T_OutputDatum, T_MetricValue], timeout: int
```
saqadri, line 94: would recommend renaming the variable to include the time unit. For e.g. `timeout_ms`

python/src/aiconfig/eval/lib.py
```diff
@@ -123,9 +123,9 @@ def _ok_with_log(res_: Result[T_MetricValue, str]) -> T_MetricValue | None:
 
 
 async def evaluate(
-    evaluation_params_list: DatasetEvaluationParams[T_InputDatum, T_OutputDatum],
+    evaluation_params_list: DatasetEvaluationParams[T_InputDatum, T_OutputDatum], eval_fn_timeout: int
```
saqadri, line 126: Same as above. `eval_fn_timeout_ms`

saqadri (approved):
LGTM, just a few comments

rholinshead:
For the description, can you update to state what the fixes are here? That will be easier for changelogs and ensures anyone can see the description in the public repo (that doc isn't public)

python/src/aiconfig/eval/lib.py
jonathanlastmileai: Added here https://docs.google.com/document/d/15gx4vVmKVgI5bDHjSDd8xCQmTGcC8RrtLU0iCa32AIQ/edit

PR #542, opened 2023-12-19
[ez] Check for nullable system prompt

rossdanlm:
[ez] Check for nullable system prompt

We should really have automated tests more in our cookbooks to catch these issues, weird

This is NOT a thorough solution, just quickly unblocking. Filed an issue to investigate deeper and make sure we have better null-checks: https://github.com/lastmile-ai/aiconfig/issues/543

### DIFF
--- a/cookbooks/Getting-Started/travel.aiconfig.json
+++ b/cookbooks/Getting-Started/travel.aiconfig.json
@@ -6,6 +6,7 @@
     "models": {
       "gpt-3.5-turbo": {
         "model": "gpt-3.5-turbo",
+        "system_prompt": null,
         "top_p": 1,
         "temperature": 1
       },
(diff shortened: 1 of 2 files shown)


rossdanlm:
test plan

| Before | After |
|---|---|
| Screenshot 2023-12-19 at 14 20 49 | Screenshot 2023-12-19 at 14 20 17 |

saqadri (approved):

cookbooks/Getting-Started/travel.aiconfig.json
```diff
@@ -6,6 +6,7 @@
     "models": {
       "gpt-3.5-turbo": {
         "model": "gpt-3.5-turbo",
+        "system_prompt": null,
```
rholinshead, line 9: This should be removed, right? @rossdanlm

rossdanlm: I supposed you are correct

PR #551, opened 2023-12-20
[editor] Only Persist Changed Model Settings

rholinshead:
# Only Persist Changed Model Settings

Fix a bug in the settings renderer copied from workbooks -- currently, just opening the settings tab and saving the workbook will persist a model settings object for the prompt. The cause of this is the use of the `useEffect` for handling changes -- `useEffect` will always run once on mount of the component so it was always persisting the empty settings on mount.

To fix, just check that the value has actually changed before persisting it to the config.

## Testing:
- Open settings & save config without touching anything. See the config has no changes
- Change system prompt and save, ensure it's persisted to the config
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/551).
* #552
* __->__ #551

### DIFF
--- a/cli/aiconfig-editor/src/shared/prompt_schemas/OpenAIChatModelParserPromptSchema.ts
+++ b/cli/aiconfig-editor/src/shared/prompt_schemas/OpenAIChatModelParserPromptSchema.ts
@@ -66,7 +66,6 @@ export const OpenAIChatModelParserPromptSchema: PromptSchema = {
         type: "integer",
         maximum: 4096,
         minimum: 16,
-        default: 4096,
       },
       n: {
         type: "integer",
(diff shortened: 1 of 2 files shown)


cli/aiconfig-editor/src/shared/prompt_schemas/OpenAIChatModelParserPromptSchema.ts
```diff
@@ -66,7 +66,6 @@ export const OpenAIChatModelParserPromptSchema: PromptSchema = {
         type: "integer",
         maximum: 4096,
         minimum: 16,
-        default: 4096,
```
rholinshead, line 69: Removing this for now, seems like a default value will be persisted if it exists

saqadri (approved):
Nice catch!

PR #555, opened 2023-12-20
[AIC-py][editor] add dev option to script which runs yarn (--mode=debug)

jonathanlastmileai:
[AIC-py][editor] add dev option to script which runs yarn (--mode=debug)

### DIFF
--- a/python/src/aiconfig/editor/README.md
+++ b/python/src/aiconfig/editor/README.md
@@ -1,39 +1,42 @@
 # Editor
 
-# Install:
+## Usage (prod)
+
+### Install:
 
 Install python-aiconfig from pip, then set `aiconfig` alias (in shell, .bashrc, .zshrc, etc.)
 
 One-liner:
 `pip install python-aiconfig; alias aiconfig="python -m 'aiconfig.scripts.aiconfig_cli'"`
 
-## Usage
+
+### Run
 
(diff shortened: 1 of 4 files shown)


python/src/aiconfig/editor/server/server.py
```diff
@@ -17,6 +17,7 @@ class EditConfig(core_utils.Record):
     server_port: int = 8888
```
rholinshead, line 17: Can we default this to 8080 as well?

rholinshead (approved):

python/src/aiconfig/editor/server/server.py
jonathanlastmileai: yes

python/src/aiconfig/editor/README.md
```diff
@@ -1,39 +1,42 @@
...
 
-```
-cd python/src/editor/client && yarn && yarn start
-```
+`pip install -e path/to/local/aiconfig/python`
+`alias aiconfig="python -m 'aiconfig.scripts.aiconfig_cli'"``
+
+2. Run backend and frontend servers:
+
+`aiconfig edit  --aiconfig-path=/my/path --server-port=8080 --server-mode=debug`
```
rossdanlm, line 37: By default server port is just 8080 right? Would it be simpler to do without it?

PR #557, opened 2023-12-20
[editor] Union Settings Property

rholinshead:
# [editor] Union Settings Property

Handle 'union' properties in the settings renderer to support properties that could be one of a union of types, rendering a `UnionPropertyControl` component whenever the property type is `union`. This required two main considerations:
1. Use a callback for setting the state/value instead of an effect; this ensures the value is set when they change instead of when the component mounts
2. Render a UnionPropertyControl with a render prop to render nested SettingsPropertyRenderers based on the property types (rendering SettingsPropertyRenderer directly in UnionPropertyControl could cause a cycle)

## Testing:
- Set both values and ensure the correct value is set in the config:

https://github.com/lastmile-ai/aiconfig/assets/5060851/05202282-f6f9-4dae-834c-be47be6aed43

### DIFF
--- a/cli/aiconfig-editor/src/shared/prompt_schemas/OpenAIChatModelParserPromptSchema.ts
+++ b/cli/aiconfig-editor/src/shared/prompt_schemas/OpenAIChatModelParserPromptSchema.ts
@@ -66,6 +66,7 @@ export const OpenAIChatModelParserPromptSchema: PromptSchema = {
         type: "integer",
         maximum: 4096,
         minimum: 16,
+        default: 4096,
       },
       n: {
         type: "integer",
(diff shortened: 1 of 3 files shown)


cli/aiconfig-editor/src/shared/prompt_schemas/OpenAIChatModelParserPromptSchema.ts
```diff
@@ -66,6 +66,7 @@ export const OpenAIChatModelParserPromptSchema: PromptSchema = {
         type: "integer",
         maximum: 4096,
         minimum: 16,
+        default: 4096,
```
rholinshead, line 69: Another benefit of using the callback instead of `useEffect` is that we can now safely set defaults in the schema without them immediately being propagated to the config

jonathanlastmileai (approved):
I don't really know what this does but I trust you

cli/aiconfig-editor/src/shared/prompt_schemas/OpenAIChatModelParserPromptSchema.ts
saqadri, line 69: Do we want to set this default btw? It's very outdated

saqadri (commented):
Looks awesome!

PR #534, opened 2023-12-19
3/n pin Gemini dependency (google api)

Ankush-lastmile:
3/n pin Gemini dependency (google api)

## What

Pin google-generative ai, the library used for Gemini

## Why
Anticipiating potential updates to the library, its prudent to pin its current version.
Additionally, it ensures that all users operate on the same, specified version of the API.

## Testplan

I ran `pip list | grep google-generativeai` to find the version of the library that was installed locally.

```
ankush@ap-mbp python % pip list | grep google-generativeai
google-generativeai                           0.3.1
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/534).
* #558
* __->__ #534
* #531
* #532

PR #532, opened 2023-12-19
1/n Gemini Update deserialize to take data in the form of completion params

Ankush-lastmile:
1/n Gemini Update deserialize to take data in the form of completion params

## What

Updated deserialize() in Gemini Model Parser to take in completion data from prompt input if it is already in the form of the Gemini Completion data

Now, you can specify input in the form of a request to gemini api, ie the parts.

## Why

Previously This was skipped due to a lack of understanding of how the api requests worked.

## Testplan
Validating 2 pieces of functionality with this test plan:
1. setting the generation config for Gemini in setings. Generation Config is gemini's way of defining settings like topP
2. Deserialize() can take in data just like the api, by setting "contents" in the prompt input.
testplan

testplan 2

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/532).
* #558
* #534
* #531
* __->__ #532

PR #558, opened 2023-12-21
4/n Update output serialization with pb

Ankush-lastmile:
4/n Update output serialization with pb

## What

When saving outputs to config, get the underlying protobuf of Gemini's response object and save as a dict

Also added some comments cc: @rossdanlm

## Why

When doing `aiconfig.save()`, pydantic is unable to deserialze the output metadata which contains a protobuf from Gemini's response. Gemini response object is a wrapper around a protobuf. Need to convert to an object that is compatible with pydantic.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/558).
* __->__ #558
* #534
* #531
* #532

### DIFF
--- a/extensions/Gemini/python/requirements.txt
+++ b/extensions/Gemini/python/requirements.txt
@@ -1,8 +1,9 @@
 #Google
-google-generativeai
+google-generativeai==0.3.1
 
 #Other
 asyncio
+copy
 
 # AIConfig
 python-aiconfig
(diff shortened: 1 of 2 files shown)


rossdanlm (approved):
image

Ankush-lastmile:
added comments cc: @rossdanlm

Ankush-lastmile:
Shipping this stack. Will fix forward if anything else comes up

PR #560, opened 2023-12-21
[AIC-py][editor] load user module for model parsers, run hello world llama

jonathanlastmileai:
[AIC-py][editor] load user module for model parsers, run hello world llama

- user passes path to file (python module) which defines `register_model_parsers()`.
- import the module and call during server setup
- add simple llama run endpoint as poc

~/my_editor_plugin.py:

```
from aiconfig import AIConfigRuntime
from llama import LlamaModelParser


def register_model_parsers() -> None:
    model_path = "/Users/jonathan/Projects/aiconfig/models/llama-2-7b-chat.Q4_K_M.gguf"
    llama_model_parser = LlamaModelParser(model_path)
    AIConfigRuntime.register_model_parser(llama_model_parser, "llama-2-7b-chat")

```

Run server:
```
aiconfig edit  --aiconfig-path=cli/aiconfig-editor/travel.aiconfig.json \
          --server-port=8080 \
          --server-mode=debug_backend \
          --parsers-module-path="~/my_editor_plugin.py" --log-level=debug

```

Request:
`curl http://localhost:8080/api/run`

Output:

```
<p>

ANSWER:
When visiting Central Park in New York City, it's a good idea to bring the following items:

- A map of the park to help you navigate.
- Comfortable walking shoes or sneakers to enjoy the many walking paths and attractions.
- Sunscreen and insect repellent to protect yourself from the sun and bugs.
- A camera to capture memories of your visit, including iconic landmarks like the Bethesda Terrace and the Loeb Boathouse.
- A picnic blanket or a cooler with sn</p>
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/560).
* #567
* #565
* #563
* __->__ #560
* #556
* #555

PR #556, opened 2023-12-20
[AIC-py][editor] bundle easy changes

jonathanlastmileai:
[AIC-py][editor] bundle easy changes

- make aiconfig runtime
- log to file
- make path relative to aiconfig root

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/556).
* #567
* #565
* #563
* #560
* __->__ #556
* #555

PR #563, opened 2023-12-21
[AIC-py][editor] implement a couple of POST endpoints and add a lot of usage and dev detail to readme

jonathanlastmileai:
[AIC-py][editor] implement a couple of POST endpoints and add a lot of usage and dev detail to readme

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/563).
* #567
* #565
* __->__ #563
* #560
* #556
* #555


python/src/aiconfig/editor/server/server.py
```diff
@@ -109,27 +110,88 @@ def _get_server_state(app: Flask) -> ServerState:
...
-    LOGGER.info("Count: %s", ss.count)
     return app.send_static_file("index.html")
 
 
-@app.route("/test")
-def test():
-    return {"key": 6}
+@app.route("/api/load", methods=["POST"])
+def load():
+    ss = _get_server_state(app)
```
saqadri, line 118: minor nit: having some meaningful variable names makes it easy to read the code down the line (maybe rename to `state`?)

python/src/aiconfig/editor/server/server.py
```diff
@@ -109,27 +110,88 @@ def _get_server_state(app: Flask) -> ServerState:
...
+    if not path:
+        return {"message": "No path provided"}, 400
+    elif not os.path.isfile(path):
+        return {"message": f"File does not exist: {path}"}, 400
+    else:
+        try:
+            ss.aiconfig_runtime = AIConfigRuntime.load(path)  # type: ignore
+            return {"message": "Done"}, 200
+        except Exception as e:
+            return {"message": f"<p>Failed to load AIConfig from {path}: {e}"}, 400
```
saqadri, line 131: `<p>` missing and end `</p>`? Do we even need the string to be wrapped in the paragraph tag?

saqadri: Same for all methods below

python/src/aiconfig/editor/server/server.py
```diff
@@ -109,27 +110,88 @@ def _get_server_state(app: Flask) -> ServerState:
...
+
+
+@app.route("/api/save", methods=["POST"])
+def save():
+    ss = _get_server_state(app)
+    request_json = request.get_json()
+    path = request_json["path"]
```
saqadri, line 138: missing `include_outputs`. I think it would be much better to stick close to the properties/data types expected by AIConfig API. Makes it easy to use this server

python/src/aiconfig/editor/server/server.py
```diff
@@ -109,27 +110,88 @@ def _get_server_state(app: Flask) -> ServerState:
...
+    prompt_name = request_json.get("prompt_name", None)
+    stream = request_json.get("stream", True)
+    LOGGER.info(f"Running prompt: {prompt_name}, {stream=}")
+    inference_options = InferenceOptions(stream=stream)
+    try:
+        result = await ss.aiconfig_runtime.run(prompt_name, options=inference_options)  # type: ignore
+        LOGGER.debug(f"Result: {result=}")
+        result_text = str(
+            ss.aiconfig_runtime.get_output_text(prompt_name)  # type: ignore
+            #
```
saqadri, line 171: Remove?

python/src/aiconfig/editor/server/server.py
```diff
@@ -109,27 +110,88 @@ def _get_server_state(app: Flask) -> ServerState:
...
+            if isinstance(result, list)
+            #
+            else result.data[0]  # type: ignore
+        )
+        return {"message": "Done", "output": result_text}, 200
```
saqadri, line 176: This shouldn't be text specific. Returning the entire output will be important

saqadri (approved):
Approving to unblock -- a few comments.

Before we implement all the endpoints, can we align on the pattern to adopt for the API request type, and responses? cc @rossdanlm -- I think a bunch of work for this has already been done in the Gradio backend (https://github.com/lastmile-ai/gradio-workbook/blob/main/gradioworkbook/backend/gradio_gradioworkbook/events.py) which we should take inspiration from.

What's currently being returned cannot be used by the frontend

jonathanlastmileai:
> Approving to unblock -- a few comments.
> 
> Before we implement all the endpoints, can we align on the pattern to adopt for the API request type, and responses? cc @rossdanlm -- I think a bunch of work for this has already been done in the Gradio backend (https://github.com/lastmile-ai/gradio-workbook/blob/main/gradioworkbook/backend/gradio_gradioworkbook/events.py) which we should take inspiration from.
> 
> What's currently being returned cannot be used by the frontend

Sure, why is this unusable and what structure should it have? I can do any endpoint naming convention and request/response structure, all should be equally easy.

python/src/aiconfig/editor/server/server.py
jonathanlastmileai: We'll fix this with standard input/output

python/src/aiconfig/editor/server/server.py
jonathanlastmileai: I'm using these # lines as a hack because otherwise it autoformats to very long lines...

PR #564, opened 2023-12-21
[extensions] gemini bump extension package version; update dependencies before publish

Ankush-lastmile:
[extensions] gemini bump extension package version; update dependencies before publish

## What

- bump pyproject toml
- Removed @override, this doesn't exist outside of python 3.12
- removed copy from requirements.txt, this is included in python; does not need to be pip installed.

## Why

To update public release with updated functionality to serialize and deserialize() for GEmini Model parser

## Testplan:

Run through the Gemini cookbook succesfully with local pip install

Screenshot 2023-12-21 at 2 18 56 PM

### DIFF
--- a/cookbooks/Gemini/Gemini.ipynb
+++ b/cookbooks/Gemini/Gemini.ipynb
@@ -54,7 +54,7 @@
     },
     {
       "cell_type": "code",
-      "execution_count": null,
+      "execution_count": 1,
       "metadata": {
         "id": "wLS1XJvyosT7"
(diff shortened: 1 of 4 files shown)


saqadri (approved):

Ankush-lastmile:
- updated requirements.txt to remove copy (already included with python not needed to be a requirement)
- removed @override annotation; only exists in python 3.12: [PEP 698](https://peps.python.org/pep-0698/)

Issue #490, opened 2023-12-14
[bug] saving outputs in the config by default


tanya-rai:
everytime i rerun a prompt with a new parameter I get this "apologies for the confusion, here is the updated response" 
it's because we save the output by default to the prompt which it then uses in context? 

the fix I need to do in the cookbook/streamlit app is to add this line of code always to make sure the output is deleted manually after config.run: 
`config.get_prompt("get_activities").outputs = []`

saqadri:
This is because we aren't deleting the output before running. @Ankush-lastmile this should be a one-line fix. Can you please ship it? https://github.com/lastmile-ai/aiconfig/pull/371

Ankush-lastmile:
Closed with #491

PR #566, opened 2023-12-21
[py] Batch Execution fix inconsistent outputs

Ankush-lastmile:
[py] Batch Execution fix inconsistent outputs

## What

1. Resolved an inconsistency between the prompt object in the deep copy and the one used for execution. Ensures the correct Prompt object is used during Batch Execution.

2. Also added an extra step of clearing outputs inside run_batch, similarly to #490

## Why

-  Python SDK checks for the existence of Prompt objects within the AIConfigRuntime, evaluating more than just their names.
- In Batch Execution, a deep copy of AIConfigRuntime is created, which also duplicates the Prompt objects. This update addresses the issue where previously a stale prompt object might have been used.

## Testplan:

- rebased onto [this commit](https://github.com/lastmile-ai/aiconfig/pull/553/files) and tested with the python notebook provided by @2timesJay. The batch execution operated correctly as intended.

- Performance comparison between batch and sequential execution showed no significant difference in speed, minus variations in GPT outputs.

Successful Batch Execution:
Screenshot 2023-12-21 at 12 15 20 PM

Comparison of sequential execution (see time):
Screenshot 2023-12-21 at 12 15 35 PM

### DIFF
--- a/python/src/aiconfig/model_parser.py
+++ b/python/src/aiconfig/model_parser.py
@@ -110,6 +110,7 @@ class ModelParser(ABC):
         for params in parameters_list:
             # Create a deep copy of the aiconfig object to prevent mutations that could affect other iterations
             aiconfig_deep_copy = copy.deepcopy(aiconfig)
+            prompt = aiconfig_deep_copy.get_prompt(prompt.name)
             # Asynchronously schedule 'run()' for execution with a set of parameters.
             # This approach enables concurrent processing of multiple aiconfigs.
             task = asyncio.create_task(self.run(prompt, aiconfig_deep_copy, options, params, **kwargs))
--- a/python/src/aiconfig/Config.py
+++ b/python/src/aiconfig/Config.py
@@ -317,6 +317,9 @@ class AIConfigRuntime(AIConfig):
         model_name = self.get_model_name(prompt_data)
         model_provider = AIConfigRuntime.get_model_parser(model_name)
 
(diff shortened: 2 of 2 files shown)


python/src/aiconfig/model_parser.py
```diff
@@ -110,6 +110,7 @@ async def run_batch(
         for params in parameters_list:
             # Create a deep copy of the aiconfig object to prevent mutations that could affect other iterations
             aiconfig_deep_copy = copy.deepcopy(aiconfig)
+            prompt = aiconfig_deep_copy.get_prompt(prompt.name)
```
saqadri, line 113: Why would this ever be different than the original?

saqadri (approved):
I don't understand why the prompt object will be different between the deep copied version and the original? Can you explain?

python/src/aiconfig/model_parser.py
Ankush-lastmile: deepcopy recursively creates a deep copy of each object.

The contents of the prompt are the same, but the **Object** itself is not the same. Inside the sdk we operate on the prompt itself. 

For example: 

```
  public async run(
    prompt: Prompt,
    aiConfig: AIConfigRuntime,
    options?: InferenceOptions,
    params?: JSONObject | undefined
  ): Promise<Output | Output[]> {  
  ```
  
`prompt.outputs = something`,  

Mutates the state of the prompt _Object_.

PR #567, opened 2023-12-21
[AIC-py][editor] add default model parsers file

jonathanlastmileai:
[AIC-py][editor] add default model parsers file

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/567).
* #573
* __->__ #567

### DIFF
--- a/python/src/aiconfig/editor/server/server.py
+++ b/python/src/aiconfig/editor/server/server.py
@@ -34,16 +34,16 @@ class EditServerConfig(core_utils.Record):
     aiconfig_path: Optional[str] = None
     log_level: str | int = "INFO"
     server_mode: str
-    parsers_module_path: Optional[str] = None
+    parsers_module_path: str = "aiconfig_model_registry.py"
 
 
 @dataclass
 class ServerState:
-    aiconfig_runtime: AIConfigRuntime | None = None
+    aiconfig: AIConfigRuntime | None = None
 
 
 @dataclass(frozen=True)
-class HttpPOSTResponse:
+class HttpPostResponse:
     message: str
     output: str | None = None
     code: int = 200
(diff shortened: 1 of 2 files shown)


python/src/aiconfig/editor/default_user_model_parsers_plugin.py
saqadri, line 1: Can we rename this file to something simpler? Maybe `model_registry.py` or `aiconfig_model_registry.py`?

python/src/aiconfig/editor/server/server.py
```diff
@@ -34,7 +34,7 @@ class EditServerConfig(core_utils.Record):
     aiconfig_path: Optional[str] = None
     log_level: str | int = "INFO"
     server_mode: str
-    parsers_module_path: Optional[str] = None
+    parsers_module_path: str = "python/src/aiconfig/editor/default_user_model_parsers_plugin.py"
```
saqadri, line 37: I don't think this default makes much sense. It should be looking for a file in the directory where the server was started from, no? Otherwise this local path is just for our own local development...

saqadri (changes requested):

python/src/aiconfig/editor/server/server.py
jonathanlastmileai: Right, good catch.

jonathanlastmileai: But the problem is if you start the server from an arbitrary directory, then we cant include the default file, which means if the user forgets to put their module in their directory, they'll get a warning on server start. This is not the end of the world, but it seems weird that the default behavior is to start up with a warning that we can prevent.

jonathanlastmileai: So I made it look in cwd for `aiconfig_model_registry.py`, which will not exist unless the user puts it there (the cwd could be outside the repo, or maybe the user doesn't even have a clone). 

This means the default server behavior is to print a warning that the file doesn't exist. I think this is fine for now. On the `load_model_parser_module` endpoint, it's stricter and logs a bad file as an error.

rholinshead (approved):

saqadri (approved):

PR #572, opened 2023-12-21
[editor] Move Over Client Files & Fix Import Paths

rholinshead:
# Move Over Client Files & Fix Import Paths

Copy over all of the client files from existing editor directory. Fix all imports to relative (we can look into making them absolute again, just wasn't able to make it work within a few minutes) so that the modules can be found.

Load the page without any JS errors (spinner showing since no config is loaded yet due to endpoint being 3000 by default):
Screenshot 2023-12-21 at 3 14 19 PM

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/572).
* __->__ #572
* #571


jonathanlastmileai (approved):
LG

PR #573, opened 2023-12-21
[AIC-py][editor] kill the frontend servers better

jonathanlastmileai:
[AIC-py][editor] kill the frontend servers better

Fix up server mode stuff
- Enum for validation
- better logging
- try to kill the frontend servers anywhere the script ends

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/573).
* __->__ #573
* #567

### DIFF
--- a/python/src/aiconfig/editor/README.md
+++ b/python/src/aiconfig/editor/README.md
@@ -59,7 +59,7 @@ e.g. `aiconfig edit --parsers-module-path="/path/to/my_editor_plugin.py"`
 ### Run backend and frontend servers:
 
 (debug mode will run the react server)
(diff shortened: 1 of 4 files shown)


rholinshead (approved):

rholinshead:
Ah, also noticing just now that if the python server has an exception spinning up (in my case, it can't find the config I'm passing, 'aiconfig/cli/aiconfig-editor/travel.aiconfig') it will still spin up the client server in the same terminal and that python exception is hidden

PR #576, opened 2023-12-21
[AIC-py][editor] return aiconfig from http endpoints

jonathanlastmileai:
[AIC-py][editor] return aiconfig from http endpoints

and some path validation fixes

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/576).
* #580
* #579
* __->__ #576
* #575

### DIFF
--- a/python/src/aiconfig/editor/server/server.py
+++ b/python/src/aiconfig/editor/server/server.py
@@ -2,12 +2,13 @@ from dataclasses import dataclass
 from enum import Enum
 import logging
 from types import ModuleType
-from typing import Any, Callable, Optional
+from typing import Any, Callable, NewType, Optional
 
 import lastmile_utils.lib.core.api as core_utils
 from flask import Flask, request
 from pydantic import field_validator
 from result import Err, Ok, Result
+import result
 
 from aiconfig.Config import AIConfigRuntime
 import importlib
@@ -29,6 +30,9 @@ formatter = logging.Formatter(core_utils.LOGGER_FMT)
 log_handler.setFormatter(formatter)
 
 LOGGER.addHandler(log_handler)
+
+UnvalidatedPath = NewType("UnvalidatedPath", str)
+ValidatedPath = NewType("ValidatedPath", str)
 
 
 class ServerMode(Enum):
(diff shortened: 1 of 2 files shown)


python/src/aiconfig/editor/server/server.py
```diff
@@ -160,62 +166,75 @@ def home():
...
+    state = _get_server_state(app)
+    if state.aiconfig is None:
+        return HttpPostResponse(message="No AIConfig in memory, nothing to save.", code=400, aiconfig=None).to_flask_format()
+    else:
+        aiconfig: AIConfigRuntime = state.aiconfig
 
-    return _http_response_with_path(_run_with_path).to_flask_format()
+        path_val = _validated_request_path(allow_create=True)
+        res_save: Result[HttpPostResponse, str] = result.do(
+            Ok(HttpPostResponse(message="Saved to disk", aiconfig=aiconfig))
```
rholinshead, line 226: nit, don't need to return the config in the response for save. Really just need to know success/fail

rholinshead (approved):

python/src/aiconfig/editor/server/server.py
```diff
@@ -63,14 +67,21 @@ class ServerState:
 @dataclass(frozen=True)
 class HttpPostResponse:
     message: str
-    output: str | None = None
+    aiconfig: AIConfigRuntime | None
     code: int = 200
```
rholinshead, line 71: We'll be able to customize this response a bit for other endpoints, right? Like list_models should just return a list of supported/registered model names

python/src/aiconfig/editor/server/server.py
jonathanlastmileai: Ok, I will clean this up in a future PR, keeping in mind that you only need the config on a load or successful update, right?

python/src/aiconfig/editor/server/server.py
jonathanlastmileai: Yes. You tell me the ideal API, i'll do it. But yes I have list_models like this:

```
+@app.route("/api/list_models", methods=["GET"])
+def list_models() -> FlaskResponse:
+    out: list[str] = ModelParserRegistry.parser_ids()  # type: ignore
+    json_obj = core_utils.JSONObject({"data": core_utils.JSONList(out)})
+    return FlaskResponse((json_obj, 200))
```

PR #579, opened 2023-12-21
[AIC-py][editor] return aiconfig from remaining existing post endpoints

jonathanlastmileai:
[AIC-py][editor] return aiconfig from remaining existing post endpoints

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/579).
* #580
* __->__ #579
* #576
* #575

PR #580, opened 2023-12-21
[AIC-py][editor] split server into 2 files

jonathanlastmileai:
[AIC-py][editor] split server into 2 files

This is getting kind of unweildy. Move most of the hleper functions into server_utils,
leaving the core server stuff (mainly endpoints).

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/580).
* __->__ #580
* #579
* #576
* #575

### DIFF
--- a/python/src/aiconfig/editor/server/server_utils.py
+++ b/python/src/aiconfig/editor/server/server_utils.py
@@ -1,0 +1,215 @@
+import importlib
+import importlib.util
+import logging
+import os
+import sys
+from dataclasses import dataclass
+from enum import Enum
+from types import ModuleType
+from typing import Any, Callable, NewType, Optional
+
+import lastmile_utils.lib.core.api as core_utils
+from aiconfig.Config import AIConfigRuntime
+from flask import Flask
+from pydantic import field_validator
+from result import Err, Ok, Result
+
+logging.getLogger("werkzeug").disabled = True
+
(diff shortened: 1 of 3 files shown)


saqadri (approved):
Nice

python/src/aiconfig/editor/server/server_utils.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+def resolve_path(path: str) -> str:
+    return os.path.abspath(os.path.expanduser(path))
+
+
+def get_validated_path(raw_path: str, allow_create: bool = False) -> Result[ValidatedPath, str]:
+    LOGGER.debug(f"{allow_create=}")
+    if not raw_path:
+        return Err("No path provided")
```
rholinshead, line 98: How do we use the aiconfig-path that's provided to the cli command? Just testing out on the client now and the client still needs to give a path to load

jonathanlastmileai: Ah, yes, this is where I think it would be helpful to disambiguate load disk -> server and load server -> frontend. init_server_state runs the former, but of course you have to request the latter. I will overload /api/load as we discussed to do either 

(a) just server -> frontend (no path), or
(b) disk -> server, server -> frontend (path)

PR #586, opened 2023-12-22
[editor] add easy to use one linear to readme

Ankush-lastmile:
[editor] add easy to use one linear to readme

adding a one line command to readme for local testing. This can be updated and removed later

### DIFF
--- a/python/src/aiconfig/editor/README.md
+++ b/python/src/aiconfig/editor/README.md
@@ -2,6 +2,10 @@
 
 ## Usage (prod)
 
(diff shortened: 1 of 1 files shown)


jonathanlastmileai (approved):
The FBI won't come after you for setting an alias :D

Ankush-lastmile:
>The FBI won't come after you for setting an alias :D

I'll tell the newbies, but they might not sleep as soundly at night 😊

saqadri (approved):

rossdanlm (approved):

PR #588, opened 2023-12-22
[py][model parser][hf] 2/n callbacks

Ankush-lastmile:
[py][model parser][hf] 2/n callbacks

## What

- Added callbacks to model parser.
- It doesn't look like there is any other issues with this specific model parser at the moment. More investigation required into logging/callbacks, parameterization as seen with mixtral issues.

## Why

This model Parser has not been updated for a while. Taking this upto par with the rest of the library

## Testplan

Ran through the getting started. Modified the config to use HuggingFace model parser and mixtral model

Note: Input prompt said generate 10 activities, got 15.

Untitled drawing

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/588).
* #597
* __->__ #588
* #587

### DIFF
--- a/python/src/aiconfig/default_parsers/hf.py
+++ b/python/src/aiconfig/default_parsers/hf.py
@@ -183,6 +183,19 @@ class HuggingFaceTextGenerationParser(ParameterizedModelParser):
         Returns:
             str: Serialized representation of the prompt and inference settings.
         """
+        await ai_config.callback_manager.run_callbacks(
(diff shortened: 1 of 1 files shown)


Ankush-lastmile:
Testplan

Untitled drawing

saqadri (approved):

PR #589, opened 2023-12-22
[RFC] Output data type

saqadri:
[RFC] Output data type

Add some structure to the outputs that allow the frontend to know how to render them.

Still need to update the Python types, and modelparsers to return data in this type.

### DIFF
--- a/python/src/aiconfig/schema.py
+++ b/python/src/aiconfig/schema.py
@@ -11,13 +11,26 @@ JSONObject = Dict[str, Any]
 InferenceSettings = JSONObject
 
 
+class OutputData(BaseModel):
+    """
+    OutputData represents the output content in a standard format.
+    """
+
+    kind: Literal["string", "file_uri", "base64"]
+    value: str
+
+
 class ExecuteResult(BaseModel):
+    """
+    ExecuteResult represents the result of executing a prompt.
+    """
+
     # Type of output
     output_type: Literal["execute_result"]
     # nth choice.
     execution_count: Union[int, None] = None
     # The result of the executing prompt.
-    data: Any
+    data: Union[Any, OutputData]
     # The MIME type of the result. If not specified, the MIME type will be assumed to be plain text.
     mime_type: Optional[str] = None
     # Output metadata
@@ -25,6 +38,10 @@ class ExecuteResult(BaseModel):
 
 
 class Error(BaseModel):
+    """
+    Error represents an error that occurred while executing a prompt.
+    """
+
     # Type of output
     output_type: Literal["error"]
     # The name of the error
@@ -182,9 +199,7 @@ class AIConfig(BaseModel):
         Adds model settings to config level metadata
         """
         if model_name in self.metadata.models:
-            raise Exception(
-                f"Model '{model_name}' already exists. Use `update_model()`."
-            )
+            raise Exception(f"Model '{model_name}' already exists. Use `update_model()`.")
         self.metadata.models[model_name] = model_settings
 
     def delete_model(self, model_name: str):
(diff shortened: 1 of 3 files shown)


schema/aiconfig.schema.json
```diff
@@ -174,27 +191,6 @@
         }
       ]
     },
-    "Attachment": {
```
rholinshead, line 177: hmm, why the change to make this inline instead of its own definition?

schema/aiconfig.schema.json
```diff
@@ -157,15 +157,32 @@
...
+                    "additionalProperties": {}
+                  }
+                },
+                "required": [
+                  "data"
+                ]
               }
+            },
+            "data": {
+              "description": "Input to the model. This can represent a single input, or multiple inputs.\nThe structure of the data object is up to the ModelParser. For example,\na multi-modal ModelParser can choose to key the data by MIME type."
```
rholinshead, line 185: Why the change to the descriptions here? I think it's pretty confusing to reference mime-types here instead of having an opinionated stance on using attachments for different mime type data -- reading these descriptions I wouldn't know which to pick and it wouldn't seem to matter?

typescript/types.ts
```diff
@@ -192,8 +192,12 @@ export type ExecuteResult = {
   /**
    * The result of executing the prompt.
    */
-  data: JSONValue;
-
+  data:
+    | JSONValue
+    | {
```
rholinshead, line 197: The thing I'm not sure of here is if we should have this data associated with an Attachment so we can know its mimetype. Even if we have file_uri, we still need to know what to render it as (image, video, audio, etc.)

typescript/types.ts
```diff
@@ -192,8 +192,12 @@ export type ExecuteResult = {
...
+        kind: "string" | "file_uri" | "base64";
+        value: string;
+      };
```
rholinshead, line 200: Ah, right, I guess we just rely on the mime_type below to know the represented type

typescript/types.ts
```diff
@@ -192,8 +192,12 @@ export type ExecuteResult = {
   /**
    * The result of executing the prompt.
    */
-  data: JSONValue;
-
+  data:
+    | JSONValue
```
rholinshead, line 196: Long-term, do we want to support arbitrary JSONValue here or just let them dump things into metadata so we can always have a kind/value?

schema/aiconfig.schema.json
saqadri: I'm not sure -- this was autogenerated schema based on the types.ts

typescript/types.ts
saqadri: We can decide that. My 2c are it's still good to support arbitrary JSONValue here, but we can revisit that. It will require a breaking change to the schema too, so we can consider this for v2 schema

schema/aiconfig.schema.json
saqadri: It looks like we may have manually updated the schema last time, and this is being updated with the `genSchema` script.

schema/aiconfig.schema.json
saqadri: I think we didn't update the schema using the script for a while. I've updated the description to remove references to MIME types

python/src/aiconfig/schema.py
```diff
@@ -11,20 +11,37 @@
...
     mime_type: Optional[str] = None
     # Output metadata
     metadata: Dict[str, Any]
 
 
 class Error(BaseModel):
+    """
+    Error represents an error that occurred while executing a prompt.
+    """
+
```
saqadri, line 44: These are the only changes -- the rest are autoformatting changes

schema/aiconfig.schema.json
rholinshead: Ah, that was definitely me. I didn't realize there's a gen script. We should update the README to reference that and the process -- just make changes to types.ts and run `yarn gen-schema`?

rholinshead: I guess the gen script removes the definition/ref for some reason. Not a big deal

rholinshead (approved):

saqadri:
@rossdanlm I'm going to ship this -- can you take up the model parser updates to satisfy these types. Can start with a couple of them (maybe Dalle and GPT) to see what it's like. cc @rholinshead

PR #592, opened 2023-12-22
[ez] Auto-format typescript files

rossdanlm:
[ez] Auto-format typescript files

Doing this before next diff to keep things separate


rholinshead (approved):
LGTM. Good call doing this separate for better reviewability

saqadri (approved):

PR #593, opened 2023-12-22
[editor] Load AIConfig on Client from Python Server

rholinshead:
# [editor] Load AIConfig on Client from Python Server

Hit the `api/load` endpoint on the server port (with temporary hard-coded file path until we can load without path param) and ensure the editor loads:

Screenshot 2023-12-22 at 3 33 46 PM

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/prompt/prompt_outputs/PromptOutputsRenderer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/prompt/prompt_outputs/PromptOutputsRenderer.tsx
@@ -16,11 +16,12 @@ const ExecuteResultOutput = memo(function ExecuteResultOutput({
 }: {
   output: ClientExecuteResult;
 }) {
-  switch (output.renderData.type) {
-    case "text":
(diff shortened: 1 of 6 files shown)


python/src/aiconfig/editor/client/src/components/prompt/prompt_outputs/PromptOutputsRenderer.tsx
```diff
@@ -16,11 +16,12 @@ const ExecuteResultOutput = memo(function ExecuteResultOutput({
 }: {
   output: ClientExecuteResult;
 }) {
-  switch (output.renderData.type) {
-    case "text":
-      return <TextRenderer content={output.renderData.text} />;
-    // TODO: Handle other types of outputs
-  }
+  return null;
+  // switch (output.renderData.type) {
```
rholinshead, line 20: Just removing output rendering for now until we get the updated output types. Otherwise, we need to do the temporary renderData hack on the python server config and that's just wasted work

saqadri (approved):
Wooohoooo

PR #594, opened 2023-12-22
[AIC-py][editor] get endpoint: list_models

jonathanlastmileai:
[AIC-py][editor] get endpoint: list_models

- Name post response better (w. aiconfig)
- add get method (list models)

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/594).
* #595
* __->__ #594

### DIFF
--- a/python/src/aiconfig/editor/server/server.py
+++ b/python/src/aiconfig/editor/server/server.py
@@ -5,8 +5,8 @@ import result
 from aiconfig.Config import AIConfigRuntime
 from aiconfig.editor.server.server_utils import (
     EditServerConfig,
-    FlaskPostResponse,
-    HttpPostResponse,
+    FlaskResponse,
+    HttpResponseWithAIConfig,
     ServerMode,
     ServerState,
     ValidatedPath,
@@ -20,6 +20,8 @@ from aiconfig.editor.server.server_utils import (
 from aiconfig.model_parser import InferenceOptions
 from flask import Flask, request
 from result import Err, Ok, Result
+
+from aiconfig.registry import ModelParserRegistry
 
 logging.getLogger("werkzeug").disabled = True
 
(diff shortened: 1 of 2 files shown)


python/src/aiconfig/editor/server/server.py
```diff
@@ -122,42 +124,48 @@ async def run() -> FlaskPostResponse:
...
+        return HttpResponseWithAIConfig(
             message=f"Failed to add prompt: {err}",
             code=400,
             aiconfig=None,
         ).to_flask_format()
 
 
+@app.route("/api/list_models", methods=["GET"])
+def list_models() -> FlaskResponse:
+    out: list[str] = ModelParserRegistry.parser_ids()  # type: ignore
```
rholinshead, line 164: Is this the same as `ModelParserRegistry.display_parsers().keys()`? parser ids vs model names is still a bit confusing to me.

Also, can we support searching by substring here? 

I had done:
```
all_models: list[core_utils.JSONValue] = list(ModelParserRegistry.display_parsers().keys())
search = request_json.get("search", "")
models = [m for m in all_models if search in m]
```
which is currently erroring on bad request format but the general concept should be there (and I can scrap it in favour of your much better python code)

jonathanlastmileai: > parser ids vs model names is still a bit confusing to me.

Me too, dont know off the top of my head. @Ankush-lastmile ?

jonathanlastmileai: Ok just checked, this returns the mapping keys, which are the model names.

rholinshead (approved):
LGTM. Discussed search filtering offline. For now we'll do that client-side. For gradio workbook we can consider server side based on hugging face models

PR #595, opened 2023-12-22
[AIC-py][editor] Basic CRUD API

jonathanlastmileai:
[AIC-py][editor] Basic CRUD API

This does some refactoring to dedup and implements the core crud API
(load, save, run, add_prompt, update_prompt, delete_prompt)

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/595).
* __->__ #595
* #594

### DIFF
--- a/python/src/aiconfig/editor/server/server.py
+++ b/python/src/aiconfig/editor/server/server.py
@@ -1,4 +1,5 @@
 import logging
+from typing import Any, Type, TypeVar
 
 import lastmile_utils.lib.core.api as core_utils
 import result
@@ -7,6 +8,8 @@ from aiconfig.editor.server.server_utils import (
     EditServerConfig,
     FlaskResponse,
     HttpResponseWithAIConfig,
+    MethodName,
+    OpArgs,
     ServerMode,
     ServerState,
     ValidatedPath,
@@ -14,14 +17,18 @@ from aiconfig.editor.server.server_utils import (
     get_server_state,
     get_validated_path,
     init_server_state,
+    make_op_run_method,
+    run_aiconfig_operation_with_op_args,
+    run_aiconfig_operation_with_request_json,
     safe_load_from_disk,
-    safe_save_to_disk,
+    safe_run_aiconfig_static_method,
 )
 from aiconfig.model_parser import InferenceOptions
 from flask import Flask, request
 from result import Err, Ok, Result
 
 from aiconfig.registry import ModelParserRegistry
+from aiconfig.schema import Prompt
 
 logging.getLogger("werkzeug").disabled = True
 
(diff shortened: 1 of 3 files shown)


python/src/aiconfig/editor/server/server.py
```diff
@@ -49,137 +88,120 @@ def home():
...
+    def _op(aiconfig: AIConfigRuntime, op_args: OpArgs) -> Result[None, str]:
+        method_name = MethodName("save")
+        return safe_run_aiconfig_method(aiconfig, method_name, op_args)
+
+    path_val = _validated_request_path(request_json, allow_create=True)
+    op_args: Result[OpArgs, str] = result.do(
+        Ok(OpArgs({"path": path_val_ok}))
+        #
+        for path_val_ok in path_val
+    )
```
rholinshead, line 149: What is this doing? Specifically, isn't path_val just a string path? I'm confusing by the for ... in

python/src/aiconfig/editor/server/server.py
```diff
@@ -49,137 +88,120 @@ def home():
...
-        LOGGER.error(f"Failed to run: {err}")
-        return HttpResponseWithAIConfig(
-            message=f"Failed to run: {err}",
-            code=400,
-            aiconfig=None,
-        ).to_flask_format()
+
+    op_args = Ok(OpArgs({"prompt_name": prompt_name, "options": inference_options}))
+
+    return run_aiconfig_operation_with_op_args(aiconfig, "run", _op, op_args)
```
rholinshead, line 183: should this be MethodName("run")?

python/src/aiconfig/editor/server/server.py
```diff
@@ -49,137 +88,120 @@ def home():
...
-        LOGGER.error(f"Failed to add prompt: {err}")
-        return HttpResponseWithAIConfig(
-            message=f"Failed to add prompt: {err}",
-            code=400,
-            aiconfig=None,
-        ).to_flask_format()
+    method_name = MethodName("add_prompt")
+    signature: dict[str, Type[Any]] = {"prompt_name": str, "prompt_data": Prompt}
 
+    return _run_method_endpoint(method_name, signature)
```
rholinshead, line 191: nit, it's really confusing when to use safe_run_aiconfig_method , run_aiconfig_operation_with_op_args, run_aiconfig_operation_with_request_json, or _run_method_endpoint

They all sound pretty similar and it's hard to make sense of the difference in the code. Can we reduce this to one helper (or two at most if needed, but with clear documentation on when to use one or the other) that can be used everywhere here for calling aiconfig methods?

python/src/aiconfig/editor/server/server.py
jonathanlastmileai: So I distinguish between "method" and more general "operation". The former is for cases where we just forward the arguments to the corresponding method, and the latter is used in cases where the API endpoint has to do more logic. In this case, the extra logic is building the InferenceOptions. I can see why this is confusing. Suggestions for simplifying?

python/src/aiconfig/editor/server/server.py
jonathanlastmileai: Point taken. It's not immediately obvious to me how to simplify this without a lot of duplication and refactoring. Let's fix after. I'll document for now.

python/src/aiconfig/editor/server/server.py
jonathanlastmileai: `path_val` might actually be invalid: it's a Result[ValidatedPath, str] where the str contains the error message. This expression basically unboxes the result and returns the expression on 146 if it's valid, otherwise it short-circuits and returns the Err value. 

You can think of it as kind of like a Result comprehension, sort of similar to a list comprehension but for Result. 

The advantage is significant: I don't have to worry about exceptions, which are not checked statically even by strict type checking like Pyright. All Python exceptions are unchecked. Turning exceptions into Results allows Pyright to find unhandled errors, which really reduces debugging time in my experience. In practice, I'm finding that if my code type checks, it runs, and a large fraction of tricky bugs are cases where I had to suppress Pyright for one reason or another.

python/src/aiconfig/editor/server/server.py
jonathanlastmileai: Ok, I got it down to 2 with a little extra duplication and added docstrings to those 2.

python/src/aiconfig/editor/server/server.py
rholinshead: It could technically be simplified down to one method and just always parse out the JSON arguments into op args (if there's a clean way to do that, e.g. `OpArgs({...request_json})`

python/src/aiconfig/editor/server/server_utils.py
```diff
@@ -213,3 +220,127 @@ def init_server_state(app: Flask, edit_config: EditServerConfig) -> Result[None,
...
+    res_op_args: Result[OpArgs, str] = cast(Result[OpArgs, str], core_utils.result_reduce_dict_all_ok(operation_args_results))
+
+    return res_op_args
+
+
+def run_aiconfig_operation_with_request_json(
+    aiconfig: AIConfigRuntime | None,
+    request_json: core_utils.JSONObject,
+    operation_name: str,
+    operation: Operation[None],
```
rholinshead, line 325: Instead of passing operation_name string, couldn't this be inferred from the operation (or set the operation name as a static variable in the Operation instance to obtain here)

rholinshead (approved):
Accepting to unblock, just a nit on getting the operation name implicitly instead of needing to pass it.

I think the helper functions are better simplified here; at this point it would be a trade-off between having a single method (`run_aiconfig_operation_with_op_args`) and always calling `_validated_op_args_from_request_json` vs calling one of the two current methods (with `run_aiconfig_operation_with_request_json` doing the above logic). I kind of think the former would be clearer -- at the cost of some extra lines of code -- but I don't feel strongly enough about it to request changing it. Should be fine as-is

python/src/aiconfig/editor/server/server_utils.py
jonathanlastmileai: Yes, it could be bundled into the Operation or inferred. I just have it like this for straight-forward logging, but yes it could be refactored. Happy to do that as a follow up https://docs.google.com/document/d/1jUsV3Khu_Gs31dIBxh1n2DnF_AbnRPiN9etZpLe1dk8/edit

python/src/aiconfig/editor/server/server.py
jonathanlastmileai: >  just always parse out the JSON arguments into op args 

Sometimes this requires logic specific to the endpoint, like checking that a string is a readable fs path. We can pass unvalidated stuff down into AIConfig and let it raise, but then root causes and response error messages will be less clear.

jonathanlastmileai:
> I think the helper functions are better simplified here; at this point it would be a trade-off between having a single method (`run_aiconfig_operation_with_op_args`) and always calling `_validated_op_args_from_request_json` vs calling one of the two current methods (with `run_aiconfig_operation_with_request_json` doing the above logic). I kind of think the former would be clearer -- at the cost of some extra lines of code -- but I don't feel strongly enough about it to request changing it. Should be fine as-is

I dont have a strong opinion on this question. My inclination is to be defensive here and not rely too much on AIConfig's internal error handling, which is why I have the generalized validate-and-run-method path. But it's not a big deal either way as long as it works.

PR #597, opened 2023-12-23
[easy][extensions][py][hf] 3/n Local Inference model parser changes

Ankush-lastmile:
[easy][extensions][py][hf] 3/n Local Inference model parser changes

## What

This diff is a copy paste of the diff below except for one difference:
1. ModelParser `id` name and class name

HuggingFace Extension contains a `local_inference` and a `remote_inference_client`.

`Remote_inference_client` and the core huggingface model parser in the sdk are the same implementation with the one difference of model parser id. The extension model parser id is `HuggingFaceTextGenerationClient`.

Note: Both of these model parsers are for Text Generation only. Other huggingface tasks are not supported on the inference-endpoint by aiconfig yet.

## Why

Some context onto why there's two extensions, one was supposed to serve as a guide and easy to write model parser. This guide has yet to be written.

## Testplan

M

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/597).
* __->__ #597
* #588
* #587

### DIFF
--- a/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/remote_inference_client/text_generation.py
+++ b/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/remote_inference_client/text_generation.py
@@ -1,6 +1,10 @@
 import copy
 from typing import TYPE_CHECKING, Any, Dict, Optional
 
+from aiconfig.default_parsers.parameterized_model_parser import ParameterizedModelParser
+from aiconfig.model_parser import InferenceOptions
+from aiconfig.util.config_utils import get_api_key_from_environment
+
 # HuggingFace API imports
 from huggingface_hub import InferenceClient
 from huggingface_hub.inference._text_generation import (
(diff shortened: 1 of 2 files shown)


saqadri (approved):
I kinda liked the old name of the class before, but very minor nit on that. No changes necessary

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/remote_inference_client/text_generation.py
```diff
@@ -93,7 +92,8 @@ def construct_stream_output(
 
         index = 0  # HF Text Generation api doesn't support multiple outputs
         delta = data
-        options.stream_callback(delta, accumulated_message, index)
+        if options and options.stream_callback:
```
rossdanlm, line 95: Nice fix!

Issue #382, opened 2023-12-01
[py] PaLM Model Parser doesn't check for API Key


Ankush-lastmile:
PaLM model parser doesn't explicitly check for API key.

It currently needs to be manually configured by doing something like the following

```
import google.generativeai as palm
import os
palm.configure(api_key=os.getenv("PALM_KEY"))   
```

In this example the key must be an environment variable under the name "PALM_KEY"

Ankush-lastmile:
note: without an api key check, palm api throws an unrecognizable error.

PR #601, opened 2023-12-23
[issue #382][AIC-py] palm parser load env api key

jonathanlastmileai:
[issue #382][AIC-py] palm parser load env api key

### DIFF
--- a/python/src/aiconfig/default_parsers/palm.py
+++ b/python/src/aiconfig/default_parsers/palm.py
@@ -2,6 +2,7 @@ from typing import TYPE_CHECKING, Dict, List, Optional
 
 import google.generativeai as palm
 from aiconfig.default_parsers.parameterized_model_parser import ParameterizedModelParser
+from aiconfig.util.config_utils import get_api_key_from_environment
 from aiconfig.util.params import resolve_parameters, resolve_prompt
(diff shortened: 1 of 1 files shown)


python/src/aiconfig/default_parsers/palm.py
```diff
@@ -138,7 +139,7 @@ async def run_inference(
             )
         )
 
-        # TODO: check api key here
+        set_api_key("PALM_KEY")
```
saqadri, line 142: Is this the standard name for the key from palm docs?

saqadri (approved):

python/src/aiconfig/default_parsers/palm.py
JLessinger: It's GOOGLE_API_KEY for gemini and YOUR_API_KEY for palm (https://ai.google.dev/palm_docs/text_quickstart)

We should standardize this anyway instead of having the user rely on a random assortment of environment variables and hope

Ankush-lastmile: We could standardize this to use the same as Gemini, especially since Gemini will be a core model parser

Ankush-lastmile:
testplan?

PR #602, opened 2023-12-23
[RFC] Move editor to its own top-level directory

saqadri:
[RFC] Move editor to its own top-level directory

Not sure why it's under `python/src/aiconfig`


jonathanlastmileai:
It's in there because the most obvious way to package the flask server with the react files was to build everything into the python-aiconfig package. This might break the packaging and we'll have to set up a new way to ship the editor.

rossdanlm:
If we still want to do this, let's do it after we're done MVP. We're touching everywhere in it now so not a good time

jonathanlastmileai:
> If we still want to do this, let's do it after we're done MVP. We're touching everywhere in it now so not a good time

Yeah, I have this as a follow-up item here: https://docs.google.com/document/d/1YwUlIQZX8CiuvtbVR4OvpFQ2nphFITRD_XZR3CDNtoE/edit#bookmark=id.8h0jqqaolx9z

Hopefully we will find another way to package this that allows sensible code structure.

PR #603, opened 2023-12-23
[typescript] Save output.data with text content instead of response data

rossdanlm:
[typescript] Save output.data with text content instead of response data

This comes after Sarmad's schema updates in https://github.com/lastmile-ai/aiconfig/pull/589. To keep diffs small and easier to review, this simply converts from model-specific outputs --> pure text. I have a diff in https://github.com/lastmile-ai/aiconfig/pull/610 which converts from pure text --> `OutputData` format.

We only needed to update the `hf.py` and `openai.py`, because `palm.py` already returns output in the form of `string | null` type.

Ran yarn automated tests, but there aren't any specifically for openai. I also ran the typescript demos to make sure that they still work. Run these commands from `aiconfig` top-level dir:
```
```

For the extensions, we only have typescript for `hf.ts` (trivial: just changed `response` to `response.generated_text`), while `llama.ts` already outputs it in text format so no changes needed

## TODO
I still need to add function call support directly to `OutputData` format. See

### DIFF
--- a/extensions/HuggingFace/typescript/hf.ts
+++ b/extensions/HuggingFace/typescript/hf.ts
@@ -14,7 +14,7 @@ import {
   ExecuteResult,
   AIConfigRuntime,
   InferenceOptions,
-  CallbackEvent
+  CallbackEvent,
 } from "aiconfig";
 import _ from "lodash";
 import * as aiconfig from "aiconfig";
@@ -211,7 +211,7 @@ export class HuggingFaceTextGenerationModelParserExtension extends Parameterized
       const response = await this.hfClient.textGenerationStream(
         textGenerationArgs
       );
-      output = await ConstructStreamOutput(
+      output = await constructStreamOutput(
         response,
         options as InferenceOptions
       );
@@ -248,11 +248,19 @@ export class HuggingFaceTextGenerationModelParserExtension extends Parameterized
     }
 
     if (output.output_type === "execute_result") {
-      return (output.data as TextGenerationOutput | TextGenerationStreamOutput)
-        .generated_text as string;
-    } else {
-      return "";
+      if (typeof output.data === "string") {
+        return output.data;
+      }
+
+      // Doing this to be backwards-compatible with old output format
+      // where we used to save the response in output.data
+      if (output.data?.hasOwnProperty("generated_text")) {
+        return (
+          output.data as TextGenerationOutput | TextGenerationStreamOutput
+        ).generated_text as string;
+      }
     }
+    return "";
   }
 }
 
@@ -262,7 +270,7 @@ export class HuggingFaceTextGenerationModelParserExtension extends Parameterized
  * @param options
  * @returns
  */
-async function ConstructStreamOutput(
+async function constructStreamOutput(
   response: AsyncGenerator<TextGenerationStreamOutput>,
   options: InferenceOptions
 ): Promise<Output> {
@@ -280,6 +288,8 @@ async function ConstructStreamOutput(
 
     output = {
       output_type: "execute_result",
+      // TODO: Investigate if we should use the accumulated message instead
+      // of delta: https://github.com/lastmile-ai/aiconfig/issues/620
       data: delta,
       execution_count: index,
       metadata: metadata,
@@ -289,14 +299,11 @@ async function ConstructStreamOutput(
 }
 
 function constructOutput(response: TextGenerationOutput): Output {
-  const metadata = {};
-  const data = response;
-
   const output = {
     output_type: "execute_result",
-    data: data,
+    data: response.generated_text,
     execution_count: 0,
-    metadata: metadata,
+    metadata: { rawResponse: response },
   } as ExecuteResult;
 
   return output;
--- a/typescript/demo/function-call-stream.ts
+++ b/typescript/demo/function-call-stream.ts
@@ -217,18 +217,20 @@ async function functionCallingWithAIConfig() {
       return;
     }
 
-    const message = output.data as ChatCompletionMessageParam;
+    const rawResponse = output.metadata?.rawResponse;
+    const functionCall = rawResponse?.function_call;
+    console.log("functionCall=", functionCall);
 
     // If there is no function call, we're done and can exit this loop
-    if (!message.function_call) {
+    if (!functionCall) {
       return;
     }
 
     // If there is a function call, we generate a new message with the role 'function'.
-    const result = await callFunction(message.function_call);
+    const result = await callFunction(functionCall);
     const newMessage = {
       role: "function" as const,
-      name: message.function_call.name!,
+      name: functionCall.name!,
       content: JSON.stringify(result),
     };
 
--- a/typescript/lib/parsers/hf.ts
+++ b/typescript/lib/parsers/hf.ts
@@ -203,7 +203,7 @@ export class HuggingFaceTextGenerationParser extends ParameterizedModelParser<Te
       const response = await this.hfClient.textGenerationStream(
         textGenerationArgs
       );
-      output = await ConstructStreamOutput(
+      output = await constructStreamOutput(
         response,
         options as InferenceOptions
       );
@@ -240,11 +240,19 @@ export class HuggingFaceTextGenerationParser extends ParameterizedModelParser<Te
     }
 
     if (output.output_type === "execute_result") {
-      return (output.data as TextGenerationOutput | TextGenerationStreamOutput)
-        .generated_text as string;
-    } else {
-      return "";
+      if (typeof output.data === "string") {
+        return output.data;
+      }
+
+      // Doing this to be backwards-compatible with old output format
+      // where we used to save the response in output.data
+      if (output.data?.hasOwnProperty("generated_text")) {
+        return (
+          output.data as TextGenerationOutput | TextGenerationStreamOutput
+        ).generated_text as string;
+      }
     }
+    return "";
   }
 }
 
@@ -254,7 +262,7 @@ export class HuggingFaceTextGenerationParser extends ParameterizedModelParser<Te
  * @param options
  * @returns
  */
-async function ConstructStreamOutput(
+async function constructStreamOutput(
   response: AsyncGenerator<TextGenerationStreamOutput>,
   options: InferenceOptions
 ): Promise<Output> {
@@ -272,25 +280,23 @@ async function ConstructStreamOutput(
 
     output = {
       output_type: "execute_result",
+      // TODO: Investigate if we should use the accumulated message instead
+      // of delta: https://github.com/lastmile-ai/aiconfig/issues/620
       data: delta,
       execution_count: index,
-      metadata: metadata,
+      metadata: { metadata },
     } as ExecuteResult;
   }
   return output;
 }
 
 function constructOutput(response: TextGenerationOutput): Output {
-  const metadata = {};
-  const data = response;
-
   const output = {
     output_type: "execute_result",
-    data: data,
+    data: response.generated_text,
     execution_count: 0,
-    metadata: metadata,
+    metadata: { rawResponse: response },
   } as ExecuteResult;
-
   return output;
 }
 
--- a/typescript/lib/parsers/openai.ts
+++ b/typescript/lib/parsers/openai.ts
@@ -170,7 +170,6 @@ export class OpenAIModelParser extends ParameterizedModelParser<CompletionCreate
 
       // Save response as Output(s) in the Prompt
       const outputs: ExecuteResult[] = [];
-      const responseWithoutChoices = _.omit(response, "choices");
       for (const choice of response.choices) {
         const output: ExecuteResult = {
           output_type: "execute_result",
@@ -179,7 +178,7 @@ export class OpenAIModelParser extends ParameterizedModelParser<CompletionCreate
           metadata: {
             finish_reason: choice.finish_reason,
             logprobs: choice.logprobs,
-            ...responseWithoutChoices,
+            rawResponse: response,
           },
         };
         outputs.push(output);
@@ -253,10 +252,26 @@ export class OpenAIModelParser extends ParameterizedModelParser<CompletionCreate
     }
 
     if (output.output_type === "execute_result") {
-      return output.data as string;
-    } else {
-      return "";
+      // TODO: Add in OutputData another way to support function calls
+      if (typeof output.data === "string") {
+        return output.data;
+      }
+
+      // Doing this to be backwards-compatible with old output format
+      // where we used to save the ChatCompletionMessageParam in output.data
+      if (
+        output.data?.hasOwnProperty("content") &&
+        output.data?.hasOwnProperty("role")
+      ) {
+        const message = output.data as Chat.ChatCompletionMessageParam;
+        if (message.content != null) {
+          return message.content;
+        } else if (message.function_call) {
+          return JSON.stringify(message.function_call);
+        }
+      }
     }
+    return "";
   }
 }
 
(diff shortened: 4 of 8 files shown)


extensions/HuggingFace/typescript/hf.ts
```diff
@@ -280,7 +279,7 @@ async function ConstructStreamOutput(
 
     output = {
       output_type: "execute_result",
-      data: delta,
+      data: accumulatedMessage,
```
rholinshead, line 282: Did this need to be changed? It was already string. It's also fairly easy to handle streamed output chunks from the delta instead of having progressively larger outputs like this. 
We could also add some metadata to denote the final chunk and/or have the final output at the end of the stream which contains the entire accumulated message once all chunks are completed (that's what we do in workbooks)

extensions/HuggingFace/typescript/hf.ts
```diff
@@ -294,7 +293,7 @@ function constructOutput(response: TextGenerationOutput): Output {
 
   const output = {
     output_type: "execute_result",
-    data: data,
+    data: data.generated_text,
     execution_count: 0,
     metadata: metadata,
```
rholinshead, line 298: Should have metadata here as well
```
const data = response.generated_text;
const metadata = _.omit(response, ["generated_text"]);
```

saqadri:
FYI to run `demo.ts`, simply run 
```
ts-node demo.ts
```

in that folder. You may need to install ts-node if you don't already have it, from here: https://www.npmjs.com/package/ts-node#installation

typescript/lib/parsers/openai.ts
```diff
@@ -378,7 +379,8 @@ export class OpenAIChatModelParser extends ParameterizedModelParser<Chat.ChatCom
               ? [
                   {
                     output_type: "execute_result",
-                    data: { ...assistantResponse },
+                    data: assistantResponse.content,
+                    metadata: { ...responseWithoutContent },
```
rholinshead, line 383: no need to spread, it's already a new object
`metadata: responseWithoutContent`

extensions/HuggingFace/typescript/hf.ts
```diff
@@ -294,7 +293,7 @@ function constructOutput(response: TextGenerationOutput): Output {
 
   const output = {
     output_type: "execute_result",
-    data: data,
+    data: data.generated_text,
```
saqadri, line 296: Is there anything else that's in `data` other than `generated_text`?

It may make sense to put the raw response in the metadata, perhaps as a `rawResponse` property

typescript/lib/parsers/hf.ts
```diff
@@ -281,16 +280,12 @@ async function ConstructStreamOutput(
...
-  const metadata = {};
-  const data = response;
-
   const output = {
     output_type: "execute_result",
-    data: data,
+    data: response.generated_text,
     execution_count: 0,
-    metadata: metadata,
+    metadata: _.omit(response, ["generated_text"]),
```
saqadri, line 287: Same comment as above

typescript/lib/parsers/openai.ts
```diff
@@ -536,13 +538,15 @@ export class OpenAIChatModelParser extends ParameterizedModelParser<Chat.ChatCom
...
+        const messageWithoutContent = _.omit(choice.message, ["content"]);
         const output: ExecuteResult = {
           output_type: "execute_result",
-          data: { ...choice.message },
+          data: choice.message?.content,
           execution_count: choice.index,
           metadata: {
             finish_reason: choice.finish_reason,
             ...responseWithoutChoices,
+            ...messageWithoutContent,
```
saqadri, line 549: I'm concerned the metadata won't make much sense if message is saved without content here. I'm guessing this would just be "role" then? Need to think about it some more but fine for now

typescript/lib/parsers/openai.ts
```diff
@@ -625,11 +633,10 @@ export class OpenAIChatModelParser extends ParameterizedModelParser<Chat.ChatCom
...
     if (output.output_type === "execute_result") {
-      const message = output.data as Chat.ChatCompletionMessageParam;
-      if (message.content != null) {
-        return message.content;
-      } else if (message.function_call) {
-        return JSON.stringify(message.function_call);
+      if (typeof output.data === "string") {
+        return output.data;
+      } else if (output.metadata?.function_call) {
+        return JSON.stringify(output.metadata?.function_call);
```
saqadri, line 639: This logic update will break existing aiconfigs. Since this library has already been released, we need to be explicit about breaking changes before making them. As such, it's best to keep the existing logic in addition to your update.

i.e.

```
if (typeof output.data === "string") {
  return output.data;
} else if (output.metadata?.function_call) {
  return JSON.stringify(output.metadata?.function_call);
} else {
  const message = output.data as Chat.ChatCompletionMessageParam;
  if (message?.content != null) {
    return message.content;
  } else if (message.function_call) {
    return JSON.stringify(message.function_call);
  } else {
    return ""
  }
```

rholinshead:
Can you make sure to test the function calling example, 

```
npx ts-node demo/function-call-stream.ts
```
I think it could be broken by the change to move it to metadata here?


Also, just curious why you chose to do this first and then move to the `OutputData` type in #610 instead of doing it in one go? I agree that using OutputData with 'string' kind is the correct way to go

typescript/lib/parsers/openai.ts
```diff
@@ -671,11 +678,17 @@ export class OpenAIChatModelParser extends ParameterizedModelParser<Chat.ChatCom
     const output = aiConfig.getLatestOutput(prompt);
     if (output != null) {
       if (output.output_type === "execute_result") {
-        const outputMessage =
-          output.data as unknown as Chat.ChatCompletionMessageParam;
         // If the prompt has output saved, add it to the messages array
-        if (outputMessage.role === "assistant") {
-          messages.push(outputMessage);
+        if (output.metadata?.role === "assistant") {
```
saqadri, line 682: Same as above, we cannot have breaking changes like this without a warning to users or some way of converting the existing format to the new.

typescript/lib/parsers/openai.ts
```diff
@@ -671,11 +678,17 @@ export class OpenAIChatModelParser extends ParameterizedModelParser<Chat.ChatCom
...
+          if (typeof output.data === "string") {
+            messages.push({
+              content: output.data,
+              role: output.metadata?.role,
+              function_call: output.metadata?.function_call,
+              name: output.metadata?.name,
+            });
+          }
+          // TODO (rossdanlm): Support function_call
```
saqadri, line 691: What breaks with function calling for this?

typescript/lib/parsers/openai.ts
saqadri: It may make sense to just have a `message`, `rawMessage` or `rawResponse` property in metadata that stores the entire raw response.

saqadri (changes requested):
Requesting changes on the breaking changes introduced here.

typescript/lib/parsers/openai.ts
rholinshead: @saqadri wouldn't that be addressed by a major version bump?

saqadri: Until we do a major version bump, we have to support both. And even when we do a major version bump, ideally we enable a "conversion" in this code path. Can discuss separately, but right now it seems like we introduced a breaking change implicitly

saqadri:
> I agree that using OutputData with 'string' kind is the correct way to go

@rholinshead is that significantly different on the frontend? It would be a much simpler JSON/YAML file for string outputs if the "string" kind was inferred if the typeof output.data was string.

extensions/HuggingFace/typescript/hf.ts
rossdanlm: In our case we don't need it because there's no other field beside `generated_text` in the response, but yea good idea for future proofing

extensions/HuggingFace/typescript/hf.ts
rossdanlm: There's nothing else

rossdanlm:
> Can you make sure to test the function calling example,
> 
> ```
> npx ts-node demo/function-call-stream.ts
> ```
> 
> I think it could be broken by the change to move it to metadata here?
> 
> Also, just curious why you chose to do this first and then move to the `OutputData` type in #610 instead of doing it in one go? I agree that using OutputData with 'string' kind is the correct way to go

cc thanks @rholinshead for flagging, you are totally correct this breaks because of the way we implement it in places like this, but moving forwards this isn't something we should continue to support and we should use the function call directly instead from `OutputData` or metadata
Screenshot 2023-12-26 at 16 00 00

@saqadri I know this breaks existing code API implementation that may be calling into this, but we can't both choose to support old formats while also wanting to make our output types easy to parse and understand. The `getOutputText()` methods are still supported, it's just if anyone is engaging with the old data structure their implementation will break

saqadri:
> > Can you make sure to test the function calling example,
> 
> > 
> 
> > ```
> 
> > npx ts-node demo/function-call-stream.ts
> 
> > ```
> 
> > 
> 
> > I think it could be broken by the change to move it to metadata here?
> 
> > 
> 
> > Also, just curious why you chose to do this first and then move to the `OutputData` type in #610 instead of doing it in one go? I agree that using OutputData with 'string' kind is the correct way to go
> 
> 
> 
> cc thanks @rholinshead for flagging, you are totally correct this breaks because of the way we implement it in places like this, but moving forwards this isn't something we should continue to support and we should use the function call directly instead from `OutputData` or metadata
> 
> Screenshot 2023-12-26 at 16 00 00
> 
> 
> 
> @saqadri I know this breaks existing code API implementation that may be calling into this, but we can't both choose to support old formats while also wanting to make our output types easy to parse and understand. The `getOutputText()` methods are still supported, it's just if anyone is engaging with the old data structure their implementation will break
> 
> 

Can you share how the function call script should be updated to support the new format? One other option can be adding output kind "function" to the schema to it's distinguished from plain string.

We cannot have breaking changes without explicitly mentioning it and having a clear major version bump. The issue is there are several implicit breaking changes introduced in this stack

rossdanlm:
> Can you share how the function call script should be updated to support the new format? One other option can be adding output kind "function" to the schema to it's distinguished from plain string.

Saw your comments, I didn't have time to update the diff summary before you replied but just did. Basically we need to do this: 
Screenshot 2023-12-26 at 16 14 14
I updated the change to the `function-call-stream.ts` file

> We cannot have breaking changes without explicitly mentioning it and having a clear major version bump. The issue is there are several implicit breaking changes introduced in this stack
Alright, I'm in favour of doing a clear major version (I haven't posted changelog yet for this reason). Let's also include in this NOT to support arbitrary JSON in OutputData schema.

rossdanlm:
Updated based on feedback, main changes:
- sending `rawResponse` as a field in metadata for all non-streaming options
- no longer using `delta` instead of `accumulatedData`, created task for it: https://github.com/lastmile-ai/aiconfig/issues/620
- Made note to support function calling more directly as part of `OutputData`. I will do this end of diff stack
- Supported all old AIConfig schemas with backwards compatbility inside of the `getOutputText()` methods

Note that this change will introduce breaking the SDK implementation. Ex: instead of check for `data.content` you now just check `data` instead. I will land changelog for today before landing these changes, so this won't be included or mentioned in that

rholinshead:
> I agree that using OutputData with 'string' kind is the correct way to go

> @rholinshead is that significantly different on the frontend? It would be a much simpler JSON/YAML file for string outputs if the "string" kind was inferred if the typeof output.data was string.

It isn't significantly different on the front end, it's just checking if the type is a string or an object first instead of just checking the keys of the object. In general, was just thinking it is a much cleaner API to get rid of the redundant union types (i.e. can represent string output in 2 supported ways) with no clear standard or recommendation.

If it's a lot nicer for the JSON/YAML format then maybe it's worth supporting raw string there, we should just consider:
- what is our recommendation / standard for how we think it should be used? All our example configs should be consistent
- do we even want to consider {kind: 'string'} then, or just remove it to have the single way (raw text as the data value for text outputs)

rholinshead:
> Let's also include in this NOT to support arbitrary JSON in OutputData schema.

Based on previous PR discussion with @saqadri I think we should still support arbitrary JSON for now and just provide the new structured types. We can revisit later to make a better informed decision with more data (e.g. once we see more examples of user-created parsers to know how they're using it)

typescript/demo/function-call-stream.ts
```diff
@@ -217,18 +209,20 @@ async function functionCallingWithAIConfig() {
       return;
     }
 
-    const message = output.data as ChatCompletionMessageParam;
+    const rawResponse = output.metadata?.rawResponse;
+    const function_call = rawResponse?.function_call;
+    console.log("function_call=", function_call);
```
rholinshead, line 214: nit, can remove

typescript/lib/parsers/openai.ts
```diff
@@ -671,11 +684,17 @@ export class OpenAIChatModelParser extends ParameterizedModelParser<Chat.ChatCom
     const output = aiConfig.getLatestOutput(prompt);
     if (output != null) {
       if (output.output_type === "execute_result") {
-        const outputMessage =
-          output.data as unknown as Chat.ChatCompletionMessageParam;
         // If the prompt has output saved, add it to the messages array
-        if (outputMessage.role === "assistant") {
-          messages.push(outputMessage);
+        if (output.metadata?.role === "assistant") {
```
rholinshead, line 688: I think we still need to handle the existing outputMessage-as-ChatCompletionMessageParam case here for backwards compatibility on deserialize. We can reserialize in our updated format, but it's possible the config being loaded has the old format

I think you can do similar check to above

```
if (typeof output.data === "string") {
   ... handle how you do here
} else if (output.data.hasOwnProperty("role")) {
  // for backwards compatibility
  const outputMessage = output.data as unknown as Chat.ChatCompletionMessageParam;
  if (outputMessage.role === "assistant") {
    messages.push(outputMessage);
  }
}

Comment (rossdanlm) [reply typescript/lib/parsers/openai.ts:688]: You are correct, thank you for catching!

File: typescript/demo/function-call-stream.ts (hunk 10 above)
Comment (rossdanlm) [reply typescript/demo/function-call-stream.ts:214]: I wanted to keep it because we have other `console.logs()` earlier and it shows more clearly in the demo how function_call is being used. Ex: https://github.com/lastmile-ai/aiconfig/blob/main/typescript/demo/function-call-stream.ts#L176

File: typescript/demo/function-call-stream.ts [hunk 12]
```diff
@@ -217,18 +209,20 @@ async function functionCallingWithAIConfig() {
       return;
     }
 
-    const message = output.data as ChatCompletionMessageParam;
+    const rawResponse = output.metadata?.rawResponse;
+    const function_call = rawResponse?.function_call;
```
saqadri, line 213: I don't think this is a good model. If the only way for someone to support function calling is by delving into the raw response, then we've missed a critical scenario.

rossdanlm: Ok after rebasing onto https://github.com/lastmile-ai/aiconfig/pull/628, I will update but do this at the end of the diff stack. I just want to keep this for string-only to make it easier to review, then the rest of diffs will be easier to followup

saqadri (approved):
Approving to unblock :). We are aligned on the acceptable breaking changes to the SDK while still supporting previous aiconfig schemas.

Also, as you land this stack please document the breaking changes explicitly in our changelog doc for next week so we can communicate it explicitly

typescript/package-lock.json
saqadri, line 1: Delete this

PR #605, opened 2023-12-24
[python][non-image]Save output.data with text content instead of response data

rossdanlm:
[python][non-image]Save output.data with text content instead of response data

Same as last diff except now for Python.

Going to do the image outputs next diff because those actually require updating the data as a key-value pair with kinds and values

For extensions,
- Gemini and Llama-Guard are already in text output format, so not need to do any changes
- updated Llama and the HuggingFace ones (non-image)

## Test plan
Updated automated tests

### DIFF
--- a/extensions/llama/python/llama.py
+++ b/extensions/llama/python/llama.py
@@ -83,7 +83,11 @@ class LlamaModelParser(ParameterizedModelParser):
                 if options:
                     options.stream_callback(data, acc, index)
             print(flush=True)
-            return ExecuteResult(output_type="execute_result", data=[acc], metadata={})
+            return ExecuteResult(
+                output_type="execute_result",
+                data=acc,
+                metadata={}
+            )
         else:
             response = llm(model_input)
             try:
@@ -91,13 +95,25 @@ class LlamaModelParser(ParameterizedModelParser):
             except TypeError:
                 texts = [response["choices"][0]["text"]]
 
-            return ExecuteResult(output_type="execute_result", data=texts, metadata={})
+            return ExecuteResult(
+                output_type="execute_result",
+                data=texts[-1],
+                metadata={}
+            )
 
     def get_output_text(
         self, prompt: Prompt, aiconfig: AIConfigRuntime, output: Output | None = None
     ) -> str:
         match output:
             case ExecuteResult(data=d):
-                return d
+                if isinstance(d, str):
+                    return d
+
+                # Doing this to be backwards-compatible with old output format
+                # where we used to save a list in output.data
+                # See https://github.com/lastmile-ai/aiconfig/issues/630
+                elif isinstance(d, list):
+                    return d
+                return ""
             case _:
                 raise ValueError(f"Unexpected output type: {type(output)}")
--- a/python/src/aiconfig/default_parsers/openai.py
+++ b/python/src/aiconfig/default_parsers/openai.py
@@ -3,17 +3,17 @@ from abc import abstractmethod
 from typing import TYPE_CHECKING, Dict, List, Optional, Union
 
 import openai
+from openai.types.chat import ChatCompletionMessage
 from aiconfig.callback import CallbackEvent
 from aiconfig.default_parsers.parameterized_model_parser import ParameterizedModelParser
 from aiconfig.model_parser import InferenceOptions
+from aiconfig.schema import ExecuteResult, Output, Prompt, PromptInput, PromptMetadata
 from aiconfig.util.config_utils import get_api_key_from_environment
 from aiconfig.util.params import (
     resolve_prompt,
     resolve_prompt_string,
     resolve_system_prompt,
 )
-
-from aiconfig.schema import ExecuteResult, Output, Prompt, PromptInput, PromptMetadata
 
 if TYPE_CHECKING:
     from aiconfig.Config import AIConfigRuntime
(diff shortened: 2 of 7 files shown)


extensions/llama/python/llama.py
```diff
@@ -83,21 +83,31 @@ async def _run_inference_helper(self, model_input, options) -> ExecuteResult:
...
             response = llm(model_input)
             try:
                 texts = [r["choices"][0]["text"] for r in response]
             except TypeError:
                 texts = [response["choices"][0]["text"]]
 
-            return ExecuteResult(output_type="execute_result", data=texts, metadata={})
+            return ExecuteResult(
+                output_type="execute_result",
+                data=texts[0],
```
saqadri, line 100: I'm not sure about this part -- what about the rest of the items in the `texts` array?

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -106,6 +106,17 @@ async def serialize(
...
 
+                assistant_output = []
+                if assistant_response is not None:
+                    assistant_content = assistant_response.pop("content", "")
+                    assistant_output = [
+                        ExecuteResult(
+                            output_type="execute_result",
+                            execution_count=None,
+                            data=assistant_content,
+                            metadata={**assistant_response},
```
saqadri, line 117: It's important to ensure the output produced by TS and Python SDKs is the same. This logic seems to produce different things in the metadata between the SDKs. If that is intentional, it should be explicitly documented

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -116,16 +127,7 @@ async def serialize(
...
-                    outputs=[
-                        ExecuteResult(
-                            output_type="execute_result",
-                            execution_count=None,
-                            data=assistant_response,
-                            metadata={},
-                        )
-                    ]
-                    if assistant_response
-                    else [],
```
saqadri, line 128: What about the case where there's no assistant_response? We still want outputs to be set to an empty array, right?

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -493,9 +496,21 @@ def add_prompt_as_message(
     output = aiconfig.get_latest_output(prompt)
     if output:
         if output.output_type == "execute_result":
-            output_message = output.data
-            if output_message["role"] == "assistant":
```
saqadri, line 497: This still needs to be checked for backwards compatibility.

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -493,9 +496,21 @@ def add_prompt_as_message(
...
-                messages.append(output_message)
+            if output.metadata.get("role", "") == "assistant":
+                output_data = output.data
+                if isinstance(output_data, str):
+                    output_message = {
+                        "content": output_data,
+                        "role": "assistant",
```
saqadri, line 504: Why is `"role": "assistant"` hardcoded here?

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -493,9 +496,21 @@ def add_prompt_as_message(
...
+                    }
+                    function_call = output.metadata.get("function_call", None)
```
saqadri, line 506: I'm worried this logic is more brittle now than the previous. If we store raw response in metadata, we could probably use that instead? But ok for now

python/src/aiconfig/default_parsers/palm.py
```diff
@@ -338,10 +341,14 @@ async def run_inference(
         response = palm.chat(**completion_data)
         outputs = []
         for i, candidate in enumerate(response.candidates):
+            # candidate is a MessageDict obj (https://shorturl.at/jKY35), 
+            # but Pydantic TypedDict breaks for Python v<3.12  so using
+            # generic Dict type
+            candidate: Dict
```
saqadri, line 347: If there's anything other than `candidate.content` in `candidate`, then we should store it in the metadata

saqadri (changes requested):
Requesting changes for the breaking changes introduced

saqadri (approved):
Approving to unblock -- many of the things we discussed in the TS diff apply here too, so trust we'll apply the same to the python side. Beyond those comments, ship it!

extensions/llama/python/llama.py
rossdanlm: This is a good callout. I dug deeper and realized that this extension stores all the data under a single prompt output, vs out other chat-bot models which stores each message as it's own individual prompt. Ex: https://github.com/lastmile-ai/aiconfig/blob/main/python/src/aiconfig/default_parsers/openai.py#L93-L130

For now I'll use `texts[-1]` to ensure we get the most recent conversation message, but in general I think this extension (and associated cookbook) will need to be refactored to match our other ways of storing message history as 1-1 mapping to prompt. Llama also supports chat completion params so this shouldn't take too long: https://github.com/abetlen/llama-cpp-python/blob/main/llama_cpp/llama.py#L791-L796

I made GH issue for it in

python/src/aiconfig/default_parsers/openai.py
rossdanlm: All our output chats can only exist if they have the `role: assistant` as defined. This means that the model is the one who "answered"

Code pointer: https://github.com/lastmile-ai/aiconfig/blob/main/python/src/aiconfig/default_parsers/openai.py#L100-L127

python/src/aiconfig/default_parsers/openai.py
rossdanlm: I made changes so that `assistant_output = []` is initialized and only updated when `assistant_response is not None`

python/src/aiconfig/default_parsers/openai.py
rossdanlm: Thanks for the callout, will update so that they're both `metadata: {rawResponse: assistantResponse}`

Also yea I know this isn't handling function calling yet, but I'll update in future diff at the end of stack. Sarmad and I discussed offline but sharing so others can see, we want metadata eventually to be something like this:

```
metadata: {rawResponse: response, ..._.omit(response, ["contents"])}
```
while using `function_call` kind in the output.data: https://github.com/lastmile-ai/aiconfig/pull/628

PR #607, opened 2023-12-24
[ez] Rename llama extension to aiconfig_extension_llama

rossdanlm:
[ez] Rename llama extension to aiconfig_extension_llama

TSIA, we shouldn't be making this as part of the generalized `python_aiconfig` module since this is an extension

Will publish this package after it's been stamped

### DIFF
--- a/cookbooks/llama/python/README.md
+++ b/cookbooks/llama/python/README.md
@@ -8,7 +8,7 @@ All GGUF LLaMA models from https://huggingface.co/TheBloke?search_models=gguf sh
 ## Install
 
 ```
-pip install python-aiconfig python-aiconfig-llama
+pip install python-aiconfig aiconfig-extension-llama
 ```
 
 ## Usage:
--- a/extensions/llama/python/pyproject.toml
+++ b/extensions/llama/python/pyproject.toml
@@ -2,7 +2,7 @@
 requires = ["setuptools", "wheel"]
 
 [project]
-name = "python-aiconfig"
+name = "aiconfig_extension_llama"
 version = "0.0.1"
 authors = [
   { name="Jonathan Lessinger", email="jonathan@lastmileai.dev" },


cookbooks/llama/python/README.md
```diff
@@ -8,7 +8,7 @@ All GGUF LLaMA models from https://huggingface.co/TheBloke?search_models=gguf sh
 ## Install
 
 ```
-pip install python-aiconfig python-aiconfig-llama
+pip install python-aiconfig aiconfig-extension-llama
```
jonathanlastmileai, line 11: I think this will just create confusion. Why do we need to rename from `python-aiconfig-llama` ?

jonathanlastmileai (changes requested):

cookbooks/llama/python/README.md
rossdanlm: This extension is not called `python-aiconfig-llama`, it is called `python-aiconfig`. We already have that for our core library and this is an extension. We have other extensions follow the format of `aiconfig-extension-XYZ`
Screenshot 2023-12-25 at 16 21 11

saqadri (approved):

cookbooks/llama/python/README.md
rossdanlm: Just for extra context, this is where the `python-aiconfig` is defined for our main python library: https://github.com/lastmile-ai/aiconfig/blob/main/python/pyproject.toml#L5

saqadri: Let's push this change. It's very straightforward and doesn't make sense to have been blocked for so long.

If we request changes, please have a fast turnaround on unblocking the diff once comments are addressed.

saqadri (approved):

cookbooks/llama/python/README.md
jonathanlastmileai: I think this will break the cookbook readme. The extension package is (for better or worse) called `python-aiconfig-llama` and there is no `aiconfig-extension-llama`. The immediate fix for extensions/llama/python/pyproject.toml would be to update the name to the actual existing pypi package, "python-aiconfig-llama". 

Also, even if we updated it to aiconfig_extension_llama and published a new package with that name, I don't think pip install aiconfig-extension-llama would work (underscore vs hyphen). 

What we can do is publish a new package consistent with the other extensions, e.g. aiconfig-extension-llama, and then update the project file and README. 

Isn't it better to postpone cosmetic changes, especially if risky, until editor ships?

jonathanlastmileai (changes requested):

cookbooks/llama/python/README.md
saqadri: @jonathanlastmileai this is the same model that every other extension uses. pyproject.toml doesn't allow hyphens, but when you install that package, it is installed as hyphens instead of underscores. This change will work fine.

Please look at other extensions for examples.

Let's ship this.

jonathanlastmileai (approved):

jonathanlastmileai:
Unblocking. Let's make sure to publish the package before merging.

rossdanlm:
Cool thanks, published this in https://pypi.org/project/aiconfig-extension-llama/

PR #608, opened 2023-12-24
[python][image] Save output.data images in OutputData format, not just strings

rossdanlm:
[python][image] Save output.data images in OutputData format, not just strings

This will make it easy for Ryan on frontend to know exactly what data type we need. This is also backwards compatible for existing AIConfigs that have data saved only as a text string

##test
Automated tests

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/608).
* __->__ #608
* #632

### DIFF
--- a/python/src/aiconfig/default_parsers/dalle.py
+++ b/python/src/aiconfig/default_parsers/dalle.py
@@ -97,7 +97,7 @@ class DalleImageGenerationParser(ParameterizedModelParser):
         ai_config: "AIConfigRuntime",
         parameters: Optional[Dict] = None,
         **kwargs,
-    ) -> Prompt:
+    ) -> List[Prompt]:
         """
         Defines how a prompt and model inference settings get serialized in the .aiconfig.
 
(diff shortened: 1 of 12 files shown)


python/tests/parsers/test_dalle_parser.py
```diff
@@ -14,11 +14,10 @@ async def test_serialize_basic(set_temporary_env_vars: None):
         "size": "1024x1024",
     }
     aiconfig = AIConfigRuntime.create()
-    serialized_prompts = await aiconfig.serialize(
+    serialized_prompt = await aiconfig.serialize(
```
saqadri, line 17: Doesn't `serialize` return an array?

python/src/aiconfig/default_parsers/dalle.py
```diff
@@ -124,7 +138,7 @@ async def serialize(
                 model=model_metadata, parameters=parameters, **kwargs
             ),
         )
-        return [prompt]
+        return prompt
```
saqadri, line 141: Hmmmm, I think we should return an array always from `serialize`

saqadri (approved):
Looks good, just fix the return type for `serialize`

python/src/aiconfig/default_parsers/dalle.py
rossdanlm: Thanks for clarifying, did diff here to fix issue: https://github.com/lastmile-ai/aiconfig/pull/632

PR #610, opened 2023-12-25
[typescript] Use OutputDataWithValue type for function calls

rossdanlm:
[typescript] Use OutputDataWithValue type for function calls

After discussion and based on comments, we are going to keep pure string form as is, but still need to support non-pure string types to make sure that they're supported. Specifically images (which we don't have on ts) and function calls.

I didn't need to update the hugging face files because those already are just pure text

##Test plan
Pass automated tests, running these commands from `aiconfig` top-level dir
```
```

Also run `yarn test` from typescript dir

### DIFF
--- a/extensions/HuggingFace/typescript/hf.ts
+++ b/extensions/HuggingFace/typescript/hf.ts
@@ -6,15 +6,16 @@ import {
 } from "@huggingface/inference";
 
 import {
-  Prompt,
-  Output,
-  PromptInput,
-  ParameterizedModelParser,
-  ModelMetadata,
-  ExecuteResult,
   AIConfigRuntime,
-  InferenceOptions,
   CallbackEvent,
+  ExecuteResult,
+  InferenceOptions,
+  ModelMetadata,
+  Output,
+  OutputDataWithValue,
+  ParameterizedModelParser,
+  Prompt,
+  PromptInput,
 } from "aiconfig";
 import _ from "lodash";
 import * as aiconfig from "aiconfig";
@@ -252,6 +253,16 @@ export class HuggingFaceTextGenerationModelParserExtension extends Parameterized
         return output.data;
       }
 
+      if (output.data?.hasOwnProperty("value")) {
+        const outputData = output.data as OutputDataWithValue;
+        if (typeof outputData.value === "string") {
+          return outputData.value;
+        }
+        // should never get here for this model parser since hugging face
+        // does not support function calling, just being safe
+        return JSON.stringify(outputData.value);
+      }
+
       // Doing this to be backwards-compatible with old output format
       // where we used to save the response in output.data
       if (output.data?.hasOwnProperty("generated_text")) {
@@ -278,21 +289,18 @@ async function constructStreamOutput(
   let output = {} as ExecuteResult;
 
   for await (const iteration of response) {
-    const data = iteration.token.text;
-    const metadata = iteration;
+    const newText = iteration.token.text;
 
-    accumulatedMessage += data;
-    const delta = data;
+    accumulatedMessage += newText;
     const index = 0;
-    options.callbacks!.streamCallback(delta, accumulatedMessage, 0);
-
+    options.callbacks!.streamCallback(newText, accumulatedMessage, index);
     output = {
       output_type: "execute_result",
       // TODO: Investigate if we should use the accumulated message instead
-      // of delta: https://github.com/lastmile-ai/aiconfig/issues/620
-      data: delta,
+      // of newText: https://github.com/lastmile-ai/aiconfig/issues/620
+      data: newText,
       execution_count: index,
-      metadata: metadata,
+      metadata: iteration,
     } as ExecuteResult;
   }
   return output;
--- a/typescript/demo/function-call-stream.ts
+++ b/typescript/demo/function-call-stream.ts
@@ -10,7 +10,12 @@ import {
 } from "openai/resources/chat";
 import * as path from "path";
 import { AIConfigRuntime } from "../lib/config";
-import { Prompt } from "../types";
+import {
+  FunctionData,
+  Prompt,
+  OutputDataWithValue,
+  OutputDataWithToolCallsValue,
+} from "../types";
 import { uniqueId } from "lodash";
 
 // This example is taken from https://github.com/openai/openai-node/blob/v4/examples/function-call-stream.ts
@@ -217,12 +222,21 @@ async function functionCallingWithAIConfig() {
       return;
     }
 
-    const rawResponse = output.metadata?.rawResponse;
-    const functionCall = rawResponse?.function_call;
+    let functionCall: FunctionData | null = null;
+    if (output.data?.hasOwnProperty("value")) {
+      const outputData = output.data as OutputDataWithValue;
+      if (outputData.kind === "tool_calls") {
+        outputData as OutputDataWithToolCallsValue;
+        // Typescript schema does not support array of function calls yet,
+        // but Python does, so extracting from an array is for
+        // forwards compatibility
+        functionCall = outputData.value[outputData.value.length - 1].function;
+      }
+    }
     console.log("functionCall=", functionCall);
 
     // If there is no function call, we're done and can exit this loop
-    if (!functionCall) {
+    if (functionCall == null) {
       return;
     }
 
--- a/typescript/lib/parsers/hf.ts
+++ b/typescript/lib/parsers/hf.ts
@@ -12,6 +12,7 @@ import {
   ExecuteResult,
   ModelMetadata,
   Output,
+  OutputDataWithValue,
   Prompt,
   PromptInput,
 } from "../../types";
@@ -244,6 +245,16 @@ export class HuggingFaceTextGenerationParser extends ParameterizedModelParser<Te
         return output.data;
       }
 
+      if (output.data?.hasOwnProperty("value")) {
+        const outputData = output.data as OutputDataWithValue;
+        if (typeof outputData.value === "string") {
+          return outputData.value;
+        }
+        // should never get here for this model parser since hugging face
+        // does not support function calling, just being safe
+        return JSON.stringify(outputData.value);
+      }
+
       // Doing this to be backwards-compatible with old output format
       // where we used to save the response in output.data
       if (output.data?.hasOwnProperty("generated_text")) {
@@ -270,21 +281,20 @@ async function constructStreamOutput(
   let output = {} as ExecuteResult;
 
   for await (const iteration of response) {
-    const data = iteration.token.text;
+    const newText = iteration.token.text;
     const metadata = iteration;
 
-    accumulatedMessage += data;
-    const delta = data;
+    accumulatedMessage += newText;
     const index = 0;
-    options.callbacks!.streamCallback(delta, accumulatedMessage, 0);
+    options.callbacks!.streamCallback(newText, accumulatedMessage, 0);
 
     output = {
       output_type: "execute_result",
       // TODO: Investigate if we should use the accumulated message instead
-      // of delta: https://github.com/lastmile-ai/aiconfig/issues/620
-      data: delta,
+      // of newText: https://github.com/lastmile-ai/aiconfig/issues/620
+      data: newText,
       execution_count: index,
-      metadata: { metadata },
+      metadata,
     } as ExecuteResult;
   }
   return output;
--- a/typescript/lib/parsers/openai.ts
+++ b/typescript/lib/parsers/openai.ts
@@ -1,10 +1,13 @@
 import { JSONObject, JSONValue } from "../../common";
 import {
-  Prompt,
-  Output,
-  PromptInput,
-  ModelMetadata,
   ExecuteResult,
+  FunctionData,
+  ModelMetadata,
+  Output,
+  OutputDataWithValue,
+  OutputDataWithToolCallsValue,
+  Prompt,
+  PromptInput,
 } from "../../types";
 import { AIConfigRuntime } from "../config";
 import { ParameterizedModelParser } from "../parameterizedModelParser";
@@ -12,11 +15,12 @@ import OpenAI, { ClientOptions } from "openai";
 import { getAPIKeyFromEnv } from "../utils";
 import { encode as gpt3Encode, decode as gpt3Decode } from "gpt-3-encoder";
 import {
-  CompletionCreateParams,
   Chat,
+  CompletionCreateParams,
   CompletionCreateParamsNonStreaming,
   CompletionCreateParamsStreaming,
 } from "openai/resources";
+import { ChatCompletionMessageParam } from "openai/resources/chat";
 import _ from "lodash";
 import { InferenceOptions } from "../modelParser";
 import { CallbackEvent } from "../callback";
@@ -173,6 +177,7 @@ export class OpenAIModelParser extends ParameterizedModelParser<CompletionCreate
       for (const choice of response.choices) {
         const output: ExecuteResult = {
           output_type: "execute_result",
+          // CompletionChoice does not support function calls
           data: choice.text,
           execution_count: choice.index,
           metadata: {
@@ -252,9 +257,17 @@ export class OpenAIModelParser extends ParameterizedModelParser<CompletionCreate
     }
 
     if (output.output_type === "execute_result") {
-      // TODO: Add in OutputDataWithValue another way to support function calls
       if (typeof output.data === "string") {
         return output.data;
+      }
+
+      if (output.data?.hasOwnProperty("value")) {
+        const outputData = output.data as OutputDataWithValue;
+        if (typeof outputData.value === "string") {
+          return outputData.value;
+        }
+        // If we get here that means it must be of kind tool_calls
+        return JSON.stringify(outputData.value);
       }
 
       // Doing this to be backwards-compatible with old output format
@@ -263,7 +276,7 @@ export class OpenAIModelParser extends ParameterizedModelParser<CompletionCreate
         output.data?.hasOwnProperty("content") &&
         output.data?.hasOwnProperty("role")
       ) {
-        const message = output.data as Chat.ChatCompletionMessageParam;
+        const message = output.data as ChatCompletionMessageParam;
         if (message.content != null) {
           return message.content;
         } else if (message.function_call) {
@@ -369,7 +382,7 @@ export class OpenAIChatModelParser extends ParameterizedModelParser<Chat.ChatCom
       const message = data.messages[i];
       if (message.role === "user" || message.role == "function") {
         // Serialize user message as prompt, and save the assistant response as an output
-        let assistantResponse: Chat.ChatCompletionMessageParam | null = null;
+        let assistantResponse: ChatCompletionMessageParam | null = null;
         if (i + 1 < data.messages.length) {
           const nextMessage = data.messages[i + 1];
           if (nextMessage.role === "assistant") {
@@ -379,7 +392,23 @@ export class OpenAIChatModelParser extends ParameterizedModelParser<Chat.ChatCom
 
         const input: PromptInput =
           message.role === "user" ? message.content ?? "" : { ...message };
-
+        let outputs: Output[] | undefined = undefined;
+        if (assistantResponse != null) {
+          const assistantOutputData: OutputDataWithValue | string | undefined =
+            buildOutputData(assistantResponse);
+          if (assistantOutputData != null) {
+            outputs = [
+              {
+                output_type: "execute_result",
+                data: assistantOutputData,
+                metadata: {
+                  rawResponse: assistantResponse,
+                  ..._.omit(assistantResponse, ["content", "function_call"]),
+                },
+              },
+            ];
+          }
+        }
         const prompt: Prompt = {
           name: `${promptName}_${prompts.length + 1}`,
           input,
@@ -388,16 +417,7 @@ export class OpenAIChatModelParser extends ParameterizedModelParser<Chat.ChatCom
             parameters: params ?? {},
             remember_chat_context: true,
           },
-          outputs:
-            assistantResponse != null
-              ? [
-                  {
-                    output_type: "execute_result",
-                    data: assistantResponse.content,
-                    metadata: { rawResponse: assistantResponse },
-                  },
-                ]
-              : undefined,
+          outputs,
         };
 
         prompts.push(prompt);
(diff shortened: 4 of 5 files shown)


extensions/HuggingFace/typescript/hf.ts
```diff
@@ -288,14 +296,15 @@ async function constructStreamOutput(
...
+    kind: "string",
+    value: response.generated_text,
+  };
   const output = {
     output_type: "execute_result",
-    data: data.generated_text,
+    data,
     execution_count: 0,
-    metadata: metadata,
+    metadata: {},
```
saqadri, line 307: Why is this emptied out?

typescript/types.ts
```diff
@@ -174,6 +174,15 @@ export type Prompt = {
...
 
+/**
+ * The output type of the result from executing a prompt.
+ * We use this the kind field to determine how to parse it.
+ */
+export type OutputData = {
+  kind: "string" | "file_uri" | "base64";
+  value: string;
+};
+
```
saqadri, line 185: Please regenerate the aiconfig schema
```
yarn genSchema
```

typescript/__tests__/parsers/hf/hf.test.ts
```diff
@@ -257,7 +257,7 @@ describe("HuggingFaceTextGeneration ModelParser", () => {
 
     const expectedOutput = {
       output_type: "execute_result",
-      data: "Test text generation",
+      data: { kind: "string", value: "Test text generation" },
```
saqadri, line 260: I feel like even though this is slightly easier for the frontend to parse, it makes the model parsers and readability of the aiconfig more annoying. It should be pretty simple to check `typeof output.data === "string"` to infer string type, no? Up to you @rholinshead

typescript/lib/parsers/hf.ts
```diff
@@ -240,7 +241,15 @@ export class HuggingFaceTextGenerationParser extends ParameterizedModelParser<Te
...
 
     if (output.output_type === "execute_result") {
-      if (typeof output.data === "string") {
+      if (output.data?.hasOwnProperty("value")) {
+        const outputData = output.data as OutputData;
+        if (typeof outputData.value === "string") {
+          return outputData.value;
+        }
+        // Sarmad + Ryan, let me know if you prefer I do something else
+        // Was basing this off existing code for function calls in openai.ts
```
rossdanlm, line 250: Lol @rholinshead + @saqadri let me know if you prefer me to return entire JSON vs. a specific field within function call (ex: we can pass in `name`: https://github.com/lastmile-ai/aiconfig/blob/main/typescript/types.ts#L182-L195)

typescript/lib/parsers/openai.ts
```diff
@@ -851,6 +913,31 @@ function multiChoiceMessageReducer(
...
+  message: Chat.ChatCompletionMessageParam | null
+): OutputData | undefined {
+  let outputData: OutputData | undefined = undefined;
+  if (message !== null) {
+    if (message.content !== null) {
+      outputData = {
+        kind: "string",
+        value: message.content,
+      };
+    } else if (message.function_call !== undefined) {
```
rossdanlm, line 926: Here I am assuming that if we have function call response, then `content` field must be null (`ChatCompletionMessageParam.content` is of type `(string | null)`. I am assuming that both fields can't be set at the same time

typescript/__tests__/parsers/hf/hf.test.ts
rossdanlm: Sg, let me do this another diff to keep things simple and easier to review!

extensions/HuggingFace/typescript/hf.ts
```diff
@@ -248,7 +249,14 @@ export class HuggingFaceTextGenerationModelParserExtension extends Parameterized
...
 
     if (output.output_type === "execute_result") {
-      if (typeof output.data === "string") {
+      if (output.data?.hasOwnProperty("value")) {
+        const outputData = output.data as OutputDataWithValue;
+        if (typeof outputData.value === "string") {
+          return outputData.value;
+        }
+        // should never get here for this model parser, just being safe
+        return JSON.stringify(outputData.value);
```
saqadri, line 258: I'm guessing we do JSON.stringify to cover the case where output.data.kind is type "tool_calls"? JSON.stringify of other `value` (which are string) will just return the string back.

typescript/demo/function-call-stream.ts
```diff
@@ -217,12 +222,20 @@ async function functionCallingWithAIConfig() {
...
+        // Typescript schema does not support array of function calls yet,
+        // but Python does so doing this for forward compatibility
+        functionCall = outputData.value[outputData.value.length - 1].function;
+      }
+    }
     console.log("functionCall=", functionCall);
 
     // If there is no function call, we're done and can exit this loop
-    if (!functionCall) {
+    if (functionCall === null) {
```
saqadri, line 238: This is incorrect. 

`if (!functionCall)` checks for both `null` or `undefined`, whereas `if (functionCall === null)` only checks for `null`.

typescript/demo/function-call-stream.ts
```diff
@@ -217,12 +222,20 @@ async function functionCallingWithAIConfig() {
...
+    let functionCall: FunctionData | null = null;
+    if (output.data?.hasOwnProperty("value")) {
+      const outputData = output.data as OutputDataWithValue;
+      if (outputData.kind === "tool_calls") {
+        outputData as OutputDataWithToolCallsValue;
+        // Typescript schema does not support array of function calls yet,
+        // but Python does so doing this for forward compatibility
+        functionCall = outputData.value[outputData.value.length - 1].function;
+      }
+    }
```
saqadri, line 234: nit: this can be simplified:

```
if (typeof output.data !== "string" && output.data.kind === "tool_calls") {
  // OutputDataWithToolCallsValue type should be automatically inferred by the TS compiler in this if block
  const outputData: OutputDataWithToolCallsValue = output.data
  if (Array.isArray(outputData)) {
    functionCall = outputData.value[outputData.value.length - 1].function;
  } else {
    functionCall = outputData.value.function;
  }
}
```

typescript/lib/parsers/hf.ts
saqadri: I think this is good. Just add a comment so it's clear what the function is trying to do.

Perhaps something like "JSON.stringify to cover the case where output.data.kind is type "tool_calls". Any string values are returned back as-is"

typescript/lib/parsers/openai.ts
```diff
@@ -552,17 +553,22 @@ export class OpenAIChatModelParser extends ParameterizedModelParser<Chat.ChatCom
       const outputs: ExecuteResult[] = [];
       const responseWithoutChoices = _.omit(response, "choices");
       for (const choice of response.choices) {
+        const outputData: OutputDataWithValue | string | undefined =
+          buildOutputData(choice.message);
+        if (outputData === undefined) {
+          continue;
+        }
```
saqadri, line 560:
```
if (!outputData) {
  continue;
}
```

to check for null or undefined? Unless this is intentional

typescript/lib/parsers/openai.ts
```diff
@@ -601,12 +607,19 @@ export class OpenAIChatModelParser extends ParameterizedModelParser<Chat.ChatCom
             /*accumulatedData*/ message,
             /*index*/ choice.index
           );
+          if (message === undefined) {
+            continue;
+          }
```
saqadri, line 612: Same as above. `if (!message)`? or `if (message == null)` (double `==` checks for both null or undefined, triple `===` checks just for null/undefined

typescript/lib/parsers/openai.ts
```diff
@@ -601,12 +607,19 @@ export class OpenAIChatModelParser extends ParameterizedModelParser<Chat.ChatCom
...
+
+          const outputData: OutputDataWithValue | string | undefined =
+            buildOutputData(message);
+          if (outputData === undefined) {
+            continue;
+          }
```
saqadri, line 618: Same as above

typescript/lib/parsers/palm.ts
```diff
@@ -232,7 +245,7 @@ function constructOutputs(
     const candidate = response.candidates[i];
     const output: ExecuteResult = {
       output_type: "execute_result",
-      data: candidate.output,
+      data: candidate.output ?? "",
```
saqadri, line 248: Maybe `?? null`? See: #631 as well

typescript/lib/parsers/openai.ts
```diff
@@ -173,6 +176,7 @@ export class OpenAIModelParser extends ParameterizedModelParser<CompletionCreate
       for (const choice of response.choices) {
         const output: ExecuteResult = {
           output_type: "execute_result",
+          // CompletionChoice does not support function calls
```
saqadri, line 179: Yes, this class is for the older OpenAI models that don't support function calling. Good comment to leave

typescript/lib/parsers/openai.ts
```diff
@@ -645,9 +658,16 @@ export class OpenAIChatModelParser extends ParameterizedModelParser<Chat.ChatCom
...
+        const outputData = output.data as OutputDataWithValue;
+        if (typeof outputData.value === "string") {
+          return outputData.value;
+        }
+        return JSON.stringify(outputData.value); // function_call
+      } else if (typeof output.data === "string") {
         return output.data;
+      } else if (output.metadata?.rawResponse?.function_call) {
+        return JSON.stringify(output.metadata?.rawResponse?.function_call);
       }
```
saqadri, line 671: What is the case where we need this? Should be safe to remove since line 661-667 should handle this, right?

typescript/lib/parsers/openai.ts
```diff
@@ -700,17 +720,32 @@ export class OpenAIChatModelParser extends ParameterizedModelParser<Chat.ChatCom
...
+            } else if (outputData.kind === "tool_calls") {
+              outputData as OutputDataWithToolCallsValue;
+              // Typescript schema does not support array of function calls yet,
+              // but Python does so doing this for forward compatibility
+              function_call =
+                outputData.value[outputData.value.length - 1].function;
+            }
+          } else if (typeof output.data === "string") {
+            content = output.data;
           }
```
saqadri, line 741: This can be simplified:

```
if (typeof output.data === "string") {
  content = output.data;
} else if (output.data.kind === "tool_calls") {
  // OutputDataWithToolCallsValue type should be automatically inferred by the TS compiler in this if block 
  const outputData: OutputDataWithToolCallsValue = output.data 
  if (Array.isArray(outputData)) { 
    functionCall = outputData.value[outputData.value.length - 1].function; 
  } else { 
    functionCall = outputData.value.function; 
  }
} else {
  // output.data.value should be inferred to be string in this block
  content = outputData.value;
}
```

typescript/lib/parsers/openai.ts
```diff
@@ -700,17 +720,32 @@ export class OpenAIChatModelParser extends ParameterizedModelParser<Chat.ChatCom
...
+
+          messages.push({
+            content,
+            role: output.metadata?.rawResponse?.role,
+            ...(function_call !== undefined && { function_call }),
```
saqadri, line 746: We should update this to `tools` and `tool_choice` based on the latest OpenAI API schema, right? https://platform.openai.com/docs/api-reference/chat/create?lang=node.js

(Can come in a future diff)

typescript/lib/parsers/openai.ts
```diff
@@ -700,17 +720,32 @@ export class OpenAIChatModelParser extends ParameterizedModelParser<Chat.ChatCom
...
+              function_call =
+                outputData.value[outputData.value.length - 1].function;
+            }
+          } else if (typeof output.data === "string") {
+            content = output.data;
           }
+
+          messages.push({
+            content,
+            role: output.metadata?.rawResponse?.role,
```
saqadri, line 745: Does metadata contain the role directly as well? 
Can we do `output.metadata?.role ?? output.metadata?.rawResponse?.role`?

typescript/lib/parsers/openai.ts
```diff
@@ -700,17 +720,32 @@ export class OpenAIChatModelParser extends ParameterizedModelParser<Chat.ChatCom
...
+            ...(function_call !== undefined && { function_call }),
+            name: output.metadata?.rawResponse?.name,
```
saqadri, line 747: What is `name`? And similar to `role`, can we do
`output.metadata?.name ?? output.metadata?.rawResponse?.name`?

saqadri: I'm not really seeing `name` in the OpenAI api

typescript/lib/parsers/openai.ts
```diff
@@ -730,6 +765,26 @@ export class OpenAIChatModelParser extends ParameterizedModelParser<Chat.ChatCom
 
     return messages;
   }
+
+  private buildAssistantOutputDataIfNeeded(
```
saqadri, line 769: Unless this is used in multiple places, would suggest inlining this function

typescript/lib/parsers/openai.ts
```diff
@@ -730,6 +765,26 @@ export class OpenAIChatModelParser extends ParameterizedModelParser<Chat.ChatCom
...
+    assistantResponse: Chat.ChatCompletionMessageParam | null
+  ): Output[] | undefined {
+    let outputs: Output[] | undefined = undefined;
+    if (assistantResponse != null) {
+      const assistantOutputData: OutputDataWithValue | string | undefined =
+        buildOutputData(assistantResponse);
+      if (assistantOutputData !== undefined) {
```
saqadri, line 776: `if (!assistantOutputData)`

typescript/lib/parsers/openai.ts
```diff
@@ -851,6 +906,28 @@ function multiChoiceMessageReducer(
   return messages;
 }
 
+function buildOutputData(
+  message: Chat.ChatCompletionMessageParam | null
+): OutputDataWithValue | string | undefined {
+  let outputData: OutputDataWithValue | string | undefined = undefined;
+  if (message !== null) {
```
saqadri, line 913: `message != null`

typescript/lib/parsers/openai.ts
```diff
@@ -851,6 +906,28 @@ function multiChoiceMessageReducer(
...
+    if (message.content !== null) {
```
saqadri, line 914: `message.content != null`

typescript/lib/parsers/openai.ts
```diff
@@ -851,6 +906,28 @@ function multiChoiceMessageReducer(
...
+      outputData = message.content;
+    } else if (message.function_call !== undefined) {
```
saqadri, line 916: `(message.function_call != null)` (checks for both null or undefined)

saqadri (changes requested):
A few subtle bugs, so requesting changes. Mostly looks good

extensions/HuggingFace/typescript/hf.ts
rossdanlm: The thing is that this is `hf.ts` so it will always output a string value in it's `response.generated_text` field

typescript/demo/function-call-stream.ts
rossdanlm: I initialized it to null so it can never be undefined in this flow, but yea I get what you're saying I can check for both to ensure it's future-proofed

typescript/demo/function-call-stream.ts
rossdanlm: > if (Array.isArray(outputData)) { functionCall = outputData.value[outputData.value.length - 1].function; } else { functionCall = outputData.value.function; }

We already set in our schema that `outputData.value.function` must always be an array (which follows what OpenAI has defined). But yea I can do this for future-proofing if you like

typescript/lib/parsers/openai.ts
rossdanlm: This is intentional. `outputData` can only be of type `OutputDataWithValue | string | undefined`. 

I did this because I did not want to set the prompt outputs to an empty array if there was no data (I feel it might be slightly unintuitive if you first serialize a prompt, don't run it, yet still see an empty `outputs` field. In our schema the `prompt.outputs` is optional so I think it makes more sense to simply not have present

typescript/lib/parsers/openai.ts
rossdanlm: This is also intentional since message can only be of type `Chat.ChatCompletionMessageParam | undefined` from the `messages.get(choice.index)`. But yea I can add null check too to future-proof

typescript/lib/parsers/openai.ts
rossdanlm: Same comment, intentionally can't be null, only `OutputDataWithValue | string | undefined`

typescript/lib/parsers/openai.ts
rossdanlm: Thanks for the catch, I just missed it from all the if-else overloading my brain

typescript/lib/parsers/openai.ts
rossdanlm: No problem, I can do this

typescript/lib/parsers/openai.ts
rossdanlm: This is a good callout: `tool_calls` field is only used for assistant output type but that's fine because we're only setting this from assistant role anyways: https://github.com/openai/openai-node/blob/master/src/resources/chat/completions.ts#L147-L151
Screenshot 2023-12-27 at 12 39 56


Just a general note so you're aware, variable `message` is of a more general interface type `openai/resources/chat/ChatCompletionMessageParams` (defined here: https://github.com/openai/openai-node/blob/master/src/resources/chat/completions.ts#L445) which does not have the `function_call` or `tool_calls` field. Ex: here is the structure for user message: 
Screenshot 2023-12-27 at 12 39 34

typescript/lib/parsers/openai.ts
rossdanlm: It's defined in all of the sub-types for `ChatCompletionMessageParam` except for the tool one: https://github.com/openai/openai-node/blob/master/src/resources/chat/completions.ts#L445

But yes that's a good point I should check first whether it's undefined or not before setting, thanks for flagging

typescript/lib/parsers/openai.ts
rossdanlm: Same comments as 
- https://github.com/lastmile-ai/aiconfig/pull/610/files/e2dbf8d858178093869d39985c140992f4dba48b..2aea134d8938ca54314125962917246a6b6c0f58#r1437166541
- https://github.com/lastmile-ai/aiconfig/pull/610/files/e2dbf8d858178093869d39985c140992f4dba48b..2aea134d8938ca54314125962917246a6b6c0f58#r1437167423

typescript/lib/parsers/openai.ts
rossdanlm: Same as https://github.com/lastmile-ai/aiconfig/pull/610/files/e2dbf8d858178093869d39985c140992f4dba48b..2aea134d8938ca54314125962917246a6b6c0f58#r1437167423

typescript/lib/parsers/openai.ts
rossdanlm: message.content can only be null in all implementations of `ChatCompletionMessageParam`, but yea sure I can future proof

typescript/lib/parsers/palm.ts
rossdanlm: Our data object must be a string. This is palm which is purely text so it cannot be a function call or any other data type

typescript/lib/parsers/openai.ts
rholinshead: I think it's just cleaner to do `!= null` if there isn't a need to explicitly distinguish between null and undefined values

typescript/lib/parsers/openai.ts
```diff
@@ -700,17 +720,32 @@ export class OpenAIChatModelParser extends ParameterizedModelParser<Chat.ChatCom
...
-            messages.push({
-              content: output.data,
-              // TODO (rossdanlm): Support function_call and don't rely on rawResponse
-              role: output.metadata?.rawResponse?.role,
-              function_call: output.metadata?.rawResponse?.function_call,
-              name: output.metadata?.rawResponse?.name,
-            });
+          output as ExecuteResult;
+          let content: string | null = null;
+          let function_call: FunctionData | undefined = undefined;
```
rholinshead, line 726:
```suggestion
          let function_call: FunctionData | undefined;
```
nit, the value will default to `undefined` without needing to set it

typescript/demo/function-call-stream.ts
saqadri: Ah, got it. In that case we can ignore the Array.isArray check. Thanks for confirming. Rest of the suggested refactor is still valid

typescript/lib/parsers/openai.ts
saqadri: If outputData is null, we also don't want to serialize it, right? Or do we want to?

If we want the same behavior whether outputData is null or undefined, then we should change to `if (!outputData)`

typescript/lib/parsers/openai.ts
saqadri: I always get worried when I see explicit null or explicit undefined checks. Usually, those are clearly documented when the behavior diverges between null or undefined.

In this case if the behavior between the two is intended to be the same, please switch to `if (!message)`

typescript/lib/parsers/openai.ts
saqadri: +1 to @rholinshead's comment. I think applies to all the places where null or undefined is being explicitly checked.

typescript/lib/parsers/openai.ts
saqadri: Yes should be fine. We can have a task for updating this, don't need to support tool and tool_calls in this diff

saqadri (approved):
Approving to unblock after the comments are addressed

typescript/lib/parsers/openai.ts
rossdanlm: `outputData` can never be null since we set it to `OutputDataWithValue | string | undefined` but yea i'll add explicit check for null too!

typescript/lib/parsers/openai.ts
rossdanlm: Btw I get this error when TS doesn't know the explicit type so I need to set it as something. Maybe I'm missing something but that's why I explicitly set the line `const outputData = output.data as OutputDataWithValue;`
Screenshot 2023-12-27 at 14 48 54

rossdanlm: But yea I'll put the explicit null-check first to make it easier to read!

typescript/demo/function-call-stream.ts
rossdanlm: Also I don't think this works because we're supporting arbitrary JSON types, so it's not guaranteed to have `output.data.kind` defined:
Screenshot 2023-12-27 at 14 56 15

typescript/lib/parsers/openai.ts
rossdanlm: Ok update, I tried doing this but couldn't import `ChatCompletionAssistantMessageParam`. COuld only work with the generalized `ChatCompletionMessageParam` object so just going to save it into the `function_call` field. Will leave comment with GH issue to track as backlog

rossdanlm:
Wow ok finally done updating feedback and ready to land

PR #611, opened 2023-12-25
[python] Use OutputDataWithValue type for function calls

rossdanlm:
[python] Use OutputDataWithValue type for function calls

This is extending https://github.com/lastmile-ai/aiconfig/pull/605 where I converted from model-specific response types --> pure strings. Now I'm going to also support function calls --> OutputData types

Same as last diff (https://github.com/lastmile-ai/aiconfig/pull/610) except now for python files instead of typescript

Don't need to do this for the image outputs since I already did that in https://github.com/lastmile-ai/aiconfig/pull/608

Actually like typescript, the only model parser we need to support for function calling is `openai.py`

### DIFF
--- a/extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
+++ b/extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
@@ -1,17 +1,27 @@
 # Define a Model Parser for LLama-Guard
 from typing import TYPE_CHECKING, Dict, List, Optional, Any
 import copy
+import json
 
 import google.generativeai as genai
-import copy
-
-from aiconfig.default_parsers.parameterized_model_parser import ParameterizedModelParser
-from aiconfig.model_parser import InferenceOptions
-from aiconfig.schema import ExecuteResult, Output, Prompt
-from aiconfig.util.params import resolve_prompt, resolve_prompt_string
-from aiconfig import CallbackEvent, get_api_key_from_environment, AIConfigRuntime, PromptMetadata, PromptInput
 from google.generativeai.types import content_types
 from google.protobuf.json_format import MessageToDict
+
+from aiconfig import (
+    AIConfigRuntime,
+    CallbackEvent,
+    get_api_key_from_environment,
+)
+from aiconfig.default_parsers.parameterized_model_parser import ParameterizedModelParser
+from aiconfig.model_parser import InferenceOptions
+from aiconfig.schema import (
+    ExecuteResult,
+    Output,
+    OutputDataWithValue,
+    Prompt,
+    PromptInput,
+)
+from aiconfig.util.params import resolve_prompt, resolve_prompt_string
 
 # Circuluar Dependency Type Hints
 if TYPE_CHECKING:
@@ -208,7 +218,6 @@ class GeminiModelParser(ParameterizedModelParser):
                 user_message_parts = user_message["parts"]
                 outputs = []
                 if i + 1 < len(contents):
-                    # TODO (rossdanlm): Support function calls
                     model_message = contents[i + 1]
                     model_message_parts = model_message["parts"]
                     # Gemini api currently only supports one candidate aka one output. Model should only be retuning one part in response.
@@ -218,8 +227,8 @@ class GeminiModelParser(ParameterizedModelParser):
                         ExecuteResult(
                             **{
                                 "output_type": "execute_result", 
-                                "data": model_message_parts[0], 
-                                "metadata": {"rawResponse": model_message}
+                                "data": model_message_parts[0],
+                                "metadata": {"rawResponse": model_message},
                             }
                         )
                     ]
@@ -359,11 +368,20 @@ class GeminiModelParser(ParameterizedModelParser):
             output_data = output.data
             if isinstance(output_data, str):
                 return output_data
-            else:
-                raise ValueError("Not Implemented")
+            if isinstance(output_data, OutputDataWithValue):
+                if isinstance(output_data.value, str):
+                    return output_data.value
 
-        else:
-            return ""
+        # We use this method in the model parser functionality itself,
+        # so we must error to let the user know non-string
+        # formats are not supported
+        error_message = \
+"""
+We currently only support chats where prior messages from a user are only a 
+single message string instance, please see docstring for more details:
+https://github.com/lastmile-ai/aiconfig/blob/v1.1.8/extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py#L31-L33
+"""
+        raise ValueError(error_message)
 
     def _construct_chat_history(
         self, prompt: Prompt, aiconfig: "AIConfigRuntime", params: Dict
--- a/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/text_generation.py
+++ b/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/text_generation.py
@@ -1,11 +1,23 @@
 import copy
-from typing import TYPE_CHECKING, Any, Dict, List, Optional
-from transformers import AutoTokenizer, Pipeline, pipeline, TextIteratorStreamer
+import json
 import threading
+from typing import TYPE_CHECKING, Any, Dict, List, Optional
+from transformers import (
+    AutoTokenizer,
+    Pipeline,
+    pipeline,
+    TextIteratorStreamer,
+)
 
 from aiconfig.default_parsers.parameterized_model_parser import ParameterizedModelParser
 from aiconfig.model_parser import InferenceOptions
-from aiconfig.schema import ExecuteResult, Output, Prompt, PromptMetadata
+from aiconfig.schema import (
+    ExecuteResult,
+    Output,
+    OutputDataWithValue,
+    Prompt,
+    PromptMetadata,
+)
 from aiconfig.util.params import resolve_prompt
 
 # Circuluar Dependency Type Hints
@@ -101,19 +113,21 @@ def construct_stream_output(
     Constructs the output for a stream response.
 
     Args:
-        streamer (TextIteratorStreamer): Streams the output. See https://huggingface.co/docs/transformers/v4.35.2/en/internal/generation_utils#transformers.TextIteratorStreamer
-        options (InferenceOptions): The inference options. Used to determine the stream callback.
+        streamer (TextIteratorStreamer): Streams the output. See:
+            https://huggingface.co/docs/transformers/v4.35.2/en/internal/generation_utils#transformers.TextIteratorStreamer
+        options (InferenceOptions): The inference options. Used to determine
+            the stream callback.
 
     """
-    accumulated_message = ""
     output = ExecuteResult(
             **{
                 "output_type": "execute_result",
-                "data": accumulated_message,
+                "data": "", # We update this below
                 "execution_count": 0, #Multiple outputs are not supported for streaming
                 "metadata": {},
             }
         )
+    accumulated_message = ""
     for new_text in streamer:
         accumulated_message += new_text
         options.stream_callback(new_text, accumulated_message, 0)
@@ -283,7 +297,13 @@ class HuggingFaceTextGenerationTransformer(ParameterizedModelParser):
         # TODO (rossdanlm): Handle multiple outputs in list
         # https://github.com/lastmile-ai/aiconfig/issues/467
         if output.output_type == "execute_result":
-            if isinstance(output.data, str):
-                return output.data
-        else:
-            return ""
+            output_data = output.data
+            if isinstance(output_data, str):
+                return output_data
+            if isinstance(output_data, OutputDataWithValue):
+                if isinstance(output_data.value, str):
+                    return output_data.value
+                # HuggingFace Text generation does not support function
+                # calls so shouldn't get here, but just being safe
+                return json.dumps(output_data.value, indent=2)
+        return ""
--- a/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/remote_inference_client/text_generation.py
+++ b/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/remote_inference_client/text_generation.py
@@ -1,9 +1,6 @@
 import copy
-from typing import TYPE_CHECKING, Any, Dict, List, Optional
-
-from aiconfig.default_parsers.parameterized_model_parser import ParameterizedModelParser
-from aiconfig.model_parser import InferenceOptions
-from aiconfig.util.config_utils import get_api_key_from_environment
+import json
+from typing import TYPE_CHECKING, Any, Iterable, List, Optional, Union
 
 # HuggingFace API imports
 from huggingface_hub import InferenceClient
@@ -12,13 +9,19 @@ from huggingface_hub.inference._text_generation import (
     TextGenerationStreamResponse,
 )
 
-from aiconfig.schema import ExecuteResult, Output, Prompt, PromptMetadata
 from aiconfig import CallbackEvent
-
+from aiconfig.default_parsers.parameterized_model_parser import ParameterizedModelParser
+from aiconfig.model_parser import InferenceOptions
+from aiconfig.schema import (
+    ExecuteResult,
+    Output,
+    OutputDataWithValue,
+    Prompt,
+    PromptMetadata,
+)
+from aiconfig.util.config_utils import get_api_key_from_environment
 from aiconfig.util.params import resolve_prompt
 
-# ModelParser Utils
-# Type hint imports
 
 # Circuluar Dependency Type Hints
 if TYPE_CHECKING:
@@ -63,7 +66,7 @@ def refine_chat_completion_params(model_settings: dict[Any, Any]) -> dict[str, A
 
 
 def construct_stream_output(
-    response: TextGenerationStreamResponse,
+    response: Union[Iterable[TextGenerationStreamResponse], Iterable[str]],
     response_includes_details: bool,
     options: InferenceOptions,
 ) -> Output:
@@ -79,26 +82,24 @@ def construct_stream_output(
     accumulated_message = ""
     for iteration in response:
         metadata = {}
-        data = iteration
+        # If response_includes_details is false, `iteration` will be a string,
+        # otherwise, `iteration` is a TextGenerationStreamResponse
+        new_text = iteration
         if response_includes_details:
             iteration: TextGenerationStreamResponse
-            data = iteration.token.text
+            new_text = iteration.token.text
             metadata = {"token": iteration.token, "details": iteration.details}
-        else:
-            data: str
 
         # Reduce
-        accumulated_message += data
+        accumulated_message += new_text
 
         index = 0  # HF Text Generation api doesn't support multiple outputs
-        delta = data
         if options and options.stream_callback:
-            options.stream_callback(delta, accumulated_message, index)
-
+            options.stream_callback(new_text, accumulated_message, index)
         output = ExecuteResult(
             **{
                 "output_type": "execute_result",
-                "data": copy.deepcopy(accumulated_message),
+                "data": accumulated_message,
                 "execution_count": index,
                 "metadata": metadata,
             }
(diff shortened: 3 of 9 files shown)


extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/remote_inference_client/text_generation.py
```diff
@@ -98,7 +100,7 @@ def construct_stream_output(
         output = ExecuteResult(
             **{
                 "output_type": "execute_result",
-                "data": copy.deepcopy(accumulated_message),
```
saqadri, line 101: Why did this do a deepcopy while the updated one doesn't?

saqadri (commented):
I'm wondering if we're better off keeping strings just as strings. Thoughts?

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/remote_inference_client/text_generation.py
rossdanlm: I think `accumulated_message` is a message so this isn't necessary. It gets updated from `data`, which itself is also a `str`

extensions/llama/python/llama.py
```diff
@@ -95,25 +105,43 @@ async def _run_inference_helper(self, model_input, options) -> List[Output]:
             except TypeError:
                 texts = [response["choices"][0]["text"]]
 
+            output_data_value : str = ''
+            last_response = texts[-1]
+            if isinstance(last_response, str):
+                output_data_value = last_response
+            else:
+                raise ValueError(f"Output {last_response} needs to be of type 'str' but is of type: {type(last_response)}")
```
saqadri, line 113: Instead of this, should we concatenate all the strings in `texts`? Isn't that the complete response?

saqadri: Or is this similar to `choices` in openai inference, where a model can return `n` different responses if asked to do so (in which case we construct a single `ExecuteResult` per choice.

python/src/aiconfig/default_parsers/hf.py
```diff
@@ -115,7 +117,7 @@ def construct_regular_output(response: TextGenerationResponse, response_includes
     output = ExecuteResult(
         **{
             "output_type": "execute_result",
-            "data": response.generated_text,
+            "data": response.generated_text or '',
```
saqadri, line 120: Looking at the original implementation of this at https://github.com/lastmile-ai/aiconfig/blob/75f00d16ff64c53d0f0fdc4734eb13692bc1b1b7/python/src/aiconfig/default_parsers/hf.py#L112-L115, I notice that in certain cases (specifically if `response_includes_details` is `false`), the entire response was assigned to "data" here.

Perhaps we should check if `isinstance(response, str)` is true, in which case we should use that, otherwise use `response.generated_text`.

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/text_generation.py
```diff
@@ -105,19 +117,20 @@ def construct_stream_output(
...
                 "execution_count": 0, #Multiple outputs are not supported for streaming
                 "metadata": {},
             }
         )
+    accumulated_message = ""
     for new_text in streamer:
-        accumulated_message += new_text
-        options.stream_callback(new_text, accumulated_message, 0)
-        output.data = accumulated_message
+        if isinstance(new_text, str):
```
rholinshead, line 130: Can this be anything other than string? These changes seem unnecessary if I'm understanding the streamer correctly

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/text_generation.py
```diff
@@ -283,7 +296,14 @@ def get_output_text(
...
+            output_data = output.data
+            if isinstance(output_data, str):
+                return output_data
+            if isinstance(output_data, OutputDataWithValue):
+                if isinstance(output_data.value, str):
+                    return output_data.value
+                # HuggingFace Text generation does not support function
+                # calls so shouldn't get here, but just being safe
+                return json.dumps(output_data.value, indent=2)
+            raise ValueError("Not Implemented")
```
rholinshead, line 308: For these cases we could technically json.dumps(output_data), right?

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -505,27 +529,48 @@ def add_prompt_as_message(
...
-                    function_call = output.metadata["rawResponse"].get("function_call", None)
-                    tool_calls = output.metadata["rawResponse"].get("tool_calls", None)
-                    if function_call is not None:
-                        output_message["function_call"] = function_call
-                    if tool_calls is not None:
-                        output_message["tool_calls"] = tool_calls
-                    messages.append(output_message)
+            role = output.metadata.get("assistant", None) or \
+                ("rawResponse" in output.metadata and
+                    output.metadata["rawResponse"].get("assistant", None))
```
saqadri, line 534: This is incorrect. Can you check if TS or Python implementations have this issue elsewhere in your stack?

```suggestion
            role = output.metadata.get("role", None) or \
                ("rawResponse" in output.metadata and
                    output.metadata["rawResponse"].get("role", None))
```

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/remote_inference_client/text_generation.py
```diff
@@ -79,26 +82,25 @@ def construct_stream_output(
     accumulated_message = ""
     for iteration in response:
         metadata = {}
-        data = iteration
+        # If response_includes_details is false, `iteration` will be a string,
+        # otherwise, `iteration` is a TextGenerationStreamResponse
```
rholinshead, line 86: Should we update `response` type to `Iterable[Union[TextGenerationStreamResponse, str]]` to reflect this?

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -505,27 +529,48 @@ def add_prompt_as_message(
...
+                    if isinstance(output_data.value, str):
+                        content = output_data.value
+                    elif output_data.kind == "tool_calls":
+                        assert isinstance(output, OutputDataWithToolCallsValue)
+                        function_call = \
+                            output_data.value[len(output_data.value) - 1] \
+                            .function
+
+                output_message["content"] = content
+                output_message["role"] = "assistant"
```
saqadri, line 552:
```suggestion
                output_message["role"] = role
```

python/src/aiconfig/default_parsers/openai.py
saqadri: Also perhaps we should default role to "assistant"

extensions/LLama-Guard/python/python/src/aiconfig_extension_llama_guard/LLamaGuard.py
```diff
@@ -242,17 +245,21 @@ async def run_inference(
         output_text = self.tokenizer.decode(
             response[0][prompt_len:], skip_special_tokens=True
         )
-
-        Output = ExecuteResult(
+        output_data_content : str = ''
```
rholinshead, line 248: nit, extra space on left of `:`

extensions/LLama-Guard/python/python/src/aiconfig_extension_llama_guard/LLamaGuard.py
```diff
@@ -242,17 +245,21 @@ async def run_inference(
...
+        if isinstance(output_text, str):
```
rholinshead, line 249: Can this actually be non-string type?

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -505,27 +529,48 @@ def add_prompt_as_message(
...
+                output_message = {}
+                content : Union[str, None] = None
+                function_call : Union[FunctionCallData, None] = None
+                if isinstance(output_data, str):
+                    content = output_data
+                elif isinstance(output_data, OutputDataWithValue):
+                    if isinstance(output_data.value, str):
+                        content = output_data.value
+                    elif output_data.kind == "tool_calls":
+                        assert isinstance(output, OutputDataWithToolCallsValue)
```
saqadri, line 546: I feel like this can be simplified:

```
if isinstance(output_data, str):
  # Do string stuff
elif isinstance(output_data, OutputDataWithToolCallsValue):
  # Do function call stuff
elif isinstance(output_data, OutputDataWithVale):
  # Do other value stuff where data.value is string
elif isinstance(output_data, ChatCompletionMessage):
  # Do chat completion stuff

File: extensions/llama/python/llama.py [hunk 12]
```diff
@@ -83,9 +87,15 @@ async def _run_inference_helper(self, model_input, options) -> List[Output]:
                 if options:
                     options.stream_callback(data, acc, index)
             print(flush=True)
+
+            output_data_value : str = ''
+            if isinstance(acc, str):
+                output_data_value = acc
+            else:
+                raise ValueError(f"Output {acc} needs to be of type 'str' but is of type: {type(acc)}")
```
rholinshead, line 95: Is this actually needed? Can't we just type `acc: str = ""` above?

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -536,3 +581,41 @@ def is_prompt_template(prompt: Prompt):
     return isinstance(prompt.input, str) or (
         hasattr(prompt.input, "data") and isinstance(prompt.input.data, str)
     )
+
+def build_output_data(
+    message: Union[ChatCompletionMessage, None],
+) -> Union[OutputDataWithValue, str, None]:
+    output_data : Union[OutputDataWithValue, str, None] = None
+    if message is not None:
```
saqadri, line 589: nit: 
```suggestion
    if not message:
      return None
```

Then everything below doesn't have to be nested in the if

extensions/llama/python/llama.py
rholinshead, line 113: Same here. Is this all needed?

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -536,3 +581,41 @@ def is_prompt_template(prompt: Prompt):
...
+        if message.get("content", None) is not None:
+            output_data = message.get("content")
+        elif message.get("tool_calls", None) is not None:
+            tool_calls = [ToolCallData(
```
saqadri, line 593: nit: this is fine for now, but we might need to update this in the future as `tool_calls` include things that aren't just function calls

saqadri (approved):
A few blocking comments, but otherwise looks good. Really nice work! Not an easy change

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -108,13 +119,16 @@ async def serialize(
 
                 assistant_output = []
                 if assistant_response is not None:
-                    assistant_content = assistant_response.get("content", "")
+                    output_data = build_output_data(assistant_response)
+                    metadata = {"rawResponse": assistant_response}
+                    if assistant_response.get("role", None) is not None:
```
rholinshead, line 124: nit, wouldn't `.get("role")` work without fallback (defaults to None already)?

extensions/llama/python/llama.py
```diff
@@ -95,25 +105,43 @@ async def _run_inference_helper(self, model_input, options) -> List[Output]:
...
+            return ""
+
+        if output.output_type == "execute_result":
+            output_data = output.data
+            if isinstance(output_data, str):
+                return output_data
+            if isinstance(output_data, OutputDataWithValue):
+                if isinstance(output_data.value, str):
+                    return output_data.value
+                # HuggingFace Text generation does not support function
```
rholinshead, line 136:
```suggestion
                # llama does not support function
```

python/src/aiconfig/default_parsers/palm.py
```diff
@@ -377,15 +383,19 @@ def get_output_text(
...
         if output.output_type == "execute_result":
-            assert isinstance(output, ExecuteResult)
             output_data = output.data
-
             if isinstance(output_data, str):
                 return output_data
+            if isinstance(output_data, OutputDataWithValue):
+                if isinstance(output_data.value, str):
+                    return output_data.value
+                # HuggingFace Text generation does not support function
```
rholinshead, line 392:
```suggestion
                # PaLM does not support function
```

rholinshead:
For testing, can you please be sure to run the tests and the relevant cookbooks/demo scripts (e.g. to test function calling)

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/text_generation.py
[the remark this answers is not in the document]
rossdanlm: No it can't happen for hugging face text generation model parser

extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
```diff
@@ -359,11 +369,14 @@ def get_output_text(
...
-
-        else:
-            return ""
+            if isinstance(output_data, OutputDataWithValue):
+                if isinstance(output_data.value, str):
+                    return output_data.value
+                # Gemini does not support function calls so shouldn't
+                # get here, but just being safe
+                return json.dumps(output_data.value, indent=2)
+            raise ValueError("Not Implemented")
```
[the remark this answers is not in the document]
rossdanlm: It should never actually get here but yea I can remove

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/text_generation.py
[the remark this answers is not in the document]
rossdanlm: I can remove the value error, but why remove the comment?

extensions/LLama-Guard/python/python/src/aiconfig_extension_llama_guard/LLamaGuard.py
rossdanlm: According to Python, this only returns `str` type (https://github.com/huggingface/transformers/blob/main/src/transformers/tokenization_utils_base.py#L3720-L3726), but I don't trust Python type hints so that why I wanted to check to be sure

python/src/aiconfig/default_parsers/openai.py
rossdanlm: Yea, just copy pasted wrong my bad, should be `role`.

> Can you check if TS or Python implementations have this issue elsewhere in your stack?

I don't have this issue anywhere else (https://github.com/lastmile-ai/aiconfig/pull/610/files#diff-e044912cdb054134cac74868d76969daa92c6dd7c1238805672e023f9b7aa92d), this was just a typo

> Also perhaps we should default role to "assistant"

We shouldn't be doing this. Role can be one of 5 different types, not just "assistant", so we need to check for that explicitly: https://github.com/openai/openai-node/blob/b595cd953a704dba2aef4c6c3fa431f83f18ccf9/src/resources/chat/completions.ts#L514

python/src/aiconfig/default_parsers/openai.py
rossdanlm: If we get to this part of code, it has to be assistant so no functional changes, but yea I can pass the variable instead

python/src/aiconfig/default_parsers/openai.py
rossdanlm: I find this more complicated and unintuitive. `OutputDataWithToolCallsValue` is a subset of `OutputDataWithValue`. I get what you're saying that you can have single else-if instead of nested if-statements, but it's jarring to go from "small subset" --> "bigger subset" while reading down through if-else logic. 

This is also more confusing because the inner `# Do string stuff`, `# Do function call stuff` etc all need to individually check for the role type being assistant before we can do anything

python/src/aiconfig/default_parsers/openai.py
rossdanlm: Good idea, I'll return early. But still going to keep `message is not None` check because if message is empty string (which is possible) it would pass `if not message`

I know I specify that message can only be of type `Union[ChatCompletionMessage, None]` as the function argument, but Python type hints are unreliable and you can actually pass in whatever you want and the function will still process it

python/src/aiconfig/default_parsers/openai.py
rossdanlm: Good callout, I'll do another check where I'll check for the type being 'function'

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/text_generation.py
rossdanlm: Streamer can actually contain anything, doesn't have to be text. Theoretically it always be str (except for the stop signal), but I'm just being safe. This is a fork of the HF source code (couldn't find it publicly on GH, but it's this: https://github.com/PaddlePaddle/PaddleNLP/blob/a55039cc16bbcfa06b204a5f21b813e98a3f65ca/paddlenlp/generation/streamers.py#L197-L206)

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/text_generation.py
rossdanlm: I should not ever be possible to get to this point, which is why I raised an error to be cautious. But yea I can just remove error and fall into the final `return ""` instead

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/remote_inference_client/text_generation.py
rossdanlm: Good suggestion!

extensions/LLama-Guard/python/python/src/aiconfig_extension_llama_guard/LLamaGuard.py
rossdanlm: Same comment as https://github.com/lastmile-ai/aiconfig/pull/611#discussion_r1437759469, but yea I can remove from this diff

extensions/llama/python/llama.py
rossdanlm: It shouldn't ever happen, but it is possible that a non-string value gets passed to the streamer. I guess we have control of our own model parser where we pass in the streamer, so this should be fine, more just for being safe

extensions/llama/python/llama.py
rossdanlm: Same comments as above, I can remove if we want

python/src/aiconfig/default_parsers/openai.py
rossdanlm: You're right, there's no functional difference, I'm just being explicit but can remove if you find that easier to read

extensions/llama/python/llama.py
rossdanlm: Oh thanks

rossdanlm:
> For testing, can you please be sure to run the tests and the relevant cookbooks/demo scripts (e.g. to test function calling)

I ran the automated tests and function calls still work (notably the `test_openai_util.py` file). I also went through the python function call cookbook just to be safe and it still works, but I'm not going to do the other 10-ish that we have
Screenshot 2023-12-28 at 11 34 16


I get what you're saying just to test the relevant cookbook, so this below isn't to disagree with your comment, just sharing general thoughts to flag:

I would be surprised if they all still worked after this output refactoring since we're changing the way we output data and thus changing how we should access it. I feel having automated testing for the cookbooks should be P0 after we finish local editor MVP. I've mentioned to Sarmad that it's simply not feasible to test them all manually. If any of them are broken now, to me having automated testing + fixing them is a separate work-project

extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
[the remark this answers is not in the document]
rossdanlm: Edit: Ok I can't remove, I added more detailed error message so it's clear why

extensions/llama/python/llama.py
rossdanlm: > in which case we construct a single ExecuteResult per choice

Yea this is what flagged in https://github.com/lastmile-ai/aiconfig/issues/630, I'll add this issue to the code comment

python/src/aiconfig/default_parsers/hf.py
rossdanlm: This is a bit confusing, same thing as https://github.com/lastmile-ai/aiconfig/pull/611#discussion_r1437763720

I will update the comments in code here to make it more clear though

PR #612, opened 2023-12-26
[editor][server endpoints][1/n]: Add delete_prompt() event

rossdanlm:
[editor][server endpoints][1/n]: Add delete_prompt() event

To test, follow the readme to run the backend server in a terminal, then go to another terminal and enter:

## Test plan
```bash
alias aiconfig="python -m 'aiconfig.scripts.aiconfig_cli'"
aiconfig edit --aiconfig-path="/Users/rossdancraig/Projects/aiconfig/cookbooks/Getting-Started/travel.aiconfig.json" --server-port=8080 --server-mode=debug_servers

# Now run this from another terminal
curl http://localhost:8080/api/run -d '{"prompt_name":"get_activities"}' -X POST -H 'Content-Type: application/json'
```

This results in `get_activities` prompt being deleted (notice that the second prompt is now the one that appears first):
Screenshot 2023-12-26 at 00 30 23
Screenshot 2023-12-26 at 00 30 49

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/612).
* #615
* #613
* __->__ #612

### DIFF
--- a/python/src/aiconfig/editor/server/server.py
+++ b/python/src/aiconfig/editor/server/server.py
@@ -19,7 +19,7 @@ from aiconfig.editor.server.server_utils import (
 )
 from aiconfig.model_parser import InferenceOptions
 from flask import Flask, request
-from flask_cors import CORS
+from flask_cors import CORS # TODO: add this to requirements.txt
 from result import Err, Ok, Result
 
 from aiconfig.registry import ModelParserRegistry
@@ -156,6 +156,26 @@ def add_prompt() -> FlaskResponse:
         LOGGER.error(f"Failed to add prompt: {err}")
         return HttpResponseWithAIConfig(
             message=f"Failed to add prompt: {err}",
+            code=400,
+            aiconfig=None,
+        ).to_flask_format()
+
+@app.route("/api/delete_prompt", methods=["POST"])
+def delete_prompt() -> FlaskResponse:
+    state = get_server_state(app)
+    request_json = request.get_json()
+    try:
+        LOGGER.info(f"Deleting prompt: {request_json}")
+        state.aiconfig.delete_prompt(**request_json)  # type: ignore
+        return HttpResponseWithAIConfig(
+            message="Done",
+            aiconfig=state.aiconfig,
+        ).to_flask_format()
+    except Exception as e:
+        err: Err[str] = core_utils.ErrWithTraceback(e)
+        LOGGER.error(f"Failed to delete prompt: {err}")
+        return HttpResponseWithAIConfig(
+            message=f"Failed to delete prompt: {err}",
             code=400,
             aiconfig=None,
         ).to_flask_format()


python/src/aiconfig/editor/server/server.py
```diff
@@ -19,7 +19,7 @@
 )
 from aiconfig.model_parser import InferenceOptions
 from flask import Flask, request
-from flask_cors import CORS
+from flask_cors import CORS # TODO: add this to requirements.txt
```
jonathanlastmileai, line 22: Please go ahead and add it in any diff

jonathanlastmileai (changes requested):
Thanks, this is consistent with the previous code and well-tested, but duplicates the endpoint added in #595. Please rebase.

Also you might find it _slightly_ easier to test with the python client code in a Jupyter-in-VSCode window. It's in the README, and you can do something similar for GET requests.

python/src/aiconfig/editor/server/server.py
rholinshead: Isn't it already in requirements.txt?

Fwiw, we may actually be able to remove it. My hunch is:
- prod implementations, client can actually send api requests to /api instead of needing to explicitly specify localhost:8080 and this should work for multiple server instances
- dev implementation, client may be able to just proxy to the right port (https://create-react-app.dev/docs/proxying-api-requests-in-development/) but TBD if that actually does avoid CORS issues

jonathanlastmileai: re: ports and CORS, @rholinshead I think you understand this the best, but I'm not sure we need any special logic for prod. Since it just serves static files at the user-specified port, doesn't the user just need to start another instance on another port and then navigate there?

rholinshead: Ya, this is just more of a note for the client side since we'll need to change it from the current implementation hard-coded to port 8080

jonathanlastmileai: Where is it hard-coded to 8080? Isn't that just for dev? For prod, isn't it just wherever the user navigates in browser?

rholinshead: https://github.com/lastmile-ai/aiconfig/blob/main/python/src/aiconfig/editor/client/src/utils/api.ts#L4 right now, not just for dev. This was from before the discussion on the weekend. Should be easy enough to update today

saqadri (approved):

jonathanlastmileai:
This duplicates an existing endpoint (https://github.com/lastmile-ai/aiconfig/blob/ea91e273a0d50ca7ce2a47fa00b0c0a1d490b7a6/python/src/aiconfig/editor/server/server.py#L215)

rossdanlm:
Closing becasue this already exists in https://github.com/lastmile-ai/aiconfig/pull/595

PR #613, opened 2023-12-26
[editor][server endpoints][2/n]: Add params arg to run command and check for undefined prompt_name

rossdanlm:
[editor][server endpoints][2/n]: Add params arg to run command and check for undefined prompt_name

TSIA, pretty simple otherwise we'd be using empty params field and that would error. For the future, I created issue https://github.com/lastmile-ai/aiconfig/issues/671 where we shouldn't even need to pass in params into run and this works implicitly

## Test Plan
Follow dev README to setup the local editor: https://github.com/lastmile-ai/aiconfig/tree/main/python/src/aiconfig/editor#dev, then run this command
```
curl http://localhost:8080/api/run -d '{"prompt_name":"get_activities"}' -X POST -H 'Content-Type: application/json'
```

Without params

https://github.com/lastmile-ai/aiconfig/assets/151060367/7233a4db-9042-4fa9-affb-0f854ae8a50e

With params

https://github.com/lastmile-ai/aiconfig/assets/151060367/6c078804-ea66-4358-824d-21d0fe26bafa

PR #615, opened 2023-12-26
[editor][server endpoints][3/n]: Proof-of-Concept for Streaming

rossdanlm:
[editor][server endpoints][3/n]: Proof-of-Concept for Streaming

The actual call in the `run` command will be a bit more complex since we'll have to attach a stream callback and probably create a text queue iterator like what we did for Gradio, but this shows that it's possible and shouldn't be too complicated.

Sorry Jonathan, I didn't use any of your super useful generic functions, but we can do that later.

A big thing I wanted to point out is that in the `yield` generator, I need explicitly return a string (not a json object), otherwise this will not work, so the frontend needs to be able to parse that result somehow. We should sync on this but I feel jsons are strings anyways and as long as we follow the same output "formula" (ex: in the `list_models` we out `data` key vs other commands we directly return the AIConfig json), it should be fine.

## Test plan

https://github.com/lastmile-ai/aiconfig/assets/151060367/4c99cec1-9342-4876-8300-ddb27eb7a2cf

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/615).
* __->__ #615
* #613
* #612

### DIFF
--- a/python/src/aiconfig/editor/server/server.py
+++ b/python/src/aiconfig/editor/server/server.py
@@ -119,12 +119,23 @@ def create() -> FlaskResponse:
 async def run() -> FlaskResponse:
     state = get_server_state(app)
     request_json = request.get_json()
-    prompt_name = request_json.get("prompt_name", None)
-    stream = request_json.get("stream", True)
+    prompt_name : str = request_json.get("prompt_name", None)
+    if prompt_name is None:
+        return HttpResponseWithAIConfig(
+            message="No prompt name provided, cannot execute `run` command",
+            code=400,
+            aiconfig=None,
+        ).to_flask_format()
+    params : str = request_json.get("params", None)
+    stream : bool = request_json.get("stream", True)
     LOGGER.info(f"Running prompt: {prompt_name}, {stream=}")
     inference_options = InferenceOptions(stream=stream)
     try:
-        result = await state.aiconfig.run(prompt_name, options=inference_options)  # type: ignore
+        result = await state.aiconfig.run( # type: ignore
+            prompt_name,
+            params,
+            options=inference_options,
+        ) 
         LOGGER.debug(f"Result: {result=}")
         return HttpResponseWithAIConfig(
             message="Done",
--- a/python/src/aiconfig/editor/server/server.py
+++ b/python/src/aiconfig/editor/server/server.py
@@ -19,7 +19,7 @@ from aiconfig.editor.server.server_utils import (
 )
 from aiconfig.model_parser import InferenceOptions
 from flask import Flask, request
-from flask_cors import CORS
+from flask_cors import CORS # TODO: add this to requirements.txt
 from result import Err, Ok, Result
 
 from aiconfig.registry import ModelParserRegistry
@@ -156,6 +156,26 @@ def add_prompt() -> FlaskResponse:
         LOGGER.error(f"Failed to add prompt: {err}")
         return HttpResponseWithAIConfig(
             message=f"Failed to add prompt: {err}",
+            code=400,
+            aiconfig=None,
+        ).to_flask_format()
+
+@app.route("/api/delete_prompt", methods=["POST"])
+def delete_prompt() -> FlaskResponse:
+    state = get_server_state(app)
+    request_json = request.get_json()
+    try:
+        LOGGER.info(f"Deleting prompt: {request_json}")
+        state.aiconfig.delete_prompt(**request_json)  # type: ignore
+        return HttpResponseWithAIConfig(
+            message="Done",
+            aiconfig=state.aiconfig,
+        ).to_flask_format()
+    except Exception as e:
+        err: Err[str] = core_utils.ErrWithTraceback(e)
+        LOGGER.error(f"Failed to delete prompt: {err}")
+        return HttpResponseWithAIConfig(
+            message=f"Failed to delete prompt: {err}",
             code=400,
             aiconfig=None,
         ).to_flask_format()
--- a/python/src/aiconfig/editor/server/server.py
+++ b/python/src/aiconfig/editor/server/server.py
@@ -1,7 +1,9 @@
 import logging
 
 import lastmile_utils.lib.core.api as core_utils
+import json
 import result
+import time
 from aiconfig.Config import AIConfigRuntime
 from aiconfig.editor.server.server_utils import (
     EditServerConfig,
@@ -18,7 +20,8 @@ from aiconfig.editor.server.server_utils import (
     safe_save_to_disk,
 )
 from aiconfig.model_parser import InferenceOptions
-from flask import Flask, request
+from aiconfig.schema import ExecuteResult, Prompt
+from flask import Flask, Response, request, stream_with_context
 from flask_cors import CORS # TODO: add this to requirements.txt
 from result import Err, Ok, Result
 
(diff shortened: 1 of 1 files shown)


python/src/aiconfig/editor/server/server.py
```diff
@@ -18,8 +20,9 @@
     safe_save_to_disk,
 )
 from aiconfig.model_parser import InferenceOptions
-from flask import Flask, request
-from flask_cors import CORS
+from aiconfig.schema import ExecuteResult, Prompt
+from flask import Flask, Response, request, stream_with_context
+from flask_cors import CORS # TODO: add this to requirements.txt
```
jonathanlastmileai, line 25: Add it! :D

python/src/aiconfig/editor/server/server.py
```diff
@@ -114,17 +117,83 @@ def create() -> FlaskResponse:
     state.aiconfig = AIConfigRuntime.create()  # type: ignore
     return HttpResponseWithAIConfig(message="Created new AIConfig", aiconfig=state.aiconfig).to_flask_format()
 
+@app.route('/api/test_streaming', methods=["POST"])
```
jonathanlastmileai, line 120: You may as well make the real endpoint. It's just us using it anyway; we're not committing to anything big here.

python/src/aiconfig/editor/server/server.py
```diff
@@ -114,17 +117,83 @@ def create() -> FlaskResponse:
...
+def test_streaming():
+    EXCLUDE_OPTIONS = {
+        "prompt_index": True,
+        "file_path": True,
+        "callback_manager": True,
+    }
```
jonathanlastmileai, line 126: Isn't there a lot of duplication with `HttpResponseWithAIConfig.to_flask_format()` ?

python/src/aiconfig/editor/server/server.py
```diff
@@ -114,17 +117,83 @@ def create() -> FlaskResponse:
...
+    state = get_server_state(app)
+    request_json = request.get_json()
+    num_stream_steps : int = request_json.get("num_stream_steps", 10)
+    print(f"{num_stream_steps=}")
+    print(f"{type(num_stream_steps)=}")
+    def generate(num_stream_steps: int):
+        prompt : Prompt = state.aiconfig.get_prompt('get_activities')
+        for i in range(num_stream_steps):
+            time.sleep(1)
```
jonathanlastmileai, line 135: why are we sleeping?

python/src/aiconfig/editor/server/server.py
```diff
@@ -114,17 +117,83 @@ def create() -> FlaskResponse:
...
-    stream = request_json.get("stream", True)
+    prompt_name : str = request_json.get("prompt_name", None)
+    if prompt_name is None:
+        return HttpResponseWithAIConfig(
+            message="No prompt name provided, cannot execute `run` command",
+            code=400,
+            aiconfig=None,
+        ).to_flask_format()
+    params : str = request_json.get("params", None)
+    stream : bool = request_json.get("stream", True)
```
jonathanlastmileai, line 188: Nit: remove space before :. You can autoformat in-place with the linter.

python/src/aiconfig/editor/server/server.py
```diff
@@ -114,6 +117,61 @@ def create() -> FlaskResponse:
...
+        prompt : Prompt = state.aiconfig.get_prompt('get_activities')
+        for i in range(num_stream_steps):
+            time.sleep(1)
+            output = ExecuteResult(
+                output_type="execute_result",
+                execution_count=0,
+                data = "Rossdan" + str(i+1),
+                metadata = {},
+            )
+            prompt.outputs = [output]
```
jonathanlastmileai, line 142: Should we be explicitly constructing this? This feels like unnecessary duplication

python/src/aiconfig/editor/server/server.py
```diff
@@ -114,6 +117,61 @@ def create() -> FlaskResponse:
...
+            print(f"Done step {i+1}/{num_stream_steps}...")
+
+            aiconfig_json = state.aiconfig.model_dump(exclude=EXCLUDE_OPTIONS)
+            # print(f"{aiconfig_json=}\n")
+            # yield_output = core_utils.JSONObject({"data": aiconfig_json})
+            # print(f"{yield_output=}")
+            print(f"{str(aiconfig_json)=}\n")
+            yield str(aiconfig_json) + "\n\n"
```
jonathanlastmileai, line 150: How does this iterate over the tokens returned? Doesn't it just yield the same fixed contents of `get_activities()` every iteration ?

rholinshead:
So the response itself would be a literal string instead of an object with {data: string}?

> A big thing I wanted to point out is that in the yield generator, I need explicitly return a string (not a json object)

Can you explain the reasoning for this? Couldn't you alternatively just yield an object instead of the string? I'm not super familiar with generators but I didn't think they were limited to primitive types?

jonathanlastmileai:
> So the response itself would be a literal string instead of an object with {data: string}?
> 
> > A big thing I wanted to point out is that in the yield generator, I need explicitly return a string (not a json object)
> 
> Can you explain the reasoning for this? Couldn't you alternatively just yield an object instead of the string? I'm not super familiar with generators but I didn't think they were limited to primitive types?

This might be a flask limitation. @rossdanlm should help clarify. I suggest that if this works and there's any question whether flask can do it another way, and it's easy to deal with on frontend, let's just do that. For non-streaming endpoints I have a pretty good understanding what's possible and have returned JSON objects.

python/src/aiconfig/editor/server/server.py
rossdanlm: Yea you're right, I was mainly just playing around and trying to get things tested quickly

rossdanlm:
> Can you explain the reasoning for this? Couldn't you alternatively just yield an object instead of the string? I'm not super familiar with generators but I didn't think they were limited to primitive types?

> This might be a flask limitation. @rossdanlm should help clarify. 

Yea I wasn't able to get regular object to be passed, so had to convert the json to a string and it worked

> I suggest that if this works and there's any question whether flask can do it another way, and it's easy to deal with on frontend, let's just do that. 

Yea I agree, unless we're able to get this unblocked in 20-30 mins of investigation

python/src/aiconfig/editor/server/server.py
rossdanlm: The `aiconfig_json` is updated with different output each time. Sorry I forgot to include test video but you can see it now :)

python/src/aiconfig/editor/server/server.py
rossdanlm: You are right lol, it's kind of not the greatest. It would be better to just overwrite output.data each time, but we'd still have to initialize it first to an `ExecuteResult` object (not too hard, just check if output is empty array first) but yea, just a proof of concept, nothing fancy this diff yet

python/src/aiconfig/editor/server/server.py
rossdanlm: Just to prove that the streaming is going as planned and I can see outputs bit-by-bit

jonathanlastmileai:
> Yielding string is fine (and probably expected) since we can stream the json chunks, something like:
> 
> ```
>     def generate(num_stream_steps: int):
>         aiconfig_json: str | None = None
>         prompt: Prompt = state.aiconfig.get_prompt(prompt_name)
>         for i in range(num_stream_steps):
>             time.sleep(1)
>             output = ExecuteResult(
>                 output_type="execute_result",
>                 execution_count=0,
>                 data="Rossdan" + str(i + 1),
>                 metadata={},
>             )
>             prompt.outputs = [output]
>             print(f"Done step {i+1}/{num_stream_steps}...")
> 
>             aiconfig_json = state.aiconfig.model_dump(exclude=EXCLUDE_OPTIONS)
>             # print(f"{aiconfig_json=}\n")
>             # yield_output = core_utils.JSONObject({"data": aiconfig_json})
>             # print(f"{yield_output=}")
>             print(f"{str(aiconfig_json)=}\n")
>             yield json.dumps({"output_chunk": output.model_dump()})
>             # yield aiconfig_json
> 
>             # HttpResponseWithAIConfig(
>             #     message=f"Done step {i+1}/{num_stream_steps}...",
>             #     aiconfig=state.aiconfig,
>             # ).to_flask_format()
> 
>         if aiconfig_json is None:
>             aiconfig_json = state.aiconfig.model_dump(exclude=EXCLUDE_OPTIONS)
>         yield json.dumps({"aiconfig": aiconfig_json})
> ```
> 
> Where each intermediate chunk is the output chunk with accumulated content and then a final output of the full aiconfig

Makes sense to me!

rholinshead:
Ok, so for the streaming to work with the client side api we'll use (oboe), we need to wrap the entire response in an array, so:
- first chunk should sent "[{output_chunk: output}, \n" with opening "["
- next chunks send just "{output_chunk: output}, \n"
- end sends "{aiconfig: aiconfig_json}]" with closing "]"

See https://github.com/lastmile-ai/aiconfig/pull/650 for example, just need some improvements there:
- shouldn't need to yield [ or ] separately (needs some python string manipulation to add to the first chunk / end config text
- can we do this for *all* run_prompt calls? i.e. even for non-streaming run_prompt, immediately just send [{output: output}, {aiconfig: aiconfig}]" and ideally we can use one endpoint/handling for all run prompt calls

rossdanlm:
Closing because we have updated streaming PR in https://github.com/lastmile-ai/aiconfig/pull/683

PR #616, opened 2023-12-26
Add prompt parameters renderer

saqadri:
Add prompt parameters renderer

Add parameters renderer component.

This is the same component we use in lastmile hosted version of AI Workbooks.

Next up will reuse the same component for global parameters.


saqadri:
https://github.com/lastmile-ai/aiconfig/assets/25641935/76ff27fe-545f-4dbb-a0c0-ec0a9e37f328

python/src/aiconfig/editor/client/src/components/ParametersRenderer.tsx
```diff
@@ -0,0 +1,253 @@
...
+          parameterName: newParameterName,
+          parameterValue: newParameterValue,
+        });
+
+        setLastParameterName(newParameterName);
+      }, 250),
+    [lastParameterName, onUpdateParameter]
+  );
+
+  const theme = useMantineTheme();
```
rholinshead, line 80: Just a note that we'll need to add the MantineProvider wrapper inside EditorContainer in order for this to pick up context; that's something we can do later with our styling/UI cleanup pass

python/src/aiconfig/editor/client/src/components/ParametersRenderer.tsx
```diff
@@ -0,0 +1,253 @@
...
+  );
+});
+
+export type ParametersArray = {
+  parameterName: string;
+  parameterValue: JSONValue;
+  key: string;
+}[];
+
+export function ParametersRenderer(props: {
```
rholinshead, line 139: nit, `export default memo(function ParametersRenderer(props: { ....`

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -69,6 +69,18 @@ export default function EditorContainer({
     [dispatch]
   );
 
+  const onUpdatePromptParameters = useCallback(
```
rholinshead, line 72: I think we should have this at the single parameter granularity -- we should have both set_parameter and delete_parameter endpoints being implemented

python/src/aiconfig/editor/client/src/components/aiconfigReducer.ts
```diff
@@ -67,5 +75,15 @@ export default function aiconfigReducer(
         },
       }));
     }
+    case "UPDATE_PROMPT_PARAMETERS": {
+      console.log(JSON.stringify(action));
```
rholinshead, line 79: nit, can remove log

python/src/aiconfig/editor/client/src/components/prompt/PromptContainer.tsx
```diff
@@ -75,6 +82,17 @@ export default memo(function PromptContainer({
...
+            const newParameters: Record<string, unknown> = {};
+            for (const paramTuple of data.newParameters ?? []) {
+              const key = paramTuple.parameterName;
+              const val = paramTuple.parameterValue;
+
+              newParameters[key] = val;
+            }
+
+            onUpdateParameters(index, newParameters);
+          }}
```
rholinshead, line 95: Should move this logic up into the `useCallback`, otherwise this inline function just makes the above memoization wasted

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
saqadri: Can do that -- should I do that in this diff or at top of stack?

python/src/aiconfig/editor/client/src/components/ParametersRenderer.tsx
saqadri: Added a TODO comment stating as such

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
saqadri: Functionally speaking, any reason for one or the other?

rholinshead: A couple things:
1. It's a more accurate mapping to the config api set/delete (single) parameter methods
2. I think it should more easily allow us to have the temporary 'empty' parameters in state in the UI here and then only send them to the config when a name/value is set

python/src/aiconfig/editor/client/src/components/ParametersRenderer.tsx
```diff
@@ -0,0 +1,254 @@
...
+          parameterName: "",
+          parameterValue: "",
+        },
+      ];
+      onUpdateParameters({ newParameters });
+      return newParameters;
+    });
+  }, [onUpdateParameters]);
+
+  // TODO: saqadri - add MantineProvider wrapper inside EditorContainer in order for this to pick up context
```
rholinshead, line 196: nit, probably inside Editor. The Editor component should be completely interchangeable between clients (e.g. here, gradio, etc.) with EditorContainer ideally completely reusable with some configuration via callbacks

rholinshead (approved):
LGTM as a starting point. We can update to single-parameter granularity in subsequent PR

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
saqadri: Resolving this comment because we discussed offline and realized that per-parameter updates won't work reliably:

Suppose I have parameter1: `{key: "test", val: "value"}`
Now I'm typing parameter2, which eventually will be `{key: "test2", val: "something"}`
While typing, parameter2's key becomes `"test"` (i.e. parameter1.key is a prefix of parameter2).
I send the setParameter request at this point, which will overwrite parameter1's value.
Eventually, I finish typing, so parameter2.key is now `"test2"`. Now "oldParameterKey" = `"test"`, and the update deletes `"test"` key and just leaves `"test2"`.
So the update would have wiped out parameter1 entirely from the parameters object.

The best option is to keep updating parameters as a single update to the entire `parameters` JSON object

rossdanlm: Why not wait until user clicks away from the editing UI component to send the request instead of `onUpdateParameters`? It seems wasteful to dispatch a server request until user is 'finished' setting their params

rholinshead: I don't think either UX changes what's needed. If we only send the request when the user clicks away we still want to send one request of all parameters and have the server set them each instead of the client waiting and setting n param update requests, right?

PR #618, opened 2023-12-26
[easy][ts] 2/n Set run output type to be Output[] instead of Union

Ankush-lastmile:
[easy][ts] 2/n Set run output type to be Output[] instead of Union

## What

Updating the types in the ts sdk to return `Output[]` instead of `Output | Output[]`

## Why

Run method should always return a array of outputs, even if its just one output.

The Prompt schema already refers to `outputs` as `outputs?: Output[];` anyways

## Testplan

Ran tests without errors. No type errors on compilation either

Screenshot 2023-12-26 at 1 16 46 PM

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/618).
* __->__ #618
* #617

### DIFF
--- a/typescript/lib/config.ts
+++ b/typescript/lib/config.ts
@@ -398,7 +398,7 @@ export class AIConfigRuntime implements AIConfig {
     promptName: string,
     params: JSONObject = {},
     options?: InferenceOptions
-  ) {
+  ): Promise<Output[]> {
     const startEvent = {
       name: "on_run_start",
       file: __filename,
@@ -452,7 +452,7 @@ export class AIConfigRuntime implements AIConfig {
     promptName: string,
     params: JSONObject = {},
     options?: InferenceOptions
-  ) {
+  ): Promise<Output[] | undefined> {
     const prompt = this.getPrompt(promptName);
     if (!prompt) {
       throw new Error(
--- a/typescript/lib/parameterizedModelParser.ts
+++ b/typescript/lib/parameterizedModelParser.ts
@@ -63,7 +63,7 @@ export abstract class ParameterizedModelParser<
     aiConfig: AIConfigRuntime,
     options?: InferenceOptions,
     params: JSONObject = {}
-  ) {
+  ): Promise<Output[] | undefined> {
     const dependencyGraph = getDependencyGraph(aiConfig);
 
     return await this.runWithDependenciesInternal(
--- a/cookbooks/HuggingFace/hf.py
+++ b/cookbooks/HuggingFace/hf.py
@@ -1,5 +1,5 @@
 import copy
-from typing import TYPE_CHECKING, Any, Dict, Optional
+from typing import TYPE_CHECKING, Any, Dict, List, Optional
 
 # HuggingFace API imports
 from huggingface_hub import InferenceClient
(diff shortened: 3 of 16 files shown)


typescript/lib/config.ts
```diff
@@ -398,7 +398,7 @@ export class AIConfigRuntime implements AIConfig {
     promptName: string,
     params: JSONObject = {},
     options?: InferenceOptions
-  ) {
+  ): Promise<Output[]> {
```
Ankush-lastmile, line 401: line 430 below guarantees the output of run() is an array of outputs

typescript/lib/parameterizedModelParser.ts
```diff
@@ -63,17 +63,23 @@ export abstract class ParameterizedModelParser<
...
-    return await this.runWithDependenciesInternal(
+    const result =  await this.runWithDependenciesInternal(
       promptName,
       aiConfig,
       params,
       dependencyGraph,
       /*alreadyExecutedPrompts*/ new Set<string>(),
       options
     );
+    if (result === undefined) {
```
saqadri, line 77: This doesn't quite make sense to me.

Are we sure runWithDependencies will always return something?

You can also do

if (!result) return []

saqadri (approved):
This mostly lgtm. One thing I'd suggest as followup is cleaning up callsites where we check for array or singular output, since now that can all be simplified

typescript/lib/parameterizedModelParser.ts
Ankush-lastmile: `runWithDependencies()` is designed to return either
1. `undefined` when [ it reaches the base case of recursion](https://github.com/lastmile-ai/aiconfig/blob/f1fb2e1965bd678f66c9ccb94e22722187231134/typescript/lib/parameterizedModelParser.ts#L138)
2. `Output[]` [once it successfully calls run()](https://github.com/lastmile-ai/aiconfig/blob/f1fb2e1965bd678f66c9ccb94e22722187231134/typescript/lib/parameterizedModelParser.ts#L135-L138)

The initial call of runWithDependencies is not expected to hit the base case; therefore, it should typically return an array of Output[]. I implemented an exception to handle unexpected base case hits, ensuring that we explicitly acknowledge this anomaly rather than silently fail or return an empty array.

Would the expected behavior be, instead, to return an empty array when encountering such case?

saqadri: If runWithDependencies returns undefined, this function should also return undefined I think

saqadri: The error being thrown is very generic, which is why it gets me worried.

Ankush-lastmile: >If runWithDependencies returns undefined, this function should also return undefined I think
runWithDependencies was untyped previously. In this diff I typed it to `Output[]`.

If it's fine for the output type to include `undefined` then this can also be undefined and send it forward. Should I go ahead and do that?

saqadri: Yes good idea, let's do that

Ankush-lastmile:
- rebase
- `runwithdependencies` output type defined as : `Output[] | undefined`
- `runWithDependencies`  output type defined as :  Output[] | undefined

typescript/lib/parameterizedModelParser.ts
Ankush-lastmile: updated

Ankush-lastmile:
> One thing I'd suggest as followup

created #645

PR #624, opened 2023-12-26
[editor] Functionality to run server on multiple configs without specifying port

Ankush-lastmile:
[editor] Functionality to run server on multiple configs without specifying port

## What:

Before creating the server process, check if the port to be used is taken. If so, find the next available one before continuing

for context: this builds on top of #619 

## Why

When running the editor, user should be able to open multiple configs.

Under the hood this means spinning up another server in an unused port. Frontend already supports dynamic ports.

## Testplan

1. yarn build in aiconfig/python/src/aiconfig/editor/client
2. Open two aiconfigs with "prod" editor"

| Getting Started aiconfig | Chain Of Verification aiconfig|
|------------|-------------|
| Screenshot 1 |  Screenshot 2|
| Screenshot 3 | Screenshot 4|

Terminals showcase the command used to open editor. Each one has a seperate config file path. No port is passed in from the user.

Chain of verification Uses Port 8080
Getting Started in a new process using Port 8081

### DIFF
--- a/python/src/aiconfig/scripts/aiconfig_cli.py
+++ b/python/src/aiconfig/scripts/aiconfig_cli.py
@@ -3,6 +3,7 @@ import logging
 import signal
 import subprocess
 import sys
+import socket
 
 import lastmile_utils.lib.core.api as core_utils
 import result
@@ -64,8 +65,33 @@ def _sigint(procs: list[subprocess.Popen[bytes]]) -> Result[str, str]:
         p.send_signal(signal.SIGINT)
     return Ok("Sent SIGINT to frontend servers.")
 
+def is_port_in_use(port: int) -> bool:
+    """ 
+    Checks if a port is in use at localhost.
+    
+    Creates a temporary connection.
+    Context manager will automatically close the connection
+    """
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+        return s.connect_ex(('localhost', port)) == 0
+
 
 def _run_editor_servers(edit_config: EditServerConfig) -> Result[list[str], str]:
+    
+    port = edit_config.server_port
+
+    while is_port_in_use(port):
+        LOGGER.warning(f"Port {port} is in use. Checking next port.")
+        port += 1
+
+    # Must reconstruct, EditServerConfig is an immutable type (frozen dataclass)
+    edit_config_dict = edit_config.model_dump()
+    edit_config_dict["server_port"] = port
+    edit_config = EditServerConfig(**edit_config_dict)
+
+    LOGGER.warning(f"Using {port}.")
+
+    # Check if server is already running
     LOGGER.info("Running editor servers")
     frontend_procs = _run_frontend_server_background() if edit_config.server_mode in [ServerMode.DEBUG_SERVERS] else Ok([])
     match frontend_procs:


rholinshead:
In the second screenshot, the terminal output shows port 8082 but browser shows 8081. Just double-checking this is just an incorrect image / I assume you ran it once more after 8081 was already in use?

rholinshead (approved):
LGTM

saqadri (approved):
Amazing, thanks a lot for the simple fix

Ankush-lastmile:
>In the second screenshot, the terminal output shows port 8082 but browser shows 8081. Just double-checking this is just an incorrect image / I assume you ran it once more after 8081 was already in use?

Yes good catch. Re-executed test plan and the ports are as expected. Updated Testplan as well

python/src/aiconfig/scripts/aiconfig_cli.py
```diff
@@ -64,8 +65,33 @@ def _sigint(procs: list[subprocess.Popen[bytes]]) -> Result[str, str]:
...
+    port = edit_config.server_port
+
+    while is_port_in_use(port):
+        LOGGER.warning(f"Port {port} is in use. Checking next port.")
+        port += 1
+
+    # Must reconstruct, EditServerConfig is an immutable type (frozen dataclass)
+    edit_config_dict = edit_config.model_dump()
+    edit_config_dict["server_port"] = port
+    edit_config = EditServerConfig(**edit_config_dict)
```
jonathanlastmileai, line 90: `model_copy` is what you want here https://docs.pydantic.dev/latest/concepts/serialization/#model_copy

python/src/aiconfig/scripts/aiconfig_cli.py
```diff
@@ -64,8 +65,33 @@ def _sigint(procs: list[subprocess.Popen[bytes]]) -> Result[str, str]:
...
+
+    LOGGER.warning(f"Using {port}.")
```
jonathanlastmileai, line 92: nit: .info

python/src/aiconfig/scripts/aiconfig_cli.py
```diff
@@ -64,8 +65,33 @@ def _sigint(procs: list[subprocess.Popen[bytes]]) -> Result[str, str]:
         p.send_signal(signal.SIGINT)
     return Ok("Sent SIGINT to frontend servers.")
 
+def is_port_in_use(port: int) -> bool:
```
jonathanlastmileai, line 68: I think this is fine for now, but there are both functional and UX problems with this approach.

Functional: It might be unlikely, but technically you can have a race condition where the port gets used after this returns but before the flask server starts.

UX: If I pass a port explicitly and it's already in use, I expect an error. I don't want the app to find the next open port for me. 

Here's what I would have done:

- If user provides port, don't do any special logic
- Otherwise, loop over all possible ports (e.g. starting at 8080) and try to start the server until one works. 


cc @saqadri , @rossdanlm

rossdanlm: Makes sense, but I think this could be seen as aan enhancement instead of a bug. I made a GH issue for this in https://github.com/lastmile-ai/aiconfig/issues/633 as a bootcamp task that others can do if they like

PR #599, opened 2023-12-23
[AIC-py] optional: add prompt at index

jonathanlastmileai:
[AIC-py] optional: add prompt at index

add p3, add p4 at index 1.

```
[Prompt(name='get_activities', input='Tell me 10 fun attractions to do in NYC.', metadata=None, outputs=[]),
 Prompt(name='p4', input='thep', metadata=None, outputs=[]),
 Prompt(name='gen_itinerary', input='Generate an itinerary ordered by {{order_by}} for these activities: {{get_activities.output}}.', metadata=PromptMetadata(model='gpt-4', tags=None, parameters={'order_by': 'geographic location'}), outputs=[]),
 Prompt(name='p3', input='thep', metadata=None, outputs=[])]
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/599).
* #600
* __->__ #599

PR #626, opened 2023-12-26
[editor] Add Prompt Button

rholinshead:
[editor] Add Prompt Button

[editor] Add Prompt Button

# [editor] Add Prompt Button

Implement the button for adding a prompt below/between prompts

Screenshot 2023-12-26 at 6 10 12 PM

https://github.com/lastmile-ai/aiconfig/assets/5060851/c3ba1699-adbb-4f7a-a001-9a6972f31cb6

Ensure the new prompt has correct name, model and settings renderer from the schema:
Screenshot 2023-12-26 at 6 11 10 PM

### DIFF
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -6,7 +6,7 @@ flask[async]
 google-generativeai
 huggingface_hub
 hypothesis==6.91.0
-lastmile-utils==0.0.13
+lastmile-utils==0.0.14
 mock
 nest_asyncio
 nltk
--- a/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
@@ -1,17 +1,46 @@
 import PromptContainer from "./prompt/PromptContainer";
-import { Container, Group, Button, createStyles } from "@mantine/core";
+import { Container, Group, Button, createStyles, Stack } from "@mantine/core";
 import { showNotification } from "@mantine/notifications";
-import { AIConfig, PromptInput } from "aiconfig";
-import { useCallback, useReducer, useState } from "react";
-import aiconfigReducer from "./aiconfigReducer";
+import { AIConfig, Prompt, PromptInput } from "aiconfig";
+import { useCallback, useReducer, useRef, useState } from "react";
+import aiconfigReducer, { AIConfigReducerAction } from "./aiconfigReducer";
 import { ClientAIConfig, clientConfigToAIConfig } from "../shared/types";
+import AddPromptButton from "./prompt/AddPromptButton";
+import { getDefaultNewPromptName } from "../utils/aiconfigStateUtils";
 
 type Props = {
   aiconfig: ClientAIConfig;
+  addPrompt: (
+    promptName: string,
+    prompt: Prompt,
+    index: number
+  ) => Promise<{ aiconfig: AIConfig }>;
   onSave: (aiconfig: AIConfig) => Promise<void>;
+  getModels: (search: string) => Promise<string[]>;
 };
 
 const useStyles = createStyles((theme) => ({
+  addPromptRow: {
+    borderRadius: "4px",
+    display: "inline-block",
+    bottom: -24,
+    left: -40,
+    "&:hover": {
+      backgroundColor:
+        theme.colorScheme === "light"
+          ? theme.colors.gray[1]
+          : "rgba(255, 255, 255, 0.1)",
+    },
+    [theme.fn.smallerThan("sm")]: {
+      marginLeft: "0",
+      display: "block",
+      position: "static",
+      bottom: -10,
+      left: 0,
+      height: 28,
+      margin: "10px 0",
+    },
+  },
   promptsContainer: {
     [theme.fn.smallerThan("sm")]: {
       padding: "0 0 200px 0",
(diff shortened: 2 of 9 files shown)


python/requirements.txt
```diff
@@ -6,7 +6,7 @@ flask[async]
 google-generativeai
 huggingface_hub
 hypothesis==6.91.0
-lastmile-utils==0.0.13
+lastmile-utils==0.0.14
```
rholinshead, line 9: Needed to bump this since `server_utils` needs `core_utils.safe_model_validate_json_object`

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -81,6 +114,43 @@ export default function EditorContainer({
     [dispatch]
   );
 
+  const onAddPrompt = useCallback(
+    async (promptIndex: number, model: string) => {
+      const promptName = getDefaultNewPromptName(stateRef.current as AIConfig);
+      const newPrompt: Prompt = {
```
rholinshead, line 120: Not a huge fan of this right now. Ideally we could just send up the model name and have the proper prompt structure inserted into the config to reconsolidate but I think we need to have the input schemas implemented in order to do so. Also, if no schema exists we will probably want to default to `{}` in the input with JSON renderer?

We can update as a follow-up

python/src/aiconfig/editor/client/src/components/aiconfigReducer.ts
```diff
@@ -50,11 +67,60 @@ function reduceReplaceInput(
...
+  };
+}
+
+function reduceConsolidateAIConfig(
+  state: ClientAIConfig,
+  action: MutateAIConfigAction,
+  responseConfig: AIConfig
+): ClientAIConfig {
+  switch (action.type) {
+    case "ADD_PROMPT_AT_INDEX": {
```
rholinshead, line 91: I don't necessarily think any reconsolidation is necessary after adding a prompt, but wanted to have this as an example. Some actions could potentially update the config in the server and we should make sure those changes are subsequently reflected on the client as well. We should still prioritize client-side input/metadata since it can be changed on the UI while the server requests are still in flight, so we don't want to overwrite the client with the server state for those values.

As a side note, I realize this is going to be problematic as-is since add_prompt on the config size doesn't support specifying an index yet. So, we can add at any index on the client but only the end on the server. I'll update the code to just have the add prompt in the UI at the bottom for now

rholinshead: Done

rholinshead:
Opened issue for saving not working: https://github.com/lastmile-ai/aiconfig/issues/627

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
saqadri: Agreed with this

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -100,15 +170,31 @@ export default function EditorContainer({
...
+                index={i}
+                prompt={prompt}
+                onChangePromptInput={onChangePromptInput}
+                onUpdateModelSettings={onUpdatePromptModelSettings}
+                onUpdateParameters={onUpdatePromptParameters}
+                defaultConfigModelName={aiconfigState.metadata.default_model}
+              />
+              {i ===
+                aiconfigState.prompts.length -
+                  1 /* TODO: Remove this check once we can add prompts at any index */ && (
```
saqadri, line 184: Hasn't this been added now? 
https://github.com/lastmile-ai/aiconfig/pull/599

python/src/aiconfig/editor/server/server.py
```diff
@@ -182,6 +182,7 @@ def _get_op_args():
 @app.route("/api/add_prompt", methods=["POST"])
 def add_prompt() -> FlaskResponse:
     method_name = MethodName("add_prompt")
+    # TODO: Support specifying index
     signature: dict[str, Type[Any]] = {"prompt_name": str, "prompt_data": Prompt}
```
saqadri, line 186: This can be added (maybe follow up diff), because #599 added it to the library.

Aside -- not sure why we've made this so convoluted to call the aiconfig api in the server. Why do we need to do `make_op_run_method` instead of just simply calling `aiconfig.add_prompt`. Looks over-engineered. cc @jonathanlastmileai  @rossdanlm. Not ship-blocking but mentioning this for the near future.

saqadri (approved):

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
rholinshead: Ah weird. I could have sworn my local copy didn't have it when I cmd+clicked in... But it is there now so it's a quick add. Will do so here

rholinshead: https://github.com/lastmile-ai/aiconfig/assets/5060851/46ecfef1-df55-4cf1-80eb-fe55dc1667fc

python/src/aiconfig/editor/server/server.py
rholinshead: Ah, whoops. Not sure how I didn't see it but added now

jonathanlastmileai: Fair enough, this could be simpler. Basically it reduces boilerplate for input JSON schema checking and constructing HTTP responses.

PR #628, opened 2023-12-27
[RFC] Add "function" as an explicit kind of output data

saqadri:
[RFC] Add "function" as an explicit kind of output data

This can help us distinguish function calling from regular strings quickly -- since function calling is now supported across multiple models, it deserves to be a top-level output kind in the schema.

For example, current models that support it:
* GPT-3.5, GPT-4, GPT-4V
* Gemini
* [Mixtral (using Anyscale endpoints)](https://docs.endpoints.anyscale.com/guides/function-calling/#:~:text=With%20Anyscale%20Endpoints%2C%20you%20can,parameters%20to%20pass%20to%20it.)
* There will be more to come

### DIFF
--- a/typescript/types.ts
+++ b/typescript/types.ts
@@ -196,6 +196,14 @@ export type ExecuteResult = {
     | {
         kind: "string" | "file_uri" | "base64";
         value: string;
+      }
+    | {
+        kind: "function";
+        value: {
+          name: string;
+          arguments: string;
+          [k: string]: any;
+        }[];
       };
   /**
    * The MIME type of the result. If not specified, the MIME type will be assumed to be plain text.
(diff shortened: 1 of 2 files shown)


rossdanlm (approved):
thanks!

typescript/types.ts
```diff
@@ -196,6 +196,14 @@ export type ExecuteResult = {
...
         kind: "string" | "file_uri" | "base64";
         value: string;
+      }
+    | {
+        kind: "function";
+        value: {
+          name: string;
+          arguments: string;
+          [k: string]: any;
+        }[];
```
rossdanlm, line 206: Oh I realize, we probably shouldn't have put a list here, will fix

Issue #627, opened 2023-12-26
[editor] Saving Fails Due to Invalid Path


rholinshead:
When clicking the save button in the editor, I get an error:
```
[INFO] 2023-12-26 18:12:04,160 server_utils.py:275: About to run operation: save, res_op_args=Ok({'path': '/Users/ryanholinshead/Projects/aiconfig/None'})
[INFO] 2023-12-26 18:12:04,160 server_utils.py:250: Running method: save, operation_args={'path': '/Users/ryanholinshead/Projects/aiconfig/None'}
[INFO] 2023-12-26 18:12:04,160 server_utils.py:233: Failed to run method (save): AIConfigRuntime.save() got an unexpected keyword argument 'path'
[INFO] 2023-12-26 18:12:04,162 server_utils.py:282: res_run=Err('\nException:\nAIConfigRuntime.save() got an unexpected keyword argument \'path\'\nTraceback (most recent call last):\n  File "/Users/ryanholinshead/Projects/aiconfig/python/src/aiconfig/editor/server/server_utils.py", line 229, in _safe_run_aiconfig_method\n    out = method(**method_args)\n          ^^^^^^^^^^^^^^^^^^^^^\nTypeError: AIConfigRuntime.save() got an unexpected keyword argument \'path\'\n')
```

rholinshead:
Closed with fix https://github.com/lastmile-ai/aiconfig/pull/629

PR #629, opened 2023-12-27
[bug fix][issue #627] pass the correct method arg name down to aiconfig save

jonathanlastmileai:
[bug fix][issue #627] pass the correct method arg name down to aiconfig save

Fix the argument name: "path" -> "config_filepath".
Also add a couple of related path fixes to make
other endpoints more robust.

Test:

Send requests to /load_model_parser_module, /load, and /save
with valid and invalid paths, and with incorrectly-named arguments.
Also make sure /load runs with no arguments.

Client code:
```
import requests

for url in [
    'http://localhost:8080/api/load',
    'http://localhost:8080/api/save',
    'http://localhost:8080/api/load_model_parser_module',
]:
# url = 'http://localhost:8080/api/load'
    for data in [
        {},
        {
            "path": "/Users/jonathan/Projects/aiconfig/my_aiconfig_model_registry.py"
        },
        {
            "config_filepath": "fake"
        },
        {
            "path": "python/src/aiconfig/editor/travel.aiconfig.json"
        },
        {
            "path": "python/src/aiconfig/editor/travel.aiconfig.json",
            "some_other_key": "value"
        }
    ]:
        print(f"\n\n{url=}, {data=}")

        print(f"\n{response=}"),

        import json
        response_json = json.loads(response.text)
        print(f"{response_json=}")
        print(f"{response_json['message']=}")
        aic = response_json.get('aiconfig', {})
        print("AIC:", aic)
```

### DIFF
--- a/python/src/aiconfig/editor/server/server.py
+++ b/python/src/aiconfig/editor/server/server.py
@@ -69,8 +69,10 @@ def run_backend_server(edit_config: EditServerConfig) -> Result[str, str]:
 
 
 def _validated_request_path(request_json: core_utils.JSONObject, allow_create: bool = False) -> Result[ValidatedPath, str]:
-    path = str(request_json.get("path", None))
-    return get_validated_path(path, allow_create=allow_create)
+    if "path" not in request_json or not isinstance(request_json["path"], str):
+        return Err("Request JSON must contain a 'path' field with a string value.")
+    else:
+        return get_validated_path(request_json["path"], allow_create=allow_create)
 
 
 @app.route("/")
@@ -102,6 +104,10 @@ def load_model_parser_module() -> FlaskResponse:
 def load() -> FlaskResponse:
     state = get_server_state(app)
     request_json = request.get_json()
+    if not request_json.keys() <= {"path"}:
+        return HttpResponseWithAIConfig(
+            message="Request JSON must contain a 'path' field with a string value, or no arguments.", code=400, aiconfig=None
+        ).to_flask_format()
     path: str | None = request_json.get("path", None)
     if path is None:
         aiconfig = state.aiconfig
@@ -129,11 +135,11 @@ def save() -> FlaskResponse:
 
     _op = make_op_run_method(MethodName("save"))
 
-    path_val = _validated_request_path(request_json, allow_create=True)
+    res_path_val: Result[ValidatedPath, str] = _validated_request_path(request_json, allow_create=True)
     op_args: Result[OpArgs, str] = result.do(
-        Ok(OpArgs({"path": path_val_ok}))
+        Ok(OpArgs({"config_filepath": path_val_ok}))
         #
-        for path_val_ok in path_val
+        for path_val_ok in res_path_val
     )
 
     return run_aiconfig_operation_with_op_args(aiconfig, "save", _op, op_args)
--- a/python/src/aiconfig/editor/server/server_utils.py
+++ b/python/src/aiconfig/editor/server/server_utils.py
@@ -107,7 +107,7 @@ def resolve_path(path: str) -> str:
     return os.path.abspath(os.path.expanduser(path))
 
 
-def get_validated_path(raw_path: str, allow_create: bool = False) -> Result[ValidatedPath, str]:
+def get_validated_path(raw_path: str | None, allow_create: bool = False) -> Result[ValidatedPath, str]:
     LOGGER.debug(f"{allow_create=}")
     if not raw_path:
         return Err("No path provided")
--- a/python/src/aiconfig/editor/travel.aiconfig.json
+++ b/python/src/aiconfig/editor/travel.aiconfig.json
@@ -1,115 +1,120 @@
 {
-    "name": "NYC Trip Planner",
-    "description": "Intrepid explorer with ChatGPT and AIConfig",
-    "schema_version": "latest",
-    "metadata": {
-        "models": {
-            "gpt-3.5-turbo": {
-                "model": "gpt-3.5-turbo",
-                "top_p": 1,
-                "temperature": 1
-            },
-            "gpt-4": {
-                "model": "gpt-4",
-                "max_tokens": 3000,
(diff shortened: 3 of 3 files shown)


saqadri (approved):

python/src/aiconfig/editor/server/server.py
```diff
@@ -129,11 +135,11 @@ def save() -> FlaskResponse:
 
     _op = make_op_run_method(MethodName("save"))
 
-    path_val = _validated_request_path(request_json, allow_create=True)
+    res_path_val: Result[ValidatedPath, str] = _validated_request_path(request_json, allow_create=True)
     op_args: Result[OpArgs, str] = result.do(
-        Ok(OpArgs({"path": path_val_ok}))
+        Ok(OpArgs({"config_filepath": path_val_ok}))
```
Ankush-lastmile, line 140: I'm not that familiar with the Result library, can you explain what this line is doing?

python/src/aiconfig/editor/server/server_utils.py
```diff
@@ -107,7 +107,7 @@ def resolve_path(path: str) -> str:
     return os.path.abspath(os.path.expanduser(path))
 
 
-def get_validated_path(raw_path: str, allow_create: bool = False) -> Result[ValidatedPath, str]:
+def get_validated_path(raw_path: str | None, allow_create: bool = False) -> Result[ValidatedPath, str]:
```
Ankush-lastmile, line 110: why is None now allowed?

python/src/aiconfig/editor/server/server.py
```diff
@@ -102,6 +104,10 @@ def load_model_parser_module() -> FlaskResponse:
 def load() -> FlaskResponse:
     state = get_server_state(app)
     request_json = request.get_json()
+    if not request_json.keys() <= {"path"}:
```
Ankush-lastmile, line 107: cool trick, is this the best/standard way to validate requests in flask? 

Just a note that FastApi handles this by making request data pydantic objects

python/src/aiconfig/editor/server/server.py
```diff
@@ -129,11 +135,11 @@ def save() -> FlaskResponse:
 
     _op = make_op_run_method(MethodName("save"))
 
-    path_val = _validated_request_path(request_json, allow_create=True)
+    res_path_val: Result[ValidatedPath, str] = _validated_request_path(request_json, allow_create=True)
```
saqadri, line 138: Approved this but this code is super confusing and looks unnecessarily over abstracted. Why not just call config.save directly?

I added a similar comment in this diff: https://github.com/lastmile-ai/aiconfig/pull/626#discussion_r1436661063

Ankush-lastmile:
> with valid and invalid paths, and with incorrectly-named arguments.

Can you add a visual for the testplan? A video or picture would be really helpful to visualize this change

python/src/aiconfig/editor/server/server.py
jonathanlastmileai: This is unrelated to Flask. I'm just taking the dict keys as a set and comparing them to the set containing "path".

python/src/aiconfig/editor/server/server_utils.py
jonathanlastmileai: You're right, I don't think it can actually be None at any of the call sites. Just expanding the accepted values to be defensive.

python/src/aiconfig/editor/server/server.py
jonathanlastmileai: Basically it's error handling with less boilerplate. Otherwise every endpoint would have to `try...except` and return a response with a not-necessarily-clear error message. It is pretty abstract and could be simplified, point taken. The thinking is to move the tricker parts into a tested and reusable module (server_utils) and then have the main server file be relatively clean.

saqadri: The pattern could instead take a callback fn which it invokes inside a standardized try/catch with standard error handling, but it's very confusing that the function tries to construct a signature and function in such an abstract way.

The callback passed in could call config.* directly then

python/src/aiconfig/editor/server/server.py
jonathanlastmileai: Sure - basically this expression evaluates to the Ok(...) expression if everything goes right, and otherwise -- as soon as there's an Err -- it short-circuits and evaluates to that Err value. Using Result[T, U] this way is a little like propagating exceptions, but statically checkable (all Python exceptions are unchecked). 

It's a bit like a Result comprehension, sort of analogous to a list comprehension.

python/src/aiconfig/editor/server/server.py
jonathanlastmileai: You're right, this can be simplified quite a bit - I can work on that after higher-pri stuff.

The motivation for this signature thing in particular was that not all AIConfig methods type check. e.g. `run()` does not type check:

```
Type of "run" is partially unknown
  Type of "run" is "(prompt_name: str, params: dict[Unknown, Unknown]
...
```

The explicit signature was supposed to be a stand-in for that. But I agree this can be simplified a lot, and it's much a better solution to fix the AIConfig method typing so we can call them directly with type checking and less abstraction.

python/src/aiconfig/editor/server/server.py
rossdanlm: Can we keep it simple and say `if len(request_json.keys()) != 0 or request_json.keys() != {"path"}`? using `__lt__` is not intuitive to read or understand. I get that this logic is "elegant", but I had to play around to figure it out
Screenshot 2023-12-28 at 14 04 28

jonathanlastmileai: Sure, added here https://docs.google.com/document/d/1YwUlIQZX8CiuvtbVR4OvpFQ2nphFITRD_XZR3CDNtoE/edit#bookmark=id.8h0jqqaolx9z

PR #631, opened 2023-12-27
Better typing for AIConfig schema

saqadri:
Better typing for AIConfig schema

We had `JSONValue` defined as possibly `unknown` type, which made the schema generator ignore that type entirely.

This change removes that requirement, while still allowing flexibility for what gets assigned to `JSONObject`.

Regenerated the schema with `yarn gen-schema` and we get better typings now.

### DIFF
--- a/schema/aiconfig.schema.json
+++ b/schema/aiconfig.schema.json
@@ -164,6 +164,7 @@
                 "type": "object",
                 "properties": {
                   "data": {
+                    "$ref": "#/definitions/JSONValue",
                     "description": "The data representing the attachment"
                   },
                   "mime_type": {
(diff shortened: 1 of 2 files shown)


saqadri:
Test Plan:
```
yarn build
```

typescript/common.ts
```diff
@@ -1,5 +1,5 @@
 // From https://github.com/microsoft/TypeScript/issues/1897
-export type JSONPrimitive = string | number | boolean | null | unknown;
+export type JSONPrimitive = string | number | boolean | null | undefined;
```
rholinshead, line 2: I think we should revisit later and see if we can remove undefined in favour of making the property optional where applicable. `undefined` is not actually a valid JSON value if we were to deserialize a JSON string

typescript/common.ts
```diff
@@ -1,5 +1,5 @@
...
 export type JSONValue = JSONPrimitive | JSONObject | JSONArray;
-export type JSONObject = { [member: string]: JSONValue };
+export type JSONObject = { [member: string]: JSONValue | any };
```
rholinshead, line 4: Hmm, why is this needed? It seems to defeat the purpose of the JSONObject typing

schema/aiconfig.schema.json
```diff
@@ -208,7 +232,79 @@
...
+                {
+                  "type": "object",
+                  "additionalProperties": {}
+                },
+                {
+                  "type": "array",
+                  "items": {
+                    "$ref": "#/definitions/JSONValue"
+                  }
+                },
```
rholinshead, line 246: Weird that this doesn't just use the JSONValue ref

typescript/common.ts
jonathanlastmileai: undefined is definitely better than unknown though, right ?

saqadri:
Going to rebase and keep this diff open so I can try getting rid of `undefined`. My 2c (as described in https://github.com/lastmile-ai/aiconfig/pull/636#discussion_r1437111185) -- I think this isn't perfect, but moves in a better direction because `unknown` typing was worse before (including for JSONObject when it relied on JSONValue)

typescript/common.ts
saqadri: Take a look at https://lastmileai.dev/workbooks/clqnz6ebq00ekpew37w20ctzg. I can't really figure out a way around this

schema/aiconfig.schema.json
saqadri: Good question

rossdanlm:
Btw https://github.com/lastmile-ai/aiconfig/pull/635 already landed (I tried rebasing off this diff but it created a duplicate copy). If we want to make changes we can fix forward

typescript/common.ts
rholinshead: Ah, darn. I think we might need to leave this for a future iteration and potentially move to using Record in place of JSONObject where applicable. JSONObject was intended to be a valid representation of what the object can be based on supporting JSON serialize/deserialize requirements but seems like it's too restrictive without the any/undefined

PR #635, opened 2023-12-27
Better typing for AIConfig schema

rossdanlm:
Better typing for AIConfig schema

We had `JSONValue` defined as possibly `unknown` type, which made the schema generator ignore that type entirely.

This change removes that requirement, while still allowing flexibility for what gets assigned to `JSONObject`.

Regenerated the schema with `yarn gen-schema` and we get better typings now.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/635).
* #636
* #634
* __->__ #635

PR #636, opened 2023-12-27
Update function call schema to mirror ChatCompletionMessageToolCall

rossdanlm:
Update function call schema to mirror ChatCompletionMessageToolCall

OpenAI seems like the standard API for now, so we're just going to follow that:
Screenshot 2023-12-27 at 03 31 20

```
cd typescript
yarn gen-schema
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/636).
* __->__ #636
* #634
* #635


typescript/types.ts
```diff
@@ -174,6 +174,47 @@ export type Prompt = {
...
+      value: string;
+    }
+  /**
+   * The following is based off of ChatCompletionMessageToolCall
+   * from openai/types/chat and is used for general tool calls.
+   */
+  | {
+      kind: "tool_calls";
+      value: {
+        id: string;
```
saqadri, line 193: nit: should we make this optional? Not sure if everyone would have an id

saqadri: Or make it nullable -- string | null

saqadri (approved):
Great point consolidating with OpenAI format. Tools can be more than just function calls in the near future

typescript/types.ts
rossdanlm: sg I'll make it optional

typescript/common.ts
```diff
@@ -1,5 +1,5 @@
 // From https://github.com/microsoft/TypeScript/issues/1897
-export type JSONPrimitive = string | number | boolean | null | unknown;
+export type JSONPrimitive = string | number | boolean | null | undefined;
 export type JSONValue = JSONPrimitive | JSONObject | JSONArray;
-export type JSONObject = { [member: string]: JSONValue };
+export type JSONObject = { [member: string]: JSONValue | any };
```
rholinshead, line 4: I left a comment in https://github.com/lastmile-ai/aiconfig/pull/631 (note, your schema typing PR was still in draft but merged) but think we should reconsider this. Ideally JSONPrimitive should not support unknown and JSONObject shouldn't support any

saqadri: The issue is that there are many assignments to properties where type checking complains unless JSONObject is Record<string, any>.

If you remove the any union from JSONObject and open openai.ts, it'll show the issues. It's still a step in the right direction from  having unknown in JSONValue

PR #637, opened 2023-12-27
Update output schema to not use {"string": string_value} as output data

rossdanlm:
Update output schema to not use {"string": string_value} as output data

See comment in https://github.com/lastmile-ai/aiconfig/pull/610#discussion_r1437099426

Now we're just going to directly check for the type being a pure string and not embedded within the `OutputData` object. I also renamed `OutputData` --> `OutputDataWithValue` to make this more explicit and easier to understand

Rebuilt the schema using `yarn gen-schema` from `aiconfig/typescript` dir

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/637).
* #610
* __->__ #637


saqadri (approved):
👨‍🍳 🚀

PR #642, opened 2023-12-27
[editor] Connect UPDATE_PROMPT_INPUT to update_prompt endpoint

rholinshead:
[editor] Connect UPDATE_PROMPT_INPUT to update_prompt endpoint

# [editor] Connect UPDATE_PROMPT_INPUT to update_prompt endpoint

Hook up the `UPDATE_PROMPT_INPUT` action to the server `update_prompt` endpoint, sending requests with a 250ms debounce so that we don't spam the server and avoid potential overlap.

We don't have server-side endpoints for updating specific parts of the prompt right now (not sure if that would be any major improvement), so we currently need to send a Prompt constructed with the updated input as a spread into the current prompt in state. We still maintain the UI state in the reducer instead of passing in this new spread prompt to the reducer from this callback since the reducer should be the source of truth for UI state and will ensure the proper immutable state flow is maintained. We can also add other state values (e.g. 'isExecuting', 'isDirty', 'isReadonly') in this reducer but leaving that out for now until it's for sure needed.

## Testing:
- Make some quick changes to the prompt input and see the proper updates and debounced requests

https://github.com/lastmile-ai/aiconfig/assets/5060851/b95d5116-9484-4d2f-88c3-de441507a140

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/642).
* #647
* #644
* #643
* __->__ #642

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
@@ -2,11 +2,12 @@ import PromptContainer from "./prompt/PromptContainer";
 import { Container, Group, Button, createStyles, Stack } from "@mantine/core";
 import { showNotification } from "@mantine/notifications";
 import { AIConfig, Prompt, PromptInput } from "aiconfig";
-import { useCallback, useReducer, useRef, useState } from "react";
+import { useCallback, useMemo, useReducer, useRef, useState } from "react";
 import aiconfigReducer, { AIConfigReducerAction } from "./aiconfigReducer";
 import { ClientAIConfig, clientConfigToAIConfig } from "../shared/types";
 import AddPromptButton from "./prompt/AddPromptButton";
 import { getDefaultNewPromptName } from "../utils/aiconfigStateUtils";
+import { debounce } from "lodash";
 
 type Props = {
   aiconfig: ClientAIConfig;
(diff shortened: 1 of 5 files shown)


python/src/aiconfig/editor/client/src/utils/aiconfigStateUtils.ts
```diff
@@ -8,7 +8,3 @@ export function getDefaultNewPromptName(aiconfig: AIConfig): string {
   }
   return `prompt_${i}`;
 }
-
-export function getPromptName(aiconfig: AIConfig, promptIndex: number): string {
```
rholinshead, line 12: This wasn't actually used/needed

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -82,16 +87,47 @@ export default function EditorContainer({
     }
   }, [aiconfigState, callbacks.save]);
 
+  const debouncedUpdatePrompt = useMemo(
```
rholinshead, line 90: Not 100% confident this debouncing should be done here vs in the Editor.tsx; I did it here since it seems like it should be 'our' responsibility to debounce external update calls instead of having to implement debounce in Editor.tsx and in the gradio root component

rholinshead:
Going to land this stack to unblock and can make any review changes on top

PR #646, opened 2023-12-27
[easy] Reference local version of aiconfig package for local editor

saqadri:
[easy] Reference local version of aiconfig package for local editor

While we iterate on the aiconfig schema and SDK, it's better to reference the latest code for the local editor (typescript side) to ensure we are in sync on the schema updates coming in.

It also means we don't need to publish to npm/pypi until we have let the changes bake.

### DIFF
--- a/python/src/aiconfig/editor/client/package.json
+++ b/python/src/aiconfig/editor/client/package.json
@@ -33,7 +33,7 @@
     "@mantine/notifications": "^6.0.7",
     "@mantine/prism": "^6.0.7",
     "@tabler/icons-react": "^2.44.0",
-    "aiconfig": "^1.1.0",
+    "aiconfig": "../../../../../typescript",
     "lodash": "^4.17.21",
     "node-fetch": "^3.3.2",
     "react": "^18",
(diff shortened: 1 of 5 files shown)


python/src/aiconfig/editor/client/package.json
```diff
@@ -33,7 +33,7 @@
     "@mantine/notifications": "^6.0.7",
     "@mantine/prism": "^6.0.7",
     "@tabler/icons-react": "^2.44.0",
-    "aiconfig": "^1.1.0",
+    "aiconfig": "../../../../../typescript",
```
saqadri, line 36: references the local folder <root>/typescript, which is the root of the aiconfig TypeScript package

rholinshead: Will we need to update this when we publish the aiconfig python package next week?

python/src/aiconfig/editor/client/src/components/prompt/model_settings/ModelSettingsConfigRenderer.tsx
```diff
@@ -1,4 +1,4 @@
-import { JSONObject } from "aiconfig/dist/common";
+import { JSONObject } from "aiconfig";
```
rholinshead, line 1: Ah, I had these same changes in https://github.com/lastmile-ai/aiconfig/pull/643 , plus using the import in the parameters renderer. Not an issue, I can rebase after this lands

python/src/aiconfig/editor/client/package.json
saqadri: I don't think so -- `yarn build` in this folder will package up everything it needs

rholinshead:
Awesome - thanks!

rholinshead (approved):

PR #647, opened 2023-12-27
[editor] Save without Requiring File Path from Client

rholinshead:
[editor] Save without Requiring File Path from Client

# [editor] Save without Requiring File Path from Client

The client currently doesn't know the file path that its config is saved at, so there's no way for the client to specify the 'path' to save the config at when sending the `save` request.

For now, updating the `save` endpoint to default to saving the config to the path specified in its associated AIConfigRuntime.

As we update the UX we should consider:
- Supporting changing config name/path from the client
- Sending the current path to the client on load

## Testing:
- Made changes to input (which are reflected in the server-side config)
- Click save, and ensure the config contents are saved to the config on disk

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/647).
* __->__ #647
* #644
* #643
* #642

### DIFF
--- a/python/src/aiconfig/editor/server/server.py
+++ b/python/src/aiconfig/editor/server/server.py
@@ -132,17 +132,23 @@ def save() -> FlaskResponse:
     state = get_server_state(app)
     aiconfig = state.aiconfig
     request_json = request.get_json()
+    path: str | None = request_json.get("path", None)
 
-    _op = make_op_run_method(MethodName("save"))
+    if path is None:
+        if aiconfig is None:
+            return HttpResponseWithAIConfig(message="No AIConfig loaded", code=400, aiconfig=None).to_flask_format()
+        else:
+            path = aiconfig.file_path
 
-    res_path_val: Result[ValidatedPath, str] = _validated_request_path(request_json, allow_create=True)
-    op_args: Result[OpArgs, str] = result.do(
-        Ok(OpArgs({"config_filepath": path_val_ok}))
-        #
-        for path_val_ok in res_path_val
-    )
+    res_path_val = get_validated_path(path, allow_create=True)
+    match res_path_val:
+        case Ok(path_ok):
+            _op = make_op_run_method(MethodName("save"))
+            op_args: Result[OpArgs, str] = result.Ok(OpArgs({"config_filepath": path_ok}))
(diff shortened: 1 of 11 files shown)


python/src/aiconfig/editor/server/server.py
```diff
@@ -132,15 +132,19 @@ def save() -> FlaskResponse:
     state = get_server_state(app)
     aiconfig = state.aiconfig
     request_json = request.get_json()
+    path: str | None = request_json.get("path", None)
 
-    _op = make_op_run_method(MethodName("save"))
+    if path is None:
+        if aiconfig is None:
+            return HttpResponseWithAIConfig(message="No AIConfig loaded", code=400, aiconfig=None).to_flask_format()
```
jonathanlastmileai, line 139: Shouldn't we do this check if there is a path, too?

python/src/aiconfig/editor/server/server.py
```diff
@@ -132,15 +132,19 @@ def save() -> FlaskResponse:
...
+        else:
+            path = aiconfig.file_path
+
+    if path is None:
+        return HttpResponseWithAIConfig(message="No path to save to", code=400, aiconfig=None).to_flask_format()
 
-    res_path_val: Result[ValidatedPath, str] = _validated_request_path(request_json, allow_create=True)
```
jonathanlastmileai, line 138: Are we skipping the path validation check now? Why?

python/src/aiconfig/editor/server/server.py
rholinshead: It is already being done in `run_aiconfig_operation_with_op_args`. Should we do it again here explicitly?

python/src/aiconfig/editor/server/server.py
rholinshead: We can't use `_validated_request_path` since it doesn't need to be in the json but thanks for pointing this out because we can/should use `get_validated_path` instead. Will update

python/src/aiconfig/editor/server/server.py
jonathanlastmileai: Right, yes, no need to duplicate.

jonathanlastmileai (approved):
I have little idea what's going on in the frontend code, but if it works, ship!

PR #648, opened 2023-12-28
Explicit str type checking in model parsers

rossdanlm:
Explicit str type checking in model parsers

See comment here for context: https://github.com/lastmile-ai/aiconfig/pull/611#discussion_r1437759469

Separating this diff from https://github.com/lastmile-ai/aiconfig/pull/611 to make it easier and get that one unblocked

### DIFF
--- a/extensions/LLama-Guard/python/python/src/aiconfig_extension_llama_guard/LLamaGuard.py
+++ b/extensions/LLama-Guard/python/python/src/aiconfig_extension_llama_guard/LLamaGuard.py
@@ -245,10 +245,15 @@ class LLamaGuardParser(ParameterizedModelParser):
         output_text = self.tokenizer.decode(
             response[0][prompt_len:], skip_special_tokens=True
         )
+        output_data_content: str = ''
+        if isinstance(output_text, str):
+            output_data_content = output_text
+        else:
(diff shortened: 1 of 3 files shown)


extensions/LLama-Guard/python/python/src/aiconfig_extension_llama_guard/LLamaGuard.py
```diff
@@ -245,10 +245,15 @@ async def run_inference(
         output_text = self.tokenizer.decode(
             response[0][prompt_len:], skip_special_tokens=True
         )
+        output_data_content : str = ''
```
rholinshead, line 248:
```suggestion
        output_data_content: str = ''
```
nit

extensions/llama/python/llama.py
```diff
@@ -87,9 +87,15 @@ async def _run_inference_helper(self, model_input, options) -> List[Output]:
                 if options:
                     options.stream_callback(data, acc, index)
             print(flush=True)
+
+            output_data_value : str = ''
```
rholinshead, line 91:
```suggestion
            output_data_value: str = ''
```
nit

rholinshead (approved):
LGTM, just nits

PR #652, opened 2023-12-28
[editor] Text Output Rendering

rholinshead:
[editor] Text Output Rendering

# [editor] Text Output Rendering

Get text output rendering for the case where `output.data` is a string. For now, defaulting to raw JSON rendering for the following cases:
- `output.data` is `null`
- `output.data` is non-object or has no `kind` property (i.e. is not `OutputDataWithValue` type)
- `kind` is `tool_calls` --> not sure what the best rendering would be for that

In subsequent PRs, will add:
- JSON rendering toggle for raw rendering of output
- mimetype data rendering

## Testing:
- Render string data:
Screenshot 2023-12-28 at 3 31 39 PM

- Render raw JSON data (no 'kind'):
Screenshot 2023-12-28 at 3 51 37 PM

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/652).
* #656
* #654
* __->__ #652


rholinshead:
Going to land this stack to unblock & follow up on any review changes on top

PR #653, opened 2023-12-28
Remove extra `python` dir from the LLama Guard extension

rossdanlm:
Remove extra `python` dir from the LLama Guard extension

I also made the READMEs clearer and easier to understand. I'll publish this after


saqadri (approved):
Awesome

PR #655, opened 2023-12-28
Remove declaration of generic T from server.py

rossdanlm:
Remove declaration of generic T from server.py

Not being used, let's get rid of it dawg!

I also ran the linter based on Jonathan's command from this comment: https://github.com/lastmile-ai/aiconfig/pull/613#discussion_r1437724642

```
fd --glob '*.py'  python/src/aiconfig/editor/server | xargs python -m 'scripts.lint' --mode=fix --files
```

^ (btw you can install fd on Mac using `brew install fd`, looks pretty cool). It actually seems pretty cool https://github.com/sharkdp/fd?tab=readme-ov-file

### DIFF
--- a/python/src/aiconfig/editor/server/server.py
+++ b/python/src/aiconfig/editor/server/server.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Any, Type, TypeVar
+from typing import Any, Type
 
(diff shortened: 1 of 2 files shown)


jonathanlastmileai (approved):
cool thanks!

PR #658, opened 2023-12-28
[ts] Change `serialize` output from `Prompt | Prompt[]` to `Prompt[]`

rossdanlm:
[ts] Change `serialize` output from `Prompt | Prompt[]` to `Prompt[]`

TSIA, already did this for Python in https://github.com/lastmile-ai/aiconfig/pull/632, just doing for ts now too

# Test plan
Run the typescript demos to make sure that they continue to work. Run these commands in terminal from the `aiconfig` top-level dir
```bash
export OPENAI_API_KEY=sk-XYZ #set your API key, which you can get from https://platform.openai.com/api-keys
```
^Make sure delete any generated files so we don't include them in your GH push requests

Now enter into the `aiconfig/typescript` dir and run these commands to make sure tests pass and yarn compiles:
```bash
yarn test
yarn build
rm -rf dist/ #Run this everytime after you run `yarn build`
```

### DIFF
--- a/extensions/llama/typescript/llama.ts
+++ b/extensions/llama/typescript/llama.ts
@@ -105,7 +105,7 @@ export class LlamaModelParser extends ParameterizedModelParser<LlamaCompletionPa
     data: LlamaCompletionParams,
     aiConfig: AIConfigRuntime,
     params?: JSONObject | undefined
-  ): Prompt | Prompt[] {
+  ): Prompt[] {
     const startEvent = {
       name: "on_serialize_start",
       file: __filename,
@@ -126,7 +126,7 @@ export class LlamaModelParser extends ParameterizedModelParser<LlamaCompletionPa
       completionParams.model
     );
 
-    let prompts: Prompt[] = [];
+    const prompts: Prompt[] = [];
 
     if (conversationHistory && conversationHistory.length > 0) {
       let i = 0;
@@ -153,7 +153,7 @@ export class LlamaModelParser extends ParameterizedModelParser<LlamaCompletionPa
       }
     }
 
-    const prompt: Prompt = {
+    const newPrompt: Prompt = {
       name: promptName,
       input,
       metadata: {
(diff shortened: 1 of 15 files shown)


extensions/llama/typescript/llama.ts
```diff
@@ -163,7 +163,7 @@ export class LlamaModelParser extends ParameterizedModelParser<LlamaCompletionPa
       },
     };
 
-    let result: Prompt | Prompt[] = prompt;
+    let result: Prompt[] = [prompt];
```
rholinshead, line 166: This can just be replaced with `prompts.push(prompt)` now and code below updated to use `prompts` instead of `result`.

This code was just for handling conditional of Prompt or Prompt[]

rholinshead: Also `prompts` should be made const above since it's not reassigned

typescript/demo/demo.ts
```diff
@@ -104,14 +105,10 @@ async function createAIConfig() {
   };
 
   const aiConfig = AIConfigRuntime.create("demo", "this is a demo AIConfig");
-  const result = await aiConfig.serialize(model, data, "demoPrompt");
+  const prompts: Prompt[] = await aiConfig.serialize(model, data, "demoPrompt");
```
rholinshead, line 108: nit, is the explicit typing actually needed here/elsewhere? Should be inferred from the `serialize` response now

typescript/demo/demo.ts
```diff
@@ -104,14 +105,10 @@ async function createAIConfig() {
...
 
-  if (Array.isArray(result)) {
-    for (const prompt of result) {
-      aiConfig.addPrompt(prompt);
-    }
-  } else {
-    aiConfig.addPrompt(result);
+  for (const prompt of prompts) {
```
rholinshead, line 110: We should add this as `addPrompts` in the config API as well (in its own PR)

rholinshead (approved):
Nice, thanks.

Just a couple small things

typescript/demo/demo.ts
rossdanlm: I like the explicit type because it makes it easier for people to know immediately what type it is

typescript/demo/demo.ts
rossdanlm: Yea that's a good idea

PR #661, opened 2023-12-29
[AIC-py][Bug fix][editor] simplify and add missing await to run endpoint

jonathanlastmileai:
[AIC-py][Bug fix][editor] simplify and add missing await to run endpoint

Currently we're not awaiting `run()`. This awaits the run method call.
Also, I'm taking the opportunity to simplify the code by removing some abstractions.

Later if we simplify more and end up with repetition we can refactor again
to dedup.

Test:

```
url = "http://localhost:8080/api/run"

for data in [
    {},
    {    "prompt_name": "get_activities"},
    {
        "prompt_name": "get_activities",
        "params": {
            "city": "karachi"
        }
    }
]:

    print(f"\n\n{url=}, {data=}")

    print(f"\n{response=}"),

    import json
    response_json = json.loads(response.text)
    print(f"{response_json=}")
    print(f"{response_json['message']=}")
    aic = response_json.get('aiconfig', {})
    print("AIC:", aic)


url='http://localhost:8080/api/run', data={}

response=<Response [400]>
response_json={'message': "Failed to run prompt: <class 'KeyError'>, 'prompt_name'"}


url='http://localhost:8080/api/run', data={'prompt_name': 'get_activities'}


response=<Response [200]>
response_json={'aiconfig': {'$schema':...
response_json['message']='Ran prompt'

AIC: {'$schema': 'https://json.schem... '1. Visit  in . It is a unique and interactive...

url='http://localhost:8080/api/run', data={'prompt_name': 'get_activities', 'params': {'city': 'karachi'}}


response=<Response [200]>
response_json={'aiconfig': {'$schema':...
response_json['message']='Ran prompt'
"1. Clifton Beach: Locat...


```

### DIFF
--- a/python/src/aiconfig/editor/server/server.py
+++ b/python/src/aiconfig/editor/server/server.py
@@ -167,26 +167,26 @@ async def run() -> FlaskResponse:
     aiconfig = state.aiconfig
     request_json = request.get_json()
 
-    _op = make_op_run_method(MethodName("run"))
-
-    def _get_op_args():
-        prompt_name = request_json.get("prompt_name", None)
-        stream = request_json.get("stream", True)
-        LOGGER.info(f"Running prompt: {prompt_name}, {stream=}")
-        inference_options = InferenceOptions(stream=stream)
-        return Ok(
-            OpArgs(
-                {
-                    "prompt_name": prompt_name,
-                    #
-                    "options": inference_options,
-                }
-            )
-        )
-
-    op_args = _get_op_args()
-
-    return run_aiconfig_operation_with_op_args(aiconfig, "run", _op, op_args)
+    try:
+        prompt_name = request_json["prompt_name"]
+        params = request_json.get("params", {})
+        stream = request_json.get("stream", False)
+        options = InferenceOptions(stream=stream)
+        run_output = await aiconfig.run(prompt_name, params, options)  # type: ignore
+        LOGGER.debug(f"run_output: {run_output}")
+        return HttpResponseWithAIConfig(
+            #
+            message="Ran prompt",
+            code=200,
+            aiconfig=aiconfig,
+        ).to_flask_format()
+    except Exception as e:
+        return HttpResponseWithAIConfig(
+            #
+            message=f"Failed to run prompt: {type(e)}, {e}",
+            code=400,
+            aiconfig=None,
+        ).to_flask_format()
 
 
 @app.route("/api/add_prompt", methods=["POST"])


Ankush-lastmile:
testplan?

python/src/aiconfig/editor/travel.aiconfig.json
rholinshead, line 1: Please revert thees changes -- we want the output for testing and the new change is also using undefined param

rossdanlm: > the new change is also using undefined param

This is because if this is not included, we should default to what the AIConfig has stored already correct? Something like:

```python
prompt_name = request_json["prompt_name"]
prompt = aiconfig.get_prompt(prompt_name)
aiconfig_params = {}
if hasattr(prompt, "metadata"):
  if hasattr(prompt.metadata, "parameters"):
    aiconfig_params = prompt.metadata.parameters
aiconfig_params = aiconfig.get_prompt(prompt_name).metadata
params = request_json.get("params", aiconfig_params)
```

rholinshead: Although that is true, I was just stating that this changes the input from NYC to a param value that isn't defined (which will just be an error if run). The main thing is the outputs are removed, which would be kind of annoying to add it back to use in testing the editor

python/src/aiconfig/editor/server/server.py
```diff
@@ -167,26 +167,27 @@ async def run() -> FlaskResponse:
...
-
-    op_args = _get_op_args()
-
-    return run_aiconfig_operation_with_op_args(aiconfig, "run", _op, op_args)
+    try:
+        prompt_name = request_json["prompt_name"]
+        params = request_json.get("params", {})
+        stream = request_json.get("stream", False)
+        options = InferenceOptions(stream=stream)
+        kwargs = request_json.get("kwargs", {})
```
rholinshead, line 175: kwargs is really specific to python and not something any client should ever need to send up -- the endpoint request should always have a clear type of what's accepted instead of being open

jonathanlastmileai: whatever kwargs is for, I can't think of a better way to allow the frontend to control it. We would have to hard-code specific kwargs that can be accepted. Thoughts @Ankush-lastmile ?

rossdanlm: I'm cool with hard-coding. If it's from the client, they should be explicit and conform to our API anyways

jonathanlastmileai: If we're hard-coding, can we change the signature of `run()` to define these hard-coded things instead? That would fix a lot of quiet problems as a nice side effect

rholinshead: I think we can just remove it here.

 I think it can be removed from the config api itself, as well. Looking at the python code for config, abstract model_parser run method doesn't support kwargs and neither should individual parser implementations. It seems to be used for `run_with_dependencies` but I think that should be part of InferenceOptions instead

jonathanlastmileai: Agreed, run_with_dependencies should be in InferenceOptions. Although, funnily enough, InferenceOptions has its own kwargs so that doesn't really solve the problem. Anyway I'll remove it from the endpoint and we'll just have to hard-code or be satisfied with not supporting extra options for now.

rholinshead: https://github.com/lastmile-ai/aiconfig/issues/664

rossdanlm: Ok cool, added an issue to track this here: https://github.com/lastmile-ai/aiconfig/issues/663. It's a good bootcamp task, let's delete from here to get this PR unblocked and landed

rholinshead (approved):
LGTM

python/src/aiconfig/editor/server/server.py
rholinshead: I see we both made an issue for it. I don't know if it's as simple as just deleting the kwargs since we need to move existing run_with_dependencies off of the kwargs usage

rossdanlm: ^You're right, I closed my issue (https://github.com/lastmile-ai/aiconfig/issues/663) and commented on yours (https://github.com/lastmile-ai/aiconfig/issues/664) because yours was more descriptive, thanks for adding!

jonathanlastmileai: What do you mean move existing run_with_dependencies off kwargs usage? I *think* adding the argument and removing kwargs should work except at call sites that give a keyword arg other than run_with_dependencies.

rossdanlm:
Heads up, I'm going to land to get Ryan and me unblocked

PR #507, opened 2023-12-15
Refactor update_model() API to be more user friendly and flexible

rossdanlm:
Refactor update_model() API to be more user friendly and flexible

See The last part of https://github.com/lastmile-ai/aiconfig/issues/503

This greatly makes the `update_model()` API way more flexible. Instead of needing to set the entire ModelSettings (which can be annoying because you have to first retrieve the settings from the prompt to even set the modelMetadata to pass into your argument), you can now do each part individually:

1. Set both the model name and model settings: `update_model(name="my_name", settings="my_settings")` --> pass in prompt_name arg to update at prompt-level, do not pass to update at AI-Config level
2. Set the model name only: `update_model(name="my_name")` --> pass in prompt_name arg to update at prompt-level, do not pass to update at AI-Config level
3. Set model settings only: `update_model(settings="my_settings", prompt_name="my_prompt")` --> must pass in prompt_name otherwise this will error since AI-Config can have multiple models so we don't know which model to update

Imo, this is much cleaner and makes working with everything a lot easier, preserving the existing settings/name for model metadata if it already exists!

PR #662, opened 2023-12-29
[editor] Model Selector

rholinshead:
[editor] Model Selector

# [editor] Model Selector

Implement the model selector and add a placeholder call to the future api endpoint for updating the model.

https://github.com/lastmile-ai/aiconfig/assets/5060851/812f3a10-aaac-4ec3-9f1f-4d21ebc99e9f

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/662).
* #667
* #666
* #665
* __->__ #662

### DIFF
--- a/python/src/aiconfig/editor/client/src/Editor.tsx
+++ b/python/src/aiconfig/editor/client/src/Editor.tsx
@@ -2,7 +2,7 @@ import EditorContainer, {
   AIConfigCallbacks,
 } from "./components/EditorContainer";
 import { Flex, Loader, MantineProvider } from "@mantine/core";
-import { AIConfig, Prompt } from "aiconfig";
+import { AIConfig, ModelMetadata, Prompt } from "aiconfig";
 import { useCallback, useEffect, useMemo, useState } from "react";
 import { ufetch } from "ufetch";
 import { ROUTE_TABLE } from "./utils/api";
@@ -66,12 +66,23 @@ export default function Editor() {
     []
   );
 
+  const updateModel = useCallback(
+    async (_promptName?: string, _modelData?: string | ModelMetadata) => {
+      // return await ufetch.post(ROUTE_TABLE.UPDATE_MODEL,
+      //   prompt_name: promptName,
+      //   model_data: modelData,
+      // });
+    },
+    []
+  );
+
   const callbacks: AIConfigCallbacks = useMemo(
     () => ({
       addPrompt,
       getModels,
       runPrompt,
       save,
+      updateModel,
       updatePrompt,
     }),
     [save, getModels, addPrompt, runPrompt]
--- a/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
@@ -1,7 +1,7 @@
 import PromptContainer from "./prompt/PromptContainer";
 import { Container, Group, Button, createStyles, Stack } from "@mantine/core";
 import { showNotification } from "@mantine/notifications";
-import { AIConfig, Prompt, PromptInput } from "aiconfig";
+import { AIConfig, ModelMetadata, Prompt, PromptInput } from "aiconfig";
 import { useCallback, useMemo, useReducer, useRef, useState } from "react";
 import aiconfigReducer, { AIConfigReducerAction } from "./aiconfigReducer";
 import {
@@ -28,6 +28,10 @@ export type AIConfigCallbacks = {
   getModels: (search: string) => Promise<string[]>;
   runPrompt: (promptName: string) => Promise<void>;
   save: (aiconfig: AIConfig) => Promise<void>;
+  updateModel: (
+    promptName?: string,
+    modelData?: string | ModelMetadata
+  ) => Promise<void /*{ aiconfig: AIConfig }*/>;
   updatePrompt: (
     promptName: string,
     promptData: Prompt
@@ -150,6 +154,16 @@ export default function EditorContainer({
     [dispatch]
   );
 
+  const debouncedUpdateModel = useMemo(
+    () =>
+      debounce(
+        (promptName?: string, modelMetadata?: string | ModelMetadata) =>
+          callbacks.updateModel(promptName, modelMetadata),
+        250
+      ),
+    [callbacks.updateModel]
+  );
+
   const onUpdatePromptModelSettings = useCallback(
     async (promptIndex: number, newModelSettings: any) => {
       dispatch({
@@ -157,9 +171,44 @@ export default function EditorContainer({
         index: promptIndex,
         modelSettings: newModelSettings,
       });
-      // TODO: Call server-side endpoint to update model settings
+      // TODO: Call server-side endpoint to update model
     },
     [dispatch]
+  );
+
+  const onUpdatePromptModel = useCallback(
+    async (promptIndex: number, newModel?: string) => {
+      dispatch({
+        type: "UPDATE_PROMPT_MODEL",
+        index: promptIndex,
+        modelName: newModel,
+      });
+
+      try {
+        const prompt = clientPromptToAIConfigPrompt(
+          aiconfigState.prompts[promptIndex]
+        );
+        const currentModel = prompt.metadata?.model;
+        let modelData: string | ModelMetadata | undefined = newModel;
+        if (newModel && currentModel && typeof currentModel !== "string") {
+          modelData = {
+            ...currentModel,
+            name: newModel,
+          };
+        }
+
+        await debouncedUpdateModel(prompt.name, modelData);
+
+        // TODO: Consolidate
+      } catch (err: any) {
+        showNotification({
+          title: "Error updating prompt model",
+          message: err.message,
+          color: "red",
+        });
+      }
+    },
+    [dispatch, debouncedUpdateModel]
   );
 
   const onUpdatePromptParameters = useCallback(
@@ -241,8 +290,6 @@ export default function EditorContainer({
 
   const { classes } = useStyles();
 
-  // TODO: Implement editor context for callbacks, readonly state, etc.
-
   return (
     <>
       <Container maw="80rem">
@@ -262,9 +309,11 @@ export default function EditorContainer({
               <PromptContainer
                 index={i}
                 prompt={prompt}
+                getModels={callbacks.getModels}
                 onChangePromptInput={onChangePromptInput}
                 onChangePromptName={onChangePromptName}
                 onRunPrompt={onRunPrompt}
+                onUpdateModel={onUpdatePromptModel}
                 onUpdateModelSettings={onUpdatePromptModelSettings}
                 onUpdateParameters={onUpdatePromptParameters}
                 defaultConfigModelName={aiconfigState.metadata.default_model}
--- a/python/src/aiconfig/editor/client/src/components/aiconfigReducer.ts
+++ b/python/src/aiconfig/editor/client/src/components/aiconfigReducer.ts
@@ -10,6 +10,7 @@ export type MutateAIConfigAction =
   | AddPromptAction
   | UpdatePromptInputAction
   | UpdatePromptNameAction
+  | UpdatePromptModelAction
   | UpdatePromptModelSettingsAction
   | UpdatePromptParametersAction;
 
@@ -35,6 +36,12 @@ export type UpdatePromptNameAction = {
   type: "UPDATE_PROMPT_NAME";
   index: number;
   name: string;
+};
+
+export type UpdatePromptModelAction = {
+  type: "UPDATE_PROMPT_MODEL";
+  index: number;
+  modelName?: string;
 };
 
 export type UpdatePromptModelSettingsAction = {
@@ -132,6 +139,20 @@ export default function aiconfigReducer(
       return reduceReplacePrompt(state, action.index, (prompt) => ({
         ...prompt,
         name: action.name,
+      }));
+    }
+    case "UPDATE_PROMPT_MODEL": {
+      return reduceReplacePrompt(state, action.index, (prompt) => ({
+        ...prompt,
+        metadata: {
+          ...prompt.metadata,
+          model: action.modelName
+            ? {
+                name: action.modelName,
+                // TODO: Consolidate settings based on schema union
+              }
+            : undefined,
+        },
       }));
     }
     case "UPDATE_PROMPT_MODEL_SETTINGS": {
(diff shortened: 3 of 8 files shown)


python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -150,14 +154,59 @@ export default function EditorContainer({
...
+
   const onUpdatePromptModelSettings = useCallback(
     async (promptIndex: number, newModelSettings: any) => {
       dispatch({
         type: "UPDATE_PROMPT_MODEL_SETTINGS",
         index: promptIndex,
         modelSettings: newModelSettings,
       });
-      // TODO: Call server-side endpoint to update model settings
+      // TODO: Call server-side endpoint to update model
```
rholinshead, line 174: Not setting this up just yet (unlike model name below) since there is a higher potential for bugs in managing the settings object & don't want to add friction to whomever is adding the server-side endpoint

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -28,6 +28,10 @@ export type AIConfigCallbacks = {
   getModels: (search: string) => Promise<string[]>;
   runPrompt: (promptName: string) => Promise<void>;
   save: (aiconfig: AIConfig) => Promise<void>;
+  updateModel: (
+    promptName?: string,
+    modelData?: string | ModelMetadata
+  ) => Promise<void /*{ aiconfig: AIConfig }*/>;
```
rossdanlm, line 34: Why not return aiconfig?

python/src/aiconfig/editor/client/src/Editor.tsx
```diff
@@ -66,12 +66,23 @@ export default function Editor() {
     []
   );
 
+  const updateModel = useCallback(
+    async (_promptName?: string, _modelData?: string | ModelMetadata) => {
+      // return await ufetch.post(ROUTE_TABLE.UPDATE_MODEL,
+      //   prompt_name: promptName,
+      //   model_data: modelData,
+      // });
```
rossdanlm, line 74: Are you commenting this out to add it later once the `update_model()` api is built on backend?

rossdanlm:
Btw not sure if you're aware, but I refactored the `update_model()` python SDK to take in model_name and model_settings as two separate function args: https://github.com/lastmile-ai/aiconfig/blob/ca59d2c1657fe6e2655f1a0c25180f9407b97cab/python/src/aiconfig/schema.py#L492C34-L492C34

PR for full details: https://github.com/lastmile-ai/aiconfig/pull/507

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -150,14 +154,59 @@ export default function EditorContainer({
     [dispatch]
   );
 
+  const debouncedUpdateModel = useMemo(
+    () =>
+      debounce(
+        (promptName?: string, modelMetadata?: string | ModelMetadata) =>
+          callbacks.updateModel(promptName, modelMetadata),
+        250
```
rossdanlm, line 162: Should we define this as a const somewhere? I see it's also 250 for `addPrompt`

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -150,14 +154,59 @@ export default function EditorContainer({
...
+      ),
+    [callbacks.updatePrompt]
```
rossdanlm, line 164: Should this be `.updateModel`?

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -150,14 +154,59 @@ export default function EditorContainer({
...
+        modelName: newModel,
+      });
+
+      try {
+        const prompt = clientPromptToAIConfigPrompt(
+          aiconfigState.prompts[promptIndex]
+        );
+        const currentModel = prompt.metadata?.model;
+        let modelData: string | ModelMetadata | undefined = newModel;
+        if (newModel && currentModel && typeof currentModel !== "string") {
```
rossdanlm, line 193: I think with the new SDK you won't have to do individual checks because we're passing `model_name` and `settings` as two separate args, hopefully this makes it easier for you

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -150,14 +154,59 @@ export default function EditorContainer({
...
+            name: newModel,
+          };
+        }
+
+        await debouncedUpdateModel(prompt.name, modelData);
+
+        // TODO: Consolidate
+      } catch (err: any) {
+        showNotification({
+          title: "Error updating prompt model",
```
rossdanlm, line 205: Would it be possible to read the http message that we return and we display it? Can be done future PR

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -150,14 +154,59 @@ export default function EditorContainer({
...
+          message: err.message,
```
rossdanlm, line 206: Oh I guess it's done here?

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -262,9 +309,11 @@ export default function EditorContainer({
               <PromptContainer
                 index={i}
                 prompt={prompt}
+                getModels={callbacks.getModels}
                 onChangePromptInput={onChangePromptInput}
                 onChangePromptName={onChangePromptName}
                 onRunPrompt={onRunPrompt}
+                onUpdateModel={onUpdatePromptModel}
                 onUpdateModelSettings={onUpdatePromptModelSettings}
```
rossdanlm, line 317: Oh ok cool, so we are keeping them as separate actions?

python/src/aiconfig/editor/client/src/components/aiconfigReducer.ts
```diff
@@ -134,6 +141,20 @@ export default function aiconfigReducer(
...
+      return reduceReplacePrompt(state, action.index, (prompt) => ({
+        ...prompt,
+        metadata: {
+          ...prompt.metadata,
+          model: action.modelName
+            ? {
+                name: action.modelName,
+                // TODO: Consolidate settings based on schema union
+              }
+            : undefined,
```
rossdanlm, line 154: Hopefully don't need to do this logic with updated SDK

python/src/aiconfig/editor/client/src/components/prompt/AddPromptButton.tsx
```diff
@@ -1,7 +1,7 @@
 import { ActionIcon, Menu, TextInput } from "@mantine/core";
 import { IconPlus, IconSearch, IconTextCaption } from "@tabler/icons-react";
-import { memo, useCallback, useEffect, useState } from "react";
-import { showNotification } from "@mantine/notifications";
+import { memo, useCallback, useState } from "react";
+import useLoadModels from "../../hooks/useLoadModels";
```
rossdanlm, line 4: Nice, much cleaner

python/src/aiconfig/editor/client/src/components/prompt/AddPromptButton.tsx
```diff
@@ -41,30 +41,14 @@ function ModelMenuItems({
 
 export default memo(function AddPromptButton({ addPrompt, getModels }: Props) {
   const [modelSearch, setModelSearch] = useState("");
```
rossdanlm, line 43: Sorry what exactly is `modelSearch` string supposed to represent?

python/src/aiconfig/editor/client/src/components/prompt/ModelSelector.tsx
```diff
@@ -0,0 +1,82 @@
...
+      }
+      filter={(searchValue: string, item: AutocompleteItem) => {
+        if (showAll) {
+          return true;
+        }
+
+        const modelName: string = item.value;
+        return modelName
+          .toLocaleLowerCase()
+          .includes(searchValue.toLocaleLowerCase().trim());
```
rossdanlm, line 65: Alright so what's the difference between this vs. the functionality in `getModels` callback: https://github.com/lastmile-ai/aiconfig/blob/9ac3b2e30992476e039f69a4c4562c4e51a22b0a/python/src/aiconfig/editor/client/src/Editor.tsx#L31-L40

python/src/aiconfig/editor/client/src/hooks/useLoadModels.ts
```diff
@@ -0,0 +1,28 @@
...
+        title: "Error loading models",
+        message: err?.message,
+        color: "red",
+      });
+    }
+  }, []);
+
+  useEffect(() => {
+    loadModels(modelSearch);
+  }, [loadModels, modelSearch]);
```
rossdanlm, line 25: How come we need to pass in `modelSearch` as a dependency here but not `getModels` as a dependency on L21? Is it because we're implicitly assuming that `getModels` should only be defined once and never change?

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
rholinshead: Ah, ok - for the updated config it seems like we can't support the following scenario:
Change model name for a prompt and don't set any settings (i.e. set model to the string name instead of `{name: <name>, settings: {}}`). This is also another case where we have multiple ways of representing the same thing in aiconfig, and might be annoying to set the model name and clear the settings but always have `settings: {}` set by the editor instead of just having `model: name` in the config?

On top of that, if we change the model and it did already have settings set, we would ideally want to maintain only the settings that are supported by the new model (e.g. say temperature, topK are used by both models) as defined in the prompt schema that is associated with the model provider, and drop anything not supported. This might need to wait until post-MVP, though. For now we can either clear the settings or keep them entirely... I was doing the former here, but open to opinions

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
rholinshead: Ya, we should keep them separate since there's different logic for them:
- update settings --> keep same model name but new model settings object
- update model --> change model name and either keep settings or clear them

python/src/aiconfig/editor/client/src/components/aiconfigReducer.ts
```diff
@@ -37,6 +38,12 @@ export type UpdatePromptNameAction = {
   name: string;
 };
 
+export type UpdatePromptModelAction = {
```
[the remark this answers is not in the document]
rholinshead: When we dispatch the "UPDATE_PROMPT_MODEL" action, that action is implicitly typed as this `UpdatePromptModelAction` type. We add it to `MutateAIConfigAction` above and that does two things:
1. Makes it part of the `AIConfigReducerAction` via the union for that type, which means our reducer below will require a `case "UPDATE_PROMPT_MODEL"` to handle it to ensure the state is update -- this is the automatic mapping you're referring to
2. Allows us to dispatch `CONSOLIDATE_AICONFIG` after we get the response, in order to update the client state with response state (not done here because we don't have an aiconfig response to do that with just yet)

python/src/aiconfig/editor/client/src/components/prompt/ModelSelector.tsx
```diff
@@ -0,0 +1,82 @@
...
+      }}
+      data={models}
+      value={autocompleteSearch}
+      onChange={(value: string) => {
+        setAutocompleteSearch(value);
+        setShowAll(false);
+        onSelectModel(value);
+        models.some((model) => {
+          if (model === value) {
+            setShowAll(true);
```
rossdanlm, line 75: Ok just to clarify, this `onChange` only occurs when a value is selected from the autocomplete options? Does this also occur when user changes the search string?

python/src/aiconfig/editor/client/src/components/aiconfigReducer.ts
rholinshead: I'll revisit this once the endpoint is implemented. We still need some UI state change here (this happens before the request is sent and before the response is returned to ensure responsive UI). Probably what will need to be done is:
- this action sets just the name
- aiconfig response comes back with full metadata, either cleared settings or updated settings to match supported settings based on the schema
- from aiconfig response, we consolidate the UI state model metadata with updated settings

python/src/aiconfig/editor/client/src/components/prompt/AddPromptButton.tsx
rholinshead: There's a search input and `modelSearch` is the text of that input

rossdanlm (approved):
This generally looks good to me. Biggest point is my comment about the SDK being refactored to take in name + settings separately, but we can fix forward instead of worrying about merge conflict. Also have question about the autocompletion filtering.

This'll also be easier for you to test once I add this endpoint on the backend for you

python/src/aiconfig/editor/client/src/components/prompt/ModelSelector.tsx
rholinshead: This is specific to this AutoComplete UI component because needs to handle the search/filtering of the provided data on its own

python/src/aiconfig/editor/client/src/components/prompt/ModelSelector.tsx
rholinshead: This is mainly copied from our model selector in Workbooks and is sort of a workaround to match our desired behaviour. By default, the component will only ever show results in the dropdown that match the text in the input box. So, if the aiconfig has model set to gpt-4 when you load it, the dropdown would only ever show models with gpt-4 substring.
This isn't what we want, though. The behaviour we want is:
- if the user is typing to 'search' for a model, then the dropdown should filter on that search substring
- otherwise, if the model is selected from the list / defaulted from the config then we want to show all the model options to change to

python/src/aiconfig/editor/client/src/hooks/useLoadModels.ts
rholinshead: Ah, good catch. `getModels` should be added above. Something must be up with my linter because it's missing these deps :(

python/src/aiconfig/editor/client/src/components/aiconfigReducer.ts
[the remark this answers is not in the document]
rossdanlm: Nice, thanks for the explanation!

python/src/aiconfig/editor/client/src/components/prompt/ModelSelector.tsx
rossdanlm: Oh, I generally assumed that the user would just click the x button on the right to see everything, hmmmm. I don't think it's a blocker but feels a bit strange to me that auto-completions don't remain filtered after selected. We can do this as P2 someday

PR #665, opened 2023-12-29
[editor] Support Deleting Prompt

rholinshead:
[editor] Support Deleting Prompt

# [editor] Support Deleting Prompt

Add the UI side for deleting a prompt, with placeholder to call the server endpoint once it's implemented

https://github.com/lastmile-ai/aiconfig/assets/5060851/b3e682dd-a3ed-45c1-b153-b4b6e2094616

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/665).
* #667
* #666
* __->__ #665
* #662

PR #666, opened 2023-12-29
[editor] Use Prompt ID Instead of Index for Client-side State Actions

rholinshead:
[editor] Use Prompt ID Instead of Index for Client-side State Actions

# [editor] Use Prompt ID Instead of Index for Client-side State Actions

We can uniquely identify prompts in client-side state using their client-side id, so let's use that instead of index for referencing them in the actions. This should be a bit more robust than using index.

## Testing:
Went through all relevant actions and made sure they work as expected

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/666).
* #667
* __->__ #666
* #665
* #662

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
@@ -96,7 +96,7 @@ export default function EditorContainer({
   const onSave = useCallback(async () => {
     setIsSaving(true);
     try {
-      await callbacks.save(clientConfigToAIConfig(aiconfigState));
+      await callbacks.save(clientConfigToAIConfig(stateRef.current));
     } catch (err: any) {
       showNotification({
         title: "Error saving",
@@ -106,7 +106,7 @@ export default function EditorContainer({
     } finally {
       setIsSaving(false);
     }
-  }, [aiconfigState, callbacks.save]);
+  }, [callbacks.save]);
 
   const debouncedUpdatePrompt = useMemo(
     () =>
@@ -119,10 +119,10 @@ export default function EditorContainer({
   );
 
   const onChangePromptInput = useCallback(
-    async (promptIndex: number, newPromptInput: PromptInput) => {
+    async (promptId: string, newPromptInput: PromptInput) => {
       const action: AIConfigReducerAction = {
         type: "UPDATE_PROMPT_INPUT",
-        index: promptIndex,
+        id: promptId,
         input: newPromptInput,
       };
 
@@ -130,7 +130,7 @@ export default function EditorContainer({
 
       try {
         const prompt = clientPromptToAIConfigPrompt(
-          aiconfigState.prompts[promptIndex]
+          getPrompt(stateRef.current, promptId)!
         );
         const serverConfigRes = await debouncedUpdatePrompt(prompt.name, {
           ...prompt,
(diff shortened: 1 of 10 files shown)


python/src/aiconfig/editor/client/src/components/aiconfigReducer.ts
```diff
@@ -108,22 +107,32 @@ function reduceConsolidateAIConfig(
...
+      },
+    } as ClientPrompt;
+  };
+
   switch (action.type) {
-    case "ADD_PROMPT_AT_INDEX":
+    case "ADD_PROMPT_AT_INDEX": {
+      return reduceReplacePrompt(
+        state,
+        action.prompt._ui.id,
```
rholinshead, line 130: Note that this action doesn't have an index explicitly passed in on dispatch, so need to grab it from the newly-inserted prompt in the action

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -96,7 +96,7 @@ export default function EditorContainer({
   const onSave = useCallback(async () => {
     setIsSaving(true);
     try {
-      await callbacks.save(clientConfigToAIConfig(aiconfigState));
+      await callbacks.save(clientConfigToAIConfig(stateRef.current));
```
rossdanlm, line 99: Why wouldn't we want to re-render if client aiconfig changes?

python/src/aiconfig/editor/client/src/components/aiconfigReducer.ts
```diff
@@ -108,22 +107,32 @@ function reduceConsolidateAIConfig(
...
 ): ClientAIConfig {
+  // Make sure prompt structure is properly updated. Client input and metadata takes precedence
+  // since it may have been updated by the user while the request was in flight
+  const consolidatePrompt = (statePrompt: ClientPrompt) => {
+    const responsePrompt = responseConfig.prompts.find(
+      (resPrompt) => resPrompt.name === statePrompt.name
+    );
+    return {
+      ...responsePrompt,
+      ...statePrompt,
```
rossdanlm, line 118: Just to confirm, this means we use the fields from `responsePrompt` first and whatever is left over that hasn't been defined by `responsePrompt` and exists within `statePrompts, that's what they're going to use?

rossdanlm (approved):
nice!

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
rholinshead: We will always re-render if aiconfigState changes. This stateRef is a stable ref to use in the callbacks since otherwise using aiconfigState in the callback means the callback logic is actually using the state from when the callback was constructed & memoized (which could technically be stale)

python/src/aiconfig/editor/client/src/components/aiconfigReducer.ts
rholinshead: Yes

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
rossdanlm: Ah ok this makes sense!

rholinshead: I don't necessarily know if it had to be update here, but it was a small enough change and related to using proper source of truth (like id instead of index)

PR #668, opened 2023-12-29
[python] Created `get_parameters` API

rossdanlm:
[python] Created `get_parameters` API

**Disclaimer:** I know this is 400+ lines changed, but ~300 of those are just adding/refactoring automated tests!!!

We already had `get_prompt_parameters`, but this isn't good enough if parameters aren't defined locally and we want to bubble up to the aiconfig defaults. See comment in https://github.com/lastmile-ai/aiconfig/pull/661#discussion_r1437941729.

This is important so that we can ensure we have valid params if not defined in the correct section of our AIConfig schema. This is similar in theme to https://github.com/lastmile-ai/aiconfig/pull/600

The only callsite we have today for this is in python in the `params.py` function, so I updated that callsite too. I don't know why it doesn't exist in Typescript (cc @rholinshead) but I also added this functionality in typescript in https://github.com/lastmile-ai/aiconfig/pull/669 (I'm not as familiar with automated testing there so will ask Ryan for a code pointer to help get me started)

## Test plan
Added a bunch of automated tests which you can run by doing going to the `aiconfig/python` dir and running `pytest`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/668).
* #693
* #692
* #691
* #690
* #688
* #670
* __->__ #668

### DIFF
--- a/python/src/aiconfig/schema.py
+++ b/python/src/aiconfig/schema.py
@@ -335,6 +335,85 @@ class AIConfig(BaseModel):
         else:
             return self.metadata
 
+
+    def get_parameters(
+        self,
+        prompt_or_prompt_name: Optional[str | Prompt] = None,
+    ) -> JSONObject:
+        """
+        Get the parameters for a prompt, using the global parameters if 
+        needed.
+        
+        Args:
+            prompt_or_prompt_name Optional[str | Prompt]: The name of the 
+                prompt or the prompt object. If not specified, use the 
+                global parameters.
+        """
+        prompt = prompt_or_prompt_name
+        if isinstance(prompt_or_prompt_name, str):
+            if prompt_or_prompt_name not in self.prompt_index:
+                raise IndexError(f"Prompt '{prompt_or_prompt_name}' not found in config, available prompts are:\n {list(self.prompt_index.keys())}")
+            prompt = self.prompt_index[prompt_or_prompt_name]
+
+        assert prompt is None or isinstance(prompt, Prompt)
+        if prompt is None or not prompt.metadata or not prompt.metadata.parameters:
+            return self.get_global_parameters()
+
+        return self.get_prompt_parameters(prompt)
+
+    # pylint: disable=W0102
+    def get_global_parameters(
+        self,
+        default_return_value: JSONObject = {},
+    ) -> JSONObject:
+        """
+        Get the global parameters for the AIConfig. If they're not defined,
+        return a default value ({} unless overridden)
+        
+        Args:
+            default_return_value JSONObject - Default value to return if 
+                global parameters are not defined.
+        """
+        return self._get_global_parameters_exact() or default_return_value
+    # pylint: enable=W0102
+
+    def _get_global_parameters_exact(self) -> JSONObject | None:
+        """
+        Get the global parameters for the AIConfig. This should be the
+        the explicit value (ie: if parameters is None, return None, not {})
+        """
+        return self.metadata.parameters
+
+    # pylint: disable=W0102
+    def get_prompt_parameters(
+        self,
+        prompt: Prompt,
+        default_return_value: JSONObject = {},
+    ) -> JSONObject:
+        """
+        Get the prompt's local parameters. If they're not defined,
+        return a default value ({} unless overridden)
+        
+        Args:
+            default_return_value JSONObject - Default value to return if 
+                prompt parameters are not defined.
+        """
+        return self._get_prompt_parameters_exact(prompt) \
+            or default_return_value
+    # pylint: enable=W0102
+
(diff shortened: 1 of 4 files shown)


python/src/aiconfig/schema.py
```diff
@@ -335,6 +335,85 @@ def get_metadata(self, prompt_name: Optional[str] = None):
...
+        """
+        prompt = prompt_or_prompt_name
+        if isinstance(prompt_or_prompt_name, str):
+            if prompt_or_prompt_name not in self.prompt_index:
+                raise IndexError(f"Prompt '{prompt_or_prompt_name}' not found in config, available prompts are:\n {list(self.prompt_index.keys())}")
+            prompt = self.prompt_index[prompt_or_prompt_name]
+
+        assert prompt is None or isinstance(prompt, Prompt)
+        if prompt is None or not prompt.metadata or not prompt.metadata.parameters:
+            return self.get_global_parameters()
```
rholinshead, line 360: It's confusing to return the global parameters when the requested prompt doesn't have any parameters, but otherwise returning only the prompt parameters if they exist. 

In the parameter resolution we actually use a union of global parameters & local parameters (overriding global if they match). Shouldn't this do the same?

rholinshead:
> We already had get_prompt_parameters, but this isn't good enough if parameters aren't defined locally and we want to bubble up to the aiconfig defaults. See comment in https://github.com/lastmile-ai/aiconfig/pull/661#discussion_r1437941729.

More generally, is this change actually needed? The bubbling up already happens when running the prompt, right? So the editor implementation doesn't need to explicitly get the local + global params because it will be run with the params set

rossdanlm:
> More generally, is this change actually needed? The bubbling up already happens when running the prompt, right? So the editor implementation doesn't need to explicitly get the local + global params because it will be run with the params set

You don't need to pass in params anymore from the client, this abstracts it away. You still can pass it in explicitly into the `run()` command as a method arg if you want, but this simplifies it a lot

python/src/aiconfig/schema.py
rossdanlm: > It's confusing to return the global parameters when the requested prompt doesn't have any parameters, but otherwise returning only the prompt parameters if they exist.

If the prompt parameters do not exist, it makes sense to check if any are defined in the global params. It's the same concept as https://github.com/lastmile-ai/aiconfig/pull/600

> In the parameter resolution we actually use a union of global parameters & local parameters (overriding global if they match)

Tbh I feel like "check the value closest in scope (prompt) and then check the next closest value in scope (aiconfig)" is more intuitive because it's something a real human would do, but yea it doesn't matter, I can also implement it this way

python/src/aiconfig/schema.py
```diff
@@ -335,6 +335,85 @@ def get_metadata(self, prompt_name: Optional[str] = None):
...
+        
+        Args:
+            default_return_value JSONObject - Default value to return if 
+                prompt parameters are not defined.
+        """
+        return self._get_prompt_parameters_exact(prompt) \
+            or default_return_value
+    # pylint: enable=W0102
+
+    def _get_prompt_parameters_exact(
```
jonathanlastmileai, line 405: Just FYI, {} is falsey in expressions like ` self._get_global_parameters_exact() or default_return_value`, just like None is.

python/src/aiconfig/schema.py
jonathanlastmileai: Rename to resolve_prameters or something?

rholinshead: Maybe `get_available_parameters` or similar? That avoids the potential for confusion about *which* parameters this refers to; `resolve_parameters` to me means to resolve the parameters that are being used within the prompt input.

And I think this should do the same logic as in `resolve_prompt_string` (less reference/input params) which:
- gets config-level params first
- gets current prompt params, overwriting matching config-level param names with prompt-level value

Otherwise it's pretty confusing to have different behaviours and removes most of the benefits you've listed about reusability / testability.

jonathanlastmileai:
ac to unblock

jonathanlastmileai (approved):

python/src/aiconfig/schema.py
rossdanlm: > And I think this should do the same logic as in resolve_prompt_string
> gets config-level params first
> gets current prompt params, overwriting matching config-level param names with prompt-level value

Yea sure, I can do that as well as change the name to `get_available_parameters`! I will fix forward to avoid merge conflicts: https://github.com/lastmile-ai/aiconfig/issues/695

PR #670, opened 2023-12-29
[python] Updated `set_parameter()` to work if parameters haven't been defined before

rossdanlm:
[python] Updated `set_parameter()` to work if parameters haven't been defined before

Before this function would have failed to set any parameters if:

1. The prompt_name was passed in but `prompt.metadata` (possible) or `prompt.metadata.parameters` (not possible because it defaults to `{}`) was `None`
2. The prompt_name was not passed in (so update aiconfig) but the `aiconfig.metadata.parameters` was `None` (this is actually never possible because we default to using `{}`, but still just good to be thorough and future-proof)

See example below
Screenshot 2023-12-29 at 01 04 48

## Test plan
Added automated tests: `pytest -vv`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/670).
* #693
* #692
* #691
* #690
* #688
* __->__ #670
* #668

### DIFF
--- a/python/src/aiconfig/schema.py
+++ b/python/src/aiconfig/schema.py
@@ -335,6 +335,85 @@ class AIConfig(BaseModel):
         else:
             return self.metadata
 
+
+    def get_parameters(
+        self,
+        prompt_or_prompt_name: Optional[str | Prompt] = None,
+    ) -> JSONObject:
+        """
+        Get the parameters for a prompt, using the global parameters if 
+        needed.
+        
+        Args:
+            prompt_or_prompt_name Optional[str | Prompt]: The name of the 
+                prompt or the prompt object. If not specified, use the 
+                global parameters.
+        """
+        prompt = prompt_or_prompt_name
+        if isinstance(prompt_or_prompt_name, str):
+            if prompt_or_prompt_name not in self.prompt_index:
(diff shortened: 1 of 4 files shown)


python/src/aiconfig/schema.py
```diff
@@ -414,25 +414,57 @@ def _get_prompt_parameters_exact(
...
+                        },
+                    }
+                Using the parameter in a prompt with handlebars syntax would 
+                look like this:
+                    "{{person.firstname}} {{person.lastname}}"
+            prompt_name (str, optional): The name of the prompt to add the 
+                parameter to. Defaults to None.
         """
         target_metadata = self.get_metadata(prompt_name)
+        if not target_metadata:
```
jonathanlastmileai, line 446: It's hard to reason about code like this -- not because of the code -- but because so many things are optional, and the structure is so nested and general. If you say this check is unnecessary, can we leave it out and make sure the schema doesn't break? 

We should have the discipline to make future versions of the schema more constrained and not less, which will make a lot of this checking and pulling apart/stitching together code no longer needed.

jonathanlastmileai (approved):
ac to unblock

PR #673, opened 2023-12-29
[1/n] Global parameters renderer

saqadri:
[1/n] Global parameters renderer

Use ParametersRenderer to modify global parameters.

Currently still updating the entire parameters object. Next up, will fix the ParametersRenderer component to update each parameter separately.

PR #674, opened 2023-12-29
[easy] Simplify update parameters callback

saqadri:
[easy] Simplify update parameters callback

Have updateParameters take a JSONObject. Makes it much cleaner to use elsewhere.

ParametersRenderer simply converts its ParametersArray to JSONObject before triggering the update.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/674).
* __->__ #674
* #673

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/GlobalParametersContainer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/GlobalParametersContainer.tsx
@@ -1,7 +1,7 @@
 import { createStyles, Container, Accordion, Text } from "@mantine/core";
 import { JSONObject } from "aiconfig";
-import { memo, useCallback, useState } from "react";
-import ParametersRenderer, { ParametersArray } from "./ParametersRenderer";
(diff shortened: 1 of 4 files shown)


python/src/aiconfig/editor/server/static/asset-manifest.json
rholinshead, line 1: Are all the changes here / below from `yarn build` / are they actually needed?

rholinshead (approved):
Code LGTM, but question about all the extra (build) files? Accepting to unblock

saqadri:
I think I ran yarn build, and forgot to remove the files. Does it matter if these are updated?

rholinshead:
> I think I ran yarn build, and forgot to remove the files. Does it matter if these are updated?

I don't *think* so. They probably should be removed/excluded from git actually, but I don't think they'll cause issues?

PR #677, opened 2023-12-29
[easy] Add production build folder to gitignore

saqadri:
[easy] Add production build folder to gitignore

TSIA

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -87,6 +87,9 @@ web_modules/
 .next
 out
 
+# Production build folder
+build
+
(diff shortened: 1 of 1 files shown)


rholinshead (approved):
Thanks -- just confirming this will pick up any build/ folder under the repo (don't need to specify exact path to the editor build folder)

saqadri:
I tested it with the build folder under `python/src/aiconfig/editor/client/build`, and it picked that up (i.e. making any changes to it didn't registered on source control)

PR #678, opened 2023-12-29
[editor] Initialize Empty Config with a First Prompt

rholinshead:
[editor] Initialize Empty Config with a First Prompt

# [editor] Initialize Empty Config with a First Prompt

Currently, starting the editor without specifying a file path will result in an empty editor UI without any prompts.

This PR updates it to start the config with a prompt whose model is the first supported model in the registry.

Screenshot 2023-12-29 at 4 59 03 PM

### DIFF
--- a/python/src/aiconfig/editor/server/server_utils.py
+++ b/python/src/aiconfig/editor/server/server_utils.py
@@ -8,6 +8,8 @@ from dataclasses import dataclass
 from enum import Enum
 from types import ModuleType
 from typing import Any, Callable, NewType, Optional, Type, TypeVar, cast
+from aiconfig.registry import ModelParserRegistry
+from aiconfig.schema import Prompt, PromptMetadata
 
 import lastmile_utils.lib.core.api as core_utils
 import result
@@ -217,6 +219,10 @@ def init_server_state(app: Flask, edit_config: EditServerConfig) -> Result[None,
                 return Err(f"Failed to load AIConfig from {edit_config.aiconfig_path}: {e}")
     else:
         aiconfig_runtime = AIConfigRuntime.create()  # type: ignore
+        model_ids = ModelParserRegistry.parser_ids()
+        if len(model_ids) > 0:
+            aiconfig_runtime.add_prompt("prompt_1", Prompt(name="prompt_1", input="", metadata=PromptMetadata(model=model_ids[0])))
+
         state.aiconfig = aiconfig_runtime
         LOGGER.info("Created new AIConfig")
         return Ok(None)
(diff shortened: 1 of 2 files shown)


python/src/aiconfig/registry.py
```diff
@@ -93,7 +94,7 @@ def clear_registry():
         ModelParserRegistry._parsers.clear()
 
     @staticmethod
-    def parser_ids() -> list:
+    def parser_ids() -> list[str]:
```
rholinshead, line 97: This was typed as list unknown before, so update to list to prevent error where I need to use it above

python/src/aiconfig/editor/server/server_utils.py
```diff
@@ -217,6 +219,9 @@ def init_server_state(app: Flask, edit_config: EditServerConfig) -> Result[None,
                 return Err(f"Failed to load AIConfig from {edit_config.aiconfig_path}: {e}")
     else:
         aiconfig_runtime = AIConfigRuntime.create()  # type: ignore
+        first_supported_model: str = ModelParserRegistry.parser_ids()[0]
+        aiconfig_runtime.add_prompt("prompt_1", Prompt(name="prompt_1", input="", metadata=PromptMetadata(model=first_supported_model)))
+
```
rholinshead, line 224: This file, and parser_ids return type are the only real changes here, the rest are from black auto-formatting

python/src/aiconfig/editor/server/server_utils.py
```diff
@@ -217,6 +219,9 @@ def init_server_state(app: Flask, edit_config: EditServerConfig) -> Result[None,
                 return Err(f"Failed to load AIConfig from {edit_config.aiconfig_path}: {e}")
     else:
         aiconfig_runtime = AIConfigRuntime.create()  # type: ignore
+        first_supported_model: str = ModelParserRegistry.parser_ids()[0]
```
saqadri, line 222: Is this guaranteed to have at least one element in it?

python/src/aiconfig/editor/server/server_utils.py
```diff
@@ -217,6 +219,9 @@ def init_server_state(app: Flask, edit_config: EditServerConfig) -> Result[None,
...
+        aiconfig_runtime.add_prompt("prompt_1", Prompt(name="prompt_1", input="", metadata=PromptMetadata(model=first_supported_model)))
```
saqadri, line 223: nit: Not for this diff and something just for UI treatment, but if textbox for prompt input is empty should we have a placeholder text in there?

saqadri (approved):

python/src/aiconfig/editor/server/server_utils.py
rholinshead: I'm not sure -- do we support removing all parsers from the registry? By default they'd have all default models so someone would need to explicitly empty out the registry to hit this, right?

rholinshead: Added a check to be safe

PR #680, opened 2023-12-30
[editor] Config Name & Description UI

rholinshead:
[editor] Config Name & Description UI

[editor] Config Name & Description UI

# [editor] Config Name & Description UI

Add the UI for changing config name / description, similar to workbook title UI:

https://github.com/lastmile-ai/aiconfig/assets/5060851/f0f49638-d0c5-4d88-8edd-c0bbb0294f7c

If config name is not set, keep the name / description in Edit state until the name is set

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/680).
* #686
* #682
* #681
* __->__ #680

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/ConfigNameDescription.tsx
+++ b/python/src/aiconfig/editor/client/src/components/ConfigNameDescription.tsx
@@ -1,0 +1,137 @@
+import {
+  createStyles,
+  Stack,
+  Text,
+  Textarea,
+  TextInput,
+  Title,
+} from "@mantine/core";
+import { useClickOutside } from "@mantine/hooks";
+import { memo, useCallback, useRef, useState } from "react";
+
+type Props = {
+  name?: string;
+  description?: string | null;
+  setDescription: (description: string) => void;
+  setName: (name: string) => void;
+};
+
+const useStyles = createStyles((theme) => ({
+  // Match name input style to corresponding Title element styles
+  // See https://github.com/mantinedev/mantine/blob/master/src/mantine-core/src/Title/Title.styles.ts
+  nameInput: {
+    ...theme.fn.fontStyles(),
+    fontFamily: theme.headings.fontFamily,
+    fontWeight: theme.headings.fontWeight as number,
+    fontSize: theme.headings.sizes.h1.fontSize,
+    lineHeight: theme.headings.sizes.h1.lineHeight,
+    width: "-webkit-fill-available",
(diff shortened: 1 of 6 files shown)


python/src/aiconfig/editor/client/src/components/aiconfigReducer.ts
```diff
@@ -192,6 +199,16 @@ export default function aiconfigReducer(
         },
       }));
     }
+    case "SET_NAME_DESCRIPTION": {
+      return {
+        ...state,
+        name: action.name ?? state.name,
+        description:
+          action.description === null
```
rholinshead, line 207: Using 'null' value to represent 'remove description'

python/src/aiconfig/editor/client/src/components/ConfigNameDescription.tsx
```diff
@@ -0,0 +1,147 @@
...
+        theme.colorScheme === "dark"
+          ? "rgba(255, 255, 255, 0.1)"
+          : theme.colors.gray[1],
+    },
+    borderRadius: theme.radius.sm,
+    width: "-webkit-fill-available",
+  },
+}));
+
+export default memo(function ConfigTitleDescription({
```
saqadri, line 46: super nit since this is the name used everywhere. Not picky at all 

```suggestion
export default memo(function ConfigNameDescription({
```

python/src/aiconfig/editor/client/src/components/aiconfigReducer.ts
```diff
@@ -192,6 +199,16 @@ export default function aiconfigReducer(
         },
       }));
     }
+    case "SET_NAME_DESCRIPTION": {
```
saqadri, line 202: Out of curiosity, why did you bundle name and description together in a single action? (It is fine as is)

saqadri (approved):

python/src/aiconfig/editor/client/src/components/aiconfigReducer.ts
rholinshead: Will update to do separately (to match endpoints), but was keeping them together to start in case we wanted to do something similar to workbooks where we save the values once the title/description component toggles from 'edit' mode to 'read' mode (instead of each change).

rholinshead:
Updated to handle setDescription and setName separately

PR #683, opened 2023-12-30
Get basic streaming to work for run

rossdanlm:
Get basic streaming to work for run

Building upon what Ryan investigated in https://github.com/lastmile-ai/aiconfig/pull/651, all the frontend changes are from him I just rebased onto his PR

This is a bit tricky becasue we have to support:

1. streaming models --> use a queue iterator for passing the output text
2. non-streaming models --> still yield, just don't use a queue iterator and wait for run command to finish

General flow:
1. Parse prompts from client
2. Define stream callback with queue iterator
3. Start thread to run aiconfig without blocking main thread from accessing queue iterator
4. Create a copy of the original AIConfig so we can write partially streamed outputs, yield and display it without risk of race conditions
5. Wait for queue iterator to start containing data, or wait until max timeout (becuase model may not support streaming)
5. Iterate through queue iterator, saving output to display config, yield display config
6. Once output is complete, wait for the original `config.run()` thread and display the output from that

Open questions/TODOs
1. [solved - use `while output_text_queue.isEmpty() and t.is_alive()`] ~~How can we check whether model supports streaming or not? Right now we just default to having a max timeout of 5s, but long-term would be better for people to explicitly mark this as a boolean flag in their model parser class~~
2. I need update the output format for streaming. I thought it was fine but guess not, will verify again. A bit annoying but also not a crazy blocker for now
3. Client needs to also support streaming, but that's fine Ryan can get unblocked with this diff now
4. Pretty complex, but streaming will break for `run_with_dependencies`. I've got a proposal to fix forward in https://github.com/lastmile-ai/gradio-workbook/pull/64 and really want people to take a look and give feedback

## Test plan
```bash
alias aiconfig="python -m 'aiconfig.scripts.aiconfig_cli'"
aiconfig edit --aiconfig-path="/Users/rossdancraig/Projects/aiconfig/cookbooks/Getting-Started/travel.aiconfig.json" --server-port=8080 --server-mode=debug_servers

# Now run this from another terminal
curl http://localhost:8080/api/run -d '{"prompt_name":"get_activities"}' -X POST -H 'Content-Type: application/json'
```

I also added this line to print output:
```
print(accumulated_output_text)
```
Streaming

https://github.com/lastmile-ai/aiconfig/assets/151060367/d8930ea6-3143-49a3-89c6-4a2668c2e9e1

Non-streaming (same as before)

https://github.com/lastmile-ai/aiconfig/assets/151060367/5aae7c7f-c273-4be7-bcb9-e96199a04076

### DIFF
--- a/python/src/aiconfig/editor/client/src/utils/oboeHelpers.ts
+++ b/python/src/aiconfig/editor/client/src/utils/oboeHelpers.ts
@@ -1,0 +1,37 @@
+import oboe, { Options } from "oboe";
+
+// Promisify Oboe - similar to this: https://stackoverflow.com/questions/54855494/rewrite-fetch-call-to-oboe-for-json-streams-with-typescript
+// Except it allows to use .node('*', fn) & only resolves on done
+// See https://medium.com/@amberlamps84/oboe-js-mongodb-express-node-js-and-the-beauty-of-streams-4a90fad5414 on using oboe vs raw streams
+// (multiple chunks can be sent in single response & we only want valid json ones)
+export async function streamingApi<T>(
+  headers: Options,
+  on: string = "*",
+  fn: (data: any) => void,
+  on2?: string,
+  fn2?: (data: any) => void,
+  on3?: string,
+  fn3?: (data: any) => void
+): Promise<T> {
+  return new Promise((resolve, reject) => {
+    if (fn2 && on2 && fn3 && on3) {
+      oboe(headers)
+        .node(on, fn)
+        .node(on2, fn2)
+        .node(on3, fn3)
+        .done((data) => resolve(data))
+        .fail((err) => reject(err.jsonBody));
+    } else if (fn2 && on2) {
+      oboe(headers)
+        .node(on, fn)
+        .node(on2, fn2)
+        .done((data) => resolve(data))
+        .fail((err) => reject(err.jsonBody));
+    } else {
+      oboe(headers)
+        .node(on, fn)
+        .done((data) => resolve(data))
+        .fail((err) => reject(err.jsonBody));
+    }
+  });
+}
--- a/python/src/aiconfig/editor/server/queue_iterator.py
+++ b/python/src/aiconfig/editor/server/queue_iterator.py
@@ -1,0 +1,44 @@
+"""Queue iterator for streaming. Can only process strings for now
+but in future will try to make it generic"""
+# import asyncio
+from queue import Queue
+
+# from typing import Generic, TypeVar
+
+# # TODO: Add generic typing for queue items
+# # (couldn't get sentinel value to work with generics)
+# T = TypeVar('T')
+STOP_STREAMING_SIGNAL = object()  # sentinel value to indicate end of stream
+
+
+class QueueIterator:
+    """In order to support text streaming, we need to store
+    the output in a queue and iterate over those values. A lot of this was
+    inspired by HuggingFace's TextIteratorStreamer object:
+
+    I know I can just use a queue directly in the callsite with
+    `iter(queue.get, None)`, but having a class makes it easier to manage
+    and abstracts it a bit more.
+    """
+
+    def __init__(self):
+        self.q = Queue()
+        self.stop_signal = STOP_STREAMING_SIGNAL
+        self.timeout = None
+
+    def __iter__(self):
+        return self
+
+    def __next__(self):
+        value = self.q.get(block=True, timeout=self.timeout)
+        if value == self.stop_signal:
+            raise StopIteration()
+        return value
+
+    def put(self, text: str, stream_end: bool = False):
+        self.q.put(text, timeout=self.timeout)
+        if stream_end:
+            self.q.put(self.stop_signal, timeout=self.timeout)
+
+    def isEmpty(self) -> bool:
+        return self.q.empty()
--- a/python/src/aiconfig/editor/server/server.py
+++ b/python/src/aiconfig/editor/server/server.py
@@ -1,10 +1,16 @@
+import asyncio
+import copy
+import json
 import logging
+import threading
+import time
 import webbrowser
 from typing import Any, Dict, Type, Union
 
 import lastmile_utils.lib.core.api as core_utils
 import result
 from aiconfig.Config import AIConfigRuntime
+from aiconfig.editor.server.queue_iterator import STOP_STREAMING_SIGNAL, QueueIterator
 from aiconfig.editor.server.server_utils import (
     EditServerConfig,
     FlaskResponse,
@@ -26,11 +32,11 @@ from aiconfig.editor.server.server_utils import (
 )
 from aiconfig.model_parser import InferenceOptions
 from aiconfig.registry import ModelParserRegistry
-from flask import Flask, request
+from flask import Flask, Response, request, stream_with_context
 from flask_cors import CORS
 from result import Err, Ok, Result
 
-from aiconfig.schema import Prompt
+from aiconfig.schema import ExecuteResult, Output, Prompt
 
 logging.getLogger("werkzeug").disabled = True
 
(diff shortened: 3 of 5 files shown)


python/src/aiconfig/editor/server/server.py
```diff
@@ -162,24 +171,113 @@ def create() -> FlaskResponse:
...
+                        "data": accumulated_output_text,
+                        # Assume streaming only supports single output
+                        # I think this actually may be wrong for PaLM or OpenAI
+                        # TODO: Need to sync with Ankush but can fix forward
+                        "execution_count": 0, 
+                        "metadata": {},
+                    }
+                )
+
+                displaying_config.add_output(prompt_name, accumulated_output, overwrite=True)
```
rossdanlm, line 260: I realized that this will fail for run_with_dependencies since we don't have a way of tracking the currently executing prompt, and just overwriting the last prompt name instead. Since we are deep copying the aiconfig vs. `displaying_config`, the final result won't be as bad, but it'll look weird during streaming and we should fix. I have a pseudo-diff in https://github.com/lastmile-ai/gradio-workbook/pull/64 with two separate proposals (backend vs. frontend) and would like feedback on it from others

python/src/aiconfig/editor/server/queue_iterator.py
```diff
@@ -0,0 +1,41 @@
+"""Queue iterator for streaming. Can only process strings for now
+but in future will try to make it generic"""
+# import asyncio
+from queue import Queue
+# from typing import Generic, TypeVar
+
+# # TODO: Add generic typing for queue items
+# # (couldn't get sentinel value to work with generics)
+# T = TypeVar('T')
+STOP_STREAMING_SIGNAL = object() #sentinel value to indicate end of stream
```
jonathanlastmileai, line 10: nit: object() is a weird sentinel value. Howa bout we make a Sentinel enum?

python/src/aiconfig/editor/server/server.py
```diff
@@ -162,24 +171,102 @@ def create() -> FlaskResponse:
...
     state = get_server_state(app)
     aiconfig = state.aiconfig
     request_json = request.get_json()
+    prompt_name: str = request_json.get("prompt_name", "get_activities") # Hard-coded
+    params = request_json.get("params", {}) #Fixed in https://github.com/lastmile-ai/aiconfig/pull/668
+    stream = request_json.get("stream", True)
+
+    # Define stream callback and queue object for streaming results
+    output_text_queue = QueueIterator()
+    def update_output_queue(data, _accumulated_data, _index) -> None:
```
jonathanlastmileai, line 189: Is this going to work for all the model parser? I thought the idea of the callback was that specific models need specific logic

python/src/aiconfig/editor/server/server.py
```diff
@@ -162,24 +171,102 @@ def create() -> FlaskResponse:
...
+        output_text_queue.put(data, should_end_stream)
+    inference_options = InferenceOptions(
+        stream=stream,
+        stream_callback=update_output_queue,
+    )
+
+    def generate():
+        # Use multi-threading so that we don't block run command from
+        # displaying the streamed output (if streaming is supported)
+        def run_async_config_in_thread():
```
jonathanlastmileai, line 200: nit: I think this runs a new event loop, which doesn't necessarily run in a new thread

python/src/aiconfig/editor/server/server.py
```diff
@@ -162,24 +171,102 @@ def create() -> FlaskResponse:
...
+        # with streaming partial outputs in the meantime. This probably isn't
+        # necessary, but just getting unblocked for now
+        displaying_config = copy.deepcopy(aiconfig)
+
+        # If model supports streaming, need to wait until streamer has at
+        # least 1 item to display. If model does not support streaming,
+        # need to wait until the aiconfig.run() thread is complete
+        SLEEP_DELAY_SECONDS = 0.1
+        wait_time_in_seconds = 0.0
+        while output_text_queue.isEmpty() and t.is_alive():
```
jonathanlastmileai, line 222: what's this loop for?

jonathanlastmileai (commented):
I came to the same conclusion about needing an inter-thread queue while playing with this on server v2. However, in that case, I was restricting myself to not defining a custom callback because my understanding was we would have to define a specific one for every model provider.

If we can pass a callback, isn't this a lot easier with a web socket? You can make the callback write the chunks directly into the socket synchronously; it doesn't matter if you block the main thread, and then when finished return the result as usual.

My verdict: If it works, ship, if still not quite there for whatever reason, let's consider v2

python/src/aiconfig/editor/server/server.py
rossdanlm: - For streaming: we would need to wait for the queue to be populated otherwise the `for text in output_text_queue` will not be able to process anything since it's empty. Therefore we need to wait
- For non-streaming, we have to wait for the thread t to be complete anyways, so we will wait on that here too

python/src/aiconfig/editor/server/server.py
rossdancraig: If your model parser supports streaming, you need to implement something that returns the streaming results into a `stream_callback` object, otherwise nobody will be able to read the results as they're happening. If you don't do this in your model parser, that's on you, not us. 

If there are ever cases where you can stream non-text data formats, we'll make a note and try to follow up later

python/src/aiconfig/editor/server/server.py
rossdancraig: From what I've heard, Python does not actually do "multi-threading", it's just syntax sugar over a way to saying 'switch between these event loop processes'

python/src/aiconfig/editor/server/server.py
```diff
@@ -162,24 +171,102 @@ def create() -> FlaskResponse:
...
+            # Yea I know time.sleep() isn't super accurate, but it's fine,
+            # we can fix later
+            time.sleep(0.1)
+            wait_time_in_seconds += SLEEP_DELAY_SECONDS
+            print(f"Output queue is currently empty. Waiting for {wait_time_in_seconds:.1f}s...")
+
+        yield "["
```
Ankush-lastmile, line 229: nit: can you add a comment explaining why we need open and close brackets to enable streaming

can be simple something like flask/oboe needs this to enable streaming

python/src/aiconfig/editor/server/server.py
Ankush-lastmile: flask doesn't use an async event loop so this should be fine

python/src/aiconfig/editor/server/queue_iterator.py
```diff
@@ -0,0 +1,44 @@
...
+    and abstracts it a bit more.
+    """
+
+    def __init__(self):
+        self.q = Queue()
+        self.stop_signal = STOP_STREAMING_SIGNAL
+        self.timeout = None
+
+    def __iter__(self):
+        return self
```
Ankush-lastmile, line 30: why do we return self on __iter__?

python/src/aiconfig/editor/client/src/utils/oboeHelpers.ts
```diff
@@ -0,0 +1,37 @@
+import oboe, { Options } from "oboe";
```
Ankush-lastmile, line 1: this isn't adding functionality to this pr right?

Ankush-lastmile (approved):
approving to unblock. cc @rossdancraig tested that this doesn't break current functionality since this is now a 1 week old pr.

python/src/aiconfig/editor/server/queue_iterator.py
rossdanlm: Created issue to follow up in https://github.com/lastmile-ai/aiconfig/issues/794

python/src/aiconfig/editor/server/queue_iterator.py
rossdanlm: Whenever we want to build something that is iterable (ie: `for some_val in MyIterableClass`), we need to define a `__iter__` value which returns self to indicate that. If it's an async iterable, this becomes `async def __aiter__(self):`

python/src/aiconfig/editor/client/src/utils/oboeHelpers.ts
rossdanlm: It was needed I think on client for Ryan to process. I kept it here from his diff in https://github.com/lastmile-ai/aiconfig/pull/651

cc @rholinshead will let you keep or delete moving forward

PR #685, opened 2023-12-30
Old: Streaming 'working' With Client-side Handling

rossdanlm:
Old: Streaming 'working' With Client-side Handling

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/685).
* __->__ #685
* #683

### DIFF
--- a/python/src/aiconfig/editor/server/async_queue_iterator.py
+++ b/python/src/aiconfig/editor/server/async_queue_iterator.py
@@ -1,0 +1,40 @@
+"""Queue iterator for streaming. Can only process strings for now
+but in future will try to make it generic"""
+import asyncio
+# from typing import Generic, TypeVar
+
+# # TODO: Add generic typing for queue items
+# # (couldn't get sentinel value to work with generics)
+# T = TypeVar('T')
+STOP_STREAMING_SIGNAL = object() #sentinel value to indicate end of stream
+
+class AsyncQueueIterator():
+    """In order to support text streaming, we need to store
+    the output in a queue and iterate over those values. A lot of this was
+    inspired by HuggingFace's TextIteratorStreamer object:
+    
+    I know I can just use a queue directly in the callsite with 
+    `iter(queue.get, None)`, but having a class makes it easier to manage 
+    and abstracts it a bit more.
+    """
+    def __init__(self):
+        self.q = asyncio.Queue()
+        self.stop_signal = STOP_STREAMING_SIGNAL
+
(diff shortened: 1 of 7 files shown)


python/src/aiconfig/editor/server/server.py
```diff
@@ -161,25 +170,260 @@ def create() -> FlaskResponse:
...
+            )
+            output_text_queue.put(STOP_STREAMING_SIGNAL)
+        t = threading.Thread(target=run_async_config_in_thread)
+        t.start()
+
+
+        # Need to wait until streamer has at least 1 stream
+        # def wait_for_stream():
+        #     while output_text_queue.empty():
+        #         time.sleep(0.1)
```
rossdanlm, line 219: This is what I was having trouble with. Gradio is slow enough that by the time we end up at this if-loop check the aiconfig has already started running and we have output in the queue

python/src/aiconfig/editor/server/async_queue_iterator.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def __aiter__(self):
+        return self
+
+    def __next__(self):
+        value = await self.q.get()
+        if value == self.stop_signal:
+            raise StopIteration()
+        else:
+            return value
```
rossdanlm, line 32: Currently learning how to convert iterators --> async iterators: https://medium.com/geekculture/asynchronous-iterators-in-python-fdf55198287d

Need to refactor this

python/src/aiconfig/editor/server/server.py
```diff
@@ -161,25 +170,260 @@ def create() -> FlaskResponse:
...
+        #             "metadata": {},
+        #         }
+        #     )
+        #     displaying_config.add_output(event.prompt_name, accumulated_output, overwrite=True)
+        #     yield displaying_config
+
+        # yield displaying_config
+        
+        print("Rossdan: About to enter the loop with output_text_queue")
+        for text in output_text_queue:
```
rossdanlm, line 244: Basically `output_text_queue` was always empty, so we never entered into here

python/src/aiconfig/editor/server/server.py
rossdanlm: NM, this could jsut be perhaps because Queue object is blocking while asyncio.Queue is non-blocking. I realized since I have this within `generator` function which is not async I can just use original queue object

PR #691, opened 2024-01-02
Create delete_parameter endpoint

rossdanlm:
Create delete_parameter endpoint

TSIA

## Test Plan
Follow dev README to setup the local editor: https://github.com/lastmile-ai/aiconfig/tree/main/python/src/aiconfig/editor#dev, then run this command
```
curl http://localhost:8080/api/delete_parameter -d '{"prompt_name":"get_activities", "parameter_name": "city"}' -X POST -H 'Content-Type: application/json'
```
Notice that the parameters are now set.

### DIFF
--- a/python/src/aiconfig/editor/server/server.py
+++ b/python/src/aiconfig/editor/server/server.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Any, Dict, Type
+from typing import Any, Dict, Type, Union
 
 import lastmile_utils.lib.core.api as core_utils
 import result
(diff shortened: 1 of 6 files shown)


python/src/aiconfig/editor/server/server.py
```diff
@@ -258,3 +258,18 @@ def set_parameter() -> FlaskResponse:
...
     return run_aiconfig_operation_with_op_args(aiconfig, "set_parameter", operation, operation_args)
+
+@app.route("/api/delete_parameter", methods=["POST"])
+def delete_parameter() -> FlaskResponse:
+    state = get_server_state(app)
+    aiconfig = state.aiconfig
+    request_json = request.get_json()
+
+    parameter_name: str | None = request_json.get("parameter_name")
+    prompt_name: str | None = request_json.get("prompt_name")
```
jonathanlastmileai, line 269: Same as above, what do we want to do with None values? Also, are we distinguishing between missing and user-supplied None?

python/src/aiconfig/editor/server/server.py
```diff
@@ -258,3 +258,18 @@ def set_parameter() -> FlaskResponse:
...
+
+    operation = make_op_run_method(MethodName("delete_parameter"))
+    operation_args: Result[OpArgs, str] = result.Ok(
+        OpArgs({"parameter_name": parameter_name, "prompt_name": prompt_name})
+    )
+    return run_aiconfig_operation_with_op_args(aiconfig, "delete_parameter", operation, operation_args)
```
jonathanlastmileai, line 275: It seems many endpoints are repeating a bunch of code including this function call. It just occurred to me we have this wrapper `run_aiconfig_operation_with_request_json()` which you can use in certain cases:

>     Use this when there is no specific logic needed to extract the OpArgs from the request JSON.
    This is true in cases where the signature is enough to do that automatically

 Is it not usable here?

jonathanlastmileai (commented):
comment, also see previous about pyright just in case

python/src/aiconfig/editor/server/server.py
rholinshead: parameter_name should be required and throw if not provided

rholinshead (approved):
Accepting to unblock but we should make sure to handle parameter_name missing with an error

PR #690, opened 2024-01-02
Create set_parameter endpoint

rossdanlm:
Create set_parameter endpoint

TSIA, this relies on https://github.com/lastmile-ai/aiconfig/pull/670 being unblocked

## Test Plan
Follow dev README to setup the local editor: https://github.com/lastmile-ai/aiconfig/tree/main/python/src/aiconfig/editor#dev, then run this command
```
curl http://localhost:8080/api/set_parameter -d '{"prompt_name":"get_activities", "parameter_name": "city", "parameter_value": "New York"}' -X POST -H 'Content-Type: application/json'
```
Notice that the parameters are now set. We already have automated testing from https://github.com/lastmile-ai/aiconfig/pull/670 so we know the functionality works as long as this is connected to the client (pretty great!)

https://github.com/lastmile-ai/aiconfig/assets/151060367/1be05d7e-63fe-4d6d-9036-6f8d4d004134

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/690).
* #693
* #692
* #691
* __->__ #690
* #688
* #670
* #668

PR #688, opened 2024-01-02
Create update_model endpoint

rossdanlm:
Create update_model endpoint

TSIA, backend implementation of `update_model()`. The other changes were from running the auto-formatter:
```
fd --glob '*.py'  python/src/aiconfig/editor/server | xargs python -m 'scripts.lint' --mode=fix --files
```

## Test Plan
Follow dev README to setup the local editor: https://github.com/lastmile-ai/aiconfig/tree/main/python/src/aiconfig/editor#dev, then run this command
```
curl http://localhost:8080/api/update_model -d curl http://localhost:8080/api/update_model -d '{"prompt_name":"get_activities", "model_name": "gpt-4", "settings": {"top_p": 0.9}}' -X POST -H 'Content-Type: application/json'
```
Notice that the model name is updated to `gpt-4`, and the `top_p` is now at 0.9

https://github.com/lastmile-ai/aiconfig/assets/151060367/3031fa59-8925-495c-a5eb-e56ec65b7fba

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/688).
* #693
* #692
* #691
* #690
* __->__ #688
* #670
* #668

PR #692, opened 2024-01-02
Create set_name endpoint

rossdanlm:
Create set_name endpoint

TSIA

## Test Plan
Follow dev README to setup the local editor: https://github.com/lastmile-ai/aiconfig/tree/main/python/src/aiconfig/editor#dev, then run this command
```
curl http://localhost:8080/api/set_name -d '{"name":"cool_new_name"}' -X POST -H 'Content-Type: application/json'
```

https://github.com/lastmile-ai/aiconfig/assets/151060367/efcb7ceb-e390-4208-93b3-3effd5d66f8b

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/692).
* #693
* __->__ #692
* #691
* #690
* #688
* #670
* #668

### DIFF
--- a/python/src/aiconfig/editor/server/server.py
+++ b/python/src/aiconfig/editor/server/server.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Any, Dict, Type
+from typing import Any, Dict, Type, Union
 
 import lastmile_utils.lib.core.api as core_utils
 import result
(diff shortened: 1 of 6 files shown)


python/src/aiconfig/editor/server/server.py
```diff
@@ -269,7 +270,18 @@ def delete_parameter() -> FlaskResponse:
     prompt_name: str | None = request_json.get("prompt_name")
 
     operation = make_op_run_method(MethodName("delete_parameter"))
-    operation_args: Result[OpArgs, str] = result.Ok(
-        OpArgs({"parameter_name": parameter_name, "prompt_name": prompt_name})
-    )
+    operation_args: Result[OpArgs, str] = result.Ok(OpArgs({"parameter_name": parameter_name, "prompt_name": prompt_name}))
     return run_aiconfig_operation_with_op_args(aiconfig, "delete_parameter", operation, operation_args)
```
jonathanlastmileai, line 274: Is this auto-linting? Can you try to do that atomically for large sets of files in their own PRs, or even better do it before submitting the PR that modifies that specific code?

python/src/aiconfig/editor/server/server.py
```diff
@@ -269,7 +270,18 @@ def delete_parameter() -> FlaskResponse:
...
+def set_name() -> FlaskResponse:
+    state = get_server_state(app)
+    aiconfig = state.aiconfig
+    request_json = request.get_json()
+
+    name: str | None = request_json.get("name")
+
+    operation = make_op_run_method(MethodName("set_name"))
+    operation_args: Result[OpArgs, str] = result.Ok(OpArgs({"name": name}))
+    return run_aiconfig_operation_with_op_args(aiconfig, "set_name", operation, operation_args)
```
jonathanlastmileai, line 287: I'm fairly sure you can use the wrapper function here (see a previous comment). If not, at this point I would probably refactor a bit to make it usable. I usually use the rule of thumb that a set of 3 cases justifies an abstraction. 

If we hadn't done that to define `make_op_run_method()` and `run_aiconfig_operation_with_op_args()` in the first place, we would have either a lot more code duplication than this or a lot less robust error handling. We can probably go a little further here.

rholinshead (approved):

python/src/aiconfig/editor/server/server.py
rossdanlm: You're right, my bad I forgot to do it last time

python/src/aiconfig/editor/server/server.py
rossdanlm: Yea you're right, I was just copy pasting stuff

PR #697, opened 2024-01-02
[AIC-py][editor] file path initialization

jonathanlastmileai:
[AIC-py][editor] file path initialization

1. path is non-optional and has a default value
2. on init, create a new instance and save to the path if the file doesn't exist
(otherwise, try to load as usual)

Run server with and without explicit path. Observe in the logs
that, in either case, when the file doesn't exist it
gets created and written to.

Run again pointing at an existing valid path (e.g. the one just created).
Observe again in the logs that it loads from the path.

https://github.com/lastmile-ai/aiconfig/assets/148090348/ee24a0c0-84d8-47aa-b37f-bc9180948b4e

### DIFF
--- a/python/src/aiconfig/editor/server/server_utils.py
+++ b/python/src/aiconfig/editor/server/server_utils.py
@@ -7,7 +7,7 @@ import typing
 from dataclasses import dataclass
 from enum import Enum
 from types import ModuleType
-from typing import Any, Callable, NewType, Optional, Type, TypeVar, cast
+from typing import Any, Callable, NewType, Type, TypeVar, cast
 
 import lastmile_utils.lib.core.api as core_utils
 import result
@@ -57,7 +57,7 @@ class ServerMode(Enum):
 
 class EditServerConfig(core_utils.Record):
     server_port: int = 8080
-    aiconfig_path: Optional[str] = None
+    aiconfig_path: str = "my_aiconfig.aiconfig.json"
     log_level: str | int = "INFO"
     server_mode: ServerMode
     parsers_module_path: str = "aiconfig_model_registry.py"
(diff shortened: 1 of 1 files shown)


python/src/aiconfig/editor/server/server_utils.py
```diff
@@ -56,7 +57,7 @@ class ServerMode(Enum):
 
 class EditServerConfig(core_utils.Record):
     server_port: int = 8080
-    aiconfig_path: Optional[str] = None
+    aiconfig_path: str = "my_aiconfig.aiconfig.json"
```
rossdanlm, line 60: can we just call this `my_aiconfig.json`. Having two separate `.` is confusing to a lot of n00bs

python/src/aiconfig/editor/server/server_utils.py
```diff
@@ -204,11 +205,10 @@ def init_server_state(app: Flask, edit_config: EditServerConfig) -> Result[None,
     state = get_server_state(app)
 
     assert state.aiconfig is None
-    if edit_config.aiconfig_path:
+    if os.path.exists(edit_config.aiconfig_path):
```
rossdanlm, line 208: Edge case, what if they pass in `my_aiconfig.json` as their EXISTING path (ex: they are using local editor one time, this file gets created and saved there, then they go cool, this path now exists, I'll pass it in next time

rossdanlm (approved):
Pls upload screenshots/videos in test plan but otherwise lgtm!

python/src/aiconfig/editor/server/server_utils.py
jonathanlastmileai: It will hit this branch and attempt to load that file. If it was created properly the first time, it will just load it.

python/src/aiconfig/editor/server/server_utils.py
jonathanlastmileai: I thought the standard was going to be .aiconfig.json? cc @saqadri ?

PR #710, opened 2024-01-02
[editor] Connect Prompt Name to /update_prompt

rholinshead:
[editor] Connect Prompt Name to /update_prompt

# [editor] Connect Prompt Name to /update_prompt

Connect the onChangePromptName callback to the server with /update_prompt request. Note the special handling here:
Since most AIConfig API methods depend on unique prompt names, we can't update the prompt name in the config on every change (or else we could end up using a not-yet-existing prompt name for other actions). Instead, we maintain the prompt name input in the client side prompt name component state and then update the client AIConfig state after the response has returned successfully.

Note: Realized as part of this that the debounce value returns the prev debounce result, so we actually need to handle the async response dispatch event with a callback to the debounce function.

## Testing:
Make sure updating prompt name works now

https://github.com/lastmile-ai/aiconfig/assets/5060851/55da5642-9b36-44f4-aab1-0c461efab7a8

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/710).
* #716
* #715
* __->__ #710

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
@@ -122,8 +122,19 @@ export default function EditorContainer({
   const debouncedUpdatePrompt = useMemo(
     () =>
       debounce(
-        (promptName: string, newPrompt: Prompt) =>
-          updatePromptCallback(promptName, newPrompt),
+        async (
+          promptName: string,
+          newPrompt: Prompt,
+          onSuccess?: (aiconfigRes: AIConfig) => void
+        ) => {
+          const serverConfigRes = await updatePromptCallback(
+            promptName,
+            newPrompt
+          );
+          if (serverConfigRes && onSuccess) {
+            onSuccess(serverConfigRes.aiconfig);
+          }
+        },
         DEBOUNCE_MS
       ),
     [updatePromptCallback]
@@ -146,22 +157,23 @@ export default function EditorContainer({
         }
         const prompt = clientPromptToAIConfigPrompt(statePrompt);
 
-        const serverConfigRes = await debouncedUpdatePrompt(prompt.name, {
-          ...prompt,
-          input: newPromptInput,
-        });
-
-        if (serverConfigRes) {
-          dispatch({
-            type: "CONSOLIDATE_AICONFIG",
-            action,
-            config: serverConfigRes.aiconfig,
-          });
-        }
+        await debouncedUpdatePrompt(
+          prompt.name,
+          {
+            ...prompt,
+            input: newPromptInput,
+          },
+          (serverConfigRes) =>
+            dispatch({
+              type: "CONSOLIDATE_AICONFIG",
+              action,
+              config: serverConfigRes.aiconfig,
+            })
+        );
       } catch (err: unknown) {
         const message = err instanceof Error ? err.message : null;
         showNotification({
-          title: "Error adding prompt to config",
+          title: "Error updating prompt input",
           message,
           color: "red",
         });
@@ -172,15 +184,39 @@ export default function EditorContainer({
 
   const onChangePromptName = useCallback(
     async (promptId: string, newName: string) => {
-      const action: AIConfigReducerAction = {
-        type: "UPDATE_PROMPT_NAME",
-        id: promptId,
-        name: newName,
-      };
+      try {
+        const statePrompt = getPrompt(stateRef.current, promptId);
+        if (!statePrompt) {
+          throw new Error(`Could not find prompt with id ${promptId}`);
+        }
+        const prompt = clientPromptToAIConfigPrompt(statePrompt);
 
-      dispatch(action);
+        await debouncedUpdatePrompt(
+          prompt.name,
+          {
+            ...prompt,
+            name: newName,
+          },
+          // PromptName component maintains local state for the name to show in the UI
+          // We cannot update client config state until the name is successfully set server-side
+          // or else we could end up referencing a prompt name that is not set server-side
+          () =>
+            dispatch({
+              type: "UPDATE_PROMPT_NAME",
+              id: promptId,
+              name: newName,
+            })
+        );
+      } catch (err: unknown) {
+        const message = err instanceof Error ? err.message : null;
+        showNotification({
+          title: "Error updating prompt name",
+          message,
+          color: "red",
+        });
+      }
     },
-    [dispatch]
+    [debouncedUpdatePrompt]
   );
 
   const updateModelCallback = callbacks.updateModel;
--- a/python/src/aiconfig/editor/client/src/utils/constants.ts
+++ b/python/src/aiconfig/editor/client/src/utils/constants.ts
@@ -1,1 +1,1 @@
-export const DEBOUNCE_MS = 250;
+export const DEBOUNCE_MS = 300;


rholinshead:
@rossdanlm - here's a video of the error I'm seeing:
Basically, there's some disconnect somewhere with the config api not knowing that the server-side state has specific prompt name?

https://github.com/lastmile-ai/aiconfig/assets/5060851/e6c4f200-1cdd-40b2-98f3-31fd164557f2

rholinshead:
> @rossdanlm - here's a video of the error I'm seeing: Basically, there's some disconnect somewhere with the config api not knowing that the server-side state has specific prompt name?
> 
>  Screen.Recording.2024-01-02.at.2.06.38.PM.mov

This error was resolved by #714

rossdanlm:
> @rossdanlm - here's a video of the error I'm seeing: Basically, there's some disconnect somewhere with the config api not knowing that the server-side state has specific prompt name?
> 
>  Screen.Recording.2024-01-02.at.2.06.38.PM.mov

Just for followup, this was fixed in https://github.com/lastmile-ai/aiconfig/pull/714

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -122,8 +122,19 @@ export default function EditorContainer({
   const debouncedUpdatePrompt = useMemo(
     () =>
       debounce(
-        (promptName: string, newPrompt: Prompt) =>
-          updatePromptCallback(promptName, newPrompt),
+        async (
+          promptName: string,
+          newPrompt: Prompt,
+          resFn?: (aiconfigRes: AIConfig) => void
```
rossdanlm, line 128: nit: can we call this something a bit more clear? `res` to me I've never really liked because it can mean either `result` or `response`. `responseCallbackHandler` or `onCompletionHandler` are some examples I'd use

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -146,22 +157,23 @@ export default function EditorContainer({
...
+          prompt.name,
+          {
+            ...prompt,
+            input: newPromptInput,
+          },
+          (serverConfigRes) =>
+            dispatch({
+              type: "CONSOLIDATE_AICONFIG",
+              action,
+              config: serverConfigRes.aiconfig,
```
rossdanlm, line 170: Very cool!

rossdanlm: nit: not specific to this diff and not important at all, but for `CONSOLIDATE_AICONFIG`, is it:

1. ONLY used for processing a response from client (ex: not used for initializing AIConfig on client)?
2. ALWAYS used after we get a response back from client?

If both the above are true, I think it would be helpful to use a more descriptive name like `"REFRESH_AICONFIG_AFTER_RESPONSE"` to make this more clear, but not a big deal

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -172,15 +184,39 @@ export default function EditorContainer({
...
-      dispatch(action);
+        await debouncedUpdatePrompt(
+          prompt.name,
+          {
+            ...prompt,
+            name: newName,
+          },
+          // PromptName component maintains local state for the name to show in the UI
+          // We cannot update client config state until the name is successfully set server-side
+          // or else we could end up referencing a prompt name that is not set server-side
```
rossdanlm, line 202: Thanks for the great comment!

python/src/aiconfig/editor/client/src/utils/constants.ts
```diff
@@ -1 +1 @@
-export const DEBOUNCE_MS = 250;
+export const DEBOUNCE_MS = 300;
```
rossdanlm, line 1: Why change this?

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
rossdanlm: nit: let's add comment of code pointer that this local state happens within the `PromptName` component: https://github.com/lastmile-ai/aiconfig/blob/cec0103bd1f20c7d36b64dcc0249ce6f3817aa21/python/src/aiconfig/editor/client/src/components/prompt/PromptName.tsx#L16

I also have another question, does this dispatch to promptName change happen every time the user changes the name on the keyboard every ~300ms due to DEBOUNCE_MS? https://github.com/lastmile-ai/aiconfig/blob/cec0103bd1f20c7d36b64dcc0249ce6f3817aa21/python/src/aiconfig/editor/client/src/components/prompt/PromptName.tsx#L20

rossdanlm (approved):
Nice! left a few questions but nothing major

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
rholinshead: The debounce call will happen on each change but the resulting dispatch will happen only after the successful response is returned after the eventual debounced function call (if there's no more input after 300ms)

python/src/aiconfig/editor/client/src/utils/constants.ts
rholinshead: I was playing around with it when testing the bug with prompt name that you fixed so this is technically not needed. However, I'd like to keep this because it doesn't result in any noticeable UI change while also potentially reducing the amount of requests sent. With debounce you sometimes need to play around with it a bit to find the sweet spot

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
rossdanlm: Oh yea, this seemed to me not clear that it was `onSuccess` or `onCompletion`. I think if you intend this to be `onSuccess` only then that makes sense!

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
rossdanlm: > just if it's possible the server state could change elsewhere via the AIConfig API

I see, so just to clarify, for example you can update the AIConfig somehow on your own without client sending a request, then the client sends requests and it notices that somehow other parts of the config have been updated (un-related to request call) so it has to also "refresh" those parts too?

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
rholinshead: > let's add comment of code pointer that this local state happens within the PromptName component

Is it ok to just keep as-is, with it already referencing `PromptName component`? I think having a code pointer can be brittle since code can easily change to move lines around (e.g. with styling / UI changes)

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
rholinshead: More specifically, if we send a request to the server then the server will use the AIConfig API to mutate the server-side AIConfig runtime. That mutation could have some side-effects (e.g. elsewhere in the config) aside from the basic mutations we already do on the client. For example, changing prompt model could change the prompt model settings to some specific values (e.g. in the future, take only the existing settings that are supported by the prompt schema). On the client, we would just know that the model changes, not necessarily need to know about the settings change until the request comes back with the update. In practice, not sure how much we need to deal with consolidation since I don't think we have many side-effects, but it's something we should support to be safe

PR #712, opened 2024-01-02
[editor] Persist Setting Params

rholinshead:
[editor] Persist Setting Params

[editor] Persist Setting Params

Update the set_parameters implementation to clear existing parameters before setting the new ones (otherwise, we end up with a ton of duplicates). Then, connect the parameters callbacks to the endpoint.

## Testing:
- Set and update global and prompt-level parameters, reload to ensure persisted values server-side:

https://github.com/lastmile-ai/aiconfig/assets/5060851/19b061a3-1cff-4e66-a2fb-43bfe662fd5b

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
@@ -46,6 +46,7 @@ export type AIConfigCallbacks = {
   save: (aiconfig: AIConfig) => Promise<void>;
   setConfigDescription: (description: string) => Promise<void>;
   setConfigName: (name: string) => Promise<void>;
+  setParameters: (parameters: JSONObject, promptName?: string) => Promise<void>;
   updateModel: (value: {
     modelName?: string;
     settings?: InferenceSettings;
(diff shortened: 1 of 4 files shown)


python/src/aiconfig/schema.py
```diff
@@ -484,6 +484,17 @@ def set_parameters(self, parameters: JSONObject, prompt_name: Optional[str] = No
                 parameters dict to. If none is provided, we update the
                 AIConfig-level parameters instead
         """
+        # Clear all existing parameters before setting new ones
```
rholinshead, line 487: cc @rossdanlm for visibility

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -280,9 +302,23 @@ export default function EditorContainer({
...
+      try {
+        const statePrompt = getPrompt(stateRef.current, promptId);
+        if (!statePrompt) {
+          throw new Error(`Could not find prompt with id ${promptId}`);
+        }
+        await debouncedSetParameters(newParameters, statePrompt.name);
+      } catch (err: unknown) {
+        const message = err instanceof Error ? err.message : null;
+        showNotification({
+          title: "Error setting prompt parameters",
```
saqadri, line 315: nit: maybe we should have the prompt name in the message just in case

saqadri (approved):
Thanks, Ryan! Looks good

python/src/aiconfig/schema.py
rossdanlm: Oh true, good catch!

python/src/aiconfig/schema.py
```diff
@@ -484,6 +484,17 @@ def set_parameters(self, parameters: JSONObject, prompt_name: Optional[str] = No
...
+        parameter_names_to_delete = []
+        if prompt_name:
+            prompt = self.get_prompt(prompt_name)
+            parameter_names_to_delete = list(self.get_prompt_parameters(prompt).keys())
```
rossdanlm, line 491: Nice, glad these helper functions are being used!

PR #715, opened 2024-01-03
[editor] Handle Run Prompt Errors

rholinshead:
[editor] Handle Run Prompt Errors

# [editor] Handle Run Prompt Errors

When running a prompt results in an error, update the running state to false to ensure it doesn't get stuck in endless 'running' state.

Also, fix handling of the error notification:
- Missing <Notification/> component in the app
- Miss-handling caught object type (somehow not instance of error, just a {message: ".."} object)

## Testing:
- See run prompt error for dall-e-2, stops the running state and shows a notification with the message

https://github.com/lastmile-ai/aiconfig/assets/5060851/d156a774-9c6c-412a-9e6c-645c269065ca

As a follow-up, we should render a nicer error output as part of the prompt output to be persisted in the UI (until rerun or reload)

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/715).
* #716
* __->__ #715
* #710

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
@@ -1,6 +1,6 @@
 import PromptContainer from "./prompt/PromptContainer";
 import { Container, Button, createStyles, Stack, Flex } from "@mantine/core";
-import { showNotification } from "@mantine/notifications";
+import { Notifications, showNotification } from "@mantine/notifications";
 import {
   AIConfig,
   InferenceSettings,
@@ -460,9 +460,19 @@ export default function EditorContainer({
           config: serverConfigRes.aiconfig,
         });
       } catch (err: unknown) {
-        const message = err instanceof Error ? err.message : null;
+        const message = (err as { message?: string }).message ?? null;
+
+        // TODO: Add ErrorOutput component to show error instead of notification
+        dispatch({
+          type: "RUN_PROMPT_ERROR",
+          id: promptId,
+          message: message ?? undefined,
+        });
+
+        const promptName = getPrompt(stateRef.current, promptId)?.name;
+
         showNotification({
-          title: "Error running prompt",
+          title: `Error running prompt${promptName ? ` ${promptName}` : ""}`,
           message,
           color: "red",
         });
@@ -540,6 +550,7 @@ export default function EditorContainer({
 
   return (
     <AIConfigContext.Provider value={contextValue}>
+      <Notifications />
       <Container maw="80rem">
         <Flex justify="flex-end" mt="md" mb="xs">
           <Button loading={isSaving} onClick={onSave}>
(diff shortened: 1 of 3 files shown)


python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -540,6 +548,7 @@ export default function EditorContainer({
 
   return (
     <AIConfigContext.Provider value={contextValue}>
+      <Notifications />
```
rossdanlm, line 551: What does this do?

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -460,7 +460,15 @@ export default function EditorContainer({
...
         });
       } catch (err: unknown) {
-        const message = err instanceof Error ? err.message : null;
+        const message = (err as { message?: string }).message ?? null;
+
+        // TODO: Add ErrorOutput component to show error instead of notification
+        dispatch({
+          type: "RUN_PROMPT_ERROR",
+          id: promptId,
+          message: message ?? undefined,
```
rossdanlm, line 469: Nit: would you recommend including the `promptId` (which I think is prompt name?) in the error message to make it more clear to user?

rossdanlm (approved):
One question I have is what exactly is the issue from the test example that you showed? I would imagine Dall-E 2 should draw a cat?

I tried running it myself and got no error? https://lastmileai.dev/workbooks/clqx7k66r006eqwd4r2up9lir
Screenshot 2024-01-02 at 22 18 43

If you're not sure, pls ping me so I can check and test this on server side!

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
rholinshead: It's the Mantine component responsible for showing the notifications that are dispatched with `showNotification`. It was just left out accidentally when copying stuff over

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
rholinshead: `promptId` is a ui-state-specific unique id for each prompt and not the same as the name -- I don't really think it would be helpful to show since we never expose the unique ids for the prompts in the UI. I can update to show the name (if it can be obtained)

rholinshead: Screenshot 2024-01-03 at 10 44 27 AM

rholinshead:
> One question I have is what exactly is the issue from the test example that you showed? I would imagine Dall-E 2 should draw a cat?
> 
> I tried running it myself and got no error? https://lastmileai.dev/workbooks/clqx7k66r006eqwd4r2up9lir Screenshot 2024-01-02 at 22 18 43
> 
> If you're not sure, pls ping me so I can check and test this on server side!

Neither of the dall-e models are working for me locally, I get the TypeError, `Cannot instantiate typing.Union`. I see Sarmad had the same error in the bug bash

PR #716, opened 2024-01-03
[editor] Fix Error Message Handling

rholinshead:
[editor] Fix Error Message Handling

# [editor] Fix Error Message Handling

Just update the error type to not rely on being an instance of `Error`, and just to use the {message?: string} type

## Testing:
- Intentionally raise error in /save `return HttpResponseWithAIConfig(message="ERROR", code=500, aiconfig=None).to_flask_format()` then attempt to save & see correct error notification

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/716).
* __->__ #716
* #715
* #710

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
@@ -1,6 +1,6 @@
 import PromptContainer from "./prompt/PromptContainer";
 import { Container, Button, createStyles, Stack, Flex } from "@mantine/core";
-import { showNotification } from "@mantine/notifications";
+import { Notifications, showNotification } from "@mantine/notifications";
 import {
   AIConfig,
   InferenceSettings,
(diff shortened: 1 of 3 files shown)


python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -58,6 +58,8 @@ export type AIConfigCallbacks = {
   ) => Promise<{ aiconfig: AIConfig }>;
 };
 
+type RequestCallbackError = { message?: string };
```
rossdanlm, line 61: I see, so this is kind of like taking a generic `Error` object and casting it to an object where the only thing we care about is the message field if it exists? Not sure how much you plan to extend it, but for now would prefer to have this be more obvious in the name. Ex: `RequestCallbackErrorWithMaybeMessage`. Kind of long and tedious naming, so you don't have to do this if you don't want to

rossdanlm (approved):
lgtm

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
rholinshead: It's a bit strange how we need to handle it -- basically, any server response that is a `HttpResponseWithAIConfig` with a code will result in a fetch request error but the caught value is not an error, it's just a raw object containing message. However -- it's still possible that an Error could be caught, or I guess some other arbitrary type. So, just typing with optional message to be safe

rholinshead: Going to leave as-is but if you feel super strongly on this please lmk and I can fix forward

PR #717, opened 2024-01-03
[editor] Add DalleImageGenerationParserPromptSchema

rholinshead:
[editor] Add DalleImageGenerationParserPromptSchema

# [editor] Add DalleImageGenerationParserPromptSchema

Add the prompt schema for dall-e-2 to match `/opt/homebrew/Caskroom/miniconda/base/envs/aiconfig/lib/python3.12/site-packages/openai/resources/images.py`:

```
 def generate(
        self,
        *,
        prompt: str,
        model: Union[str, Literal["dall-e-2", "dall-e-3"], None] | NotGiven = NOT_GIVEN,
        n: Optional[int] | NotGiven = NOT_GIVEN,
        quality: Literal["standard", "hd"] | NotGiven = NOT_GIVEN,
        response_format: Optional[Literal["url", "b64_json"]] | NotGiven = NOT_GIVEN,
        size: Optional[Literal["256x256", "512x512", "1024x1024", "1792x1024", "1024x1792"]] | NotGiven = NOT_GIVEN,
        style: Optional[Literal["vivid", "natural"]] | NotGiven = NOT_GIVEN,
        user: str | NotGiven = NOT_GIVEN,
        # Use the following arguments if you need to pass additional parameters to the API that aren't available via kwargs.
        # The extra values given here take precedence over values defined on the client or passed to this method.
        extra_headers: Headers | None = None,
        extra_query: Query | None = None,
        extra_body: Body | None = None,
        timeout: float | httpx.Timeout | None | NotGiven = NOT_GIVEN,
    ) -> ImagesResponse:
```

And supported keys specified in dalle.py:
```
supported_keys = {"model", "n", "quality", "response_format", "size", "style"}
```

Screenshot 2024-01-02 at 7 20 29 PM

### DIFF
--- a/python/src/aiconfig/editor/client/src/shared/prompt_schemas/DalleImageGenerationParserPromptSchema.ts
+++ b/python/src/aiconfig/editor/client/src/shared/prompt_schemas/DalleImageGenerationParserPromptSchema.ts
@@ -1,0 +1,36 @@
+import { PromptSchema } from "../../utils/promptUtils";
+
+export const DalleImageGenerationParserPromptSchema: PromptSchema = {
+  // See /opt/homebrew/Caskroom/miniconda/base/envs/aiconfig/lib/python3.12/site-packages/openai/resources/images.py
+  // for supported settings
+
+  input: {
+    type: "string",
+  },
+  model_settings: {
+    type: "object",
+    properties: {
+      n: {
+        type: "integer",
+        minimum: 1,
+        maximum: 10,
+      },
+      quality: {
+        type: "string",
+        enum: ["standard", "hd"],
+      },
+      response_format: {
+        type: "string",
+        enum: ["url", "b64_json"],
+      },
+      size: {
+        type: "string",
+        enum: ["256x256", "512x512", "1024x1024", "1792x1024", "1024x1792"],
+      },
+      style: {
+        type: "string",
+        enum: ["vivid", "natural"],
+      },
+    },
+  },
+};
--- a/python/src/aiconfig/editor/client/src/utils/promptUtils.ts
+++ b/python/src/aiconfig/editor/client/src/utils/promptUtils.ts
@@ -1,6 +1,7 @@
 import { JSONObject, JSONValue, Prompt } from "aiconfig";
 import { OpenAIChatModelParserPromptSchema } from "../shared/prompt_schemas/OpenAIChatModelParserPromptSchema";
 import { OpenAIChatVisionModelParserPromptSchema } from "../shared/prompt_schemas/OpenAIChatVisionModelParserPromptSchema";
+import { DalleImageGenerationParserPromptSchema } from "../shared/prompt_schemas/DalleImageGenerationParserPromptSchema";
 
 /**
  * Get the name of the model for the specified prompt. The name will either be specified in the prompt's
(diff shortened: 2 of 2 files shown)


rossdanlm:
Not blocking, to do in another PR, but think it would be nice to have a little info button that when you click it:
1. Has short description of what it means
2. Links to the official API docs: https://platform.openai.com/docs/api-reference/images/create

Screenshot 2024-01-02 at 22 26 35

python/src/aiconfig/editor/client/src/shared/prompt_schemas/DalleImageGenerationParserPromptSchema.ts
```diff
@@ -0,0 +1,35 @@
+import { PromptSchema } from "../../utils/promptUtils";
+
+export const DalleImageGenerationParserPromptSchema: PromptSchema = {
+  // "n", "quality", "response_format", "size", "style"
```
rossdanlm, line 4: delete?

rossdanlm: Let's also add comment in code saying this is to match the types from `/opt/homebrew/Caskroom/miniconda/base/envs/aiconfig/lib/python3.12/site-packages/openai/resources/images.py:`

Separate note: this also kind of fits with general theme (not P1 or even P2) of us should maybe define schema on backend in 1 centralized place?

python/src/aiconfig/editor/client/src/shared/prompt_schemas/DalleImageGenerationParserPromptSchema.ts
```diff
@@ -0,0 +1,35 @@
...
+  input: {
+    type: "string",
+  },
+  model_settings: {
+    type: "object",
+    properties: {
+      n: {
+        type: "integer",
+        minimum: 1,
+        maximum: 10,
```
rossdanlm, line 15: Just a note, Dall-E 3 only supports n=1. This we already flag as an error on the backend, but yea could be a P2 UX guide if we want to make it a bit more intuitive to users

Screenshot 2024-01-02 at 22 32 12

I can't quite remember but I think the response just errors. I don't think it's a big deal, but worth testing just to sanity check that error propagates properly to UI. If not, please flag to me and I'll fix the model parser to raise this error explicitly in this situation!

rossdanlm (approved):
nice!

rholinshead:
> Not blocking, to do in another PR, but think it would be nice to have a little info button that when you click it:
> 
> 1. Has short description of what it means
> 2. Links to the official API docs: https://platform.openai.com/docs/api-reference/images/create
> 
> Screenshot 2024-01-02 at 22 26 35

I think we can augment it with a 'description' text for each item at the least, but will think on having an info link -- it would have to be part of the PromptSchema spec

python/src/aiconfig/editor/client/src/shared/prompt_schemas/DalleImageGenerationParserPromptSchema.ts
rholinshead: Neither dall-e model works for me -- I get, `Failed to run prompt: <class 'TypeError'>, Cannot instantiate typing.Union` for both. In this case, I was just ensuring the correct settings are propagated to the server from the client

rossdanlm: Ok I created a starter issue for this in https://github.com/lastmile-ai/aiconfig/issues/720. I will give to someone if they're interested in helping!

PR #723, opened 2024-01-03
[editor][hotfix] Fix Debounced Update Prompt Callback

rholinshead:
# [editor][hotfix] Fix Debounced Update Prompt Callback

Was incorrectly indexing .aiconfig from the resulting aiconfig (which types fine since AIConfig is completely open)


python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -169,7 +169,7 @@ export default function EditorContainer({
             dispatch({
               type: "CONSOLIDATE_AICONFIG",
               action,
-              config: serverConfigRes.aiconfig,
+              config: serverConfigRes,
```
saqadri, line 172: Sigh, easy to make mistakes like this with the typings

PR #724, opened 2024-01-03
[editor] Debounced Error Handling for Prompt Name and Input

rholinshead:
[editor] Debounced Error Handling for Prompt Name and Input

# [editor] Debounced Error Handling for Prompt Name and Input

`debounced` callbacks won't propagate thrown errors so we should use an `onError` callback to handle those. This PR handles them for updating prompt name and input.

## Testing:
Added the following in /update_prompt endpoint to return an error:
```
return HttpResponseWithAIConfig(
        #
        message=f"Failed to update prompt",
        code=500,
        aiconfig=None,
    ).to_flask_format()
```

See notifications:

https://github.com/lastmile-ai/aiconfig/assets/5060851/04270541-eca9-4d41-8b7c-1e456d992814

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/724).
* #725
* __->__ #724

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
@@ -127,14 +127,19 @@ export default function EditorContainer({
         async (
           promptName: string,
           newPrompt: Prompt,
-          onSuccess?: (aiconfigRes: AIConfig) => void
+          onSuccess: (aiconfigRes: AIConfig) => void,
+          onError: (err: unknown) => void
         ) => {
-          const serverConfigRes = await updatePromptCallback(
-            promptName,
-            newPrompt
-          );
-          if (serverConfigRes && onSuccess) {
-            onSuccess(serverConfigRes.aiconfig);
+          try {
+            const serverConfigRes = await updatePromptCallback(
+              promptName,
+              newPrompt
+            );
+            if (serverConfigRes?.aiconfig) {
+              onSuccess(serverConfigRes.aiconfig);
+            }
+          } catch (err: unknown) {
+            onError(err);
           }
         },
         DEBOUNCE_MS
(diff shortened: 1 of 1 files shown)


python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -127,14 +127,21 @@ export default function EditorContainer({
...
-          );
-          if (serverConfigRes && onSuccess) {
-            onSuccess(serverConfigRes.aiconfig);
+          try {
+            const serverConfigRes = await updatePromptCallback(
+              promptName,
+              newPrompt
+            );
+            if (serverConfigRes && onSuccess) {
+              onSuccess(serverConfigRes.aiconfig);
```
saqadri, line 139: Checking if this should be `serverConfigRes`? Or if `.aiconfig` indexing here is valid

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -152,6 +159,15 @@ export default function EditorContainer({
 
       dispatch(action);
 
+      const onError = (err: unknown) => {
+        const message = (err as RequestCallbackError).message ?? null;
```
saqadri, line 163: Is this guaranteed to be `RequestCallbackError`? Or should this be the same as the onChangePromptName's `onError`:
```
const message = err instanceof Error ? err.message : null;
```

saqadri (approved):

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
rholinshead: When the flask server returns an error, it results in a caught error but the actual value caught is a raw object with message -- it's not an instanceof Error. Since Error instances can also have message, this should handle both cases

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
rholinshead: aiconfig indexing is valid for our server responses. I can add the extra check to be sure

PR #734, opened 2024-01-03
[editor] Autosave at 15s Interval

rholinshead:
[editor] Autosave at 15s Interval

# [editor] Autosave at 15s Interval

Set up an interval to autosave the config every 15s from the client. Note that the interval is technically cleared & recreated each time the `isDirty` state changes, but that's fine because this will also handle the case where changes are made right when a successful save response comes back, or if a save fails (doesn't clear dirty state) another attempt will be made automatically after 15s.

https://github.com/lastmile-ai/aiconfig/assets/5060851/dfcb57bd-282d-4622-bd10-ec4a281c213c

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/734).
* #735
* __->__ #734
* #733

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
@@ -15,7 +15,14 @@ import {
   Prompt,
   PromptInput,
 } from "aiconfig";
-import { useCallback, useMemo, useReducer, useRef, useState } from "react";
+import {
+  useCallback,
+  useEffect,
+  useMemo,
+  useReducer,
+  useRef,
+  useState,
+} from "react";
 import aiconfigReducer, { AIConfigReducerAction } from "./aiconfigReducer";
 import {
   ClientPrompt,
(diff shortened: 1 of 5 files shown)


python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -626,6 +633,16 @@ export default function EditorContainer({
...
+  useEffect(() => {
+    if (!isDirty) {
+      return;
+    }
+
+    // Save every 15 seconds if there are unsaved changes
+    const saveInterval = setInterval(onSave, AUTOSAVE_INTERVAL_MS);
+
+    return () => clearInterval(saveInterval);
+  }, [isDirty, onSave]);
```
saqadri, line 645: Does `isDirty` being changed retrigger an instant save?

saqadri (approved):

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
rholinshead: No, it happens at the end of the interval, so 15s after becoming dirty

PR #735, opened 2024-01-03
[editor] Handle CMD/CTRL + S to Save

rholinshead:
# [editor] Handle CMD/CTRL + S to Save

A common action when using an editor UI is to hit CMD+s or CTRL+s to save. In browsers, this will by default open up the browser save window to save the web page itself. In this PR, we override this default behaviour and instead call the /save endpoint for the config if it's dirty

https://github.com/lastmile-ai/aiconfig/assets/5060851/a2bf656b-2209-49e0-a404-31b23b836abd

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/735).
* __->__ #735
* #734
* #733


saqadri (approved):
Looks great!~

PR #730, opened 2024-01-03
Create AnyScale Endpoint modelparser

saqadri:
Create AnyScale Endpoint modelparser

Simply extend the OpenAIInference modelparser and change the api key and base_url to point to anyscale.

Added utility classes for all the supported models for Anyscale Endpoints: https://docs.endpoints.anyscale.com/category/supported-models

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/730).
* #752
* #748
* #746
* #745
* #743
* __->__ #730

PR #743, opened 2024-01-03
[TypeScript] Default modelparser for AnyScale endpoints

saqadri:
[TypeScript] Default modelparser for AnyScale endpoints

Add a ModelParser for Anyscale Endpoints

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/743).
* #752
* #748
* #746
* #745
* __->__ #743
* #730

### DIFF
--- a/typescript/lib/config.ts
+++ b/typescript/lib/config.ts
@@ -20,6 +20,7 @@ import { PaLMTextParser } from "./parsers/palm";
 import { extractOverrideSettings } from "./utils";
 import { HuggingFaceTextGenerationParser } from "./parsers/hf";
 import { CallbackEvent, CallbackManager } from "./callback";
+import { AnyscaleEndpointModelParser } from "./parsers/anyscale";
 
 /**
  * Options for saving an AIConfig to a file.
@@ -57,6 +58,11 @@ ModelParserRegistry.registerModelParser(new OpenAIChatModelParser(), [
   "gpt-3.5-turbo-0613",
   "gpt-3.5-turbo-16k-0613",
 ]);
+
+ModelParserRegistry.registerModelParser(new AnyscaleEndpointModelParser(), [
+  "AnyscaleEndpoint",
+]);
+
 ModelParserRegistry.registerModelParser(new HuggingFaceTextGenerationParser());
 ModelParserRegistry.registerModelParser(new PaLMTextParser());
 
--- a/typescript/lib/parsers/anyscale.ts
+++ b/typescript/lib/parsers/anyscale.ts
@@ -1,0 +1,33 @@
+import { JSONObject } from "../../common";
+import { Output, Prompt } from "../../types";
+import { AIConfigRuntime } from "../config";
+import OpenAI from "openai";
+import { InferenceOptions } from "../modelParser";
+import { OpenAIChatModelParser } from "./openai";
+
+export class AnyscaleEndpointModelParser extends OpenAIChatModelParser {
+  public async run(
+    prompt: Prompt,
+    aiConfig: AIConfigRuntime,
+    options?: InferenceOptions,
+    params?: JSONObject | undefined
+  ): Promise<Output[]> {
+    if (!this.openai) {
+      const apiKey =
+        process.env.ANYSCALE_ENDPOINT_API_KEY ?? process.env.OPENAI_API_KEY;
+      if (!apiKey) {
+        throw new Error(
+          "Missing API key ANYSCALE_ENDPOINT_API_KEY or OPENAI_API_KEY in environment. Please set one of these environment variables to utilize Anyscale Endpoints. You can get your API key from https://docs.endpoints.anyscale.com/guides/authenticate"
+        );
+      }
+
+      this.openai = new OpenAI({
+        apiKey,
+        baseURL: "https://api.endpoints.anyscale.com/v1",
+        ...(this.openaiOptions || {}),
+      });
+    }
+
+    return super.run(prompt, aiConfig, options, params);
+  }
+}
--- a/python/src/aiconfig/Config.py
+++ b/python/src/aiconfig/Config.py
@@ -5,6 +5,7 @@ from typing import Any, Dict, List, Literal, Optional, Tuple
 
 import requests
 from aiconfig.callback import CallbackEvent, CallbackManager
+from aiconfig.default_parsers.anyscale_endpoint import DefaultAnyscaleEndpointParser
 from aiconfig.default_parsers.openai import DefaultOpenAIParser
 from aiconfig.default_parsers.palm import PaLMChatParser, PaLMTextParser
 from aiconfig.model_parser import InferenceOptions, ModelParser
@@ -35,6 +36,7 @@ gpt_models = [
 ]
 for model in gpt_models:
     ModelParserRegistry.register_model_parser(DefaultOpenAIParser(model))
+ModelParserRegistry.register_model_parser(DefaultAnyscaleEndpointParser("AnyscaleEndpoint"))
 ModelParserRegistry.register_model_parser(PaLMChatParser())
 ModelParserRegistry.register_model_parser(PaLMTextParser())
 ModelParserRegistry.register_model_parser(HuggingFaceTextGenerationParser())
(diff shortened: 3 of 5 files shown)


typescript/lib/config.ts
```diff
@@ -31,35 +31,47 @@ export type SaveOptions = {
...
+    "gpt-3.5-turbo-0613",
+    "gpt-3.5-turbo-16k-0613",
+  ]);
+
+  ModelParserRegistry.registerModelParser(
+    new OpenAIChatModelParser({
+      apiKey: getAPIKeyFromEnv("ANYSCALE_ENDPOINT_API_KEY"),
+      baseURL: "https://api.endpoints.anyscale.com/v1",
+    }),
+    ["AnyscaleEndpoint"]
```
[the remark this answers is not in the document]
saqadri: The way the anyscale endpoint works is we pass the "model" to it. So the settings in the aiconfig would be something like:
```
{
      "name": "gen_itinerary",
      "input": "Generate an itinerary ordered by {{order_by}} for these activities: {{get_activities.output}}.",
      "metadata": {
        "model": {
          "name": "AnyscaleEndpoint",
          "settings": {
            "model": "mistralai/Mixtral-8x7B-Instruct-v0.1"
          }
        },
        "parameters": {
          "order_by": "geographic location"
        }
      }
    }
```

Alternatively, we create a `model_parsers` mapping in the aiconfig, and use those model IDs directly:
```
"model_parsers": {
      "meta-llama/Llama-2-7b-chat-hf": "AnyscaleEndpoint",
      "meta-llama/Llama-2-13b-chat-hf": "AnyscaleEndpoint",
      "Meta-Llama/Llama-Guard-7b": "AnyscaleEndpoint",
      "meta-llama/Llama-2-70b-chat-hf": "AnyscaleEndpoint",
      "Open-Orca/Mistral-7B-OpenOrca": "AnyscaleEndpoint",
      "codellama/CodeLlama-34b-Instruct-hf": "AnyscaleEndpoint",
      "HuggingFaceH4/zephyr-7b-beta": "AnyscaleEndpoint",
      "mistralai/Mistral-7B-Instruct-v0.1": "AnyscaleEndpoint",
      "mistralai/Mixtral-8x7B-Instruct-v0.1": "AnyscaleEndpoint",
      "thenlper/gte-large": "AnyscaleEndpoint"
    },
```

I think this model parser works similarly to a HF task -- i.e. it aggregates different models under a single model parser. We will need to think about how we handle this in the UI (ideally we do it the same way we will handle HF tasks).

I was thinking a ModelParser can have a `list_models` API which returns the list of model IDs that it supports. Then the `list_models` endpoint for the local editor will be able to enumerate those properly.

saqadri:
Test Plan: 

Ran `get_activities` prompt for the following aiconfig:

```
{
  "name": "NYC Trip Planner",
  "description": "Intrepid explorer with ChatGPT and AIConfig",
  "schema_version": "latest",
  "metadata": {
    "model_parsers": {
      "meta-llama/Llama-2-7b-chat-hf": "AnyscaleEndpoint",
      "meta-llama/Llama-2-13b-chat-hf": "AnyscaleEndpoint",
      "Meta-Llama/Llama-Guard-7b": "AnyscaleEndpoint",
      "meta-llama/Llama-2-70b-chat-hf": "AnyscaleEndpoint",
      "Open-Orca/Mistral-7B-OpenOrca": "AnyscaleEndpoint",
      "codellama/CodeLlama-34b-Instruct-hf": "AnyscaleEndpoint",
      "HuggingFaceH4/zephyr-7b-beta": "AnyscaleEndpoint",
      "mistralai/Mistral-7B-Instruct-v0.1": "AnyscaleEndpoint",
      "mistralai/Mixtral-8x7B-Instruct-v0.1": "AnyscaleEndpoint",
      "thenlper/gte-large": "AnyscaleEndpoint"
    },
    "models": {
      "meta-llama/Llama-2-7b-chat-hf": {
        "model": "meta-llama/Llama-2-7b-chat-hf",
        "top_p": 1,
        "temperature": 1
      },
      "meta-llama/Llama-2-70b-chat-hf": {
        "model": "meta-llama/Llama-2-70b-chat-hf",
        "max_tokens": 3000,
        "system_prompt": "You are an expert travel coordinator with exquisite taste."
      }
    },
    "default_model": "meta-llama/Llama-2-7b-chat-hf"
  },
  "prompts": [
    {
      "name": "get_activities",
      "input": "Tell me 10 fun attractions to do in NYC."
    },
    {
      "name": "gen_itinerary",
      "input": "Generate an itinerary ordered by {{order_by}} for these activities: {{get_activities.output}}.",
      "metadata": {
        "model": {
          "name": "AnyscaleEndpoint",
          "settings": {
            "model": "mistralai/Mixtral-8x7B-Instruct-v0.1"
          }
        },
        "parameters": {
          "order_by": "geographic location"
        }
      }
    }
  ]
}
```

https://github.com/lastmile-ai/aiconfig/assets/25641935/45e7def9-9d2c-43e8-a52a-6e7f864811e3

typescript/lib/config.ts
```diff
@@ -88,6 +101,8 @@ export class AIConfigRuntime implements AIConfig {
     this.schema_version = schemaVersion;
     this.metadata = metadata || {};
     this.prompts = prompts || [];
+
+    registerDefaultModelParsers();
```
rholinshead, line 105: I think this could be problematic in two cases:
- User currently registers custom parsers before instantiating an AIConfigRuntime; this could overwrite their custom registry
- Even if they instantiate an AIConfigRuntime *before* customizing the registry, it can still be overwritten the next time they construct an AIConfigRuntime instance

saqadri: Ah, good point. My main issue was the AnyscaleEndpoint model parser takes in an API key in the constructor, which may not have loaded from the env at the start when the registration is done statically.

rholinshead: Maybe we can just do a simple check here to see if there are already model parsers registered (registry list is nonempty), and only register these defaults if not?

typescript/lib/config.ts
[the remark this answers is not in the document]
rholinshead: I still think it's a bit strange how we're able to conflate model and model parser in the config itself -- this is another way we are able to represent the same concept in multiple different ways in the config & makes it pretty hard to generalize / reason about when 'model' is sometimes a model but sometimes a model parser.

For example, this case:
```
{
      "name": "gen_itinerary",
      "input": "Generate an itinerary ordered by {{order_by}} for these activities: {{get_activities.output}}.",
      "metadata": {
        "model": {
          "name": "AnyscaleEndpoint",
          "settings": {
            "model": "mistralai/Mixtral-8x7B-Instruct-v0.1"
          }
        },
        "parameters": {
          "order_by": "geographic location"
        }
      }
    }
```

Is not really necessary to support if the parser itself just handles the serialization of model name in a consistent way (always as the model name instead of in settings)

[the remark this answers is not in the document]
saqadri: I agree, I was thinking the same thing when I was working on this. Let me create an issue and we can work on improving this (and ideally having a single way)

typescript/lib/config.ts
saqadri: I think that can cause other issues, because if a user has already registered custom model parsers, it won't register the default ones. I've updated this by creating an AnyscaleEndpointModelParser that overrides OpenAIChatModelParser and just updates the openAI client in the `run` function itself.

rholinshead (approved):
LGTM

typescript/lib/parsers/anyscale.ts
```diff
@@ -0,0 +1,33 @@
...
+import { OpenAIChatModelParser } from "./openai";
+
+export class AnyscaleEndpointModelParser extends OpenAIChatModelParser {
+  public async run(
+    prompt: Prompt,
+    aiConfig: AIConfigRuntime,
+    options?: InferenceOptions,
+    params?: JSONObject | undefined
+  ): Promise<Output[]> {
+    if (!this.openai) {
```
rholinshead, line 15: Nice, this is much cleaner than dealing with the parser registry edge cases

PR #744, opened 2024-01-03
[editor] Image Output Rendering

rholinshead:
# [editor] Image Output Rendering

Get image output rendering with the updated output types, handling both file_uri and base64 encoding.

## Testing:
- Tested both dall-e-2 and 3 models to ensure outputs render, toggle between uri/base64 and raw JSON:

https://github.com/lastmile-ai/aiconfig/assets/5060851/0f73b765-d509-4305-a3bd-98ddb3005ff9

- Made sure the gpt4-v input rendering still works with updated MimeTypeRenderer component:
Screenshot 2024-01-03 at 4 31 52 PM


saqadri (approved):
Niiiice

PR #751, opened 2024-01-04
[editor] Temporarily Remove 'Add New Model' Option in AddPromptButton

rholinshead:
# [editor] Temporarily Remove 'Add New Model' Option in AddPromptButton

Custom model support is a fast-follow post-MVP so remove the button that will eventually show the info modal for adding custom model parsers

## Before:
Screenshot 2024-01-04 at 10 45 12 AM

## After:
Screenshot 2024-01-04 at 10 40 21 AM

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/prompt/AddPromptButton.tsx
+++ b/python/src/aiconfig/editor/client/src/components/prompt/AddPromptButton.tsx
@@ -79,8 +79,9 @@ export default memo(function AddPromptButton({ addPrompt, getModels }: Props) {
           collapseLimit={5}
           onSelectModel={onAddPrompt}
         />
+        {/* TODO: Add back once we have custom model parsers fully supported
         <Menu.Divider />
-        <Menu.Item icon={<IconPlus size="16" />}>Add New Model</Menu.Item>
+        <Menu.Item icon={<IconPlus size="16" />}>Add New Model</Menu.Item> */}
       </Menu.Dropdown>
     </Menu>
   );


python/src/aiconfig/editor/client/src/components/prompt/AddPromptButton.tsx
```diff
@@ -79,8 +79,9 @@ export default memo(function AddPromptButton({ addPrompt, getModels }: Props) {
           collapseLimit={5}
           onSelectModel={onAddPrompt}
         />
+        {/* TODO: Add back once we have custom model parsers fully supported
         <Menu.Divider />
-        <Menu.Item icon={<IconPlus size="16" />}>Add New Model</Menu.Item>
+        <Menu.Item icon={<IconPlus size="16" />}>Add New Model</Menu.Item> */}
```
saqadri, line 84: This makes sense. Also my feedback for this (once we add it back) is that it's confusing. It's not immediately clear what "Add new model" should do here. Wondering if this should belong in a toolbar at the top alongside save. cc @zakariassoul

saqadri (approved):

python/src/aiconfig/editor/client/src/components/prompt/AddPromptButton.tsx
rholinshead: @saqadri the button wasn't actually implemented fully -- designs show that it will show an info modal on how to add a custom model parser which will then show up in the actual model list once added following the instructions. I just didn't implement the modal yet because it would not have correct info (since we don't support custom model parsers in the editor just yet)

PR #754, opened 2024-01-04
[editor] PaLMChatParserPromptSchema

rholinshead:
[editor] PaLMChatParserPromptSchema

# [editor] PaLMChatParserPromptSchema

Implement the Prompt Schema for PaLM Chat, using the supported params with defaults from https://cloud.google.com/vertex-ai/docs/generative-ai/model-reference/text-chat, selecting those that are defined in the `refine_chat_completion_params` of palm.py:
```
def refine_chat_completion_params(model_settings):
    # completion parameters to be used for Palm's chat completion api
    # messages handled seperately
    supported_keys = {
        "candidate_count",
        "examples",
        "model",
        "temperature",
        "top_k",
        "top_p",
        "context",
    }
 ```

Screenshot 2024-01-04 at 1 59 24 PM

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/754).
* #756
* #755
* __->__ #754

PR #755, opened 2024-01-04
[editor] HuggingFaceTextGenerationParserPromptSchema

rholinshead:
[editor] HuggingFaceTextGenerationParserPromptSchema

# [editor] HuggingFaceTextGenerationParserPromptSchema

Adding the PromptSchema for HuggingFaceTextGenerationParser, with supported properties from `refine_chat_completion_params` in the parser implementation. Types obtained from https://github.com/huggingface/huggingface_hub/blob/a331e82aad1bc63038194611236db28fa013814c/src/huggingface_hub/inference/_client.py#L1206 and defaults obtained from https://huggingface.co/docs/api-inference/detailed_parameters if they're listed

Screenshot 2024-01-04 at 3 19 05 PM

Note, this is for the default/core parser which uses inference API. We'll need to add the other prompt schemas for the extension models for gradio.

Will update the hf prompt UX to support actual 'model' in a subsequent diff

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/755).
* #756
* __->__ #755
* #754

### DIFF
--- a/python/src/aiconfig/editor/client/src/shared/prompt_schemas/HuggingFaceTextGenerationParserPromptSchema.ts
+++ b/python/src/aiconfig/editor/client/src/shared/prompt_schemas/HuggingFaceTextGenerationParserPromptSchema.ts
@@ -1,0 +1,73 @@
+import { PromptSchema } from "../../utils/promptUtils";
+
+export const HuggingFaceTextGenerationParserPromptSchema: PromptSchema = {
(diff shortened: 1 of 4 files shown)


rholinshead:
Landing to unblock bug bashing, will fix forward as needed

PR #756, opened 2024-01-04
[editor] Fix Model Search to Ignore Casing

rholinshead:
# [editor] Fix Model Search to Ignore Casing

Previously, the model search was doing exact match, including capitalization. So, searching 'hu' would not show the "HuggingFaceTextGenerationParser".

Fix by making both search and model names lowercase when filtering:
Screenshot 2024-01-04 at 3 35 46 PM

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/756).
* __->__ #756
* #755
* #754


rholinshead:
Landing to unblock bug bashing, will fix forward as needed

PR #759, opened 2024-01-04
[editor] Fix Prompt Model Name Check

rholinshead:
# [editor] Fix Prompt Model Name Check

Noticed in bug bash for the example aiconfig file, modifying the model settings for the first prompt fails.
The error was thrown if setting the model settings when the prompt has no model set; when this happens we need to set the model name to the default model in the config and then we can set the remaining settings.

## Testing:
```
aiconfig edit --aiconfig-path=python/src/aiconfig/editor/travel.aiconfig.json --server-mode='debug_servers'
```
Edit settings for the first prompt and make sure it succeeds and sets correct config data:
```
      "name": "get_activities",
      "input": "Tell me 10 fun attractions to do in NYC.",
      "metadata": {
        "model": {
          "name": "gpt-3.5-turbo",
          "settings": {
            "system_prompt": "test"
          }
        },
        "parameters": {}
      },
```

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
@@ -299,7 +299,10 @@ export default function EditorContainer({
         if (!statePrompt) {
           throw new Error(`Could not find prompt with id ${promptId}`);
         }
-        const modelName = getPromptModelName(statePrompt);
+        const modelName = getPromptModelName(
+          statePrompt,
+          stateRef.current.metadata.default_model
+        );
         if (!modelName) {
           throw new Error(`Could not find model name for prompt ${promptId}`);
         }


python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -299,7 +299,10 @@ export default function EditorContainer({
         if (!statePrompt) {
           throw new Error(`Could not find prompt with id ${promptId}`);
         }
-        const modelName = getPromptModelName(statePrompt);
+        const modelName = getPromptModelName(
+          statePrompt,
+          stateRef.current.metadata.default_model
+        );
```
rossdanlm, line 305: Wow what a move!

rossdanlm:
> The error was thrown if setting the model settings when the prompt has no model set

Do you mean that this occurs when the system model settings + default model are defined, but prompt model is not? Also why would

rossdanlm (approved):
You sir have earned an arthur_delivers macro
3d349795-639d-467e-b4a8-862a8e6493db

rholinshead:
> > The error was thrown if setting the model settings when the prompt has no model set
> 
> Do you mean that this occurs when the system model settings + default model are defined, but prompt model is not? Also why would

When we have a prompt in the config with no 'model' in its metadata (i.e. it will use default model of the config), and then we try to set model settings for that prompt, we need to update the prompt to have model metadata (specifically, `ModelMetadata` object with `name: <model name>`). The code was previously using setting that model name in the metadata based on existing prompt model metadata (which doesn't exist if using default model), so need to get the name from the config level

PR #761, opened 2024-01-04
[aiconfig] Remove Duplicate GPT-4 Model from Default Registry

rholinshead:
# [aiconfig] Remove Duplicate GPT-4 Model from Default Registry

The default model parser registry was registering GPT-4 with capitals and was only used in this example config used for testing. Let's just fix that config to use "gpt-4" (lowercase) and remove the duplicate/confusing capitalized model.

## Testing:
`pytest`
```
-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
====================================================================================== 103 passed, 38 warnings in 1.59s =======================================================================================
```

### DIFF
--- a/python/src/aiconfig/Config.py
+++ b/python/src/aiconfig/Config.py
@@ -22,7 +22,6 @@ from .util.config_utils import is_yaml_ext
 
 gpt_models = [
     "gpt-4",
-    "GPT-4",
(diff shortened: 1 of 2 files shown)


rossdanlm (approved):
bang! get out of here capital letters, positive vibes only!

PR #762, opened 2024-01-04
[editor] Server Status Heartbeat

rholinshead:
[editor] Server Status Heartbeat

# [editor] Server Status Heartbeat

Implement a `server_status` endpoint to use as a heartbeat for the client to know if the server is down; if the request fails, we'll show an error banner at the top of the config to warn the user that the server's in a bad state & some remediation steps (copy important work & restart editor)

Screenshot 2024-01-04 at 6 58 48 PM

QUESTION: Once the server status is marked as error, do we want to keep the heartbeat going (is it possible for the server to recover without user manually restarting it?)

## Testing:
- Load editor, ensure good status and no banner shown
- CMD+C in CLI to shut down the editor. Make sure client shows the error banner
- Start editor again. Client reloads and shows no error

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
@@ -5,7 +5,9 @@ import {
   createStyles,
   Stack,
   Flex,
+  Text,
   Tooltip,
+  Alert,
 } from "@mantine/core";
 import { Notifications, showNotification } from "@mantine/notifications";
 import {
@@ -40,9 +42,14 @@ import PromptMenuButton from "./prompt/PromptMenuButton";
 import GlobalParametersContainer from "./GlobalParametersContainer";
 import AIConfigContext from "./AIConfigContext";
 import ConfigNameDescription from "./ConfigNameDescription";
-import { AUTOSAVE_INTERVAL_MS, DEBOUNCE_MS } from "../utils/constants";
+import {
+  AUTOSAVE_INTERVAL_MS,
+  DEBOUNCE_MS,
+  SERVER_HEARTBEAT_INTERVAL_MS,
+} from "../utils/constants";
 import { getPromptModelName } from "../utils/promptUtils";
 import { IconDeviceFloppy } from "@tabler/icons-react";
+import CopyButton from "./CopyButton";
 
 type Props = {
   aiconfig: AIConfig;
@@ -57,6 +64,7 @@ export type AIConfigCallbacks = {
   ) => Promise<{ aiconfig: AIConfig }>;
   deletePrompt: (promptName: string) => Promise<void>;
   getModels: (search: string) => Promise<string[]>;
+  getServerStatus?: () => Promise<{ status: "OK" | "ERROR" }>;
   runPrompt: (promptName: string) => Promise<{ aiconfig: AIConfig }>;
   save: (aiconfig: AIConfig) => Promise<void>;
   setConfigDescription: (description: string) => Promise<void>;
@@ -110,6 +118,7 @@ export default function EditorContainer({
   callbacks,
 }: Props) {
   const [isSaving, setIsSaving] = useState(false);
+  const [serverStatus, setServerStatus] = useState<"OK" | "ERROR">("OK");
   const [aiconfigState, dispatch] = useReducer(
     aiconfigReducer,
     aiConfigToClientConfig(initialAIConfig)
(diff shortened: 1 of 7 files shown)


rossdanlm:
> QUESTION: Once the server status is marked as error, do we want to keep the heartbeat going (is it possible for the server to recover without user manually restarting it?)

Let's just send a link to FAQ docs. cc @tanya-rai-lm

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -665,9 +669,45 @@ export default function EditorContainer({
...
+    }
+
+    const interval = setInterval(async () => {
+      try {
+        const res = await getServerStatusCallback();
+        setServerStatus(res.status);
+      } catch (err: unknown) {
+        setServerStatus("ERROR");
+      }
+    }, 3000);
```
rossdanlm, line 686: Can we store this in the same place with the DEBOUNCE param? Will make it easier to see our macros

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -665,9 +669,45 @@ export default function EditorContainer({
...
+    if (!getServerStatusCallback) {
+      return;
+    }
+
+    const interval = setInterval(async () => {
+      try {
+        const res = await getServerStatusCallback();
+        setServerStatus(res.status);
+      } catch (err: unknown) {
+        setServerStatus("ERROR");
```
rossdanlm, line 684: Should we propagate the error message to display in the alert?

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -665,9 +669,45 @@ export default function EditorContainer({
...
+
+    return () => clearInterval(interval);
+  }, [getServerStatusCallback]);
+
   return (
     <AIConfigContext.Provider value={contextValue}>
       <Notifications />
+      {serverStatus !== "OK" && (
+        <>
+          {/* // Simple placeholder block div to make sure the banner does not overlap page contents until scrolling past its height */}
```
rossdanlm, line 696: This is actually incredible. Very few people know the small nuances like this and even less actually document it well so others know it easily!

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
rossdanlm: Also is there a best practice for how frequent heartbeats should be? Or similar to debounce it's just more of a "play around and find out" kind of deal?

rossdanlm (approved):
r/MemeTemplatesOfficial[624 × 722](https://www.google.com/url?sa=i&url=https%3A%2F%2Fwww.reddit.com%2Fr%2FMemeTemplatesOfficial%2Fcomments%2Fet9qoe%2Fcall_the_ambulance_but_not_for_me_template%2F&psig=AOvVaw2pdH4vB2-au6fHnQ9dCSOe&ust=1704530733740000&source=images&cd=vfe&opi=89978449&ved=0CBMQjRxqFwoTCIjt-vztxYMDFQAAAAAdAAAAABAK)

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -665,9 +669,45 @@ export default function EditorContainer({
...
+          <Alert
+            color="red"
+            title="Server Connection Error"
+            w="100%"
+            style={{ position: "fixed", top: 0, zIndex: 999 }}
+          >
+            <Text>
+              There is a problem with the editor server connection. Please copy
+              important changes somewhere safe and try restarting the editor.
+            </Text>
```
saqadri, line 707: What do you think of allowing people to save the json from the frontend state? Or at least a "copy" button for it?

saqadri: Can be in follow-up diff

saqadri (approved):
Awesome! I had the same idea, so you beat me to it :) 

One question about allowing save or copy from the frontend state as a fallback if server fails

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
rholinshead: There isn't really an error message -- it's either "OK" or not, since it will only error if the server is not able to return a response

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
rholinshead: Lol. There's probably a cleaner way to do it with css (and/or js, to calculate alert height) but this is my "if it works..." solution

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
rholinshead: I think it's usually order of seconds but depends on the importance. In our case there's a potential for lost work so shouldn't be a large interval

rholinshead:
> > QUESTION: Once the server status is marked as error, do we want to keep the heartbeat going (is it possible for the server to recover without user manually restarting it?)
> 
> Let's just send a link to FAQ docs. cc @tanya-rai-lm , @tanya-rai

Will add as a follow-up once we have that

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
rholinshead: Added a copy button here; client-based save can be a follow-up if it will work, just need to dedicate time to looking into it

rholinshead: Screenshot 2024-01-05 at 2 26 46 PM

PR #752, opened 2024-01-04
[bug] Ensure that Output is constructed consistently across streaming and non-streaming

saqadri:
[bug] Ensure that Output is constructed consistently across streaming and non-streaming

Currently the openai modelparser calls `build_output_data` to properly construct the response from the inference endpoint into an AIConfig OutputData type.

However, it only does this when stream = False. This change adds the same logic to stream = True (and also ensures that the metadata is set consistently and in a loss-less way.

Also updated the same logic for Anyscale Endpoint model parser.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/752).
* __->__ #752
* #748
* #746
* #745
* #743
* #730

PR #766, opened 2024-01-05
Check tool call type instead of function field for tool call data support

rossdanlm:
Check tool call type instead of function field for tool call data support

Just a small nit I noticed from Sarmad's changes in https://github.com/lastmile-ai/aiconfig/pull/752. I also ran the autoformatter command: `fd --glob '*.py'  python/src/aiconfig/default_parsers | xargs python -m 'scripts.lint' --mode=fix --files`.

I also updated the 3 ipynb notebooks and found + fixed some bugs from that! I also cleared the outputs here, not sure if that's desired, but tested that it works so should be good!

## Test Plan
Run the anyscale, openai, and function call cookbooks to make sure they still work

Screenshot 2024-01-04 at 22 13 11
Screenshot 2024-01-04 at 22 13 14
Screenshot 2024-01-04 at 22 35 08

### DIFF
--- a/cookbooks/Function-Calling-OpenAI/python/Function_Calling_with_OpenAI.ipynb
+++ b/cookbooks/Function-Calling-OpenAI/python/Function_Calling_with_OpenAI.ipynb
@@ -20,7 +20,7 @@
       "outputs": [],
       "source": [
         "# Install AIConfig package\n",
-        "!pip install python-aiconfig\n",
+        "!pip3 install python-aiconfig\n",
         "\n",
         "# Create .env file at aiconfig/.env containing the following line: \n",
         "# OPENAI_API_KEY=<your key here>\n",
@@ -242,8 +242,8 @@
         "import json\n",
         "\n",
         "def call_function(function_call):\n",
-        "    args = json.loads(function_call['arguments'])\n",
-        "    name = function_call['name']\n",
+        "    args = json.loads(function_call.arguments)\n",
+        "    name = function_call.name\n",
         "\n",
         "    if name == 'list':\n",
         "        return list(args['genre'])\n",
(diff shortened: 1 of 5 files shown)


cookbooks/Function-Calling-OpenAI/python/Function_Calling_with_OpenAI.ipynb
```diff
@@ -267,10 +267,9 @@
         "params = {\"book\":\"Where the Crawdads Sing\"}\n",
         "inference_options = InferenceOptions(stream=True)\n",
         "\n",
-        "completion = await config.run(\"recommend_book\", params, options=inference_options)\n",
-        "\n",
-        "func_call = completion[0].data.get('function_call')\n",
-        "print(func_call)\n"
+        "result = await config.run(\"recommend_book\", params, options=inference_options)\n",
+        "tool_calls = result[0].data.value\n",
+        "func_call = tool_calls[0].function\n"
```
rossdanlm, line 272: These changes actually fix something that was broken in the cookbook

saqadri (commented):
Please also update anyscale_endpoint.py with the same changes as in OpenAI.py

saqadri (approved):
Looks good, just the one comment to update anyscale_endpoint.py with the same changes as in OpenAI.py

rossdanlm:
> Please also update anyscale_endpoint.py with the same changes as in OpenAI.py

This is actually already done because anyscale inherits from OpenAI:

> class AnyscaleEndpoint(OpenAIInference):

We can probably actually just make `build_output_data` a class method inside of `OpenAIInference` file so that we dont' have to redo two callsites on changes. Made bootcamp task for it in https://github.com/lastmile-ai/aiconfig/issues/767

Issue #768, opened 2024-01-05
Automated Testing for HuggingFace text generation core parser


rossdanlm:
See comments in

PR #769, opened 2024-01-05
Fix issue where hugging face response was not correct type

rossdanlm:
Fix issue where hugging face response was not correct type

Turned out to be a typing issue that python did not actually help with.

I also had to adjust the way we load in hugging face key because I got rate limited. We NEED to have more clear way to set API keys in `aiconfig/.env` file though, or else people won't know how to use it! cc @tanya-rai let's add an FAQ/README similar to what we have in the cookbooks: https://github.com/lastmile-ai/aiconfig/blob/3dcf70bcddda7f7041c3bfb6e5d4b6c9a7abd4ab/cookbooks/Getting-Started/getting_started.ipynb?short_path=dcb1340#L23-L25

We need to do this for keys
- HUGGING_FACE_API_TOKEN
- ANYSCALE_ENDPOINT_API_KEY
- OPENAI_API_KEY
- GOOGLE_API_KEY

## Testing
Connected to local editor (you'll have to modify the `max_tokens` setting) and runs now. We should also have automated testing for this hugging face model parser. Created task for it in https://github.com/lastmile-ai/aiconfig/issues/768

Screenshot 2024-01-05 at 00 47 51

### DIFF
--- a/cookbooks/HuggingFace/hf.py
+++ b/cookbooks/HuggingFace/hf.py
@@ -128,7 +128,7 @@ class HuggingFaceTextParser(ParameterizedModelParser):
     A model parser for HuggingFace text generation models.
     """
 
-    def __init__(self, model_id: str = None, use_api_token=False):
+    def __init__(self, model_id: str = None, use_api_token=True):
         """
         Args:
             model_id (str): The model ID of the model to use.
--- a/python/src/aiconfig/default_parsers/hf.py
+++ b/python/src/aiconfig/default_parsers/hf.py
@@ -99,7 +99,7 @@ def construct_stream_output(
     return output
 
 
-def construct_regular_output(response: TextGenerationResponse, response_includes_details: bool) -> Output:
+def construct_regular_output(response: str, response_includes_details: bool) -> Output:
     metadata = {"raw_response": response}
     if response_includes_details:
         metadata["details"] = response.details
@@ -107,7 +107,7 @@ def construct_regular_output(response: TextGenerationResponse, response_includes
     output = ExecuteResult(
         **{
             "output_type": "execute_result",
-            "data": response.generated_text or "",
+            "data": response,
             "execution_count": 0,
             "metadata": metadata,
         }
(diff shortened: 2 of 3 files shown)


rholinshead (approved):
Nice, thanks for fixing

cookbooks/HuggingFace/hf.py
```diff
@@ -153,7 +153,9 @@ def __init__(self, model_id: str = None, use_api_token=False):
         token = None
 
         if use_api_token:
-            token = maybe_get_api_key_from_environment("HUGGING_FACE_API_TOKEN")
+            # You are allowed to use Hugging Face for a bit before you get
+            # rate limited, in which case you will receive a clear error
+            token = maybe_get_api_key_from_environment("HUGGING_FACE_API_TOKEN", required=False)
```
saqadri, line 158: why is required=False?

python/src/aiconfig/default_parsers/hf.py
```diff
@@ -120,7 +120,7 @@ class HuggingFaceTextGenerationParser(ParameterizedModelParser):
     A model parser for HuggingFace text generation models.
     """
 
-    def __init__(self, model_id: str = None, use_api_token=False):
+    def __init__(self, model_id: str = None, use_api_token=True):
```
saqadri, line 123: Why was this false before and why is it set to True now?

saqadri (approved):

cookbooks/HuggingFace/hf.py
rossdanlm: I left a comment, but it's because Hugging Face allows you to use their API for a bit before you get rate-limited. When you get rate limited, we get a very clear error from Hugging Face saying to login and use an API key

python/src/aiconfig/default_parsers/hf.py
rossdanlm: It's because after using it for a bit, you end up getting rate-limited, so this needs to be true so that you can by default fetch your API key

PR #772, opened 2024-01-05
Make `get_api_key_from_environment` return nullable

rossdanlm:
Make `get_api_key_from_environment` return nullable

Sometimes key is not defined or set, we need to do this to unblock https://github.com/lastmile-ai/aiconfig/pull/769

The functionality is still unchanged because the optional param `required` defaults to true

## Test Plan
Pass all automated tests (which it didn't before)

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/772).
* #769
* __->__ #772

### DIFF
--- a/python/src/aiconfig/util/config_utils.py
+++ b/python/src/aiconfig/util/config_utils.py
@@ -1,6 +1,7 @@
 import copy
+import dotenv
 import os
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Union
 
 if TYPE_CHECKING:
     pass
@@ -10,10 +11,28 @@ if TYPE_CHECKING:
     from ..schema import AIConfig
 
 
-def get_api_key_from_environment(api_key_name: str):
-    if api_key_name not in os.environ:
-        raise Exception("Missing API key '{}' in environment".format(api_key_name))
+def get_api_key_from_environment(
+    api_key_name: str,
+    required: bool = True) -> Union[str, None]:
+    """Get the API key if it exists, return None or error if it doesn't
 
+    Args:
+        api_key_name (str): The keyname that we're trying to import from env variable
+        required (bool, optional): If this is true, we raise an error if the 
+            key is not found
+
+    Returns:
+        Union[str, None]: the value of the key. If `required` is false, this can be None
+    """
+    dotenv.load_dotenv()
+    if required:
+        _get_api_key_from_environment_required(api_key_name)
+    return os.getenv(api_key_name)
+
+
+def _get_api_key_from_environment_required(api_key_name: str) -> str:
+    if api_key_name not in os.environ:
+        raise KeyError(f"Missing API key '{api_key_name}' in environment")
     return os.environ[api_key_name]
 
 
--- a/extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
+++ b/extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
@@ -1,10 +1,8 @@
 # Define a Model Parser for LLama-Guard
 from typing import TYPE_CHECKING, Dict, List, Optional, Any
 import copy
-import json
 
 import google.generativeai as genai
-from google.generativeai.types import content_types
 from google.protobuf.json_format import MessageToDict
 
 from aiconfig import (
@@ -108,6 +106,9 @@ class GeminiModelParser(ParameterizedModelParser):
     def __init__(self, id: str = "gemini-pro"):
         super().__init__()
         self.model = genai.GenerativeModel(id)
+        # TODO: Define API key here so we only call get_api_key_from_environment once
+        # Pass it in explicit to genai.configure(api_key=self.api_key), similar to 
+        # https://github.com/zhayujie/chatgpt-on-wechat/blob/eb809055d49efcc2036f7d89088aea7b50097162/bot/gemini/google_gemini_bot.py#L37
 
     def id(self) -> str:
         """
--- a/python/tests/test_util/test_config_util.py
+++ b/python/tests/test_util/test_config_util.py
@@ -1,7 +1,7 @@
 from aiconfig.util.config_utils import get_api_key_from_environment
 
 
-def test_get_api_key_from_environment():
+def test_maybe_get_api_key_from_environment():
     key = "TEST_API_KEY"
     try:
         get_api_key_from_environment(key)


python/src/aiconfig/util/config_utils.py
```diff
@@ -10,10 +11,28 @@
...
+
+    Args:
+        api_key_name (str): The keyname that we're trying to import from env variable
+        required (bool, optional): If this is true, we raise an error if the 
+            key is not found
+
+    Returns:
+        Union[str, None]: the value of the key. If `required` is false, this can be None
+    """
+    dotenv.load_dotenv()
```
rholinshead, line 27: Nice, I always thought it was strange that we had to manually use dotenv everywhere instead of just having it done once here. Is there a reason this wouldn't be wanted?

python/src/aiconfig/util/config_utils.py
```diff
@@ -10,10 +11,28 @@
     from ..schema import AIConfig
 
 
-def get_api_key_from_environment(api_key_name: str):
-    if api_key_name not in os.environ:
-        raise Exception("Missing API key '{}' in environment".format(api_key_name))
+def maybe_get_api_key_from_environment(
```
rholinshead, line 14: nit, generally not a fan of using 'maybe' in function names when it can be derived from type. And here it is a bit confusing because by default it actually has to either return a value or throw -- there's no 'maybe' case until you specify required false

rholinshead (approved):
Accepting to unblock but a few comments:
- just want to double-check it's ok to do the dotenv here for all cases (i.e. it won't cause issues if the user has no .env file or a .env file in different location or some other scenario like that?)
- consider changing the name to remove 'maybe' to reduce confusion around default behaviour

python/src/aiconfig/util/config_utils.py
saqadri: agreed

python/src/aiconfig/util/config_utils.py
saqadri: Wouldn't this load dotenv every time this function is called? Is that what we want?

python/src/aiconfig/util/config_utils.py
```diff
@@ -10,10 +11,28 @@
...
+        required (bool, optional): If this is true, we raise an error if the 
+            key is not found
+
+    Returns:
+        Union[str, None]: the value of the key. If `required` is false, this can be None
+    """
+    dotenv.load_dotenv()
+    if required:
+        _get_api_key_from_environment(api_key_name)
+    return os.getenv(api_key_name)
```
saqadri, line 30: Why does the other function do `os.environ[api_key_name]` but this uses `getenv`?

python/src/aiconfig/util/config_utils.py
saqadri, line 14: please rename back to original

saqadri (changes requested):
Requesting changed to understand if load dotenv is necessary each time.

python/src/aiconfig/util/config_utils.py
rossdanlm: `os.environ[api_key_name]` errors if the key isn't defined. `getenv` returns None

python/src/aiconfig/util/config_utils.py
rossdanlm: > Wouldn't this load dotenv every time this function is called?

Even if we did call this every time, I don't think it's an issue?

This is only called on the `init()` for the hugging face model parsers, and only if the api key is not set in OpenAI (which it will be after calling this function once): https://github.com/search?q=repo%3Alastmile-ai%2Faiconfig%20get_api_key_from_environment&type=code

I'll edit the Gemini callsite in this diff to only call this one time just like OpenAI: https://github.com/lastmile-ai/aiconfig/blob/5b28259ff44dd3afd03c40206888ec06774ccaed/extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py#L329

python/src/aiconfig/util/config_utils.py
jonathanlastmileai: Spoke to @rossdanlm . We're going to do optional here and a follow-up PR to switch to Result, because None is actually a valid value.

python/src/aiconfig/util/config_utils.py
```diff
@@ -10,10 +11,28 @@
...
 
+
+def _get_api_key_from_environment(api_key_name: str) -> str:
+    if api_key_name not in os.environ:
+        raise KeyError(f"Missing API key '{api_key_name}' in environment")
```
jonathanlastmileai, line 35: Don't raise and return Optional. You can just do this:

```
def get_api_key_from_environment(api_key_name: str) -> str | None:
    try:
         return os.environ[api_key_name]
    except Exception:
         # you want to do the logging here because you're about to lose the message forever by returning `None`.
         LOGGER.error(...)
          return None
```

jonathanlastmileai (changes requested):

python/src/aiconfig/util/config_utils.py
rossdanlm: ^Gemini fixed in https://github.com/lastmile-ai/aiconfig/pull/786

saqadri: Ok sounds good, thanks!

saqadri (approved):

rossdanlm:
Updated with comments to change the name. Did separate PRs for:
- Making Gemini model parser not have to keep checking the key every time it calls `run_inference`: https://github.com/lastmile-ai/aiconfig/pull/786
- Changing this to Result-OK format: https://github.com/lastmile-ai/aiconfig/pull/787

python/src/aiconfig/util/config_utils.py
```diff
@@ -10,10 +11,28 @@
     from ..schema import AIConfig
 
 
-def get_api_key_from_environment(api_key_name: str):
-    if api_key_name not in os.environ:
-        raise Exception("Missing API key '{}' in environment".format(api_key_name))
+def maybe_get_api_key_from_environment(
+    api_key_name: str,
+    required: bool = True) -> Union[str, None]:
```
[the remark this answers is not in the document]
jonathanlastmileai: I don't think the "required" logic belongs here. This function should be responsible for 1 thing: get it if it's there, otherwise return None. The required or not branch should happen above this and just not call this function.

Also please type this as str | None. Union[str, None] is definitely not the standard. Optional[str] is the old but reasonable way.

python/src/aiconfig/util/config_utils.py
```diff
@@ -10,10 +11,28 @@
...
+    """
+    dotenv.load_dotenv()
+    if required:
+        _get_api_key_from_environment_required(api_key_name)
+    return os.getenv(api_key_name)
 
+
+def _get_api_key_from_environment_required(api_key_name: str) -> str:
+    if api_key_name not in os.environ:
+        raise KeyError(f"Missing API key '{api_key_name}' in environment")
```
jonathanlastmileai, line 35: I'm confused why we're checking and then immediately reraising and not catching it. KeyError is just going to propagate

jonathanlastmileai (changes requested):

jonathanlastmileai (approved):
accepting to unblock assuming this fixes a P0 bug and is tested. Let's talk about error handling and required vs. not as a follow up

PR #777, opened 2024-01-05
[AiConfig Branded Editor]

zakariassoul:
[AiConfig Branded Editor]

I styled the local editor to match ai-config brand colors. This is just iteration 1...more to come


zakariassoul:
https://github.com/lastmile-ai/aiconfig/assets/17841936/8fbd4b8b-0eb5-41d5-a16f-89ba066b6143

cookbooks/Getting-Started/travel.aiconfig.json
rholinshead, line 1: Can you revert the changes to this config file please? We want to keep it as close to the getting started one as possible for testing

python/src/aiconfig/editor/client/src/Editor.tsx
```diff
@@ -141,8 +141,115 @@ export default function Editor() {
...
 
   return (
-    <div>
-      <MantineProvider withGlobalStyles withNormalizeCSS>
+    <div className="editorBackground">
+      <MantineProvider
+        withGlobalStyles
+        withNormalizeCSS
+        theme={{
+          colorScheme: "dark",
```
rholinshead, line 149: Shouldn't this be based on the user's system theme?

python/src/aiconfig/editor/client/src/components/prompt/ModelSelector.tsx
```diff
@@ -40,12 +40,15 @@ export default memo(function ModelSelector({
     <Autocomplete
       placeholder="Select model"
       limit={100}
+      className="ghost"
+      variant="unstyled"
```
rholinshead, line 44: Just noting that we've had user feedback about unstyled inputs before (discord comment about parameter name input which was previously unstyled in lastmile workbooks)

python/src/aiconfig/editor/client/src/components/prompt/RunPromptButton.tsx
```diff
@@ -8,28 +8,19 @@ type Props = {
...
-  const { classes } = useStyles();
   return (
     <Button
       onClick={runPrompt}
       disabled={isRunning}
       p="xs"
       size="xs"
       fullWidth={size === "full"}
-      className={classes.executeButton}
+      className="primaryPinkButton"
```
rholinshead, line 23: Can we make this `runPromptButton` ? It's going to be yellow in gradio, so having pink in the name is confusing

python/src/aiconfig/editor/client/src/Editor.tsx
```diff
@@ -141,8 +141,115 @@ export default function Editor() {
...
+
+          colors: {
+            // Add your color
+            deepBlue: ["#E9EDFC", "#C1CCF6", "#99ABF0" /* ... */],
+            // or replace default theme color
+            // gray: ["#E9EDFC", "#C1CCF6", "#99ABF0" /* ... */],
+            blue: ["#E9EDFC", "#C1CCF6", "#99ABF0" /* ... */],
```
rholinshead, line 156: Are these actually needed or just copied from the example?

These should be an array of 10 elements since mantine core could reference theme.colors.blue[8] for example and that will probably result in an unhandled error that could break rendering

zakariassoul: not needed. Will remove

python/src/aiconfig/editor/client/src/components/prompt/ModelSelector.tsx
zakariassoul: These ghost buttons behave differently. hover shows an input outline and cursor changes.

python/src/aiconfig/editor/client/src/Editor.tsx
zakariassoul: For starters We want to force the AiConfig Brand until i add the light mode colors.

cookbooks/Getting-Started/travel.aiconfig.json
zakariassoul: sure thing.

python/src/aiconfig/editor/client/src/components/prompt/RunPromptButton.tsx
rholinshead: We will use the same class name ('.runPromptButton') with different styles. This RunPromptButton code is going to be the same in gradio and local editor so we can't change the className on the component itself, we can only change the styles for that classname

rholinshead (approved):
Accepting to unblock once the following changes are made:
- update prompt run button class name to be more generic
- move aiconfig image up to root Editor.tsx component; this is the local editor-specific component. The <EditorContainer> component will be shared between local editor, gradio and readonly lastmile renderer

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -676,12 +685,20 @@ export default function EditorContainer({
           <Tooltip
             label={isDirty ? "Save changes to config" : "No unsaved changes"}
           >
-            <div>
+            <div className={classes.toolbar}>
+              <Image
```
rholinshead, line 689: Can you do this image/toolbar in the EditorContainer instead? That will solve 2 things:
- as is, this will have the AIConfig image show up in gradio when we use this component there
- should fix the grey bar issue

zakariassoul: Thanks, will amend a change.

zakariassoul: Done

python/src/aiconfig/editor/client/src/components/prompt/RunPromptButton.tsx
zakariassoul: great

PR #780, opened 2024-01-05
[extensions][py][hf] 2/n ASR model parser impl

Ankush-lastmile:
[extensions][py][hf] 2/n ASR model parser impl

Model Parser for the Automatic Speech Recognition task on huggingface.

Decisions made while implementing:
- manual impl to parse input attachments
- - threw exceptions on every unexpected step. Not sure if this is the direction we want to go with this.
- This diff does not implement serialize() for the model parser (will implement on diff ontop)

## Testplan

Created an mp3 file that says "hi". Used aiconfig to run asr on it.

|Screenshot 2024-01-09 at 7 14 47 PM

|Screenshot 2024-01-09 at 5 54 33 PM|

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/780).
* __->__ #780
* #779

### DIFF
--- a/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/__init__.py
+++ b/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/__init__.py
@@ -5,6 +5,7 @@ from .local_inference.text_generation import HuggingFaceTextGenerationTransforme
 from .local_inference.text_summarization import HuggingFaceTextSummarizationTransformer
 from .local_inference.text_translation import HuggingFaceTextTranslationTransformer
 from .remote_inference_client.text_generation import HuggingFaceTextGenerationParser
+from .local_inference.automatic_speech_recognition import HuggingFaceAutomaticSpeechRecognitionTransformer
 
 
 LOCAL_INFERENCE_CLASSES = [
@@ -15,6 +16,7 @@ LOCAL_INFERENCE_CLASSES = [
     "HuggingFaceText2SpeechTransformer",
     "HuggingFaceAutomaticSpeechRecognition",
     "HuggingFaceImage2TextTransformer",
+    "HuggingFaceAutomaticSpeechRecognitionTransformer",
 ]
 REMOTE_INFERENCE_CLASSES = ["HuggingFaceTextGenerationParser"]
 __ALL__ = LOCAL_INFERENCE_CLASSES + REMOTE_INFERENCE_CLASSES
--- a/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/automatic_speech_recognition.py
+++ b/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/automatic_speech_recognition.py
@@ -1,8 +1,11 @@
-from typing import Any, Coroutine, Dict, Optional, List, TYPE_CHECKING
-from aiconfig import ParameterizedModelParser, InferenceOptions, AIConfig
+from typing import Any, Dict, Literal, Optional, List, TYPE_CHECKING
+from aiconfig import ParameterizedModelParser, InferenceOptions
+from aiconfig.callback import CallbackEvent
+from pydantic import BaseModel
+import torch
+from aiconfig.schema import Prompt, Output, ExecuteResult, Attachment
 
-from aiconfig.schema import Prompt, Output
-from transformers import Pipeline
+from transformers import pipeline, Pipeline
 
 if TYPE_CHECKING:
     from aiconfig import AIConfigRuntime
@@ -11,7 +14,7 @@ Model Parser for HuggingFace ASR (Automatic Speech Recognition) models.
 """
 
 
-class HuggingFaceAutomaticSpeechRecognition(ParameterizedModelParser):
+class HuggingFaceAutomaticSpeechRecognitionTransformer(ParameterizedModelParser):
     def __init__(self):
         """
         Returns:
@@ -24,13 +27,13 @@ class HuggingFaceAutomaticSpeechRecognition(ParameterizedModelParser):
                 config.register_model_parser(parser)
         """
         super().__init__()
-        self.generators: dict[str, Pipeline] = {}
+        self.pipelines: dict[str, Pipeline] = {}
 
     def id(self) -> str:
         """
         Returns an identifier for the Model Parser
         """
-        return "HuggingFaceAutomaticSpeechRecognition"
+        return "HuggingFaceAutomaticSpeechRecognitionTransformer"
 
     async def serialize(
         self,
@@ -38,10 +41,10 @@ class HuggingFaceAutomaticSpeechRecognition(ParameterizedModelParser):
         data: Any,
         ai_config: "AIConfigRuntime",
         parameters: Optional[Dict[str, Any]] = None,
-        **completion_params,
     ) -> List[Prompt]:
         """
         Defines how a prompt and model inference settings get serialized in the .aiconfig.
+        Assume input in the form of input(s) being passed into an already constructed pipeline.
 
         Args:
             prompt (str): The prompt to be serialized.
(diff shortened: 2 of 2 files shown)


extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/automatic_speech_recognition.py
```diff
@@ -0,0 +1,152 @@
+from typing import Any, Dict, Optional, List, TYPE_CHECKING
+from aiconfig import ParameterizedModelParser, InferenceOptions
+from aiconfig.callback import CallbackEvent
+import torch
+from aiconfig.schema import Prompt, Output, ExecuteResult, Attachment
+
+from transformers import pipeline, Pipeline
+
+if TYPE_CHECKING:
+    from aiconfig import AIConfigRuntime
```
jonathanlastmileai, line 10: Totally separate discussion: do we have to do this?

jonathanlastmileai (approved):

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/__init__.py
```diff
@@ -1,8 +1,9 @@
 from .local_inference.text_2_image import HuggingFaceText2ImageDiffusor
 from .local_inference.text_generation import HuggingFaceTextGenerationTransformer
-from .remote_inference_client.text_generation import HuggingFaceTextGenerationClient
+# from .remote_inference_client.text_generation import HuggingFaceTextGenerationClient
```
saqadri, line 3: Remove?

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/automatic_speech_recognition.py
```diff
@@ -0,0 +1,152 @@
...
+    - no attachments are found
+    - operation fails for any reason
+    """
+
+    if not hasattr(prompt.input, "attachments") or len(prompt.input.attachments) == 0:
+        raise ValueError(f"No attachments found in input for prompt {prompt.name}. Please add an audio attachment to the prompt input.")
+
+    audio_uris: list[str] = []
+
+    for i, attachment in enumerate(prompt.input.attachments):
```
saqadri, line 143: Let's match the type of attachment to be the same as
 
```
class OutputDataWithStringValue(BaseModel):
    """
    This represents the output content that is storied as a string, but we use
    both the `kind` field here and the `mime_type` in ExecuteResult to convert
    the string into the output format we want.
    """

    kind: Literal["file_uri", "base64"]
    value: str
```

So instead of data being the uri itself, would be great for data to be {"kind": "file_uri", "value": "<uri>"}

We can add proper typings for attachments later, but having that type for this model parser will help us standardize on this format

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/__init__.py
rossdanlm: Why does this keep erroring or everyone lol?

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/automatic_speech_recognition.py
rossdanlm: For more context, you can see how we did it for (look up `kind=` in these file changes):
- image output: https://github.com/lastmile-ai/aiconfig/pull/608/commits/71749931e095a8cdf1e07e2432f36401f647cf7a
- tool call output: https://github.com/lastmile-ai/aiconfig/pull/611/files

The output schema is defined in https://github.com/lastmile-ai/aiconfig/blob/main/python/src/aiconfig/schema.py#L71-L74

Ankush-lastmile:
Added a couple of changes:

1. Defined `AttachmentInputDataWithStringValue`. This will need to be incorporated into the AIConfig sdk but for now we will define this inside this model parser and enforce it here. TODO: Task #

1.5. AIConfig for this will look slightly different. See updated testplan

2. Updated Model Parser to be able to customize pipeline creation. Note: for now this customizability is only possible the first time the pipeline is created in memory

3. Updated Model Parser completion params with certain subset

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/__init__.py
Ankush-lastmile: will remove after rebase

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/automatic_speech_recognition.py
Ankush-lastmile: from before, this was needed. Keeping it for now since we don't have automated tests running

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/automatic_speech_recognition.py
Ankush-lastmile: as per additional converstion offline, added custom input type `AttachmentInputDataWithStringValue` inside this model parser. Enforces file attachments to be of this type containing a `kind` value. 

Added todo's to cleanup and add to aiconfig sdk

Ankush-lastmile:
rebase, and fix init.py

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/automatic_speech_recognition.py
```diff
@@ -52,14 +55,224 @@ async def serialize(
...
+            # TODO: @ankush-lastmile update AIConfig sdk schema to include this type (can also remove this try catch once done)
+            _ = AttachmentInputDataWithStringValue(**attachment.data)
+        except Exception as e:
+            # pylint: disable=W0707
+            raise ValueError(
+                f"Attachment #{i} for prompt {prompt.name} is not a valid attachment for Automatic Speech Recognition model parser. "
+                f"Error: {str(e)}. Please specify a kind and value for the attachment data."
+            )
+        # TODO: once previous todo gets resolved, modify this line to `.kind`` instead of .data.get("kind"). It will be a pydantic base class.
+        if not attachment.data.get("kind") == "file_uri":
```
rholinshead, line 165: We'll need to update the client rendering and prompt schema format to support this. And we'll need to be consistent with this type in the image-to-text parser as well

rholinshead: Can we just have this be a string and migrate to the object with kind and value afterwards? My concerns here are:
- client rendering is currently adding attachments as {data: <url>, mime_type}
- image-to-text parser is currently not doing this {kind/value} stuff and that inconsistency is going to cause a lot more work on the client

Ankush-lastmile: discussed offline with @rholinshead, reverted change to input data type. For ease of use we will have data as just a string and implement structured input later

Ankush-lastmile:
discussed offline with @rholinshead, reverted change to input data type. For ease of use we will have data as just a string and implement structured input later

updated testplan to show as well

rholinshead (approved):

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/automatic_speech_recognition.py
```diff
@@ -0,0 +1,257 @@
...
+
+        prompt.outputs = outputs
+        await aiconfig.callback_manager.run_callbacks(CallbackEvent("on_run_complete", __name__, {"result": prompt.outputs}))
+        return prompt.outputs
+
+    def _get_device(self) -> str:
+        if torch.cuda.is_available():
+            return "cuda"
+        # Mps backend is not supported for all asr models. Seen when spinning up a default asr pipeline which uses facebook/wav2vec2-base-960h 55bb623
+        return "cpu"
```
rossdanlm, line 121: We should use mac if we have it. See `_get_device` from text_2_image.py

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/automatic_speech_recognition.py
```diff
@@ -0,0 +1,257 @@
...
+
+    def get_output_text(self, response: dict[str, Any]) -> str:
+        raise NotImplementedError("get_output_text is not implemented for HuggingFaceAutomaticSpeechRecognition")
```
rossdanlm, line 124: Why are we not implementing?

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/automatic_speech_recognition.py
```diff
@@ -0,0 +1,257 @@
...
+    Note: This doesn't support base pipeline params like `num_workers`
+    TODO: Figure out how to find which params are supported.
+    """
+
+    supported_keys = {
+        # inputs
+        "return_timestamps",
+        "generate_kwargs",
+        "max_new_tokens",
+    }
```
rossdanlm, line 222: There's actually way more because we're using `generate_kwargs` which takes in from the generalized text generation params. Pls add this to the task for unifying this in prompt schema, and link to task here

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/automatic_speech_recognition.py
```diff
@@ -0,0 +1,257 @@
...
+            completion_params[key.lower()] = unfiltered_completion_params[key]
+
+    return completion_params
+
+
+def construct_outputs(response: list[Any]) -> list[Output]:
+    """
+    Constructs an output from the response of the HF ASR pipeline.
+
+    Response from pipeline could contain multiple outputs and time stamps
```
rossdanlm, line 236: Can you provide link to API so it's easier to tell what examples they can be?

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/automatic_speech_recognition.py
```diff
@@ -0,0 +1,257 @@
...
+    """
+    outputs: list[Output] = []
+
+    if not isinstance(response, list):
+        # response contains a single output. Found by testing variations of the asr pipeline
+        response = [response]
+
+    for i, result in enumerate(response):
+        # response is expected to be a dict containing the text output and timestamps if specified. Could not find docs for this.
```
rossdanlm, line 245: rip, ok pls mention can't find docs in the method docstrings

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/automatic_speech_recognition.py
```diff
@@ -0,0 +1,257 @@
...
+        result: dict[str, Any]
+        output = ExecuteResult(
+            **{
+                "output_type": "execute_result",
+                "data": result.get("text"),
```
rossdanlm, line 250: So response is always guaranteed to have the text attribute? I think it would be safe just to have another line above saying `text_output: str = result.get("text") if "text" in result and isinstance(result, dict) else result` or whatever just to future proof

rossdanlm: The problem is without knowing whta `result` type is, we can't know for sure and better to be safe just to ensure it can't break

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/automatic_speech_recognition.py
```diff
@@ -0,0 +1,257 @@
...
+
+        model_settings = self.get_model_settings(prompt, aiconfig)
+        [pipeline_creation_data, _] = refine_pipeline_creation_params(model_settings)
+        model_name = aiconfig.get_model_name(prompt)
+
+        if isinstance(model_name, str) and model_name not in self.pipelines:
+            device = self._get_device()
+            if pipeline_creation_data.get("device", None) is None:
+                pipeline_creation_data["device"] = device
+            self.pipelines[model_name] = pipeline(task="automatic-speech-recognition", **pipeline_creation_data)
```
rossdanlm, line 103: nit: no change, just annoying that they called this `automatic-speech-recognition` instead of `audio-to-text` like everything else they did smh

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/automatic_speech_recognition.py
```diff
@@ -0,0 +1,257 @@
...
+from transformers import pipeline, Pipeline
+
+if TYPE_CHECKING:
+    from aiconfig import AIConfigRuntime
+"""
+Model Parser for HuggingFace ASR (Automatic Speech Recognition) models.
+"""
+
+
+class HuggingFaceAutomaticSpeechRecognition(ParameterizedModelParser):
```
rossdanlm, line 17: nit: add "Transformer" suffix at end to meet same format as the others

rossdanlm (approved):
pls link relevant comments to relevant issues so we can track in the future

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/automatic_speech_recognition.py
Ankush-lastmile: added implementation

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/automatic_speech_recognition.py
Ankush-lastmile: updated

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/automatic_speech_recognition.py
Ankush-lastmile: updated

Ankush-lastmile:
- added some comments
- renamed to `HuggingFaceAutomaticSpeechRecognitionTransformer` from `HuggingFaceAutomaticSpeechRecognitionTransformer`
- added get output text impl (copied from other model parser impls

Ankush-lastmile:
rebase

Issue #782, opened 2024-01-05
Have a consistent way to distinguish model parsers from models


saqadri:
A single ModelParser can be used to run inference for multiple models. For example, the `AnyscaleEndpoint` model parser (https://github.com/lastmile-ai/aiconfig/blob/main/python/src/aiconfig/default_parsers/anyscale_endpoint.py) supports inference for LLaMA, Mistral, CodeLLaMA, LLaMAGuard, and other models.

Similarly, Hugging Face task model parsers can each support inference for thousands of models.

In aiconfig today, it's hard to distinguish between models and model parsers. A "model parser" is really a "model inference provider" for all models that can be run in the same way.

For example, the following prompt is valid:
```
  {
      "name": "gen_itinerary",
      "input": "Generate an itinerary ordered by {{order_by}} for these activities: {{get_activities.output}}.",
      "metadata": {
        "model": {
          "name": "AnyscaleEndpoint",
          "settings": {
            "model": "mistralai/Mixtral-8x7B-Instruct-v0.1",
            "max_tokens": 3000,
            "system_prompt": "You are an expert travel coordinator with exquisite taste."
          }
        },
        "parameters": {
          "order_by": "geographic location"
        }
      }
    }
```

But, if we add a `model_parsers` mapping to the global metadata
```
    "model_parsers": {
      "meta-llama/Llama-2-7b-chat-hf": "AnyscaleEndpoint",
   }
```

then the same prompt can be expressed as:
```
    {
      "name": "gen_itinerary",
      "input": "Generate an itinerary ordered by {{order_by}} for these activities: {{get_activities.output}}.",
      "metadata": {
        "model": {
          "name": "mistralai/Mixtral-8x7B-Instruct-v0.1",
          "settings": {
            "model": "mistralai/Mixtral-8x7B-Instruct-v0.1",
            "max_tokens": 3000,
            "system_prompt": "You are an expert travel coordinator with exquisite taste."
          }
        },
        "parameters": {
          "order_by": "geographic location"
        }
      }
    }
```

Some things we need to clean up and simplify:
1. confusion between `model.name` and `model.settings.model` -- they are often duplicated for no good reason.
2. having cleaner distinguishability between the concept of "model parser" and the "model id"

rholinshead:
A couple more thoughts here:
- Ideally we could also just express as
```
{
      "name": "gen_itinerary",
[...]

PR #783, opened 2024-01-05
Explicitly set "model" property for models that support it

saqadri:
Explicitly set "model" property for models that support it

This change allows us to automatically set the "model" for the following model parsers:
* OpenAI
* Anyscale Endpoint
* Hugging Face text generation
* PaLM
* Dall-E

Without this change the local editor breaks because if a user doesn't explicitly specify the "model" property in settings metadata, inference fails for (some of) the above model parsers. This isn't ideal, because the user has already selected the model they want to use from the model selector, so shouldn't need to additionally specify a model id as well.

There is still a larger issue of how we distinguish a model ID from a model parser ID. Filed #782 to come up with a proper fix for that.

### DIFF
--- a/cookbooks/HuggingFace/hf.py
+++ b/cookbooks/HuggingFace/hf.py
@@ -209,7 +209,7 @@ class HuggingFaceTextParser(ParameterizedModelParser):
         Returns:
             dict: Model-specific completion parameters.
         """
-        resolved_prompt = resolve_prompt(prompt, params, aiconfig)
+        resolved_prompt = resolve_prompt(prompt, params if params is not None else {}, aiconfig)
 
         # Build Completion data
         model_settings = self.get_model_settings(prompt, aiconfig)
--- a/python/src/aiconfig/model_parser.py
+++ b/python/src/aiconfig/model_parser.py
@@ -175,10 +175,10 @@ class ModelParser(ABC):
             # Merge config and prompt settings with prompt settings taking precedent
             model_settings = {}
             global_settings = aiconfig.get_global_settings(model_metadata.name)
-            prompt_setings = prompt.metadata.model.settings if prompt.metadata.model.settings is not None else {}
+            prompt_settings = prompt.metadata.model.settings if prompt.metadata.model.settings is not None else {}
 
             model_settings.update(global_settings)
-            model_settings.update(prompt_setings)
+            model_settings.update(prompt_settings)
 
             return model_settings
 
--- a/python/src/aiconfig/default_parsers/dalle.py
+++ b/python/src/aiconfig/default_parsers/dalle.py
@@ -25,7 +25,7 @@ if TYPE_CHECKING:
 
 
 # TODO: Centralize this into utils function with the HuggingFaceTextParser class
-def refine_image_completion_params(model_settings):
+def refine_image_completion_params(model_settings, aiconfig, prompt):
     """
     Refines the completion params for the Dall-E request API. Removes any unsupported params.
     The supported keys were found by looking at the OpenAI Dall-E API: https://platform.openai.com/docs/api-reference/images/create?lang=python`
@@ -36,6 +36,11 @@ def refine_image_completion_params(model_settings):
     for key in model_settings:
         if key.lower() in supported_keys:
             completion_data[key.lower()] = model_settings[key]
+
+    # Explicitly set the model to use if not already specified
+    if completion_data.get("model") is None:
+        model_name = aiconfig.get_model_name(prompt)
+        completion_data["model"] = model_name
 
     return completion_data
 
@@ -141,11 +146,11 @@ class DalleImageGenerationParser(ParameterizedModelParser):
             dict: Model-specific completion parameters.
         """
         # Get inputs from aiconfig
-        resolved_prompt = resolve_prompt(prompt, params, aiconfig)
+        resolved_prompt = resolve_prompt(prompt, params if params is not None else {}, aiconfig)
         model_settings = self.get_model_settings(prompt, aiconfig)
 
         # Build Completion data
-        completion_data = refine_image_completion_params(model_settings)
+        completion_data = refine_image_completion_params(model_settings, aiconfig, prompt)
         completion_data["prompt"] = resolved_prompt
         return completion_data
 
--- a/python/src/aiconfig/default_parsers/openai.py
+++ b/python/src/aiconfig/default_parsers/openai.py
@@ -160,7 +160,7 @@ class OpenAIInference(ParameterizedModelParser):
         # Build Completion params
         model_settings = self.get_model_settings(prompt, aiconfig)
 
-        completion_params = refine_chat_completion_params(model_settings)
+        completion_params = refine_chat_completion_params(model_settings, aiconfig, prompt)
 
         # In the case thhat the messages array weren't saves as part of the model settings, build it here. Messages array is used for conversation history.
         if not completion_params.get("messages"):
@@ -407,10 +407,10 @@ def multi_choice_message_reducer(messages: Union[Dict[int, dict], None], chunk: 
     return messages
 
 
-def refine_chat_completion_params(model_settings):
+def refine_chat_completion_params(model_settings, aiconfig, prompt):
     # completion parameters to be used for openai's chat completion api
     # system prompt handled separately
-    # streaming handled seperatly.
+    # streaming handled separately.
     # Messages array built dynamically and handled separately
     supported_keys = {
         "frequency_penalty",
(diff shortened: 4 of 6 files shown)


cookbooks/HuggingFace/hf.py
```diff
@@ -60,6 +60,11 @@ def refine_chat_completion_params(model_settings):
         if key.lower() in supported_keys:
             completion_data[key.lower()] = model_settings[key]
 
+    # Explicitly set the model to use if not already specified
+    if completion_data.get("model") is None:
+        model_name = aiconfig.get_model_name(prompt)
+        completion_data["model"] = model_name
```
saqadri, line 66: @rholinshead any thoughts on this? I feel like this should be left empty for HF unless explicitly specified? Thoughts?

cookbooks/HuggingFace/hf.py
```diff
@@ -219,20 +215,18 @@ async def deserialize(
         Returns:
             dict: Model-specific completion parameters.
         """
-        resolved_prompt = resolve_prompt(prompt, params, aiconfig)
+        resolved_prompt = resolve_prompt(prompt, params if params is not None else {}, aiconfig)
```
saqadri, line 218: This was a pyright error

pyrightconfig.json
saqadri, line 1: Will revert

cookbooks/HuggingFace/hf.py
rholinshead: Agreed, for hf inference API the model is optional so we should only set it if explicitly set

saqadri: So this is safe to remove for this model parser, right?

rossdanlm:
This feels really weird to make the model parsers do this instead of adding it as a centralized SDK functionality. See comments in here for how I think we should do it: https://github.com/lastmile-ai/aiconfig/pull/765#issuecomment-1878297875

Do you want me to take over this task? I can do it

saqadri:
> This feels really weird to make the model parsers do this instead of adding it as a centralized SDK functionality. See comments in here for how I think we should do it: [#765 (comment)](https://github.com/lastmile-ai/aiconfig/pull/765#issuecomment-1878297875)
> 
> Do you want me to take over this task? I can do it

Chatted offline with @rossdanlm, we are good to go with this change. Different model parsers can have different properties so it isn't possible to have a centralized place to do this.

jonathanlastmileai:
I think this is **probably** the right solution currently, just need to make sure it's well tested.

Note that this takes the stance that an AIConfig can be **valid** even with a "dangling foreign key" to the global model mapping, and it's the parser's responsibility to fill in the gap. IMO it would be better to constrain AIConfigs to not have dangling pointers, but I don't know if that can actually be enforced currently via the CRUD API.

cookbooks/HuggingFace/hf.py
rholinshead: Yes

python/src/aiconfig/model_parser.py
```diff
@@ -174,11 +174,13 @@ def get_model_settings(self, prompt: Prompt, aiconfig: "AIConfigRuntime") -> Dic
         else:
             # Merge config and prompt settings with prompt settings taking precedent
             model_settings = {}
-            global_settings = aiconfig.get_global_settings(model_metadata.name)
-            prompt_setings = prompt.metadata.model.settings if prompt.metadata.model.settings is not None else {}
+            prompt_settings = prompt.metadata.model.settings if prompt.metadata.model.settings is not None else {}
+
+            model_name = prompt_settings.get("model", model_metadata.name)
+            global_settings = aiconfig.get_global_settings(model_name)
```
rholinshead, line 180: Hmm, is this required as part of the change? I don't think there's any requirement (currently) that the `model` in the arbitrary model settings has any correlation with global metadata, unlike the prompt metadata model

rholinshead (changes requested):
Parser changes look good but I think we should refrain from pulling 'model' from model.settings as part of the sdk since there's no guarantee it's related to any model parser; it could technically be some other meaning

python/src/aiconfig/model_parser.py
Ankush-lastmile: is this the only blocking change requested?

python/src/aiconfig/schema.py
```diff
@@ -289,7 +289,8 @@ def get_model_name(self, prompt: Union[str, Prompt]) -> str:
             return prompt.metadata.model
         else:
             # Expect a ModelMetadata object
-            return prompt.metadata.model.name
+            prompt_settings = prompt.metadata.model.settings if prompt.metadata.model.settings is not None else {}
+            return prompt_settings.get("model", prompt.metadata.model.name)
```
rholinshead, line 293: same here

python/src/aiconfig/model_parser.py
saqadri: I can make this change in a separate diff, but the motivation for this change was this example aiconfig:

```
{
  "name": "NYC Trip Planner",
  "description": "Intrepid explorer with ChatGPT and AIConfig",
  "schema_version": "latest",
  "metadata": {
    "models": {
      "mistralai/Mixtral-8x7B-Instruct-v0.1": {
        "top_p": 1,
        "temperature": 1
        "max_tokens": 3000,
        "system_prompt": "You are an expert travel coordinator with exquisite taste."
      }
    },
    "default_model": "meta-llama/Llama-2-7b-chat-hf"
  },
  "prompts": [
    {
      "name": "gen_itinerary",
      "input": "Generate an itinerary ordered by {{order_by}} for some activities.",
      "metadata": {
        "model": {
          "name": "AnyscaleEndpoint",
          "settings": {
            "model": "mistralai/Mixtral-8x7B-Instruct-v0.1",
            "max_tokens": 3000,
            "system_prompt": "You are an expert travel coordinator with exquisite taste."
          }
        },
        "parameters": {
          "order_by": "geographic location"
        }
      }
    }
  ]
}
```

In this example, we won't be able to read any of the global model settings for the `gen_itinerary` prompt.

Alternative way of expressing the same config would be:
```
{
  "name": "NYC Trip Planner",
  "description": "Intrepid explorer with ChatGPT and AIConfig",
  "schema_version": "latest",
  "metadata": {
    "model_parsers": {
      "mistralai/Mixtral-8x7B-Instruct-v0.1": "AnyscaleEndpoint"
    },
    "models": {
      "mistralai/Mixtral-8x7B-Instruct-v0.1": {
        "top_p": 1,
        "temperature": 1
        "max_tokens": 3000,
        "system_prompt": "You are an expert travel coordinator with exquisite taste."
      }
    },
    "default_model": "meta-llama/Llama-2-7b-chat-hf"
  },
  "prompts": [
    {
      "name": "gen_itinerary",
      "input": "Generate an itinerary ordered by {{order_by}} for some activities.",
      "metadata": {
        "model": {
          "name": "mistralai/Mixtral-8x7B-Instruct-v0.1",
          "settings": {
            "max_tokens": 3000,
            "system_prompt": "You are an expert travel coordinator with exquisite taste."
          }
        },
        "parameters": {
          "order_by": "geographic location"
        }
      }
    }
  ]
}
```

If we are fine with supporting the latter and not the former, then we can keep this code simple and as it was before.

Thoughts @rholinshead @Ankush-lastmile?

python/src/aiconfig/schema.py
saqadri: Replied to comment above. Reverting this for this diff, but I don't think global settings will work in the UI with the current approach.

model name will be set to something like "AnyscaleEndpoint", and there are no corresponding model settings for it. That's fine for now if all prompt settings are going to be in the prompt itself, but we'll need to figure this out.

saqadri: Hmmm, actually this change is needed

python/src/aiconfig/model_parser.py
rholinshead: IMO the latter is correct and the former I actually wouldn't expect to work since the AnyscaleEndpoint 'model' name doesn't match the global settings model -- it's essentially like using a model that hasn't been registered, right?

python/src/aiconfig/schema.py
rholinshead: For the parsers that have 'model' as a setting right now we already expose 'model' in the settings drawer to be set and that takes precedence over the model.name

saqadri: Ah I see -- so in that case, would we add model_parsers mapping for those models? Or do we expect that registration to have happened outside already?

saqadri:
Test plan: 

https://github.com/lastmile-ai/aiconfig/assets/25641935/32975ac4-096e-46df-8e03-51a0f2b93dae

Also tested with dalle3 (not shown)

rholinshead (approved):
LGTM

PR #786, opened 2024-01-05
Set the Gemini api key in init instead of re-running every inference call

rossdanlm:
Set the Gemini api key in init instead of re-running every inference call

See comments from https://github.com/lastmile-ai/aiconfig/pull/772#discussion_r1443387445

## Test Plan
Run the Gemini cookbook to make sure it still works

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/786).
* #787
* __->__ #786
* #772

### DIFF
--- a/extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
+++ b/extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
@@ -1,10 +1,8 @@
 # Define a Model Parser for LLama-Guard
 from typing import TYPE_CHECKING, Dict, List, Optional, Any
 import copy
-import json
 
 import google.generativeai as genai
-from google.generativeai.types import content_types
 from google.protobuf.json_format import MessageToDict
 
 from aiconfig import (
@@ -108,6 +106,9 @@ class GeminiModelParser(ParameterizedModelParser):
     def __init__(self, id: str = "gemini-pro"):
         super().__init__()
         self.model = genai.GenerativeModel(id)
+        # TODO: Define API key here so we only call get_api_key_from_environment once
+        # Pass it in explicit to genai.configure(api_key=self.api_key), similar to 
+        # https://github.com/zhayujie/chatgpt-on-wechat/blob/eb809055d49efcc2036f7d89088aea7b50097162/bot/gemini/google_gemini_bot.py#L37
 
     def id(self) -> str:
         """
--- a/extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
+++ b/extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
@@ -106,9 +106,12 @@ class GeminiModelParser(ParameterizedModelParser):
     def __init__(self, id: str = "gemini-pro"):
         super().__init__()
         self.model = genai.GenerativeModel(id)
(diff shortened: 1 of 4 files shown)


Ankush-lastmile:
> in init instead of re-running every inference call

I followed the comment chain but I wasn't able to understand why this change is needed.

Checking for api key inside init will not allow the user to execute serialize and deserialize if they don't have an api key. This was something we wanted to allow previously

extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
```diff
@@ -109,6 +107,12 @@ def __init__(self, id: str = "gemini-pro"):
         super().__init__()
         self.model = genai.GenerativeModel(id)
 
+        # Note: we don't need to explicitly set the key as long as this is set
+        # as an env var genai.configure() will pick up the env var
+        # `GOOGLE_API_KEY`, it's just that we prefer not to call
+        # `get_api_key_from_environment` multiple times if we don't need to
+        self.api_key = get_api_key_from_environment("GOOGLE_API_KEY", required=False)
+
```
saqadri, line 115: FYI, this was intentionally done in `run_inference`. Reason: The other functions (serialize, deserialize) don't need the API key.

Since you've set required=False, this is fine.

python/tests/test_util/test_config_util.py
```diff
@@ -1,7 +1,7 @@
 from aiconfig.util.config_utils import get_api_key_from_environment
 
 
-def test_get_api_key_from_environment():
+def test_maybe_get_api_key_from_environment():
```
saqadri, line 4:
```suggestion
def test_get_api_key_from_environment():
```

python/src/aiconfig/util/config_utils.py
```diff
@@ -10,10 +11,28 @@
...
+
+    Args:
+        api_key_name (str): The keyname that we're trying to import from env variable
+        required (bool, optional): If this is true, we raise an error if the 
+            key is not found
+
+    Returns:
+        Union[str, None]: the value of the key. If `required` is false, this can be None
+    """
+    dotenv.load_dotenv()
```
saqadri, line 27: Not sure if I had this comment elsewhere or forgot to publish, but why is this called every time instead of once?

saqadri (approved):

python/src/aiconfig/util/config_utils.py
saqadri: Nvm, I see it in the other diff. Thanks!

python/tests/test_util/test_config_util.py
rossdanlm: I just deleted the entire file because it wasn't being used

extensions/Gemini/python/src/aiconfig_extension_gemini/Gemini.py
rossdanlm: Thanks for context. I think this is another reason why having this optional `required` param is helpful!

rossdanlm: I'll leave a comment in code!

PR #787, opened 2024-01-05
Make `get_api_key_from_environment` use Result-Ok

rossdanlm:
Make `get_api_key_from_environment` use Result-Ok

Nice, cool tips from Jonathan

Also deleted `test_config_util.py` because it wasn't being referenced anywhere

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/787).
* __->__ #787
* #786
* #772


jonathanlastmileai (approved):
accepting to unblock assuming this fixes a P0 bug and is tested. Let's talk about error handling and required vs. not as a follow up

PR #789, opened 2024-01-05
Handle cancellation of tasks

saqadri:
Handle cancellation of tasks

Still WIP

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
@@ -25,6 +25,7 @@ import {
   useRef,
   useState,
 } from "react";
+import { v4 as uuidv4 } from "uuid";
 import aiconfigReducer, { AIConfigReducerAction } from "./aiconfigReducer";
 import {
   ClientPrompt,
@@ -65,7 +66,10 @@ export type AIConfigCallbacks = {
   deletePrompt: (promptName: string) => Promise<void>;
   getModels: (search: string) => Promise<string[]>;
   getServerStatus?: () => Promise<{ status: "OK" | "ERROR" }>;
-  runPrompt: (promptName: string) => Promise<{ aiconfig: AIConfig }>;
+  runPrompt: (
+    promptName: string,
+    cancellationToken?: string
+  ) => Promise<{ aiconfig: AIConfig }>;
   save: (aiconfig: AIConfig) => Promise<void>;
   setConfigDescription: (description: string) => Promise<void>;
   setConfigName: (name: string) => Promise<void>;
@@ -528,9 +532,11 @@ export default function EditorContainer({
   const runPromptCallback = callbacks.runPrompt;
   const onRunPrompt = useCallback(
     async (promptId: string) => {
+      const cancellationToken = uuidv4();
       const action: AIConfigReducerAction = {
         type: "RUN_PROMPT",
         id: promptId,
+        cancellationToken,
       };
 
       dispatch(action);
@@ -541,7 +547,10 @@ export default function EditorContainer({
           throw new Error(`Could not find prompt with id ${promptId}`);
         }
         const promptName = statePrompt.name;
-        const serverConfigRes = await runPromptCallback(promptName);
+        const serverConfigRes = await runPromptCallback(
+          promptName,
+          cancellationToken
+        );
 
         dispatch({
           type: "CONSOLIDATE_AICONFIG",
--- a/python/src/aiconfig/editor/server/server.py
+++ b/python/src/aiconfig/editor/server/server.py
@@ -1,5 +1,8 @@
+import asyncio
+import copy
 import logging
 import webbrowser
+import uuid
 from typing import Any, Dict, Type, Union
 
 import lastmile_utils.lib.core.api as core_utils
@@ -65,7 +68,7 @@ def run_backend_server(edit_config: EditServerConfig) -> Result[str, str]:
             LOGGER.info("Initialized server state")
             debug = edit_config.server_mode in [ServerMode.DEBUG_BACKEND, ServerMode.DEBUG_SERVERS]
             LOGGER.info(f"Running in {edit_config.server_mode} mode")
-            app.run(port=edit_config.server_port, debug=debug, use_reloader=debug)
+            app.run(port=edit_config.server_port, debug=debug, use_reloader=debug, threaded=True)
             return Ok("Done")
         case Err(e):
             LOGGER.error(f"Failed to initialize server state: {e}")
@@ -177,8 +180,14 @@ async def run() -> FlaskResponse:
     state = get_server_state(app)
     aiconfig = state.aiconfig
     request_json = request.get_json()
+    cancellation_token_id: Union[str, None] = None
+    aiconfig_deep_copy: Union[AIConfigRuntime, None] = None
 
     try:
+        cancellation_token_id = request_json.get("cancellation_token_id")
+        if not cancellation_token_id:
+            cancellation_token_id = str(uuid.uuid4())
+
         prompt_name: Union[str, None] = request_json.get("prompt_name")
         if prompt_name is None:
             return HttpResponseWithAIConfig(
@@ -194,7 +203,21 @@ async def run() -> FlaskResponse:
         params = request_json.get("params", aiconfig.get_parameters(prompt_name))  # type: ignore
         stream = request_json.get("stream", False)
         options = InferenceOptions(stream=stream)
-        run_output = await aiconfig.run(prompt_name, params, options)  # type: ignore
+
+        # Deepcopy the aiconfig prior to run so we can restore it in the case the run operation is cancelled or encounters some error
+        aiconfig_deep_copy = copy.deepcopy(aiconfig)
+
+        # Create a task to run the prompt, and associate the task with the cancellation token
+        # so that we can cancel the task if /cancel is invoked
+        task = asyncio.create_task(aiconfig.run(prompt_name, params, options))  # type: ignore
+        state.tasks[cancellation_token_id] = task  # type: ignore
+
+        # Wait for the task to complete, or for the task to be cancelled
+        run_output = await task  # type: ignore
+
+        # Remove the task from the cancellation token map if the task is complete
+        state.tasks.pop(cancellation_token_id, None)
+
         LOGGER.debug(f"run_output: {run_output}")
         return HttpResponseWithAIConfig(
             #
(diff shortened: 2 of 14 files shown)


python/src/aiconfig/editor/server/server.py
```diff
@@ -194,15 +203,42 @@ async def run() -> FlaskResponse:
...
+
+        # Return a 499 Client Closed Request status code to indicate that the client cancelled the request
+        return HttpResponseWithAIConfig(
+            message="The task was cancelled.",
+            payload={"cancellation_token_id": cancellation_token_id},
+            code=499,
+            aiconfig=state.aiconfig,
+        ).to_flask_format()
     except Exception as e:
+        # TODO: saqadri - should we restore the aiconfig to aiconfig_deep_copy in the case of an exception?
```
jonathanlastmileai, line 240: @saqadri , FYI https://en.wikipedia.org/wiki/Memento_pattern#:~:text=The%20memento%20pattern%20is%20implemented,originator%20for%20a%20memento%20object.

python/src/aiconfig/editor/server/server.py
```diff
@@ -194,15 +203,42 @@ async def run() -> FlaskResponse:
...
         stream = request_json.get("stream", False)
         options = InferenceOptions(stream=stream)
-        run_output = await aiconfig.run(prompt_name, params, options)  # type: ignore
+
+        # Deepcopy the aiconfig prior to run so we can restore it in the case the run operation is cancelled or encounters some error
+        aiconfig_deep_copy = copy.deepcopy(aiconfig)
+
+        # Create a task to run the prompt, and associate the task with the cancellation token
+        # so that we can cancel the task if /cancel is invoked
+        task = asyncio.create_task(aiconfig.run(prompt_name, params, options))  # type: ignore
```
jonathanlastmileai, line 212: I don't think this is going to work unless `run()` periodically yields control through an await, which it does not. I think we have to go heavy-handed and use threads.

python/src/aiconfig/editor/server/server.py
```diff
@@ -177,8 +180,14 @@ async def run() -> FlaskResponse:
     state = get_server_state(app)
     aiconfig = state.aiconfig
     request_json = request.get_json()
+    cancellation_token_id: Union[str, None] = None
```
jonathanlastmileai, line 183: nit: str | None or the older Optional[str] is ok. Union[str, None] is nonstandard

python/src/aiconfig/editor/server/server.py
```diff
@@ -177,8 +180,14 @@ async def run() -> FlaskResponse:
...
+    aiconfig_deep_copy: Union[AIConfigRuntime, None] = None
```
jonathanlastmileai, line 184: I'm a little worried this is going to make the state even harder to reason about. We already have the file, the in-memory copy, and the frontend copy. This becomes an even bigger concern if we're allowing concurrent mutation requests (are we doing that here as a consequence of allowing cancellation?).

python/src/aiconfig/editor/server/server.py
```diff
@@ -65,7 +68,7 @@ def run_backend_server(edit_config: EditServerConfig) -> Result[str, str]:
             LOGGER.info("Initialized server state")
             debug = edit_config.server_mode in [ServerMode.DEBUG_BACKEND, ServerMode.DEBUG_SERVERS]
             LOGGER.info(f"Running in {edit_config.server_mode} mode")
-            app.run(port=edit_config.server_port, debug=debug, use_reloader=debug)
+            app.run(port=edit_config.server_port, debug=debug, use_reloader=debug, threaded=True)
```
jonathanlastmileai, line 71: If the intention is to now allow concurrent requests, we should be exclusive-locking the aiconfig before doing any mutation. Otherwise you can get inconsistent states because the API does not support concurrent operations.

python/src/aiconfig/editor/server/server.py
```diff
@@ -211,6 +247,45 @@ async def run() -> FlaskResponse:
...
+def cancel() -> FlaskResponse:
+    state = get_server_state(app)
+    request_json = request.get_json()
+
+    cancellation_token_id: Union[str, None] = request_json.get("cancellation_token_id")
+    if cancellation_token_id is not None:
+        task = state.tasks.get(cancellation_token_id)
+        if task is not None:
+            # TODO: saqadri - specify a custom message here if needed
+            task.cancel()
```
jonathanlastmileai, line 260: I don't think this is going to work unless the task being canceled awaits frequently. In order to actually yield control back, it has to await something.

jonathanlastmileai (commented):
Just a heads up, this kind of thing is really tricky to get right. The state management is already very complex and bug-prone because HTTP is stateless and we're hacking global state without any synchronization.

On top of that, we are already non-robustly dealing with reading and writing to disk. For example, two concurrent writes to the file could corrupt it.

I think (?) we are currently somewhat protected against concurrent mutation by flask's single-threadedness, and multithreading will open us up corrupting the AIConfig.

Adding a backup copy for rollback as well as scheduling mutations as tasks adds yet a lot more complexity. Furthermore, scheduling all our mutations via create_task() effectively deletes the order of mutations requested by the frontend. Not only can mutations happen concurrently, they can happen in an arbitrary order.

Think about how tricky this is even without an HTTP server or a frontend to worry about. How would you do it if you were just in a script running mutations on the instance?

saqadri:
Test Plan:

https://github.com/lastmile-ai/aiconfig/assets/25641935/3b844283-94c1-489f-800c-ff4ae0da499c

python/src/aiconfig/editor/server/server.py
saqadri: You're right :). It's been fixed with threads, thanks to @rossdanlm's changes for streaming

python/src/aiconfig/editor/server/server.py
saqadri: Fixed

python/src/aiconfig/editor/server/server.py
saqadri: You're right :). Updated implementation

python/src/aiconfig/editor/server/server.py
saqadri: Ah I like this pattern. We could add this to Config.py perhaps? (not as part of this change)

python/src/aiconfig/editor/server/server.py
```diff
@@ -71,7 +75,7 @@ def run_backend_server(edit_config: EditServerConfig) -> Result[str, str]:
             LOGGER.info("Initialized server state")
             debug = edit_config.server_mode in [ServerMode.DEBUG_BACKEND, ServerMode.DEBUG_SERVERS]
             LOGGER.info(f"Running in {edit_config.server_mode} mode")
-            app.run(port=edit_config.server_port, debug=debug, use_reloader=debug)
+            app.run(port=edit_config.server_port, debug=debug, use_reloader=debug, threaded=True)
```
jonathanlastmileai, line 78: I don't think you need threaded=True actually. not totally sure

python/src/aiconfig/editor/server/server.py
```diff
@@ -207,28 +217,62 @@ def run() -> FlaskResponse:
...
+            PyThreadState_SetAsyncExc: This is a C API function in Python which is used to raise an exception in the context
+            of the specified thread.
+
+            SystemExit: This is the exception we'd like to raise in the target thread.
+            """
+            if thread_id is None:
+                # Nothing to do
+                return
+
+            response = ctypes.pythonapi.PyThreadState_SetAsyncExc(ctypes.c_long(thread_id), ctypes.py_object(SystemExit))
```
jonathanlastmileai, line 267: Ideally we shouldn't be this low-level. I think if this works it's ok for now, but we should figure out the best supported way to do this. create_task(asyncio.to_thread()) might be it, might not, TBD

jonathanlastmileai (approved):
Naturally not so easy to reason about, but if it's tested, let's ship. unblocking

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -594,14 +617,53 @@ export default function EditorContainer({
...
               dispatch({
                 type: "CONSOLIDATE_AICONFIG",
                 action,
                 config: event.data,
               });
             }
           },
-          enableStreaming
+          (event) => {
+            console.log(
```
saqadri, line 638: Maybe I should dispatch a run prompt error in this case where it isn't a cancellation? Currently we aren't propagating errors from the aiconfig.run call to here as far as I can tell

python/src/aiconfig/editor/client/src/components/prompt/PromptActionBar.tsx
```diff
@@ -44,6 +46,21 @@ export default memo(function PromptActionBar({
   const modelSettingsSchema = promptSchema?.model_settings;
   const promptMetadataSchema = promptSchema?.prompt_metadata;
 
+  const onCancel = useCallback(async () => {
+    if (prompt._ui.cancellationToken) {
+      console.log(
```
saqadri, line 51: Will clean up all the log statements before pushing

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
saqadri: Will do at top of stack. Currently no error is propagated other than cancellation anyways

python/src/aiconfig/editor/server/server.py
saqadri: Going to leave as-is -- couldn't figure out a great way to terminate a thread in Python, other than this. Can revisit if needed

jonathanlastmileai: This is fine for now if it works. Whenever we're ready to migrate to V2, I'm pretty sure it'll just work using asyncio to_thread() and cancel, for any operation (not just run).

saqadri: Does v2 have streaming set up? I don't think we should spend any time on v2 until we get critical things done first.

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -594,14 +617,53 @@ export default function EditorContainer({
...
+              `Error running prompt ${promptName}: ${JSON.stringify(event)}`
+            );
+            if (event.type === "error") {
+              //throw new Error(event.data.message);
+
+              if (event.data.code === 499) {
+                // This is a cancellation
+                // Reset the aiconfig to the state before we started running the prompt
```
rholinshead, line 646: IMO we shouldn't reset the config. If you cancel while streaming it should just stop the output where it is instead of clearing it

rholinshead (commented):
Nice, LGTM in general, just something we can update later is handling the cancellation by not reverting state and just keeping the output stream stopped where it was

rholinshead:
> Nice, LGTM in general, just something we can update later is handling the cancellation by not reverting state and just keeping the output stream stopped where it was

Oh, I guess that *is* what happens, just the comment was confusing (I thought it was reverting the output entirely in the comment but video shows what I would actually expect)

jonathanlastmileai:
> > Nice, LGTM in general, just something we can update later is handling the cancellation by not reverting state and just keeping the output stream stopped where it was
> 
> Oh, I guess that _is_ what happens, just the comment was confusing (I thought it was reverting the output entirely in the comment but video shows what I would actually expect)

I think later on we can have the frontend store the source of truth and control commits performed by the server to the in-memory instance and disk. We can sort of make a mini database system that supports mutation requests in the order requested, including commit and rollback. That way the frontend can decide exactly what to commit - whether before, during, or after a run.

PR #790, opened 2024-01-05
[editor][server] 1/n endpoint clear outputs

Ankush-lastmile:
[editor][server] 1/n endpoint clear outputs

server endpoint to clear outputs from server side aiconfig state (changes not persisted to disk).

## Testplan

1. Run Prompt
Screenshot 2024-01-08 at 10 43 49 AM

2. Clear outputs.

Screenshot 2024-01-08 at 10 54 12 AM

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/790).
* #791
* __->__ #790

### DIFF
--- a/python/src/aiconfig/editor/server/server.py
+++ b/python/src/aiconfig/editor/server/server.py
@@ -457,3 +457,30 @@ def set_description() -> FlaskResponse:
     operation = make_op_run_method(MethodName("set_description"))
     operation_args: Result[OpArgs, str] = result.Ok(OpArgs({"description": description}))
     return run_aiconfig_operation_with_op_args(aiconfig, "set_description", operation, operation_args)
+
+@app.route("/api/clear_outputs", methods=["POST"])
+def clear_outputs() -> FlaskResponse:
+    """
+    Clears all outputs in the server state's AIConfig.
+    """
+    state = get_server_state(app)
+    aiconfig = state.aiconfig
+    request_json = request.get_json()
+
+    if aiconfig is None:
+        LOGGER.info("No AIConfig in memory, can't run clear outputs.")
+        return HttpResponseWithAIConfig(
+            message="No AIConfig in memory, can't run clear outputs.",
+            code=400,
+            aiconfig=None,
+        ).to_flask_format()
+
+    def _op(aiconfig_runtime: AIConfigRuntime, _op_args: OpArgs) -> Result[None, str]:
+        for prompt in aiconfig_runtime.prompts:
+            prompt_name = prompt.name
+            # fn name `delete_output`` is misleading. TODO: Rename to `delete_outputs`` in AIConfig API
+            aiconfig_runtime.delete_output(prompt_name)
+        return Ok(None)
+
+    signature: dict[str, Type[Any]] = {}
+    return run_aiconfig_operation_with_request_json(aiconfig, request_json, f"method_", _op, signature)


python/src/aiconfig/editor/server/server.py
```diff
@@ -457,3 +457,30 @@ def set_description() -> FlaskResponse:
...
+        return HttpResponseWithAIConfig(
+            message="No AIConfig in memory, can't run clear outputs.",
+            code=400,
+            aiconfig=None,
+        ).to_flask_format()
+
+    for prompt in aiconfig.prompts:
+        prompt_name = prompt.name
+        # fn name `delete_output`` is misleading. TODO: Rename to `delete_outputs`` in AIConfig API
+        aiconfig.delete_output(prompt_name)
```
rossdanlm, line 480: nit: would prefer this be part of the SDK itself, but we can make a bootcamp issue for this, not a blocker

python/src/aiconfig/editor/server/server.py
```diff
@@ -457,3 +457,30 @@ def set_description() -> FlaskResponse:
     operation = make_op_run_method(MethodName("set_description"))
     operation_args: Result[OpArgs, str] = result.Ok(OpArgs({"description": description}))
     return run_aiconfig_operation_with_op_args(aiconfig, "set_description", operation, operation_args)
+
+@app.route("/api/clear_outputs", methods=["POST"])
```
rossdanlm, line 461: To extend, we can take in list of prompt_names, and if none specified, default to all of them (like what we have below)

python/src/aiconfig/editor/server/server.py
```diff
@@ -457,3 +457,30 @@ def set_description() -> FlaskResponse:
...
+            aiconfig=None,
+        ).to_flask_format()
+
+    for prompt in aiconfig.prompts:
+        prompt_name = prompt.name
+        # fn name `delete_output`` is misleading. TODO: Rename to `delete_outputs`` in AIConfig API
+        aiconfig.delete_output(prompt_name)
+
+
+    return HttpResponseWithAIConfig(
```
rossdanlm, line 483: Jonathan already built `run_aiconfig_operation_with_op_args()`, let's use that

rossdanlm (approved):
left a few comments

python/src/aiconfig/editor/server/server.py
rholinshead: That would be handy for clearing single outputs later, but can be done at that point, I think

python/src/aiconfig/editor/server/server.py
jonathanlastmileai: I think run_aiconfig_operation_with_op_args() is not quite general enough.

 This is fine, but it turns out this should be easy enough like this:
 
 ```
 def _op(aiconfig_runtime: AIConfigRuntime, _op_args: OpArgs) -> Result[None, str]:
        for prompt in aiconfig_runtime.prompts:
            prompt_name = prompt.name
            # fn name `delete_output`` is misleading. TODO: Rename to `delete_outputs`` in AIConfig API
            aiconfig_runtime.delete_output(prompt_name)
        return Ok(None)
    signature: dict[str, Type[Any]] = {}
    return run_aiconfig_operation_with_request_json(aiconfig, request_json, f"method_{method_name}", _op, signature)
 ```

jonathanlastmileai (approved):

python/src/aiconfig/editor/server/server.py
Ankush-lastmile: updated. Thanks @jonathanlastmileai for the help, used `run_aiconfig_operation_with_request_json`.

Ankush-lastmile:
updated to use `run_aiconfig_operation_with_request_json`. Tested with similar testplan as before, request now required to have a valid body. Also tested with frontend on diff on top

PR #791, opened 2024-01-05
[edtior][client] 2/n clear outputs functionality

Ankush-lastmile:
[edtior][client] 2/n clear outputs functionality

added a button that clears all outputs.

- added an action
- call to server api

## Testplan

- Run prompt
- validate config has an output
- clear outputs
- validate config cleared outputs.

https://github.com/lastmile-ai/aiconfig/assets/141073967/f1622528-5a77-49cd-80a3-581520229a7f

### DIFF
--- a/python/src/aiconfig/editor/client/src/Editor.tsx
+++ b/python/src/aiconfig/editor/client/src/Editor.tsx
@@ -68,6 +68,11 @@ export default function Editor() {
     });
   }, []);
 
+  const clearOutputs = useCallback(async () => {
+    return await ufetch.post(ROUTE_TABLE.CLEAR_OUTPUTS, {
+    });
+  }, []);
+
   const runPrompt = useCallback(
     async (
       promptName: string,
@@ -153,6 +158,7 @@ export default function Editor() {
   const callbacks: AIConfigCallbacks = useMemo(
     () => ({
       addPrompt,
+      clearOutputs,
       deletePrompt,
       getModels,
       getServerStatus,
@@ -166,6 +172,7 @@ export default function Editor() {
     }),
     [
       addPrompt,
+      clearOutputs,
       deletePrompt,
       getModels,
       getServerStatus,
--- a/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
@@ -794,7 +794,7 @@ export default function EditorContainer({
       <Container maw="80rem">
         <Flex justify="flex-end" mt="md" mb="xs">
           <Group>
-            <Button loading={undefined} onClick={onClearOutputs} size="xs" variant = "gradient">
+            <Button loading={undefined} onClick={onClearOutputs} size="xs" variant="gradient">
               Clear Outputs
             </Button>
           
--- a/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
@@ -8,6 +8,7 @@ import {
   Text,
   Tooltip,
   Alert,
+  Group,
 } from "@mantine/core";
 import { Notifications, showNotification } from "@mantine/notifications";
 import {
@@ -71,11 +72,8 @@ type RunPromptStreamEvent =
 export type RunPromptStreamCallback = (event: RunPromptStreamEvent) => void;
 
 export type AIConfigCallbacks = {
-  addPrompt: (
-    promptName: string,
-    prompt: Prompt,
-    index: number
-  ) => Promise<{ aiconfig: AIConfig }>;
+  addPrompt: (promptName: string, prompt: Prompt, index: number) => Promise<{ aiconfig: AIConfig }>;
+  clearOutputs: () => Promise<{ aiconfig: AIConfig }>;
   deletePrompt: (promptName: string) => Promise<void>;
   getModels: (search: string) => Promise<string[]>;
   getServerStatus?: () => Promise<{ status: "OK" | "ERROR" }>;
@@ -543,6 +541,26 @@ export default function EditorContainer({
     [deletePromptCallback, dispatch]
   );
 
+  const clearOutputsCallback = callbacks.clearOutputs;
+  const onClearOutputs = useCallback(
+    async () => {
+      dispatch({
+        type: "CLEAR_OUTPUTS",
+      });
+      try {
+        await clearOutputsCallback();
+      } catch (err: unknown) {
+        const message = (err as RequestCallbackError).message ?? null;
+        showNotification({
+          title: "Error clearing outputs",
+          message,
+          color: "red",
+        });
+      }
+    },
+    [clearOutputsCallback, dispatch]
+  );
+
   const runPromptCallback = callbacks.runPrompt;
 
   const onRunPrompt = useCallback(
@@ -775,6 +793,11 @@ export default function EditorContainer({
       )}
       <Container maw="80rem">
         <Flex justify="flex-end" mt="md" mb="xs">
+          <Group>
+            <Button loading={undefined} onClick={onClearOutputs} size="xs" variant = "gradient">
+              Clear Outputs
+            </Button>
+          
           <Tooltip
             label={isDirty ? "Save changes to config" : "No unsaved changes"}
           >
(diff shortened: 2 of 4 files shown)


python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -773,6 +790,11 @@ export default function EditorContainer({
       )}
       <Container maw="80rem">
         <Flex justify="flex-end" mt="md" mb="xs">
+          <div>
+            <Button loading={undefined} onClick={onClearOutputs}>
+              Clear Outputs
+            </Button>
+          </div>
```
Ankush-lastmile, line 797: I noticed save button had updated styling. Not sure what is required for this button.

rossdanlm: Not related to this diff, but let's update "save" button colour to be green or blue, definitely not red

python/src/aiconfig/editor/client/src/Editor.tsx
```diff
@@ -68,6 +68,11 @@ export default function Editor() {
     });
   }, []);
 
+  const clearOutputs = useCallback(async () => {
+    return await ufetch.post(ROUTE_TABLE.CLEAR_OUTPUTS, {
```
rossdanlm, line 72: cc @rholinshead should we add a confirmation dialog that asks them if they're sure before deleting? I'm guessing no (especially if we have undo/redo actions in the future)

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -543,6 +540,26 @@ export default function EditorContainer({
...
+          type: "CLEAR_OUTPUTS",
+        });
+        await clearOutputsCallback();
+      } catch (err: unknown) {
+        const message = (err as RequestCallbackError).message ?? null;
+        showNotification({
+          title: "Error clearing outputs",
+          message,
+          color: "red",
+        });
```
rossdanlm, line 557: Add a TODO to handle error boundaries, similar to what Ryan did in https://github.com/lastmile-ai/aiconfig/pull/799

rossdanlm (commented):
Generally lgtm but will let Ryan do final stamp

python/src/aiconfig/editor/client/src/Editor.tsx
rholinshead: I don't think we should since it's pretty explicit and not completely disruptive

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
rholinshead: Error boundaries should only be used for catching rendering errors. Catching request exception should be done as is done here with try/catch and showing notification

python/src/aiconfig/editor/client/src/components/aiconfigReducer.ts
```diff
@@ -215,6 +220,17 @@ export default function aiconfigReducer(
...
     }
+    case "CLEAR_OUTPUTS": {
+      // This action happens before the server request is even sent; This is to maximize responsiveness of the client state.
+      // Once the server response comes in we can use the CONSOLIDATE_AICONFIG action to update the client with the response and that action would trigger a re-render.
+      return {
+        ...dirtyState,
+        prompts: dirtyState.prompts.map((prompt) => ({
+          ...prompt,
+          outputs: [],
+        })),
```
rholinshead, line 231: Thinking on this a bit: should we do empty outputs array [] or just delete outputs entirely? If the former, just realized we should not do this for any prompt that doesn't have outputs defined because this will end up adding empty outputs array to the config for the prompt when it never had outputs key in the first place.

So maybe something like:
```
const prompts = state.prompts.map((prompt) => {
  if (prompt.outputs) {
    return {
       ...prompt,
        outputs: [] // or undefined
     }
  } else {
    return prompt;
  }
});

// If we want to remove instead of empty array, can do:
// for (const prompt in prompts) {
//   if (prompt.outputs) {
//    delete prompt.outputs;
//  }
// if we don't want to delete outputs here, just inline the prompts array below as before, but with the 
// map logic from above

return {
  ...dirtyState,
  prompts,
};
}
```

rholinshead (changes requested):
Just a few things

python/src/aiconfig/editor/client/src/components/aiconfigReducer.ts
Ankush-lastmile: I would say +1 to delete the outputs because as you pointed out, empty array is consistent with our design to have no outputs if outputs aren't defined

Ankush-lastmile: left a comment on slack but to reiterate:

Makes sense to update here but the server side will still have prompt.outputs as [], even though client here might have it deleted.

For this change to be persisted, I see two options, which one should I do?

1. Update AIConfig sdk method delete_output to forcefully delete the attribute? Although I'm not sure if its expected to be [] anywhere right now.

2. Alternatively Delete the outputs field inside the server endpoint in api/clear_outputs and AIConfig sdk doesn't need to be touched (easy change, set field to None after calling delete_output)

Ankush-lastmile:
- Replaced div with Group, containing both clear outputs button and save button
- moved dispatch() from inside try/catch to right before it (inside the callback)
- removed comment as requested

Awaiting response from team before impl for clearing empty outputs from config

cookbooks/Chain-of-Verification/cove_template_config.json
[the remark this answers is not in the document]
Ankush-lastmile: thanks for catching, removed

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -543,6 +540,26 @@ export default function EditorContainer({
     [deletePromptCallback, dispatch]
   );
 
+  const clearOutputsCallback = callbacks.clearOutputs;
+  const onClearOutputs = useCallback(
+    async () => {
+      try {
+        dispatch({
+          type: "CLEAR_OUTPUTS",
+        });
```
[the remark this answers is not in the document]
Ankush-lastmile: moved

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -773,6 +790,11 @@ export default function EditorContainer({
       )}
       <Container maw="80rem">
         <Flex justify="flex-end" mt="md" mb="xs">
+          <div>
```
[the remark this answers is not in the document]
Ankush-lastmile: updated with group

python/src/aiconfig/editor/client/src/components/aiconfigReducer.ts
```diff
@@ -215,6 +220,17 @@ export default function aiconfigReducer(
     case "ADD_PROMPT_AT_INDEX": {
       return reduceInsertPromptAtIndex(dirtyState, action.index, action.prompt);
     }
+    case "CLEAR_OUTPUTS": {
+      // This action happens before the server request is even sent; This is to maximize responsiveness of the client state.
+      // Once the server response comes in we can use the CONSOLIDATE_AICONFIG action to update the client with the response and that action would trigger a re-render.
```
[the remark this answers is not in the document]
Ankush-lastmile: removed comment

python/src/aiconfig/editor/client/src/components/aiconfigReducer.ts
rholinshead: Should we remove them here on the client and just do the sdk change separately in a later PR? Then we won't need to come back here to change it later

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -773,6 +791,11 @@ export default function EditorContainer({
       )}
       <Container maw="80rem">
         <Flex justify="flex-end" mt="md" mb="xs">
+          <Group>
+            <Button loading={undefined} onClick={onClearOutputs} size="xs" variant = "gradient">
```
rholinshead, line 795:
```suggestion
            <Button loading={undefined} onClick={onClearOutputs} size="xs" variant="gradient">
```

rholinshead (approved):
Accepting to unblock but let's remove the outputs here and do the SDK change separate in a subsequent PR

Ankush-lastmile:
- rebase
- inside "CLEAR_OUTPUTS" action, set prompt outputs to undefined, to "delete" outputs from client state

PR #798, opened 2024-01-06
[editor] Streaming Outputs

rholinshead:
# [editor] Streaming Outputs

Add client-side handling for streaming outputs. We currently derive whether the request should stream by checking settings. Not a huge fan of this since:
- it couples client with config runtime logic for getting the 'stream' value
- assumes 'stream' setting is used by the models

This is sufficient for now, I think. Note that I do think it's nice to allow the editor user to toggle between stream/no stream since it will be useful for debugging their model parser

## Testing:
- Error handling: added `return HttpResponseWithAIConfig(message="No AIConfig loaded", code=400, aiconfig=None).to_flask_format()` in /run with 'stream' set to true and ensured error is shown as notification and run state is stopped on the client
- Non-streaming model (dall-e) run correctly
- Streaming works for models with stream setting set to tru/false:

https://github.com/lastmile-ai/aiconfig/assets/5060851/5242f49f-0003-45a6-af19-1e29c611eca4

### DIFF
--- a/python/src/aiconfig/editor/client/src/Editor.tsx
+++ b/python/src/aiconfig/editor/client/src/Editor.tsx
@@ -1,11 +1,19 @@
 import EditorContainer, {
   AIConfigCallbacks,
+  RunPromptStreamCallback,
 } from "./components/EditorContainer";
 import { Flex, Loader, MantineProvider, Image } from "@mantine/core";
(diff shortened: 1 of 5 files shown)


jonathanlastmileai (approved):
nice!

PR #799, opened 2024-01-06
[editor] ErrorBoundary Renderer for PromptInput

rholinshead:
[editor] ErrorBoundary Renderer for PromptInput

# [editor] ErrorBoundary Renderer for PromptInput

With the ability to set arbitrary JSON for the prompt input, if we have a PromptSchema which states what the input should look like, we should be able to do some basic validation that the input at least matches the general type (for now, compare 'string' PromptInput vs object -- in the future, we could do some basic validation against the schema).

If there is any error with rendering the schema-based rendering for the input, we can use ErrorBoundary to fallback to an error message informing the user that the format is invalid and to toggle to JSON editor to fix.

Screenshot 2024-01-05 at 11 01 07 PM

Toggling to the JSON editor will clear the error state and try to render the correct Schema renderer when toggling back.

Note that in dev mode, these errors are still propagated to a top-level error screen which you can dismiss to show the fallback. In prod, that won't happen and will just show the fallback and an error in the console.

We will do the same for the SettingsRenderer in a subsequent PR

## Testing:
- Updated `OpenAIChatModelParserPromptSchema` input to match that of `OpenAIChatVisionModelParserPromptSchema` while the input is still string. See the error fallback
- Toggle to JSON editor and back, ensure error fallback still shows
- Toggle to JSON editor and update to valid input
```
{
        "attachments": [
          {
            "data": "https://s3.amazonaws.com/files.uploads.lastmileai.com/uploads/cldxsqbel0000qs8owp8mkd0z/2023_12_1_21_23_24/942/Screenshot 2023-11-28 at 11.11.25 AM.png",
            "mime_type": "image/png"
          },
          {
            "data": "https://s3.amazonaws.com/files.uploads.lastmileai.com/uploads/cldxsqbel0000qs8owp8mkd0z/2023_12_1_21_23_24/8325/Screenshot 2023-11-28 at 1.51.52 PM.png",
            "mime_type": "image/png"
          }
        ],
        "data": "What do these images show?"
      }
```
and see proper schema rendering

https://github.com/lastmile-ai/aiconfig/assets/5060851/1343a201-e5eb-46a0-a7f5-a4bdb228d120

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/799).
* #803
* #802
* #801
* #800
* __->__ #799

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/prompt/prompt_input/PromptInputRenderer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/prompt/prompt_input/PromptInputRenderer.tsx
@@ -3,9 +3,12 @@ import { memo, useState } from "react";
 import { PromptInputSchema } from "../../../utils/promptUtils";
 import PromptInputSchemaRenderer from "./schema_renderer/PromptInputSchemaRenderer";
 import PromptInputConfigRenderer from "./PromptInputConfigRenderer";
-import { ActionIcon, Flex, Tooltip } from "@mantine/core";
-import { IconBraces, IconBracesOff } from "@tabler/icons-react";
+import { Flex } from "@mantine/core";
 import PromptInputJSONRenderer from "./PromptInputJSONRenderer";
+import { ErrorBoundary, useErrorBoundary } from "react-error-boundary";
+import { Text } from "@mantine/core";
+import JSONRenderer from "../../JSONRenderer";
+import JSONEditorToggleButton from "../../JSONEditorToggleButton";
 
 type Props = {
   input: PromptInput;
@@ -13,17 +16,50 @@ type Props = {
   onChangeInput: (value: PromptInput) => void;
 };
 
+type ErrorFallbackProps = {
+  input: PromptInput;
+  toggleJSONEditor: () => void;
+};
+
+function InputErrorFallback({ input, toggleJSONEditor }: ErrorFallbackProps) {
+  const { resetBoundary: clearRenderError } = useErrorBoundary();
+  return (
+    <Flex direction="column">
+      <Text color="red" size="sm">
+        Invalid input format for model. Toggle JSON editor to update
+      </Text>
+      <JSONRenderer content={input} />
+      <Flex justify="flex-end">
+        <JSONEditorToggleButton
+          isRawJSON={false}
+          setIsRawJSON={() => {
+            clearRenderError();
+            toggleJSONEditor();
+          }}
+        />
+      </Flex>
+    </Flex>
+  );
+}
+
 export default memo(function PromptInputRenderer({
   input,
   schema,
   onChangeInput,
 }: Props) {
   const [isRawJSON, setIsRawJSON] = useState(false);
-  return (
+  const rawJSONToggleButton = (
+    <Flex justify="flex-end">
+      <JSONEditorToggleButton
+        isRawJSON={isRawJSON}
+        setIsRawJSON={setIsRawJSON}
+      />
+    </Flex>
+  );
+
+  const nonJSONRenderer = (
     <>
-      {isRawJSON ? (
-        <PromptInputJSONRenderer input={input} onChangeInput={onChangeInput} />
-      ) : schema ? (
+      {schema ? (
         <PromptInputSchemaRenderer
           input={input}
           schema={schema}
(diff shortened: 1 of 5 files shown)


python/src/aiconfig/editor/client/src/components/prompt/prompt_input/PromptInputRenderer.tsx
```diff
@@ -3,49 +3,95 @@ import { memo, useState } from "react";
...
   onChangeInput: (value: PromptInput) => void;
 };
 
+type ErrorFallbackProps = {
+  input: PromptInput;
+  toggleJSONEditor: () => void;
+};
+
+function InputErrorFallback({ input, toggleJSONEditor }: ErrorFallbackProps) {
+  const { resetBoundary: clearRenderError } = useErrorBoundary();
```
rossdanlm, line 25: Does this mean it only will process the code below if there's an error present?

python/src/aiconfig/editor/client/src/components/prompt/prompt_input/PromptInputRenderer.tsx
```diff
@@ -3,49 +3,95 @@ import { memo, useState } from "react";
...
-      <Flex justify="flex-end">
-        <Tooltip label="Toggle JSON editor" withArrow>
-          <ActionIcon onClick={() => setIsRawJSON((curr) => !curr)}>
-            {isRawJSON ? (
-              <IconBracesOff size="1rem" />
+        <ErrorBoundary
+          fallbackRender={() => (
+            <InputErrorFallback
+              input={input}
+              toggleJSONEditor={() => setIsRawJSON(true)}
```
rossdanlm, line 74: This is confusing because it only gets called during the `setIsRawJSON` prop in lines 35-27, but we're immediately setting this back to true?

Can we rename this to make ti more intuitive, or leave a comment in code explaining what's happening

python/src/aiconfig/editor/client/src/components/prompt/prompt_input/PromptInputRenderer.tsx
```diff
@@ -3,49 +3,95 @@ import { memo, useState } from "react";
...
+              <PromptInputConfigRenderer
+                input={input}
+                onChangeInput={onChangeInput}
+              />
             )}
-          </ActionIcon>
-        </Tooltip>
-      </Flex>
+            {rawJSONToggleButton}
+          </>
```
rossdanlm, line 92: Nit: lots of nested ternary expressions. Could we move this into it's own helper component like `<PromptInputRendererForNonJson>` or something?

python/src/aiconfig/editor/client/src/components/prompt/prompt_input/PromptInputRenderer.tsx
rossdanlm: It's implicitly assuming that the current state of `isRawJson` is false due to check on L61. Let's either 

1. Add a comment why we're setting this to `true` so it's obvious
2. Pass in `!isRawJSON` instead (though I guess this might become weird if code gets rearranged and we're already somehow in raw json format and asking user to switch to raw json). Maybe we should do an `assert` statement here between L71 and 72 to ensure that `isRawJson is false` before proceeding?

rossdanlm (approved):
Cool, learned about `react-error-boundary` today: https://github.com/bvaughn/react-error-boundary

python/src/aiconfig/editor/client/src/components/prompt/prompt_input/PromptInputRenderer.tsx
rholinshead: We do use it -- this just renames it to `clearRenderError` since I found `resetBoundary` to be a bit unclear

python/src/aiconfig/editor/client/src/components/prompt/prompt_input/PromptInputRenderer.tsx
rholinshead: I'll leave a comment in the code.

> This is confusing because it only gets called during the setIsRawJSON prop in lines 35-27, but we're immediately setting this back to true?
It's only ever rendered if the toggle is not showing raw JSON already (and there's a render error), so we explicitly want the fallback toggle to only be able to toggle to showing the raw JSON editor.

rholinshead:
Changes from review: added comment, pulled non-JSON renderer to a const for clarity

PR #800, opened 2024-01-06
[editor] ErrorBoundary Renderer for ModelSettings

rholinshead:
[editor] ErrorBoundary Renderer for ModelSettings

# [editor] ErrorBoundary Renderer for ModelSettings

With the ability to set arbitrary JSON for the model settings, if we have a PromptSchema which states what the setting types should be, the settings renderer will assume those types match the config settings JSON. If the JSON doesn't actually match the schema types, it's possible for the settings renderer to throw an error. If that happens, we can use an ErrorBoundary to fallback to an error message informing the user that the format is invalid and to toggle the JSON editor to fix.

Screenshot 2024-01-06 at 3 56 50 PM

Toggling to the JSON editor will clear the error state and try to render the correct Schema renderer when toggling back.

Note that in dev mode, these errors are still propagated to a top-level error screen which you can dismiss to show the fallback. In prod, that won't happen and will just show the fallback and an error in the console.

## Testing:
- Use JSON editor for model settings to set frequency_penalty to a string; expected value is a number so the settings renderer will error. See the error fallback
- Toggle to JSON editor and back, ensure error fallback still shows
- Toggle to JSON editor and update to valid settings and then toggle back to see proper schema rendering

https://github.com/lastmile-ai/aiconfig/assets/5060851/060ddb20-129d-450e-97ab-cf2939269f90

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/800).
* #803
* #802
* #801
* __->__ #800
* #799

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/prompt/model_settings/ModelSettingsRenderer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/prompt/model_settings/ModelSettingsRenderer.tsx
@@ -1,10 +1,11 @@
 import ModelSettingsSchemaRenderer from "./ModelSettingsSchemaRenderer";
 import { GenericPropertiesSchema } from "../../../utils/promptUtils";
-import { ActionIcon, Flex, Tooltip, createStyles } from "@mantine/core";
+import { Flex, Text, createStyles } from "@mantine/core";
 import { JSONObject } from "aiconfig";
 import { memo, useState } from "react";
-import { IconBraces, IconBracesOff } from "@tabler/icons-react";
 import JSONRenderer from "../../JSONRenderer";
+import JSONEditorToggleButton from "../../JSONEditorToggleButton";
+import { ErrorBoundary, useErrorBoundary } from "react-error-boundary";
 
 type Props = {
   settings?: JSONObject;
@@ -22,6 +23,35 @@ const useStyles = createStyles(() => ({
   },
 }));
 
+type ErrorFallbackProps = {
+  settings?: JSONObject;
+  toggleJSONEditor: () => void;
+};
+
+function SettingsErrorFallback({
+  settings,
+  toggleJSONEditor,
+}: ErrorFallbackProps) {
+  const { resetBoundary: clearRenderError } = useErrorBoundary();
+  return (
+    <Flex direction="column">
+      <Text color="red" size="sm">
+        <Flex justify="flex-end">
+          <JSONEditorToggleButton
+            isRawJSON={false}
+            setIsRawJSON={() => {
+              clearRenderError();
+              toggleJSONEditor();
+            }}
+          />
+        </Flex>
+        Invalid settings format for model. Toggle JSON editor to update
+      </Text>
+      <JSONRenderer content={settings} />
+    </Flex>
+  );
+}
+
 export default memo(function ModelSettingsRenderer({
   settings,
   schema,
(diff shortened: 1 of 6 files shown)


python/src/aiconfig/editor/client/src/components/prompt/model_settings/ModelSettingsRenderer.tsx
```diff
@@ -22,6 +23,35 @@ const useStyles = createStyles(() => ({
...
+type ErrorFallbackProps = {
+  settings?: JSONObject;
+  toggleJSONEditor: () => void;
+};
+
+function SettingsErrorFallback({
+  settings,
+  toggleJSONEditor,
+}: ErrorFallbackProps) {
+  const { resetBoundary: clearRenderError } = useErrorBoundary();
```
rossdanlm, line 35: Nit: Delete `resetBoundary`

python/src/aiconfig/editor/client/src/components/prompt/model_settings/ModelSettingsRenderer.tsx
```diff
@@ -30,37 +60,45 @@ export default memo(function ModelSettingsRenderer({
...
-              {isRawJSON ? (
-                <IconBracesOff size="1rem" />
-              ) : (
-                <IconBraces size="1rem" />
-              )}
-            </ActionIcon>
-          </Tooltip>
-        </Flex>
-      )}
       {isRawJSON || !schema ? (
```
rossdanlm, line 74: Nit: this is a bit different in format from the prompt input (we do something like `isRawJson ? showJson : <ErrorBoundary>...</ErrorBoundary>`. Doesn't have to be this diff but if possible, let's make a note to unify the "format" so the codebase becomes eaiser to read

python/src/aiconfig/editor/client/src/components/prompt/model_settings/ModelSettingsRenderer.tsx
```diff
@@ -30,37 +60,45 @@ export default memo(function ModelSettingsRenderer({
...
-        <JSONRenderer
-          content={settings}
-          onChange={(val) =>
-            onUpdateModelSettings(val as Record<string, unknown>)
-          }
-          // schema={schema} TODO: Add schema after fixing z-index issue
-        />
+        <>
+          {/* // Only show the toggle if there is a schema to toggle between JSON and custom schema renderer */}
```
rossdanlm, line 76: Oh ok cool, nm

python/src/aiconfig/editor/client/src/components/prompt/model_settings/ModelSettingsRenderer.tsx
```diff
@@ -22,6 +23,35 @@ const useStyles = createStyles(() => ({
...
+              clearRenderError();
+              toggleJSONEditor();
+            }}
+          />
+        </Flex>
+        Invalid settings format for model. Toggle JSON editor to update
+      </Text>
+      <JSONRenderer content={settings} />
+    </Flex>
+  );
```
rossdanlm, line 52: nit: can also make a follow up issue to centralize this with the similar Fallback function from prompt input but change the text. Not a blocker

rossdanlm (approved):
Cool, lgtm just a few small nit followups

python/src/aiconfig/editor/client/src/components/prompt/model_settings/ModelSettingsRenderer.tsx
rholinshead: They are a bit different. The input currently renders `PromptInputConfigRenderer` when no schema exists for the prompt model. It's not fully implemented but when it is, we should have the config renderer do some rendering based on the following logic:
- if input exists, implicitly try to match it to the nice renderers that are associated with schema types
- otherwise, render as JSON editor

For settings, we should just default to JSON editor if no schema exists because settings are basically a generic object of any possible key/value pairs

python/src/aiconfig/editor/client/src/components/prompt/model_settings/ModelSettingsRenderer.tsx
rholinshead: Main difference is input has the JSON toggle on the bottom right and settings has it top-right; IMO, refactor to support dynamic toggle placement and callbacks for this 14-line component would just be overkill

python/src/aiconfig/editor/client/src/components/prompt/model_settings/ModelSettingsRenderer.tsx
rholinshead: This is just renaming it for a clarity

PR #801, opened 2024-01-06
[editor][easy] Add 'model' to relevant Prompt Schemas

rholinshead:
[editor][easy] Add 'model' to relevant Prompt Schemas

# [editor][easy] Add 'model' to relevant Prompt Schemas

There is still some investigation needed to know how to handle model vs parser nicely in the editor UX. For now, let's just expose the 'model' option in the settings for those parsers that support it to ensure the usage of those parsers won't be blocked.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/801).
* #803
* #802
* __->__ #801
* #800
* #799

### DIFF
--- a/python/src/aiconfig/editor/client/src/shared/prompt_schemas/DalleImageGenerationParserPromptSchema.ts
+++ b/python/src/aiconfig/editor/client/src/shared/prompt_schemas/DalleImageGenerationParserPromptSchema.ts
@@ -10,6 +10,9 @@ export const DalleImageGenerationParserPromptSchema: PromptSchema = {
   model_settings: {
     type: "object",
     properties: {
+      model: {
+        type: "string",
+      },
       n: {
         type: "integer",
         minimum: 1,
--- a/python/src/aiconfig/editor/client/package.json
+++ b/python/src/aiconfig/editor/client/package.json
@@ -41,6 +41,7 @@
     "oboe": "^2.1.5",
     "react": "^18",
     "react-dom": "^18",
+    "react-error-boundary": "^4.0.12",
     "react-markdown": "^8.0.6",
     "react-scripts": "5.0.1",
     "remark-gfm": "^4.0.0",
--- a/python/src/aiconfig/editor/client/src/components/JSONEditorToggleButton.tsx
+++ b/python/src/aiconfig/editor/client/src/components/JSONEditorToggleButton.tsx
@@ -1,0 +1,20 @@
+import { ActionIcon, Tooltip } from "@mantine/core";
+import { IconBraces, IconBracesOff } from "@tabler/icons-react";
+
+type Props = {
+  isRawJSON: boolean;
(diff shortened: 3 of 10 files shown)


python/src/aiconfig/editor/client/src/shared/prompt_schemas/DalleImageGenerationParserPromptSchema.ts
```diff
@@ -10,6 +10,9 @@ export const DalleImageGenerationParserPromptSchema: PromptSchema = {
   model_settings: {
     type: "object",
     properties: {
+      model: {
+        type: "string",
+      },
```
rossdanlm, line 15: I don't understand what these are used for. Why do we need to have this `type: "string"` field?

rholinshead: This determines how the setting property 'model' is rendered in the nice settings renderer:
Screenshot 2024-01-08 at 10 42 56 AM


The PromptSchema is a way for model parsers to define what their input, model settings and metadata are structured like, because they are otherwise entirely generic. Without this static schema we have no way of knowing what to render for these prompts

rossdanlm: Can we add description in the info text so people know what this is for? I'm still not really sure. How is this different from the `select model` top right selector on each prompt?

rholinshead: We can add a description but the descriptions are specific to each parser's prompt schema implementation. For openai for example, model description from https://platform.openai.com/docs/api-reference/chat/create is 

> ID of the model to use. See the [model endpoint compatibility](https://platform.openai.com/docs/models/model-endpoint-compatibility) table for details on which models work with the Chat API.

Description tooltips don't support markdown so there's no way to nicely add an external link as hyperlink (also, don't know if we would want to support that since custom parsers could then technically add malicious hyperlinks ?)



> How is this different from the select model top right selector on each prompt

The select model top right selector is technically not even a model right now since AIConfig allows us to register arbitrary ids for models when registering model parsers. Take HuggingFaceTextGenerationParser for example, we register "HuggingFaceTextGenerationParser": "HuggingFaceTextGenerationParser" as parser and model id. So if we don't have a 'model' option in the settings there is no way to specify what model to actually use for the API call. Same for AnyscaleEndpoint. 

Also, technically the case for gpt-4, etc (until #783 lands), since we register the model/parser as "gpt-4": "gpt-4" but pull the actual model for the API from the settings (not from model name in the prompt metadata)

PR #806, opened 2024-01-08
Make all outputs in "streaming" format

rossdanlm:
Make all outputs in "streaming" format

Doing this to make it easier for Ryan to parse with a singular unified output format, regardless of whether the model parser actually supports streaming or not.

- moved `runPromptStream` -> `runPrompt`, overriding the old definition of `runPrompt` by now passing in an `enableStreaming` flag
- still relying on `isStreamingSupported()` function to set the `enableStreaming` param to true or false
- default to `stream` param being `True` on backend server (however this has no effect for non-streaming models like Dall-E)

## Test Plan
Test that both streaming and non-streaming settings work as expected, as well as test that a model which does not support streaming (ex: Dall-E) still works. Now when user hasn't explicitly clicked the "stream" setting, it will default to streaming. However if they explictly toggle it turns off. Follow up we should have the "stream" button auto-enabled to reflect this (doesn't have to actually be in the config, we should just have the UI show it as on by default to match user expectation)

Update: Updated this so that the default value for `stream` is now `true` inside of `OpenAIChatModelParserPromptSchema`, `HuggingFaceTextGenerationParserPromptSchema` and `AnyscaleEndpointPromptSchema`. Couldn't see it defined in `PaLMTextParserPromptSchema`

https://github.com/lastmile-ai/aiconfig/assets/151060367/34214a66-0cea-4774-a917-9476359f172c

### DIFF
--- a/python/src/aiconfig/editor/client/src/Editor.tsx
+++ b/python/src/aiconfig/editor/client/src/Editor.tsx
@@ -68,21 +68,22 @@ export default function Editor() {
     });
   }, []);
 
-  const runPrompt = useCallback(async (promptName: string) => {
-    return await ufetch.post(ROUTE_TABLE.RUN_PROMPT, {
-      prompt_name: promptName,
-    });
-  }, []);
-
-  const runPromptStream = useCallback(
-    async (promptName: string, onStream: RunPromptStreamCallback) => {
-      await streamingApi(
+  const runPrompt = useCallback(
+    async (
+      promptName: string,
+      onStream: RunPromptStreamCallback,
+      enableStreaming: boolean = true
+    ) => {
+      // Note: We run the streaming API even for
+      // non-streaming runs so that we can unify
+      // the way we process data on the client
+      return await streamingApi<{ aiconfig?: AIConfig } | void>(
         {
           url: ROUTE_TABLE.RUN_PROMPT,
           method: "POST",
           body: {
             prompt_name: promptName,
-            stream: true,
+            stream: enableStreaming,
           },
         },
         "output_chunk",
@@ -156,7 +157,6 @@ export default function Editor() {
       getModels,
       getServerStatus,
       runPrompt,
-      runPromptStream,
       save,
       setConfigDescription,
       setConfigName,
@@ -170,7 +170,6 @@ export default function Editor() {
       getModels,
       getServerStatus,
       runPrompt,
-      runPromptStream,
       save,
       setConfigDescription,
       setConfigName,
--- a/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
@@ -36,8 +36,8 @@ import {
 import AddPromptButton from "./prompt/AddPromptButton";
 import {
   getDefaultNewPromptName,
+  getModelSettingsStream,
   getPrompt,
-  isStreamingSupported,
 } from "../utils/aiconfigStateUtils";
 import { debounce, uniqueId } from "lodash";
 import PromptMenuButton from "./prompt/PromptMenuButton";
@@ -79,11 +79,11 @@ export type AIConfigCallbacks = {
   deletePrompt: (promptName: string) => Promise<void>;
   getModels: (search: string) => Promise<string[]>;
   getServerStatus?: () => Promise<{ status: "OK" | "ERROR" }>;
-  runPrompt: (promptName: string) => Promise<{ aiconfig: AIConfig }>;
-  runPromptStream: (
+  runPrompt: (
     promptName: string,
-    onStream: RunPromptStreamCallback
-  ) => Promise<void>;
+    onStream: RunPromptStreamCallback,
+    enableStreaming?: boolean
+  ) => Promise<{ aiconfig?: AIConfig } | void>;
   save: (aiconfig: AIConfig) => Promise<void>;
   setConfigDescription: (description: string) => Promise<void>;
   setConfigName: (name: string) => Promise<void>;
@@ -544,7 +544,6 @@ export default function EditorContainer({
   );
 
   const runPromptCallback = callbacks.runPrompt;
-  const runPromptStreamCallback = callbacks.runPromptStream;
 
   const onRunPrompt = useCallback(
     async (promptId: string) => {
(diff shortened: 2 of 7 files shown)


python/src/aiconfig/editor/client/src/Editor.tsx
```diff
@@ -68,21 +68,22 @@ export default function Editor() {
...
-      prompt_name: promptName,
-    });
-  }, []);
-
-  const runPromptStream = useCallback(
-    async (promptName: string, onStream: RunPromptStreamCallback) => {
+  const runPrompt = useCallback(
+    async (
+      promptName: string,
+      enableStreaming: boolean = true,
```
rossdanlm, line 74: cc @rholinshead I can remove default value if you prefer

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -79,9 +79,9 @@ export type AIConfigCallbacks = {
   deletePrompt: (promptName: string) => Promise<void>;
   getModels: (search: string) => Promise<string[]>;
   getServerStatus?: () => Promise<{ status: "OK" | "ERROR" }>;
-  runPrompt: (promptName: string) => Promise<{ aiconfig: AIConfig }>;
-  runPromptStream: (
+  runPrompt: (
     promptName: string,
+    enableStreaming: boolean,
     onStream: RunPromptStreamCallback
   ) => Promise<void>;
```
rholinshead, line 86: Let's make this return `Promise<{aiconfig?: AIConfig}>;` so that external implementations don't necessarily need to implement their execution via streaming

python/src/aiconfig/editor/client/src/Editor.tsx
rholinshead: Let's keep it. I think we should default to true so that we 'stream' by default for everything and change to regular request/response only if the model settings specifically states not to (stream: false). External implementations can default to false here if they want. We probably need to swap `onStream` to be before `enableStreaming` and make enableStreaming optional

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -79,9 +79,9 @@ export type AIConfigCallbacks = {
   deletePrompt: (promptName: string) => Promise<void>;
   getModels: (search: string) => Promise<string[]>;
   getServerStatus?: () => Promise<{ status: "OK" | "ERROR" }>;
-  runPrompt: (promptName: string) => Promise<{ aiconfig: AIConfig }>;
-  runPromptStream: (
+  runPrompt: (
     promptName: string,
+    enableStreaming: boolean,
```
rholinshead, line 84:
```suggestion
    enableStreaming?: boolean,
```

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -562,34 +561,26 @@ export default function EditorContainer({
...
-        } else {
-          const serverConfigRes = await runPromptCallback(promptName);
-
-          dispatch({
-            type: "CONSOLIDATE_AICONFIG",
-            action,
-            config: serverConfigRes.aiconfig,
-          });
-        }
+        const enableStreaming = isStreamingSupported(
```
rholinshead, line 564: Can we change `isStreamingSupported` to something like `getModelSettingsStream` or something similar, and have it return either a boolean or undefined? Thinking here is:
- by default, models that support streaming but don't have 'stream' value set in the settings should still stream in editor to have the best UX by default
- if user explicitly states 'no' for stream value in settings, then we shouldn't stream
- our server implementation can technically support 'stream' style response for all models (so we can handle the undefined 'enableStreaming' param in Editor.tsx `runPrompt` function by defaulting to true. But other external implementations might not, so they can handle the undefined case however they want

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -562,34 +561,26 @@ export default function EditorContainer({
...
+          statePrompt!,
+          stateRef.current
+        );
+
+        await runPromptCallback(promptName, enableStreaming, (event) => {
```
rholinshead, line 569:
```suggestion
        const serverConfigRes = await runPromptCallback(promptName, enableStreaming, (event) => {
```

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -562,34 +561,26 @@ export default function EditorContainer({
...
+              output: event.data,
+            });
+          } else if (event.type === "aiconfig") {
+            dispatch({
+              type: "CONSOLIDATE_AICONFIG",
+              action,
+              config: event.data,
+            });
+          }
+        });
```
rholinshead, line 583:
```suggestion
        });
        
        if (serverConfigRes) {
          dispatch({
            type: "CONSOLIDATE_AICONFIG",
            action,
            config: serverConfigRes.aiconfig,
          });
        }
```

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
rholinshead: Let's handle non-streaming still, in case external implementations don't implement streaming

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -562,10 +561,14 @@ export default function EditorContainer({
         }
 
         const promptName = statePrompt.name;
-        const isStream = isStreamingSupported(statePrompt, stateRef.current);
+        const enableStreaming: boolean | undefined = getModelSettingsStream(
+          statePrompt!,
```
rholinshead, line 565: ! shouldn't be needed with the above check

python/src/aiconfig/editor/client/src/components/EditorContainer.tsx
```diff
@@ -579,15 +582,15 @@ export default function EditorContainer({
...
             }
-          });
-          return;
-        } else {
-          const serverConfigRes = await runPromptCallback(promptName);
+          },
+          enableStreaming
+        );
 
+        if (serverConfigResponse.aiconfig) {
```
rholinshead, line 589: Does this need to be `serverConfigResponse?.aiconfig`?


Edit: Oh, I guess it's because we have required object with optional aiconfig but maybe it should be `Promise<{aiconfig?: aiconfig} | void>`?

rholinshead (approved):
LGTM, just two minor things

PR #812, opened 2024-01-08
[Local Editor Styling Updates]

zakariassoul:
[Local Editor Styling Updates]

### DIFF
--- a/python/src/aiconfig/editor/client/src/Editor.tsx
+++ b/python/src/aiconfig/editor/client/src/Editor.tsx
@@ -69,8 +69,7 @@ export default function Editor() {
   }, []);
 
   const clearOutputs = useCallback(async () => {
-    return await ufetch.post(ROUTE_TABLE.CLEAR_OUTPUTS, {
-    });
+    return await ufetch.post(ROUTE_TABLE.CLEAR_OUTPUTS, {});
   }, []);
 
   const runPrompt = useCallback(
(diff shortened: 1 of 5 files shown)


python/src/aiconfig/editor/client/src/Editor.tsx
```diff
@@ -242,6 +254,64 @@ export default function Editor() {
...
+              bottom: "0",
+              margin: "16px 0 0 0",
+              background: "#ff1cf7",
+              "&:hover": {
+                background: "#ff46f8",
+              },
+            },
+            ".mantine-InputWrapper-label": {
+              display: "none",
+            },
```
rholinshead, line 301: these are going to apply to every mantine input, right? I'm pretty sure we don't want that -- is this intended only for the prompt name / input label?

zakariassoul: I moved it to be only applied to input fields within the cell container

zakariassoul:
this should pass and have all fixes

python/src/aiconfig/editor/client/src/Editor.tsx
```diff
@@ -276,6 +291,9 @@ export default function Editor() {
               color: "white",
               borderRadius: "0",
               height: "auto",
+              "&:hover": {
+                background: "#ff46f8",
```
rholinshead, line 295: One thing to look into afterwards -- we can set a [primaryColory](https://mantine.dev/theming/colors/) with the 10 different shades and it will automatically apply to all button backgrounds, hover, highlight around text area, etc. without needing to manually do it.

rholinshead (approved):
LGTM

PR #814, opened 2024-01-08
[editor][easy] Add Descriptions to Prompt Schemas

rholinshead:
[editor][easy] Add Descriptions to Prompt Schemas

# [editor][easy] Add Descriptions to Prompt Schemas

Add descriptions to the other prompt schemas for the core models. Also, fix some of the hugging face properties to have correct ranges based on the documentation (tested those ranges are valid by setting the values to 80 or so and running HF prompt).

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/814).
* #818
* #815
* __->__ #814

PR #815, opened 2024-01-09
[editor] HuggingFaceText2ImageDiffusorPromptSchema

rholinshead:
[editor] HuggingFaceText2ImageDiffusorPromptSchema

# [editor] HuggingFaceText2ImageDiffusorPromptSchema

Add the PromptSchema for HuggingFaceText2ImageDiffusor. Note that there are several settings which are going to have to just be supported programatically since modelling them with schema types would be pretty hard:
generator, latents, prompt_embeds, negative_prompt_embeds, callback

Each is of a python class-derived type

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/815).
* #818
* __->__ #815
* #814

### DIFF
--- a/python/src/aiconfig/editor/client/src/shared/prompt_schemas/HuggingFaceText2ImageDiffusorPromptSchema.ts
+++ b/python/src/aiconfig/editor/client/src/shared/prompt_schemas/HuggingFaceText2ImageDiffusorPromptSchema.ts
@@ -1,0 +1,102 @@
+import { PromptSchema } from "../../utils/promptUtils";
+
+export const HuggingFaceText2ImageDiffusorPromptSchema: PromptSchema = {
+  // See /Users/ryanholinshead/Projects/aiconfig/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/text_2_image.py
+  // for supported settings.
+  // See https://huggingface.co/docs/diffusers/main/en/api/pipelines/stable_diffusion/text2img#diffusers.StableDiffusionPipeline.__call__
+  // for descriptions.
+
+  // This is currently for Kandinsky 2.2
+
+  // The following supported properties cannot be cleanly modeled in the schema and must be implemented programatically:
+  // generator, latents, prompt_embeds, negative_prompt_embeds, callback
+
+  input: {
+    type: "string",
+  },
+  model_settings: {
+    type: "object",
+    properties: {
+      height: {
+        type: "integer",
+        description: `The height in pixels of the generated image.`,
+      },
+      width: {
+        type: "integer",
+        description: `The width in pixels of the generated image.`,
+      },
+      num_inference_steps: {
+        type: "integer",
+        description: `The number of denoising steps. More denoising steps usually lead to a higher 
+        quality image at the expense of slower inference.`,
+      },
+      guidance_scale: {
+        type: "number",
+        description: `A higher guidance scale value encourages the model to generate images closely linked 
+        to the text prompt at the expense of lower image quality. Guidance scale is enabled when guidance_scale > 1.`,
+      },
+      // TODO: Convert to anyOf once renderer supports it
+      negative_prompt: {
+        type: "union",
+        types: [
+          {
+            type: "string",
+          },
+          {
+            type: "array",
+            items: {
+              type: "string",
+            },
+          },
+        ],
+      },
+      num_images_per_prompt: {
+        type: "integer",
+        description: `The number of images to generate per prompt.`,
+      },
+      eta: {
+        type: "number",
+        description: `Corresponds to parameter eta (η) from the DDIM paper. 
+        Only applies to the DDIMScheduler, and is ignored in other schedulers.`,
+      },
+      output_type: {
+        type: "string",
+        enum: ["pil", "array"],
+        description: `The output format of the generated image. Choose between PIL.Image or np.array.`,
+      },
+      return_dict: {
+        type: "boolean",
+        description: `Whether or not to return a StableDiffusionPipelineOutput instead of a plain tuple.`,
+      },
+      cross_attention_kwargs: {
+        type: "map",
+        keys: {
+          type: "string",
+        },
+        items: {
+          type: "string",
+        },
+        description: `A kwargs dictionary that if specified is passed along to the AttentionProcessor as defined in self.processor.`,
+      },
+      guidance_rescale: {
+        type: "number",
+        description: `Guidance rescale factor from Common Diffusion Noise Schedules and Sample Steps are Flawed.
+         Guidance rescale factor should fix overexposure when using zero terminal SNR.`,
+      },
+      clip_skip: {
+        type: "integer",
+        description: `Number of layers to be skipped from CLIP while computing the prompt embeddings. 
+        A value of 1 means that the output of the pre-final layer will be used for computing the prompt embeddings.`,
+      },
+      requires_safety_checker: {
+        type: "boolean",
+        description: `Whether or not the model requires a safety checker to be used.`,
+      },
+      callback_steps: {
+        type: "integer",
+        description: `The frequency at which the callback function will be called. 
+        If not specified, the callback will be called at every step.`,
+      },
+    },
+  },
+};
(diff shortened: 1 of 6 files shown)


saqadri (approved):

python/src/aiconfig/editor/client/src/shared/prompt_schemas/HuggingFaceText2ImageDiffusorPromptSchema.ts
```diff
@@ -0,0 +1,100 @@
+import { PromptSchema } from "../../utils/promptUtils";
+
+export const HuggingFaceText2ImageDiffusorPromptSchema: PromptSchema = {
+  // See /Users/ryanholinshead/Projects/aiconfig/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/text_2_image.py
+  // for supported settings.
+  // See https://huggingface.co/docs/diffusers/main/en/api/pipelines/stable_diffusion/text2img#diffusers.StableDiffusionPipeline.__call__
+  // for descriptions.
+
+  // The following supported properties cannot be cleanly modeled in the schema and must be implemented programatically:
+  // generator, latents, prompt_embeds, negative_prompt_embeds, callback
```
saqadri, line 10: Do you have thoughts on how this would be modeled? Will the model parser need to handle them?

saqadri (approved):

python/src/aiconfig/editor/client/src/shared/prompt_schemas/HuggingFaceText2ImageDiffusorPromptSchema.ts
rossdanlm: This is because they're all torch objects right?

What we do now is first filter for generator params, and then we do completion params. We could do something similar where we have two sets of input schemas for images

python/src/aiconfig/editor/client/src/shared/prompt_schemas/HuggingFaceText2ImageDiffusorPromptSchema.ts
```diff
@@ -0,0 +1,100 @@
...
+        description: `The number of images to generate per prompt.`,
+      },
+      eta: {
+        type: "number",
+        description: `Corresponds to parameter eta (η) from the DDIM paper. 
+        Only applies to the DDIMScheduler, and is ignored in other schedulers.`,
+      },
+      output_type: {
+        type: "string",
+        enum: ["pil", "array"],
```
rossdanlm, line 62: make default `pil` since I haven't added support for `np array` yet

python/src/aiconfig/editor/client/src/shared/prompt_schemas/HuggingFaceText2ImageDiffusorPromptSchema.ts
```diff
@@ -0,0 +1,100 @@
...
+        description: `Guidance rescale factor from Common Diffusion Noise Schedules and Sample Steps are Flawed.
+         Guidance rescale factor should fix overexposure when using zero terminal SNR.`,
+      },
+      clip_skip: {
+        type: "integer",
+        description: `Number of layers to be skipped from CLIP while computing the prompt embeddings. 
+        A value of 1 means that the output of the pre-final layer will be used for computing the prompt embeddings.`,
+      },
+      requires_safety_checker: {
+        type: "boolean",
```
rossdanlm, line 90: Also default this to false, because I've found setting this to true is anoying and blacks out so many images

python/src/aiconfig/editor/client/src/shared/prompt_schemas/HuggingFaceText2ImageDiffusorPromptSchema.ts
```diff
@@ -0,0 +1,100 @@
...
+        type: "integer",
+        description: `The width in pixels of the generated image.`,
+      },
+      num_inference_steps: {
+        type: "integer",
+        description: `The number of denoising steps. More denoising steps usually lead to a higher 
+        quality image at the expense of slower inference.`,
+      },
+      guidance_scale: {
+        type: "number",
```
rossdanlm, line 32: default to 7.5 which is what they use in code

python/src/aiconfig/editor/client/src/shared/prompt_schemas/HuggingFaceText2ImageDiffusorPromptSchema.ts
```diff
@@ -0,0 +1,100 @@
...
+      height: {
+        type: "integer",
+        description: `The height in pixels of the generated image.`,
+      },
+      width: {
+        type: "integer",
+        description: `The width in pixels of the generated image.`,
+      },
+      num_inference_steps: {
+        type: "integer",
```
rossdanlm, line 27: default to 50 which is what's used in code

python/src/aiconfig/editor/client/src/shared/prompt_schemas/HuggingFaceText2ImageDiffusorPromptSchema.ts
```diff
@@ -0,0 +1,100 @@
...
+  },
+  model_settings: {
+    type: "object",
+    properties: {
+      height: {
+        type: "integer",
+        description: `The height in pixels of the generated image.`,
+      },
+      width: {
+        type: "integer",
```
rossdanlm, line 23: Is it possible for these to be cleared/reset to None? Out of scope of this diff but would be good to have a "Reset" to put everything back to defaults in each prompt setting

python/src/aiconfig/editor/client/src/shared/prompt_schemas/HuggingFaceText2ImageDiffusorPromptSchema.ts
```diff
@@ -0,0 +1,100 @@
...
+        Only applies to the DDIMScheduler, and is ignored in other schedulers.`,
+      },
+      output_type: {
+        type: "string",
+        enum: ["pil", "array"],
+        description: `The output format of the generated image. Choose between PIL.Image or np.array.`,
+      },
+      return_dict: {
+        type: "boolean",
+        description: `Whether or not to return a StableDiffusionPipelineOutput instead of a plain tuple.`,
```
rossdanlm, line 67: Just a note, it can also be `StableDiffusionXLPipeline`, which takes on slightly different params in it's call: https://huggingface.co/docs/diffusers/api/pipelines/stable_diffusion/stable_diffusion_xl#diffusers.StableDiffusionXLPipeline.__call__

python/src/aiconfig/editor/client/src/shared/prompt_schemas/HuggingFaceText2ImageDiffusorPromptSchema.ts
rholinshead: I would need to spend more time looking into it to know, but I assume they are serialized as python objects in the config? It's just hard to map generic python objects to schema-supported types

python/src/aiconfig/editor/client/src/shared/prompt_schemas/HuggingFaceText2ImageDiffusorPromptSchema.ts
rholinshead: In general I don't really want to use 'default' because:
- UI shows empty state and that's actually what the config is
- currently having a default means changing any other unrelated config value and saving will also save the default to the config, which is annoying

rossdanlm (approved):
Text2Image is a bit tricky since there isn't really a singular "schema" for all the models that support that task. There are 3 of the most commonly used classes (Stable Diffusion v1.5, Stable Diffusion XL, and Kandinsky 2.2: https://huggingface.co/docs/diffusers/using-diffusers/conditional_image_generation#popular-models) which all take slightly different completion params. I can add the remaining completion params from Stable Diffusion XL and Kandinsky 2.2 so that everything will be available. 

But this could also lead to some confusion that if you're playing around with a Kandinsky-only setting but testing a Stable Diffusion 1.5 model, it won't have any effect on your result. We can figure this out later (ex: add note in description saying this is for `Kandinsky 2.2` models only). What do you think?

python/src/aiconfig/editor/client/src/shared/prompt_schemas/HuggingFaceText2ImageDiffusorPromptSchema.ts
rholinshead: You can toggle to the JSON editor to clear

PR #820, opened 2024-01-09
[editor] Renaming Components Prior to Packaging

rholinshead:
# [editor] Renaming Components Prior to Packaging

We'll want to package the editor client component as 'AIConfigEditor' for reuse in other places, instead of having it named 'EditorContainer'.

To prevent confusion, let's also rename our root component to 'LocalEditor' here instead of just 'Editor'.

Used `sl mv` to preserve history

## Testing:
`aiconfig edit --aiconfig-path=python/src/aiconfig/editor/travel.aiconfig.json --server-mode='debug_servers'` and see that the editor renders correctly


saqadri (approved):
Ship it! Good idea

PR #827, opened 2024-01-09
Styling changes for Gradio Workbook

saqadri:
Styling changes for Gradio Workbook

Plus publish script for test-pypi


python/pyproject.toml
```diff
@@ -2,8 +2,8 @@
 requires = ["setuptools", "wheel"]
 
 [project]
-name = "python-aiconfig"
-version = "1.1.9"
+name = "python-aiconfig-test"
+version = "1.1.25"
```
rossdanlm, line 6: For future, we don't need to keep changing this. We can have two separate repos and simply when we do pip install we explicitly pull from our test repo:

`pip3 install --index-url https://test.pypi.org/project/python-aiconfig-test/`

^We can use `python-aiconfig` instead of `python-aiconfig-test`

I supposed if we are going to update the the version number to match pytest anyways, it's not that much different to change package name so that's fine too
cc @jonathanlastmileai

python/testpypi-publish.sh
```diff
@@ -0,0 +1,27 @@
...
+ cd python 
+ rm -rf ./dist
+ source /opt/homebrew/Caskroom/miniconda/base/etc/profile.d/conda.sh && conda activate aiconfig
+ python3 -m pip install --upgrade build
+ python3 -m build
+ python3 -m pip install --upgrade twine
+
+ # If you want to upload to prod , run pypipublish.sh.
+
+ python3 -m twine upload --repository testpypi dist/*
```
rossdanlm, line 25: cc @jonathanlastmileai can you change your PR from "draft" in https://github.com/lastmile-ai/aiconfig/pull/807 so we can land and keep changes separate here?

zakariassoul:
@saqadri  can you post a video or screenshots of the editor?

PR #834, opened 2024-01-09
[editor] Audio Output Renderer

rholinshead:
# [editor] Audio Output Renderer

This is basically a copy/paste from workbook audio outputs in lastmile, so it should work the same.


saqadri (approved):

rholinshead:
Me to this renderer:

PR #842, opened 2024-01-10
[editor] HuggingFaceTextGenerationTransformerPromptSchema

rholinshead:
[editor] HuggingFaceTextGenerationTransformerPromptSchema

### DIFF
--- a/python/src/aiconfig/editor/client/src/utils/promptUtils.ts
+++ b/python/src/aiconfig/editor/client/src/utils/promptUtils.ts
@@ -7,6 +7,9 @@ import { PaLMChatParserPromptSchema } from "../shared/prompt_schemas/PaLMChatPar
 import { HuggingFaceTextGenerationParserPromptSchema } from "../shared/prompt_schemas/HuggingFaceTextGenerationParserPromptSchema";
 import { AnyscaleEndpointPromptSchema } from "../shared/prompt_schemas/AnyscaleEndpointPromptSchema";
 import { HuggingFaceText2ImageDiffusorPromptSchema } from "../shared/prompt_schemas/HuggingFaceText2ImageDiffusorPromptSchema";
+import { HuggingFaceTextGenerationTransformerPromptSchema } from "../shared/prompt_schemas/HuggingFaceTextGenerationTransformerPromptSchema";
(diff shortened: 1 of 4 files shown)


python/src/aiconfig/editor/client/src/utils/promptUtils.ts
```diff
@@ -84,6 +86,12 @@ export const PROMPT_SCHEMAS: Record<string, PromptSchema> = {
 
   // Local HuggingFace Parsers
   HuggingFaceText2ImageDiffusor: HuggingFaceText2ImageDiffusorPromptSchema,
+  TextGeneration: HuggingFaceTextGenerationTransformerPromptSchema,
+  Text2Image: HuggingFaceText2ImageDiffusorPromptSchema,
+  AutomaticSpeechRecognition: HuggingFaceAutomaticSpeechRecognitionPromptSchema,
```
rossdanlm, line 91: Nice yea, we'll just hardcode these names for now

python/src/aiconfig/editor/client/src/utils/promptUtils.ts
```diff
@@ -84,6 +86,12 @@ export const PROMPT_SCHEMAS: Record<string, PromptSchema> = {
...
+  // Text2Speech: HuggingFaceText2SpeechTransformerPromptSchema,
+  // HuggingFaceTextTranslationTransformer: HuggingFaceTextTranslationTransformerPromptSchema,
```
rossdanlm, line 93: Let's just use `HuggingFaceTextGenerationTransformerPromptSchema` for these two since both of these transformers are sub-classes of TextGeneration so I know at minimum text-generation input schema will be a sub-set of these two completion params

rossdanlm (approved):
Nice, thanks!

python/src/aiconfig/editor/client/src/utils/promptUtils.ts
rholinshead: We'll want to change them to whatever the hf package is using when registering them, but not sure what that final copy will be just yet

cc @saqadri for visibility

python/src/aiconfig/editor/client/src/utils/promptUtils.ts
rholinshead: Thanks. I made separate prompt schemas that just spread the text one in for now. We can revisit if we see actual differences, otherwise can later rename to something more general for use in these 3

PR #844, opened 2024-01-10
Action Button Replacement

zakariassoul:
Action Button Replacement

moved the action button so it does not jump around when streaming.

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/prompt/PromptActionBar.tsx
+++ b/python/src/aiconfig/editor/client/src/components/prompt/PromptActionBar.tsx
@@ -102,12 +102,12 @@ export default memo(function PromptActionBar({
               )}
             </Tabs>
           </Container>
-          <RunPromptButton
+          {/* <RunPromptButton
             isRunning={prompt._ui.isRunning}
             cancel={onCancel}
             runPrompt={onRunPrompt}
             size="full"
-          />
+          /> */}
         </>
       ) : (
         <Flex direction="column" justify="space-between" h="100%">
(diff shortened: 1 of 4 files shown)


zakariassoul:
Screenshot 2024-01-09 at 7 06 43 PM

python/src/aiconfig/editor/client/src/components/prompt/PromptActionBar.tsx
```diff
@@ -102,12 +102,12 @@ export default memo(function PromptActionBar({
               )}
             </Tabs>
           </Container>
-          <RunPromptButton
+          {/* <RunPromptButton
```
saqadri, line 105: Why commented out?

python/src/aiconfig/editor/client/src/components/prompt/PromptActionBar.tsx
```diff
@@ -116,12 +116,12 @@ export default memo(function PromptActionBar({
               <IconClearAll />
             </ActionIcon>
           </Flex>
-          <RunPromptButton
+          {/* <RunPromptButton
```
saqadri, line 119: Why is this commented out?

saqadri (commented):
I think it clashes a bit with the json button next to it since they aren't the same size. Will defer to @rholinshead for feedback

rholinshead:
> I think it clashes a bit with the json button next to it since they aren't the same size. Will defer to @rholinshead for feedback

Ya, I think we should keep the JSON editor toggle above for consistency with settings view

rholinshead:
How does this run button look for long inputs?

zakariassoul:
@rholinshead & @saqadri in a previous iteration, i made the JSON toggle in the next row. It's still discoverable, but not really critical to take a front and center place. I tried to push this change with this PR but for the life of me i could not figure out how the template was put together, i spent too much time to justify moving a button so i just pushed as is. 

@rholinshead  take a look at the screenshot below and if all good, can you please make the change in a different diff?

Screenshot 2024-01-09 at 9 52 22 PM

zakariassoul:
@rholinshead  this is how it would look like for long input 
Screenshot 2024-01-09 at 9 55 39 PM

rholinshead:
Thanks @zakariassoul  -- https://github.com/lastmile-ai/aiconfig/pull/857/files is what I was able to get to match those screenshots as best I could

PR #845, opened 2024-01-10
[editor] Get Automatic Speech Recognition Working

rholinshead:
[editor] Get Automatic Speech Recognition Working

# [editor] Get Automatic Speech Recognition Working

Screenshot 2024-01-09 at 10 25 21 PM

https://github.com/lastmile-ai/aiconfig/assets/5060851/fba30e0e-2160-4a84-b657-759f8689ed9e

Some fixes were needed for handling uploading attachments:
1. CORS is currently preventing policy request to lastmile endpoint from localhost; we'll need to fix this but for now we don't need the policy
2. Handling case of no attachments needed some fixes for how attachments are added to the end of the attachments array in `onAddAttachments` and we also need the default input structure to be an object to match the schema (previously was defaulting to string)

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/845).
* #848
* __->__ #845
* #817

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/prompt/prompt_input/attachments/AttachmentUploader.tsx
+++ b/python/src/aiconfig/editor/client/src/components/prompt/prompt_input/attachments/AttachmentUploader.tsx
@@ -15,11 +15,60 @@ type Props = {
   onCancel?: () => void;
 };
 
-async function uploadFile(_file: File) {
-  // TODO: Implement
-  return {
-    url: "https://s3.amazonaws.com/files.uploads.lastmileai.com/uploads/cldxsqbel0000qs8owp8mkd0z/2023_12_1_21_23_24/942/Screenshot 2023-11-28 at 11.11.25 AM.png",
-  };
+// s3 file uris cannot have '+' character, so replace with '_'
+function sanitizeFileName(name: string) {
+  return name.replace(/[_+]/g, "_");
+}
+
+export function getTodayDateString(): string {
+  const date = new Date();
+  const dateString = `${date.getFullYear()}_${
+    date.getMonth() + 1
+  }_${date.getDate()}`;
+  const timeString = `${date.getUTCHours()}_${date.getUTCMinutes()}_${date.getUTCSeconds()}`;
+  return `${dateString}_${timeString}`;
+}
+
+// TODO: Make this configurable for external deployments
+async function uploadFile(file: File): Promise<{ url: string }> {
+  const randomPath = Math.round(Math.random() * 10000);
+  const policyResponse = await fetch(
+    "https://lastmileai.dev/api/upload/publicpolicy"
+  );
+  const policy = await policyResponse.json();
+  const uploadUrl = "https://s3.amazonaws.com/lastmileai.aiconfig.public/";
+  const uploadKey = `uploads/${getTodayDateString()}/${randomPath}/${sanitizeFileName(
+    file.name
+  )}`;
+
+  const formData = new FormData();
+  formData.append("key", uploadKey);
+  formData.append("acl", "public-read");
+  formData.append("Content-Type", file.type);
+  formData.append("AWSAccessKeyId", policy.AWSAccessKeyId);
+  formData.append("success_action_status", "201");
+  formData.append("Policy", policy.s3Policy);
+  formData.append("Signature", policy.s3Signature);
+  formData.append("file", file);
+
+  // See this about changing to use XMLHTTPRequest to show upload progress as well
+  // https://medium.com/@cpatarun/tracking-file-upload-progress-to-amazon-s3-from-the-browser-71be6712c63d
+  const rawRes = await fetch(uploadUrl, {
+    method: "POST",
+    mode: "cors",
+    cache: "no-cache",
+    body: formData,
+    headers: {
+      Authorization: "",
+    },
+  });
+
+  if (rawRes.ok && rawRes.status === 201) {
+    // Dont really need to parse xml s3 response, just use the keys, etc. that were passed
+    return { url: `${uploadUrl}${uploadKey}` };
+  } else {
+    throw new Error("Error uploading to S3!");
+  }
 }
 
 function getSupportedFileTypes(schema: PromptInputObjectAttachmentsSchema) {
@@ -66,7 +115,7 @@ Props) {
       const attachments: Attachment[] = uploads.map((upload) => {
         return {
           data: upload.url,
-          mimeType: upload.mimeType,
+          mime_type: upload.mimeType,
         };
       });
 
--- a/python/src/aiconfig/editor/client/src/components/prompt/prompt_input/attachments/AttachmentUploader.tsx
+++ b/python/src/aiconfig/editor/client/src/components/prompt/prompt_input/attachments/AttachmentUploader.tsx
@@ -32,10 +32,11 @@ export function getTodayDateString(): string {
 // TODO: Make this configurable for external deployments
 async function uploadFile(file: File): Promise<{ url: string }> {
   const randomPath = Math.round(Math.random() * 10000);
-  const policyResponse = await fetch(
-    "https://lastmileai.dev/api/upload/publicpolicy"
-  );
-  const policy = await policyResponse.json();
+  // TODO: Add back once CORS is resolved
+  // const policyResponse = await fetch(
+  //   "https://lastmileai.dev/api/upload/publicpolicy"
+  // );
+  // const policy = await policyResponse.json();
   const uploadUrl = "https://s3.amazonaws.com/lastmileai.aiconfig.public/";
   const uploadKey = `uploads/${getTodayDateString()}/${randomPath}/${sanitizeFileName(
     file.name
@@ -45,10 +46,10 @@ async function uploadFile(file: File): Promise<{ url: string }> {
   formData.append("key", uploadKey);
   formData.append("acl", "public-read");
   formData.append("Content-Type", file.type);
-  formData.append("AWSAccessKeyId", policy.AWSAccessKeyId);
+  // formData.append("AWSAccessKeyId", policy.AWSAccessKeyId);
   formData.append("success_action_status", "201");
-  formData.append("Policy", policy.s3Policy);
-  formData.append("Signature", policy.s3Signature);
+  // formData.append("Policy", policy.s3Policy);
+  // formData.append("Signature", policy.s3Signature);
   formData.append("file", file);
 
   // See this about changing to use XMLHTTPRequest to show upload progress as well
@@ -72,7 +73,7 @@ async function uploadFile(file: File): Promise<{ url: string }> {
 }
 
 function getSupportedFileTypes(schema: PromptInputObjectAttachmentsSchema) {
-  return schema.items.mime_types;
+  return schema.items.mime_types.join(", ");
 }
 
 export default memo(function AttachmentUploader({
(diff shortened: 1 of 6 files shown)


python/src/aiconfig/editor/client/src/components/prompt/prompt_input/attachments/AttachmentUploader.tsx
```diff
@@ -161,10 +162,10 @@ Props) {
           }}
           // TODO: Get these from schema,
           // maxSize={MAX_IMAGE_FILE_SIZE}
-          // accept={}
+          accept={schema.items.mime_types}
```
rossdanlm, line 165: Oh cool, just like curl

python/src/aiconfig/editor/client/src/utils/promptUtils.ts
```diff
@@ -180,3 +180,21 @@ function isTextInputModality(prompt: Prompt): boolean {
...
+      if (schema.input.properties.data) {
+        return {
+          data: schema.input.properties.data.type === "string" ? "" : {},
+        };
+      }
+      return {};
+    }
+  }
+
+  return "";
```
rossdanlm, line 199: Ok yea this is wild, standardizing inputs would be a lot more helpful

python/src/aiconfig/editor/client/src/components/prompt/prompt_input/schema_renderer/PromptInputSchemaRenderer.tsx
```diff
@@ -42,16 +42,18 @@ function SchemaRenderer({ input, schema, onChangeInput }: SchemaRendererProps) {
 
   return (
     <Flex direction="column">
-      <DataRenderer
-        schema={dataSchema}
-        data={data}
-        onChangeData={onChangeData}
-      />
+      {dataSchema && (
+        <DataRenderer
```
rossdanlm, line 46: Why do we need another `DataRenderer` class instead of re-using `AttachmentsRenderer` for audio files?

python/src/aiconfig/editor/client/src/components/prompt/prompt_input/schema_renderer/PromptInputAttachmentsSchemaRenderer.tsx
```diff
@@ -106,18 +114,18 @@ export default memo(function PromptInputAttachmentsSchemaRenderer({
...
-              onUpdateAttachment(attachment, numAttachments + 1)
-            }
-            onAddAttachments={(addedAttachments) =>
-              onAddAttachments(addedAttachments, numAttachments + 1)
-            }
-          />
-        ))}
+      {(schema.max_items == null || numAttachments < schema.max_items) && (
+        <EditableAttachmentRenderer
+          key={numAttachments}
```
rossdanlm, line 119: shouldn't the key be something unique? Or is this created once and never changed?

python/src/aiconfig/editor/client/src/components/prompt/prompt_input/schema_renderer/PromptInputSchemaRenderer.tsx
rholinshead: DataRenderer is for the data key. Some prompts, like GPT-4V will have both data (string prompt) and attachments (images)

python/src/aiconfig/editor/client/src/components/prompt/prompt_input/schema_renderer/PromptInputAttachmentsSchemaRenderer.tsx
```diff
@@ -68,13 +68,21 @@ export default memo(function PromptInputAttachmentsSchemaRenderer({
...
+    if (index > attachments.length) {
+      updatedAttachments = [...attachments, ...addedAttachments];
+    } else {
+      updatedAttachments = attachments.reduce((acc, attachment, i) => {
+        if (i === index) {
+          // Replace the attachment at the given index with the new attachments
+          return [...acc, ...addedAttachments];
+        }
+        return [...acc, attachment];
+      }, [] as InputAttachment[]);
```
rossdanlm, line 83: Why not just do `attachments[index] = attachment` since we already know that `index <= attachments.length`?

python/src/aiconfig/editor/client/src/components/prompt/prompt_input/schema_renderer/PromptInputAttachmentsSchemaRenderer.tsx
rholinshead: Ah, let me add a comment. This is actually kind of a react trick to force each new EditableAttachmentRenderer to be freshly mounted instead of maintaining state with its previous render (because the attachment now exists and is rendered above)

python/src/aiconfig/editor/client/src/components/prompt/prompt_input/attachments/AttachmentUploader.tsx
```diff
@@ -15,15 +15,65 @@ type Props = {
...
+    method: "POST",
+    mode: "cors",
+    cache: "no-cache",
+    body: formData,
+    headers: {
+      Authorization: "",
+    },
+  });
+
+  if (rawRes.ok && rawRes.status === 201) {
```
rossdanlm, line 67: So it's 201, not 200?

python/src/aiconfig/editor/client/src/components/prompt/prompt_input/attachments/AttachmentUploader.tsx
```diff
@@ -15,15 +15,65 @@ type Props = {
...
+    // Dont really need to parse xml s3 response, just use the keys, etc. that were passed
+    return { url: `${uploadUrl}${uploadKey}` };
+  } else {
+    throw new Error("Error uploading to S3!");
```
rossdanlm, line 71: TODO is add ErrorBoundary right?

python/src/aiconfig/editor/client/src/components/prompt/prompt_input/attachments/AttachmentUploader.tsx
```diff
@@ -15,15 +15,65 @@ type Props = {
...
+  const uploadKey = `uploads/${getTodayDateString()}/${randomPath}/${sanitizeFileName(
+    file.name
+  )}`;
+
+  const formData = new FormData();
+  formData.append("key", uploadKey);
+  formData.append("acl", "public-read");
+  formData.append("Content-Type", file.type);
+  // formData.append("AWSAccessKeyId", policy.AWSAccessKeyId);
+  formData.append("success_action_status", "201");
```
rossdanlm, line 50: Why 201 and not 200?

python/src/aiconfig/editor/client/src/components/prompt/prompt_input/attachments/AttachmentUploader.tsx
```diff
@@ -15,15 +15,65 @@ type Props = {
...
+  // const policy = await policyResponse.json();
+  const uploadUrl = "https://s3.amazonaws.com/lastmileai.aiconfig.public/";
+  const uploadKey = `uploads/${getTodayDateString()}/${randomPath}/${sanitizeFileName(
+    file.name
+  )}`;
+
+  const formData = new FormData();
+  formData.append("key", uploadKey);
+  formData.append("acl", "public-read");
+  formData.append("Content-Type", file.type);
```
rossdanlm, line 48: Cool that this is handled automatically

python/src/aiconfig/editor/client/src/components/prompt/prompt_input/attachments/AttachmentUploader.tsx
```diff
@@ -15,15 +15,65 @@ type Props = {
...
+
+// TODO: Make this configurable for external deployments
+async function uploadFile(file: File): Promise<{ url: string }> {
+  const randomPath = Math.round(Math.random() * 10000);
+  // TODO: Add back once CORS is resolved
+  // const policyResponse = await fetch(
+  //   "https://lastmileai.dev/api/upload/publicpolicy"
+  // );
+  // const policy = await policyResponse.json();
+  const uploadUrl = "https://s3.amazonaws.com/lastmileai.aiconfig.public/";
```
rossdanlm, line 40: I know there's no practical difference, but it feels like it makes sense to include `uploads` at the end of this

It's not a big deal, I get that this is already an endpoint we created so not easily changeable, not a blocker

python/src/aiconfig/editor/client/src/components/prompt/prompt_input/attachments/AttachmentUploader.tsx
```diff
@@ -15,15 +15,65 @@ type Props = {
...
-  };
+// s3 file uris cannot have '+' character, so replace with '_'
+function sanitizeFileName(name: string) {
+  return name.replace(/[_+]/g, "_");
+}
+
+export function getTodayDateString(): string {
+  const date = new Date();
+  const dateString = `${date.getFullYear()}_${
+    date.getMonth() + 1
```
rossdanlm, line 26: month returns 0 for January?

python/src/aiconfig/editor/client/src/components/prompt/prompt_input/attachments/AttachmentUploader.tsx
```diff
@@ -15,15 +15,65 @@ type Props = {
...
+  }_${date.getDate()}`;
+  const timeString = `${date.getUTCHours()}_${date.getUTCMinutes()}_${date.getUTCSeconds()}`;
```
rossdanlm, line 28: Do you feel like adding the timeString for seconds is overkill and we can just use the date instead? Not a blocker

rossdanlm (approved):
It's not too familiar to me, but generally looks good? Idk I would say if any bugs we can fix forward

python/src/aiconfig/editor/client/src/components/prompt/prompt_input/schema_renderer/PromptInputAttachmentsSchemaRenderer.tsx
rholinshead: Hmm, good question. For some reason I was thinking it had to be a new array but would need to revisit later to see if that is the case

python/src/aiconfig/editor/client/src/components/prompt/prompt_input/attachments/AttachmentUploader.tsx
rholinshead: I guess so -- copied from workbooks and it works :p

python/src/aiconfig/editor/client/src/components/prompt/prompt_input/attachments/AttachmentUploader.tsx
rholinshead: No, this throws and is caught in the try/catch below. We can try/catch any errors that happen from async operations like callbacks and handle them nicely with UI notifications, etc. ErrorBoundary is for errors thrown in the render itself

python/src/aiconfig/editor/client/src/components/prompt/prompt_input/attachments/AttachmentUploader.tsx
rholinshead: Copied from lastmile -- not entirely sure

python/src/aiconfig/editor/client/src/components/prompt/prompt_input/attachments/AttachmentUploader.tsx
rholinshead: Copied from lastmile. Ya, it's 0-indexed apparently

Screenshot 2024-01-09 at 11 15 07 PM

python/src/aiconfig/editor/client/src/components/prompt/prompt_input/attachments/AttachmentUploader.tsx
rholinshead: Eh, I think it's fine to be extra safe. We could also look at adding some additional 'unique' identifier based on session or something later

PR #849, opened 2024-01-10
[editor] HuggingFaceText2SpeechTransformerPromptSchema

rholinshead:
# [editor] HuggingFaceText2SpeechTransformerPromptSchema

Screenshot 2024-01-09 at 11 54 12 PM

### DIFF
--- a/python/src/aiconfig/editor/client/src/shared/prompt_schemas/HuggingFaceText2SpeechTransformerPromptSchema.ts
+++ b/python/src/aiconfig/editor/client/src/shared/prompt_schemas/HuggingFaceText2SpeechTransformerPromptSchema.ts
@@ -1,0 +1,108 @@
+import { PromptSchema } from "../../utils/promptUtils";
+
+export const HuggingFaceText2SpeechTransformerPromptSchema: PromptSchema = {
+  // See https://huggingface.co/transformers/v4.11.3/main_classes/model.html and
+  // https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py
+
+  input: {
+    type: "string",
+  },
+  model_settings: {
+    type: "object",
+    properties: {
+      torch_dtype: {
+        type: "string",
+        description: `Override the default torch.dtype and load the model under this dtype. 
+        If "auto" is passed the dtype will be automatically derived from the model’s weights.`,
+      },
+      force_download: {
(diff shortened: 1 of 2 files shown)


python/src/aiconfig/editor/client/src/shared/prompt_schemas/HuggingFaceText2SpeechTransformerPromptSchema.ts
```diff
@@ -0,0 +1,108 @@
...
+  // See https://huggingface.co/transformers/v4.11.3/main_classes/model.html and
+  // https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py
+
+  input: {
+    type: "string",
+  },
+  model_settings: {
+    type: "object",
+    properties: {
+      torch_dtype: {
```
saqadri, line 13: Do we need a "model" property as well to be able to specify the model ID?

saqadri (approved):

python/src/aiconfig/editor/client/src/shared/prompt_schemas/HuggingFaceText2SpeechTransformerPromptSchema.ts
rholinshead: It's not currently supported but if we'll be adding it then I can add it here to require one less change later?

saqadri: I can add it too. I was thinking all HF model setting schemas accept a "model". Then the model parsers use it as appropriate.

rholinshead: I can add to each in next PR

rossdanlm: > I was thinking all HF model setting schemas accept a "model". Then the model parsers use it as appropriate.

This also fits in with this general task of separating out the pipeline generation params from the actual completion params: https://github.com/lastmile-ai/aiconfig/issues/824

cc @Ankush-lastmile

PR #860, opened 2024-01-10
[bug fix][AIC-py] sort of do exception handling during streaming

jonathanlastmileai:
[bug fix][AIC-py] sort of do exception handling during streaming

This piggybacks on the cancellation during stream flow, which works as far as I've tested.
The frontend gets the message.

image

### DIFF
--- a/python/src/aiconfig/editor/server/server.py
+++ b/python/src/aiconfig/editor/server/server.py
@@ -234,12 +234,19 @@ def run() -> FlaskResponse:
         # Use multi-threading so that we don't block run command from
         # displaying the streamed output (if streaming is supported)
         def run_async_config_in_thread():
-            asyncio.run(aiconfig.run(prompt_name=prompt_name, params=params, run_with_dependencies=False, options=inference_options))  # type: ignore
+            try:
+                asyncio.run(aiconfig.run(prompt_name=prompt_name, params=params, run_with_dependencies=False, options=inference_options))  # type: ignore
+            except Exception as e:
+                output_text_queue.put(e)
+
             output_text_queue.put(STOP_STREAMING_SIGNAL)  # type: ignore
 
-        def create_cancellation_payload():
+        def create_error_payload(message: str, code: int):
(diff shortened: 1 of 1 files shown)


python/src/aiconfig/editor/server/server.py
```diff
@@ -307,7 +311,11 @@ def kill_thread(thread_id: int | None):
                     yield from handle_cancellation()
                     return
 
-                if isinstance(text, str):
+                if isinstance(text, Exception):
+                    yield "["
+                    yield json.dumps({"message": "exception", "output": str(text)})
```
saqadri, line 316: The structure of the payload should match the same as what `create_cancellation_payload` does:
```
def create_error_payload():
            aiconfig_json = aiconfig_deep_copy.model_dump(exclude=EXCLUDE_OPTIONS) if aiconfig_deep_copy is not None else None
            return json.dumps({"error": {"message": str(text), "code": 500, "data": aiconfig_json}})
```

python/src/aiconfig/editor/server/server.py
```diff
@@ -234,12 +234,16 @@ def generate(cancellation_token_id: str):  # type: ignore
...
-            asyncio.run(aiconfig.run(prompt_name=prompt_name, params=params, run_with_dependencies=False, options=inference_options))  # type: ignore
+            try:
+                asyncio.run(aiconfig.run(prompt_name=prompt_name, params=params, run_with_dependencies=False, options=inference_options))  # type: ignore
+            except Exception as e:
+                output_text_queue.put(e)
+
             output_text_queue.put(STOP_STREAMING_SIGNAL)  # type: ignore
 
-        def create_cancellation_payload():
+        def create_cancellation_payload(message="The task was cancelled.", code=499):
```
saqadri, line 244:
```suggestion
        def create_cancellation_payload():
            return create_error_payload(message="The task was cancelled.", code=499)
        def create_error_payload(message: str, code: Int):
```

PR #866, opened 2024-01-10
Fix parameterization issues

saqadri:
Fix parameterization issues

Now that we have non-text input model parsers, we can see some issues in our parameterization.

1) ASR and Image-to-Text modelparsers should NOT be `ParameterizedModelParser` instances, since their inputs cannot be parameterized.

2) Parameter resolution logic shouldn't throw errors in the case where it's parsing prompts belonging to regular model parsers.

Test Plan:
Ran the translation prompt for this aiconfig:

```
prompts: [
{
      "name": "image_caption",
      "input": {
        "attachments": [
          {
            "data": "https://s3.amazonaws.com/lastmileai.aiconfig.public/uploads/2024_1_10_18_41_31/1742/bear_with_honey.png",
            "mime_type": "image/png"
          },
          {
            "data": "https://s3.amazonaws.com/lastmileai.aiconfig.public/uploads/2024_1_10_18_41_31/7275/fox_in_forest.png",
            "mime_type": "image/png"
          }
        ]
      },
      "metadata": {
        "model": {
          "name": "HuggingFaceImage2TextTransformer",
          "settings": {
            "max_new_tokens": 10,
            "model": "Salesforce/blip-image-captioning-base"
          }
        },
        "parameters": {}
      },
      "outputs": [
        {
          "output_type": "execute_result",
          "execution_count": 0,
          "data": "a bear sitting on a rock eating honey",
          "metadata": {}
        },
        {
          "output_type": "execute_result",
          "execution_count": 1,
          "data": "a red fox in the woods",
          "metadata": {}
        }
      ]
    },
    {
      "name": "translation",
      "input": "Once upon a time, in a lush and vibrant forest, there lived a magnificent creature known as the Quick Brown Fox. This fox was unlike any other, possessing incredible speed and agility that awed all the animals in the forest. With its fur as golden as the sun and its eyes as sharp as emeralds, the Quick Brown Fox was admired by everyone, from the tiniest hummingbird to the mightiest bear. The fox had a kind heart and would often lend a helping paw to those in need. The Quick Brown Fox had a particular fondness for games and challenges. It loved to test its skills against others, always seeking new adventures to satisfy its boundless curiosity. Its favorite game was called \"The Great Word Hunt,\" where it would embark on a quest to find hidden words scattered across the forest. \n\nOne day it got very old and died",
      "metadata": {
        "model": {
          "name": "HuggingFaceTextTranslationTransformer",
          "settings": {
[...]

PR #863, opened 2024-01-10
Support custom specified models for HF model parsers

saqadri:
Support custom specified models for HF model parsers

The HF prompt schema has been updated in #850 to support a "model" property, which can specify the model to use for a given HF task.

This change updates the model parsers to use this setting.

Test Plan:

Tested with the following prompt:
```
    {
      "name": "image_caption",
      "input": {
        "attachments": [
          {
            "data": "https://s3.amazonaws.com/lastmileai.aiconfig.public/uploads/2024_1_10_18_41_31/1742/bear_with_honey.png",
            "mime_type": "image/png"
          },
          {
            "data": "https://s3.amazonaws.com/lastmileai.aiconfig.public/uploads/2024_1_10_18_41_31/7275/fox_in_forest.png",
            "mime_type": "image/png"
          }
        ]
      },
      "metadata": {
        "model": {
          "name": "HuggingFaceImage2TextTransformer",
          "settings": {
            "max_new_tokens": 10,
            "model": "Salesforce/blip-image-captioning-base"
          }
        },
        "parameters": {}
      },
      "outputs": [
        {
          "output_type": "execute_result",
          "execution_count": 0,
          "data": "a bear sitting on a rock eating honey",
          "metadata": {}
        },
        {
          "output_type": "execute_result",
          "execution_count": 1,
          "data": "a red fox in the woods",
          "metadata": {}
        }
      ]
    },
```

Validated the the model setting was respected and worked.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/863).
* #826
* #866
* __->__ #863

PR #826, opened 2024-01-09
Gradio Local Editor cookbook

saqadri:
Gradio Local Editor cookbook

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/826).
* __->__ #826
* #866
* #863

PR #867, opened 2024-01-10
Default to stream for HF model parsers

saqadri:
Default to stream for HF model parsers

Default stream to True if no inference options are specified.

This is the expected behavior.

Test Plan:
* Tried model parsers without this change --> no streaming

* After this change --> streaming for the same aiconfig

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/867).
* #826
* __->__ #867
* #866
* #863

### DIFF
--- a/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/__init__.py
+++ b/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/__init__.py
@@ -6,7 +6,9 @@ from .local_inference.text_summarization import HuggingFaceTextSummarizationTran
 from .local_inference.text_translation import HuggingFaceTextTranslationTransformer
 from .remote_inference_client.text_generation import HuggingFaceTextGenerationParser
 from .local_inference.automatic_speech_recognition import HuggingFaceAutomaticSpeechRecognitionTransformer
+from .local_inference.util import get_hf_model
 
+UTILS = [get_hf_model]
 
 LOCAL_INFERENCE_CLASSES = [
     "HuggingFaceText2ImageDiffusor",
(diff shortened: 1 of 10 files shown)


rholinshead (approved):

rholinshead (changes requested):
Actually, will this throw if options has no stream callback?

rholinshead (commented):
Should probably do an options.stream_callback existence check before calling it

rossdanlm:
I don't think this should be possible if we're testing from local editor. We purposefully made sure not to have streaming support if `option` isn't provided, because then that means that the InferenceOptions are malformed and not useful (ex: won't have a stream callback defined like what Ryan said). I got streaming to work yesterday on all the HF model parsers when testing using the editor, so feel free to VC and sanity check that everything is hooked up correctly!


## Technical details
Streaming is enabled as true by default from the client if it's not explicitly set in AIConfig or prompt-level settings (`enableStreaming` is undefined from https://github.com/lastmile-ai/aiconfig/blob/main/python/src/aiconfig/editor/client/src/utils/aiconfigStateUtils.ts#L21). Streaming will default to true from the editor: https://github.com/lastmile-ai/aiconfig/blob/a0362540c17a09f8481dda17ff7caf69aabfc7d6/python/src/aiconfig/editor/client/src/LocalEditor.tsx#L81

This then gets read by the server (which also defaults to true anyways if for some reason server params aren't readable) where the option is passed into the run command: https://github.com/lastmile-ai/aiconfig/blob/main/python/src/aiconfig/editor/server/server.py#L224-L238

rossdanlm (changes requested):
See comment above

saqadri:
Spoke offline with Rossdan, going to close this diff -- we will investigate if the issue comes up again

PR #869, opened 2024-01-10
[AIC-py][editor] server: persistent local cfg + telemetry flag, endpoints

jonathanlastmileai:
[AIC-py][editor] server: persistent local cfg + telemetry flag, endpoints

Test:

- Start the server with and without flag and with no file, see that it gets created with correct contents
- start the server with and without flag with existing file, with both true and false
  - hit both endpoints, check responses and that the file contents are changed

https://github.com/lastmile-ai/aiconfig/assets/148090348/b36c3c00-c14b-4cde-8449-60386f2be1eb

### DIFF
--- a/python/src/aiconfig/editor/server/server.py
+++ b/python/src/aiconfig/editor/server/server.py
@@ -15,6 +15,7 @@ import result
 from aiconfig.Config import AIConfigRuntime
 from aiconfig.editor.server.queue_iterator import STOP_STREAMING_SIGNAL, QueueIterator
 from aiconfig.editor.server.server_utils import (
+    AIConfigRC,
     EditServerConfig,
     FlaskResponse,
     HttpResponseWithAIConfig,
@@ -57,7 +58,7 @@ app = Flask(__name__, static_url_path="")
 CORS(app, resources={r"/api/*": {"origins": "*"}})
 
 
-def run_backend_server(edit_config: EditServerConfig) -> Result[str, str]:
+def run_backend_server(edit_config: EditServerConfig, aiconfigrc_path: str) -> Result[str, str]:
     LOGGER.setLevel(edit_config.log_level)
     LOGGER.info("Edit config: %s", edit_config.model_dump_json())
     LOGGER.info(f"Starting server on http://localhost:{edit_config.server_port}")
@@ -68,7 +69,7 @@ def run_backend_server(edit_config: EditServerConfig) -> Result[str, str]:
         LOGGER.warning(f"Failed to open browser: {e}. Please open http://localhost:{port} manually.")
 
     app.server_state = ServerState()  # type: ignore
-    res_server_state_init = init_server_state(app, edit_config)
+    res_server_state_init = init_server_state(app, edit_config, aiconfigrc_path)
     match res_server_state_init:
         case Ok(_):
             LOGGER.info("Initialized server state")
@@ -566,3 +567,34 @@ def clear_outputs() -> FlaskResponse:
 
     signature: dict[str, Type[Any]] = {}
     return run_aiconfig_operation_with_request_json(aiconfig, request_json, f"method_", _op, signature)
+
+
+@app.route("/api/get_aiconfigrc", methods=["GET"])
+def get_aiconfigrc() -> FlaskResponse:
+    state = get_server_state(app)
+
+    yaml_mapping: Result[AIConfigRC, str] = core_utils.read_text_file(state.aiconfigrc_path).and_then(AIConfigRC.from_yaml)
+    match yaml_mapping:
+        case Ok(yaml_mapping_ok):
+            return FlaskResponse((yaml_mapping_ok.model_dump(), 200))
+        case Err(e):
+            return FlaskResponse(({"message": f"Failed to load aiconfigrc: {e}"}, 400))
+
+
+@app.route("/api/set_aiconfigrc", methods=["POST"])
+def set_aiconfigrc() -> FlaskResponse:
+    state = get_server_state(app)
+    request_json = request.get_json()
+    # TODO:
+    # We might not need to implement this at all.
+    #
+    # If so:
+    # Assuming request_json["aiconfigrc"] is a yaml-formatted string
+    # (possibly with comments)
+    # Note that the file might already exist and have contents.
+    #
+    # here's how to write it to a file:
+    # from ruamel.yaml import YAML
+    # yaml = YAML()
+    # with open(state.aiconfigrc_path, "w") as f:
+    #     yaml.dump(request_json["aiconfigrc"], f)
--- a/python/src/aiconfig/editor/server/server_utils.py
+++ b/python/src/aiconfig/editor/server/server_utils.py
@@ -6,9 +6,10 @@ import sys
 import typing
 from dataclasses import dataclass, field
 from enum import Enum
+from textwrap import dedent
+from threading import Event
 from types import ModuleType
 from typing import Any, Callable, NewType, Type, TypeVar, cast
-from threading import Event
 
 import lastmile_utils.lib.core.api as core_utils
 import result
@@ -17,6 +18,7 @@ from aiconfig.registry import ModelParserRegistry
 from flask import Flask
 from pydantic import field_validator
 from result import Err, Ok, Result
+from ruamel.yaml import YAML
 
 from aiconfig.schema import Prompt, PromptMetadata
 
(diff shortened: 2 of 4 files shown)


rossdanlm:
Can you add screenshot/video/text outputs from test plan? Thanks

python/src/aiconfig/scripts/aiconfig_cli.py
```diff
@@ -7,13 +7,16 @@
...
+import yaml
 from aiconfig.editor.server.server import run_backend_server
 from aiconfig.editor.server.server_utils import EditServerConfig, ServerMode
 from result import Err, Ok, Result
 
 
 class AIConfigCLIConfig(core_utils.Record):
     log_level: str | int = "WARNING"
+    config_path: str = "aiconfig_cli.config.yaml"
+    allow_usage_data_sharing: bool = False
```
Ankush-lastmile, line 19: is this supposed to be used anywhere in this pr? I don't see it being referenced anywhere

python/src/aiconfig/scripts/aiconfig_cli.py
```diff
@@ -46,12 +49,13 @@ def run_subcommand(argv: list[str]) -> Result[str, str]:
...
         LOGGER.debug(f"{res_edit_config.is_ok()=}")
-        res_servers = res_edit_config.and_then(_run_editor_servers)
         out: Result[str, str] = result.do(
             #
             Ok(",".join(res_servers_ok))
             #
-            for res_servers_ok in res_servers
+            for edit_config in res_edit_config
+            for cli_config in res_cli_config
+            for res_servers_ok in _run_editor_servers(edit_config, cli_config.config_path)
```
Ankush-lastmile, line 58: trying to understand this code block, what is going on here? Do `edit_config` and `cli_config` get used?

python/src/aiconfig/editor/server/server.py
```diff
@@ -57,7 +58,7 @@
 CORS(app, resources={r"/api/*": {"origins": "*"}})
 
 
-def run_backend_server(edit_config: EditServerConfig) -> Result[str, str]:
+def run_backend_server(edit_config: EditServerConfig, cli_config_path: str) -> Result[str, str]:
```
rossdanlm, line 61: Can we call this `editor_config_path`. It will contain settings that aren't only specific to CLI

python/src/aiconfig/scripts/aiconfig_cli.py
```diff
@@ -116,6 +120,24 @@ def _run_editor_servers(edit_config: EditServerConfig) -> Result[list[str], str]
...
+    file_contents = yaml.dump({"allow_usage_data_sharing": cli_config.allow_usage_data_sharing})
+    try:
+        with open(config_path, "x") as f:
+            f.write(file_contents)
+    except FileExistsError:
+        with open(config_path, "r") as f:
+            existing_file_contents = f.read()
+
+        existing_config = yaml.safe_load(existing_file_contents)
+        existing_config["allow_usage_data_sharing"] = cli_config.allow_usage_data_sharing
```
Ankush-lastmile, line 134: what happens if `cli_config.allow_usage_data_sharing` isn't defined?

python/src/aiconfig/editor/server/server.py
```diff
@@ -566,3 +567,48 @@ def _op(aiconfig_runtime: AIConfigRuntime, _op_args: OpArgs) -> Result[None, str
...
     return run_aiconfig_operation_with_request_json(aiconfig, request_json, f"method_", _op, signature)
+
+
+@app.route("/api/get_cli_config", methods=["GET"])
+def get_cli_config() -> FlaskResponse:
+    state = get_server_state(app)
+
+    def _yaml_deserialize(yaml_str: str) -> Result[dict[str, Any], str]:
+        try:
+            out = yaml.safe_load(yaml_str)
```
Ankush-lastmile, line 578: is this `out` the yaml data in string form? and that's what is returned by the api?

python/src/aiconfig/scripts/aiconfig_cli.py
jonathanlastmileai: What do you mean by isnt defined?

python/src/aiconfig/scripts/aiconfig_cli.py
jonathanlastmileai: Yes they get passed into _run_editor_servers. Specifically cli_config.config_path.

python/src/aiconfig/scripts/aiconfig_cli.py
jonathanlastmileai: _process_cli_config()

python/src/aiconfig/editor/server/server.py
jonathanlastmileai: sure

python/src/aiconfig/editor/server/server.py
jonathanlastmileai: Yeah it's a dict, i.e. the python equivalent of a json object.

The flow goes _yaml_deserialize (last step on line 586) then if it's a success, return http response with that object, otherwise return an err. see 589 and 591

python/src/aiconfig/scripts/aiconfig_cli.py
rossdanlm: It defaults to false in `aiconfig_cli.py`

python/src/aiconfig/scripts/aiconfig_cli.py
jonathanlastmileai: I'm confused

python/src/aiconfig/scripts/aiconfig_cli.py
rholinshead: Should we default it to true and only set it to false only if the user specifies false? My 2c here:
- if we default it to false and rely on the user to explicitly set it to true, we will never get metrics. People won't go out of their way just to enable logging but anyone who cares enough will do the extra (easy) manual step to prevent it
- in our case, we want to make it easy enough for users to opt out manually and we're not hiding the fact that we are obtaining metrics. IMO this is a good enough standard of "trust" for this open souce project

jonathanlastmileai: Personally, I would be annoyed if it was on by default unless it made it very clear and easy to turn off. I think we need a UI dialog either way. If it defaults to on, it will be unclear, and if it defaults to off we'll never get metrics like you said.

python/src/aiconfig/editor/server/server_utils.py
```diff
@@ -75,6 +75,7 @@ def convert_to_mode(cls, value: Any) -> ServerMode:  # pylint: disable=no-self-a
 
 @dataclass
 class ServerState:
+    aiconfigrc_path: str = ".aiconfigrc"
```
Ankush-lastmile, line 78: add $HOME to path

python/src/aiconfig/scripts/aiconfig_cli.py
```diff
@@ -116,6 +149,28 @@ def _run_editor_servers(edit_config: EditServerConfig) -> Result[list[str], str]
 
 def _process_cli_config(cli_config: AIConfigCLIConfig) -> Result[bool, str]:
     LOGGER.setLevel(cli_config.log_level)
+    try:
```
Ankush-lastmile, line 152: doc string would help

Ankush-lastmile (approved):
discussed new changes offline, logic flow makes sense to me

python/src/aiconfig/scripts/aiconfig_cli.py
```diff
@@ -114,8 +128,43 @@ def _run_editor_servers(edit_config: EditServerConfig) -> Result[list[str], str]
...
+
+    It returns Ok(True) if everything went well. Currently, it never returns Ok(False).
+    As usual, we return an error with a message if something went wrong.
+    """
+    aiconfigrc_path = cli_config.aiconfigrc_path
+
     LOGGER.setLevel(cli_config.log_level)
+    try:
+        with open(aiconfigrc_path, "x") as f:
+            YAML().dump(DEFAULT_AICONFIGRC, f)
```
jonathanlastmileai, line 145: @rossdanlm, if we check that the file doesn't exist and then open it for writing, it's possible for another process to write the file in between those two steps.

rossdanlm: Ahhhhhhh, I see. Can you pls comment this in the code in #870 so that this is clear?

PR #870, opened 2024-01-10
[AIC-py] cli rage

jonathanlastmileai:
[AIC-py] cli rage

This just tells the user how to open an issue. I don't want to do anything automatically
because (a) user trust, and (b) it's much harder to implement without additional bugs.

This is mostly bedside manner. If the user is raging, they are angry.
Troll them a little bit to lighten the mood first.

https://github.com/lastmile-ai/aiconfig/assets/148090348/2369306f-3af7-49f5-96bb-0ac9be323b51

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/870).
* #903
* __->__ #870
* #869

### DIFF
--- a/python/src/aiconfig/scripts/aiconfig_cli.py
+++ b/python/src/aiconfig/scripts/aiconfig_cli.py
@@ -13,6 +13,8 @@ from aiconfig.editor.server.server import run_backend_server
 from aiconfig.editor.server.server_utils import DEFAULT_AICONFIGRC, EditServerConfig, ServerMode
 from result import Err, Ok, Result
 
+import aiconfig.scripts.rage.rage as rage
+
 
 class AIConfigCLIConfig(core_utils.Record):
     log_level: str | int = "WARNING"
@@ -36,7 +38,7 @@ async def main_with_args(argv: list[str]) -> int:
 
 def run_subcommand(argv: list[str]) -> Result[str, str]:
     LOGGER.info("Running subcommand")
-    subparser_record_types = {"edit": EditServerConfig}
+    subparser_record_types = {"edit": EditServerConfig, "rage": rage.RageConfig}
     main_parser = core_utils.argparsify(AIConfigCLIConfig, subparser_record_types=subparser_record_types)
 
     # Try to parse the CLI args into a config.
@@ -57,6 +59,14 @@ def run_subcommand(argv: list[str]) -> Result[str, str]:
         LOGGER.debug(f"{edit_config.is_ok()=}")
         out = _run_editor_servers_with_configs(edit_config, cli_config)
         return out
+    elif subparser_name == "rage":
+        res_rage_config = core_utils.parse_args(main_parser, argv[1:], rage.RageConfig)
+        res_rage = res_rage_config.and_then(rage.rage)
+        match res_rage:
+            case Ok(msg):
+                return Ok(f"Rage complete: {msg}")
+            case Err(e):
+                return Err(e)
     else:
         return Err(f"Unknown subparser: {subparser_name}")
 
@@ -143,6 +153,8 @@ def _set_log_level_and_create_default_yaml(cli_config: AIConfigCLIConfig) -> Res
     try:
         with open(aiconfigrc_path, "x") as f:
             YAML().dump(DEFAULT_AICONFIGRC, f)
+
+        return Ok(True)
     except FileExistsError:
         try:
 
@@ -159,13 +171,12 @@ def _set_log_level_and_create_default_yaml(cli_config: AIConfigCLIConfig) -> Res
                     if k not in contents:
                         contents[k] = v
 
-                YAML().dump(contents, f)
+                YAML().dump(contents, f)  # type: ignore[yaml library untyped]
+                return Ok(True)
         except Exception as e:
             return core_utils.ErrWithTraceback(e)
     except Exception as e:
         return core_utils.ErrWithTraceback(e)
-
-    return Ok(True)
 
 
 def _run_frontend_server_background() -> Result[list[subprocess.Popen[bytes]], str]:
--- a/python/src/aiconfig/scripts/rage/rage.py
+++ b/python/src/aiconfig/scripts/rage/rage.py
@@ -1,0 +1,75 @@
+import os
+import lastmile_utils.lib.core.api as core_utils
+import sys
+import time
+import itertools
+import numpy as np
+from result import Ok, Result
+
+logger = core_utils.get_logger(__name__)
+
+
+class RageConfig(core_utils.Record):
+    log_level: str | int = "WARNING"
+
+
+def rage(config: RageConfig) -> Result[None, str]:
+    logger.setLevel(config.log_level)
+    print("Raging...")
+    spin(2)
+    print("Please hold. Your call is important to us.\nA representative will be with you shortly.")
+    spin(3, type="music")
+    print("Looking for your server logs...")
+    spin(4, type="music")
+    print("Turning up the heat...")
+    spin(2)
+    print("If I had a dollar for every time I've seen this error...")
+    spin(5)
+    print("I'm glad we're finally spending time together.")
+    spin(4, type="music")
+    print("Please continue holding. We appreciate your continued support, or whatever.")
+    spin(3, type="music")
+
+    print("\n\n\n\n............\n\n")
+    print("Please open an issue! :) Here are some tips on how to do that:")
+    for logfile in ["editor_flask_server.log", "aiconfig.log"]:
+        print()
+        if os.path.exists(logfile):
+            print(f"Found {logfile}! Please include its contents in your bug report.")
+            print(f"Full path: {os.path.abspath(logfile)}")
+        else:
+            print(f"No {logfile} found. This might be another bug :)")
+            print("For now, please include your terminal output in your bug report.")
+
+    print("\nPlease run the following commands and also include their output:")
+    print("\nwhich pip; which pip3; which python; which python3; pip3 list | grep aiconfig; python --version; python3 --version")
+
+    print("\nPlease open an issue here: https://github.com/lastmile-ai/aiconfig/issues/new")
+    print("Our sincerest apologies and gratitude. We will comment on the issue as soon as possible.")
+    print("\n\n")
+    print("Done raging! :)")
+    print("\n\n")
+
+    return Ok(None)
+
+
+def spin(seconds: int, type: str = "spinner"):
+    """This is just for fun."""
+
+    assert type in ["spinner", "music"]
+
+    spinning = itertools.cycle(["-", "/", "|", "\\"])
+
+    def get_animation():
+        if type == "spinner":
+            return next(spinning)
+        else:
+            return "".join(np.random.choice(["♩", "♫", "♬", "♪"]) for _ in range(5))
+
+    end_time = time.time() + seconds
+
+    while time.time() < end_time:
+        sys.stdout.write(get_animation())
+        sys.stdout.flush()
+        time.sleep(0.1)
+        sys.stdout.write("\b" * 5)


python/src/aiconfig/editor/server/server.py
```diff
@@ -57,7 +58,7 @@
 CORS(app, resources={r"/api/*": {"origins": "*"}})
 
 
-def run_backend_server(edit_config: EditServerConfig) -> Result[str, str]:
+def run_backend_server(edit_config: EditServerConfig, cli_config_path: str) -> Result[str, str]:
```
rossdanlm, line 61: These two names are confusing since at surface level they seem to be very similar functionality. Can we call `cli_config_path` something like `settings.yaml`, or modify `edit_config` to be something more specific like `server_settings`?

python/src/aiconfig/editor/server/server.py
```diff
@@ -566,3 +567,48 @@ def _op(aiconfig_runtime: AIConfigRuntime, _op_args: OpArgs) -> Result[None, str
...
+        try:
+            out = yaml.safe_load(yaml_str)
+            if out is None:
+                return Err("YAML string not parsed")
+            else:
+                return Ok(out)
+        except Exception as e:
+            return core_utils.ErrWithTraceback(e)
+
+    cli_config_object = core_utils.read_text_file(state.cli_config_path).and_then(_yaml_deserialize)
```
rossdanlm, line 586: Cool is `and_then` another python quirk?

python/src/aiconfig/editor/server/server.py
```diff
@@ -566,3 +567,48 @@ def _op(aiconfig_runtime: AIConfigRuntime, _op_args: OpArgs) -> Result[None, str
...
+
+@app.route("/api/get_cli_config", methods=["GET"])
+def get_cli_config() -> FlaskResponse:
+    state = get_server_state(app)
+
+    def _yaml_deserialize(yaml_str: str) -> Result[dict[str, Any], str]:
+        try:
+            out = yaml.safe_load(yaml_str)
+            if out is None:
+                return Err("YAML string not parsed")
```
rossdanlm, line 580: I know we chain this, but can since `_yaml_deserialize` is only used it one place, can we call this something like `_parse_yaml_file(filename: str)` and return a more user friendly error?

`Err(f"Cannot parse string from YAML file '{filename}' into YAML format")

python/src/aiconfig/editor/server/server.py
```diff
@@ -566,3 +567,48 @@ def _op(aiconfig_runtime: AIConfigRuntime, _op_args: OpArgs) -> Result[None, str
...
+        case Err(e):
+            return FlaskResponse(({"message": f"failed to get cli_config: {e}", "data": None}, 400))
+
+
+@app.route("/api/set_cli_config", methods=["POST"])
+def set_cli_config() -> FlaskResponse:
+    state = get_server_state(app)
+    request_json = request.get_json()
+    try:
+        new_contents = request_json.get("new_contents")
```
rossdanlm, line 599: nit: I think `contents` is fine, but easy fix not a blocker

python/src/aiconfig/editor/server/server.py
```diff
@@ -566,3 +567,48 @@ def _op(aiconfig_runtime: AIConfigRuntime, _op_args: OpArgs) -> Result[None, str
...
+        serialized = yaml.dump(new_contents)
```
rossdanlm, line 600: Can we pls call this something else? Like `yaml_contents`

python/src/aiconfig/editor/server/server_utils.py
```diff
@@ -75,6 +75,7 @@ def convert_to_mode(cls, value: Any) -> ServerMode:  # pylint: disable=no-self-a
 
 @dataclass
 class ServerState:
+    cli_config_path: str = "aiconfig_cli.config.yaml"
```
rossdanlm, line 78: I don't want to use `cli`, especially since we already have an `aiconfig_cli.py`, it's overloaded. Let's use a more clear name instead like `settings_path`

python/src/aiconfig/editor/server/server.py
jonathanlastmileai: nope, it's a Result quirk :)

https://github.com/rustedpy/result

In turn it's basically borrowed from Haskell. [Some of the theory](https://wiki.haskell.org/Monad)

python/src/aiconfig/scripts/aiconfig_cli.py
```diff
@@ -33,7 +37,7 @@ async def main_with_args(argv: list[str]) -> int:
 
 def run_subcommand(argv: list[str]) -> Result[str, str]:
     LOGGER.info("Running subcommand")
-    subparser_record_types = {"edit": EditServerConfig}
+    subparser_record_types = {"edit": EditServerConfig, "rage": rage.RageConfig}
```
rossdanlm, line 40: Q: why is this called `subparser_...`?

python/src/aiconfig/scripts/aiconfig_cli.py
```diff
@@ -46,14 +50,33 @@ def run_subcommand(argv: list[str]) -> Result[str, str]:
...
         return out
+    elif subparser_name == "rage":
+        LOGGER.debug("Running rage subcommand")
+        print("Raging...")
+        rage.spin(2)
+        print("Please hold. Your call is important to us.\nA representative will be with you shortly.")
+        rage.spin(3, type="music")
+        print("Turning up the heat...")
+        rage.spin(2)
+        print("Looking for your flipping server logs...")
```
rossdanlm, line 70: lol I would remove flipping becuase at this point users are already probably annoyed if submitting rage request

python/src/aiconfig/scripts/aiconfig_cli.py
```diff
@@ -46,14 +50,33 @@ def run_subcommand(argv: list[str]) -> Result[str, str]:
...
         LOGGER.debug(f"{res_edit_config.is_ok()=}")
-        res_servers = res_edit_config.and_then(_run_editor_servers)
         out: Result[str, str] = result.do(
             #
             Ok(",".join(res_servers_ok))
             #
-            for res_servers_ok in res_servers
+            for edit_config in res_edit_config
+            for cli_config in res_cli_config
+            for res_servers_ok in _run_editor_servers(edit_config, cli_config.config_path)
```
rossdanlm, line 59: What does this do?

python/src/aiconfig/scripts/aiconfig_cli.py
```diff
@@ -100,7 +123,7 @@ def _run_editor_servers(edit_config: EditServerConfig) -> Result[list[str], str]
             return Err(e)
 
     results: list[Result[str, str]] = []
-    backend_res = run_backend_server(edit_config)
+    backend_res = run_backend_server(edit_config, cli_config_path)
```
rossdanlm, line 126: nit: I really do not like using `res`. Can we pls use `response` or `result` so it's clear which it is?

python/src/aiconfig/scripts/aiconfig_cli.py
```diff
@@ -46,14 +50,33 @@ def run_subcommand(argv: list[str]) -> Result[str, str]:
...
+        print("Please hold. Your call is important to us.\nA representative will be with you shortly.")
+        rage.spin(3, type="music")
+        print("Turning up the heat...")
+        rage.spin(2)
+        print("Looking for your flipping server logs...")
+        rage.spin(4, type="music")
+        print("Do you know how many Zuckerbucks I should be charging for this?")
+        rage.spin(5)
+        print("I'm glad we're finally spending time together.")
+        rage.spin(4, type="music")
```
rossdanlm, line 75: Does this end up repeating after we get to the end?

python/src/aiconfig/scripts/aiconfig_cli.py
jonathanlastmileai: Each one defines the struct that a particular subcommand corresponds to. Each subcommand gets one struct, and that subcommand then automatically gets arguments corresponding to the fields in that struct, which are then automatically populated.

python/src/aiconfig/scripts/aiconfig_cli.py
```diff
@@ -46,14 +50,33 @@ def run_subcommand(argv: list[str]) -> Result[str, str]:
...
+
+        res_rage_config = core_utils.parse_args(main_parser, argv[1:], rage.RageConfig)
```
rossdanlm, line 77: Ok so this adds the "rage" arg to the rage.RageConfig object? https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.parse_args

python/src/aiconfig/scripts/aiconfig_cli.py
jonathanlastmileai: This is admittedly arcane syntax that unpacks the result of _run_editor_servers (if it's an Ok). You can think of it kind of like a Result comprehension, sort of like a list comprehension that flattens a 2d list.

Does that answer the question? What would you name differently?

python/src/aiconfig/scripts/aiconfig_cli.py
jonathanlastmileai: Sure, actually we should pick a convention that doesn't do the "maybe_..." thing that @rholinshead pointed out.

One suggestion:

```
backend_server_outcome = ... 
match backend_server_outcome:
    Ok(backend_server_outcome_ok)
        ... 
    Err(e)
        ... 

```

python/src/aiconfig/scripts/aiconfig_cli.py
```diff
@@ -46,14 +50,33 @@ def run_subcommand(argv: list[str]) -> Result[str, str]:
...
+        res_rage = res_rage_config.and_then(rage.rage)
```
rossdanlm, line 78: Why are we making `rage.rage()` a class method which requires us to create a `res_rage_config` instead of using a static method? Is it becuase we want to re-use state/object in case user calls multiple rages in the same session?

python/src/aiconfig/scripts/aiconfig_cli.py
jonathanlastmileai: Repeating what?

python/src/aiconfig/scripts/rage/rage.py
```diff
@@ -0,0 +1,60 @@
...
+            print(f"Full path: {os.path.abspath(logfile)}")
+        else:
+            print(f"No {logfile} found. This might be another bug :)")
+            print("For now, please include your terminal output in your bug report.")
+
+    print("\nPlease run the following commands and also include their output:")
+    print("\nwhich pip; which pip3; which python; which python3; pip3 list | grep aiconfig; python --version; python3 --version")
+
+    print("\nPlease open an issue here: https://github.com/lastmile-ai/aiconfig/issues/")
+    print("Our sincerest gratitude. We will comment on the issue as soon as possible.")
```
rossdanlm, line 33: I think `sincerest gratitute` is kind of overkill, especially with `Checking your flipping server logs` the tone isn't really consistent

python/src/aiconfig/scripts/rage/rage.py
```diff
@@ -0,0 +1,60 @@
...
+
+class RageConfig(core_utils.Record):
+    log_level: str | int = "WARNING"
+
+
+def rage(config: RageConfig) -> Result[None, str]:
+    logger.setLevel(config.log_level)
+    print("\n\n\n\n............\n\n")
+    print("Please open an issue! :) Here are some tips on how to do that:")
+    for logfile in ["editor_flask_server.log", "aiconfig.log"]:
```
rossdanlm, line 20: Why do we need to do two logfiles? I think this could be confusing to users having to submit 2 separate logfiles + issues

rossdanlm: Would be better to merge together

python/src/aiconfig/scripts/rage/rage.py
```diff
@@ -0,0 +1,60 @@
...
+        print()
+        if os.path.exists(logfile):
+            print(f"Found {logfile}! Please include its contents in your bug report.")
+            print(f"Full path: {os.path.abspath(logfile)}")
+        else:
+            print(f"No {logfile} found. This might be another bug :)")
+            print("For now, please include your terminal output in your bug report.")
+
+    print("\nPlease run the following commands and also include their output:")
+    print("\nwhich pip; which pip3; which python; which python3; pip3 list | grep aiconfig; python --version; python3 --version")
```
rossdanlm, line 30: Can be followup diff, but think we should do this ourselves and add to one of the logfiles

python/src/aiconfig/scripts/aiconfig_cli.py
```diff
@@ -116,6 +139,24 @@ def _run_editor_servers(edit_config: EditServerConfig) -> Result[list[str], str]
...
     LOGGER.setLevel(cli_config.log_level)
+
+    config_path = cli_config.config_path
+    file_contents = yaml.dump({"allow_usage_data_sharing": cli_config.allow_usage_data_sharing})
+    try:
+        with open(config_path, "x") as f:
+            f.write(file_contents)
+    except FileExistsError:
+        with open(config_path, "r") as f:
+            existing_file_contents = f.read()
```
rossdanlm, line 150: I get what you're doing, but I feel "x" is pretty niche, and it's more intuitive to do 

1. Read first from file
2. If FileNotFoundError, create a new one with the "w" command

python/src/aiconfig/scripts/aiconfig_cli.py
jonathanlastmileai: See do notation in the readme: https://github.com/rustedpy/result

python/src/aiconfig/scripts/aiconfig_cli.py
```diff
@@ -116,6 +139,24 @@ def _run_editor_servers(edit_config: EditServerConfig) -> Result[list[str], str]
 
 def _process_cli_config(cli_config: AIConfigCLIConfig) -> Result[bool, str]:
```
rossdanlm, line 140: Like I said, not crazy about the name, but we can update later

python/src/aiconfig/scripts/aiconfig_cli.py
jonathanlastmileai: This is a handy wrapper that uses argparse under the hood to return a struct directly.

It can be made a little clearer: e.g. we have to manually give argv[1:] and specify the config type twice, once inside subparser_record_types and once in `parse_args()`.

python/src/aiconfig/scripts/aiconfig_cli.py
jonathanlastmileai: >  we want to re-use state/object

I would never do such a thing - how dare you.

it's just a module with a function. rage module has rage function. 

`def rage(config: RageConfig) -> Result[None, str]:
`

python/src/aiconfig/scripts/rage/rage.py
jonathanlastmileai: I think we should probably remove the trolling

python/src/aiconfig/scripts/aiconfig_cli.py
```diff
@@ -116,6 +139,24 @@ def _run_editor_servers(edit_config: EditServerConfig) -> Result[list[str], str]
...
+            f.write(file_contents)
+    except FileExistsError:
+        with open(config_path, "r") as f:
+            existing_file_contents = f.read()
+
+        existing_config = yaml.safe_load(existing_file_contents)
+        existing_config["allow_usage_data_sharing"] = cli_config.allow_usage_data_sharing
+        file_contents = yaml.dump(existing_config)
+        with open(config_path, "w") as f:
+            f.write(file_contents)
```
rossdanlm, line 156: Sorry what is this doing exactly in this function? Taking a step back out of thie catch block, we are:

1. Check if file exists --> if it does, use that param
2. If file does not exist --> write to that file

^Would this be cleaner to do? is there logic that I'm missing? Can we also have brief docstrings to explain this? Thanks

python/src/aiconfig/scripts/rage/rage.py
jonathanlastmileai: Yeah this is just laziness / it's a P1 to write the server logs to aiconfig.log. Once we do that, we can just use the one log file. I think it's acceptable for a first cut

python/src/aiconfig/scripts/rage/rage.py
jonathanlastmileai: We could, but it's really easy for the user to just copy/paste this one-liner. I don't want to create room for new bugs during the flow for reporting bugs. It should be simple reliable and predictable.

python/src/aiconfig/scripts/rage/rage.py
```diff
@@ -0,0 +1,60 @@
...
+import time
+import itertools
+import numpy as np
+from result import Ok, Result
+
+logger = core_utils.get_logger(__name__)
+
+
+class RageConfig(core_utils.Record):
+    log_level: str | int = "WARNING"
```
rossdanlm, line 13: Do we need this? I don't see it being used

Or is it an abstract field we need to override for `core_utils.Record`?

python/src/aiconfig/scripts/aiconfig_cli.py
jonathanlastmileai: Why? This is actually the CLI config and the job is to process the entire CLI config. It does more than handle the settings path.

python/src/aiconfig/scripts/rage/rage.py
```diff
@@ -0,0 +1,60 @@
...
+
+    return Ok(None)
+
+
+def spin(seconds: int, type: str = "spinner"):
+    """This is just for fun."""
+
+    assert type in ["spinner", "music"]
+
+    spinning = itertools.cycle(["-", "/", "|", "\\"])
```
rossdanlm, line 59: This is cool!

Also typo on last "\\"?

python/src/aiconfig/scripts/rage/rage.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def get_animation():
+        if type == "spinner":
+            return next(spinning)
+        else:
+            return "".join(np.random.choice(["♩", "♫", "♬", "♪"]) for _ in range(5))
```
rossdanlm, line 65: Let's not use `np` if we can avoid and just use regular `from Random import random`. I get that it's in C so slightly faster, but this isn't needed

rossdanlm: Also lol this is pretty not-necessary at all, but it would be cute if we kept playing musical notes as we wait:

```
import numpy as np
from time import sleep

music_notes = [np.random.choice(["♩", "♫", "♬", "♪"]) for _ in range(50)]
# print(f"{music_notes=}")

for note in music_notes:
    sleep(0.5)
    print(note, end="")
```

https://github.com/lastmile-ai/aiconfig/assets/151060367/bd629e63-6d92-4b4f-a8aa-b5495dd9be86

python/src/aiconfig/scripts/rage/rage.py
```diff
@@ -0,0 +1,73 @@
...
+
+    end_time = time.time() + seconds
+
+    while time.time() < end_time:
+        sys.stdout.write(get_animation())
+        sys.stdout.flush()
+        time.sleep(0.1)
```
rossdanlm, line 72: Will sleep get in the way of the other rage operations?

python/src/aiconfig/editor/server/server.py
rossdanlm: lol fun
Screenshot 2024-01-10 at 21 33 50

python/src/aiconfig/scripts/aiconfig_cli.py
rossdanlm: Sorry just to clarify, what are subcommands, how do we anticipate client to pass this into to server?

python/src/aiconfig/scripts/aiconfig_cli.py
rossdanlm: Let me take a look and read to understand, I've never seen 3 for loops list comprehension within the same inline (still not super familiar how inline tabs work in Python tbh, so it's just not clear to me what parts are nested or not)

python/src/aiconfig/scripts/aiconfig_cli.py
rossdanlm: Ohhhhhh, so we're just like spending 20s trolling the people who are upset and want to rage shake? We're not actually "waiting" on anything? lmao ok uh.... hmmmmmmmm, I don't know if that's the best user experience. Either way I think we can remove if we really want to and just get this unblocked for now

python/src/aiconfig/scripts/aiconfig_cli.py
rossdanlm: Why does this cause a race condition?

python/src/aiconfig/scripts/rage/rage.py
rossdanlm: Ok I realize now this isn't actually waiting on anything, we're just purposefully blue-balling the user lmao. I feel we should probably delete but we can do that in fix forward, not a blocker for now

python/src/aiconfig/scripts/rage/rage.py
jonathanlastmileai: I'm going to use this in a subsequent PR. It's nice to have every subcommand have a controllable logging level.

python/src/aiconfig/scripts/rage/rage.py
jonathanlastmileai: ?

python/src/aiconfig/scripts/rage/rage.py
jonathanlastmileai: @rossdanlm , I get that numpy is heavier than we need here but it's very handy and easy to use - it's not about efficiency in this case. Let's not block on that

python/src/aiconfig/scripts/rage/rage.py
jonathanlastmileai: Yeah I'm just trolling the user

python/src/aiconfig/scripts/aiconfig_cli.py
jonathanlastmileai: This is #869, but the race condition is:
1. check that file exists, it does not
2. File gets written
3. Overwrite file unintentionally

python/src/aiconfig/scripts/aiconfig_cli.py
jonathanlastmileai: #869 , going to resolve this

python/src/aiconfig/scripts/aiconfig_cli.py
jonathanlastmileai: Counterintuitively, I actually think this might be better UX than just doing it. I really think `rage` might be mostly bedside manner and a little engineering.

python/src/aiconfig/scripts/rage/rage.py
```diff
@@ -0,0 +1,75 @@
...
+    logger.setLevel(config.log_level)
+    print("Raging...")
+    spin(2)
+    print("Please hold. Your call is important to us.\nA representative will be with you shortly.")
+    spin(3, type="music")
+    print("Looking for your server logs...")
+    spin(4, type="music")
+    print("Turning up the heat...")
+    spin(2)
+    print("If I had a dollar for every time I've seen this error...")
```
saqadri, line 26: Just wanted to mention I love this kind of playfulness. I hope we do more of this in the future! It brings a bit of personality and a piece of ourselves in the products we build. Really nice idea @jonathanlastmileai !

python/src/aiconfig/scripts/aiconfig_cli.py
rossdanlm: Oh thanks, this example makes it super clear to me!

python/src/aiconfig/scripts/rage/rage.py
jonathanlastmileai: Glad you like it! I felt like this was a risk. I volunteer to server as the LastMile [Secretary of Keeping it Real](https://getyarn.io/yarn-clip/b357e424-7ff7-4fbc-9b2c-607531a46a92).

python/src/aiconfig/scripts/aiconfig_cli.py
rossdanlm: > Overwrite file unintentionally

How come this overwrites the file untintentionally if it doesn't exist before?

python/src/aiconfig/scripts/rage/rage.py
rossdanlm: Can we move it to subsequent PR then until it's needed? Keeps things easier to logic together and review

rossdanlm: If you want we can keep the line just commented out

python/src/aiconfig/scripts/rage/rage.py
rossdanlm: Oh nm, \ is a special char

python/src/aiconfig/scripts/aiconfig_cli.py
jonathanlastmileai: This is unrelated to 870. I'll move this to 869.

python/src/aiconfig/scripts/rage/rage.py
jonathanlastmileai: I actually agree now and I did a lot of this in #903

python/src/aiconfig/scripts/rage/rage.py
jonathanlastmileai: Wait, I am using it

python/src/aiconfig/scripts/aiconfig_cli.py
rossdanlm: Questions answered in https://github.com/lastmile-ai/aiconfig/pull/869#discussion_r1450809345, makes sense even though it's an edge case

rossdanlm (approved):
Thanks for syncing on all questions. Please add comment in code why we need to use `"x"` filemode to avoid potential race condition

jonathanlastmileai:
> Thanks for syncing on all questions. Please add comment in code why we need to use `"x"` filemode to avoid potential race condition

added https://docs.google.com/document/d/1YwUlIQZX8CiuvtbVR4OvpFQ2nphFITRD_XZR3CDNtoE/edit#bookmark=id.8h0jqqaolx9z

PR #874, opened 2024-01-11
[docs] updated readme for local editor

tanya-rai:
[docs] updated readme for local editor

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,17 +26,21 @@
   </a>
 </p>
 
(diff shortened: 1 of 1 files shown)


rossdanlm (approved):
In the future, I think we should showcase the power of Multimodal formats and use the Gradio cookbook (or even replace that to HuggingFace cookbook with that one tbh). That looks way cooler and more exciting

PR #880, opened 2024-01-11
Revert "[HF][fix] Use ParameterizedModelParser instead of ModelParser"

rossdanlm:
Reverts lastmile-ai/aiconfig#877

Whoops, ok so I realized that the reason we use parameterizedModelParser is that sometimes you can't add parameters to inputs (ex: image and audio). This means that the fix instead should be that the ModelParser.run() method should accept generic `**kwargs` so it doesn't error. Omg wow what a mistake, the 1 time I do not get approval for review and it causes a SEV

Error that is now solved
Screenshot 2024-01-11 at 04 49 30


rossdanlm:
Fixed in https://github.com/lastmile-ai/aiconfig/pull/881

PR #881, opened 2024-01-11
Revert "[HF][fix] Use ParameterizedModelParser instead of ModelParser"

rossdanlm:
Revert "[HF][fix] Use ParameterizedModelParser instead of ModelParser"

Reverts https://github.com/lastmile-ai/aiconfig/pull/877

This caused a bug because image and audio inputs can't be parametrized. See https://github.com/lastmile-ai/aiconfig/pull/882 for a better fix instead

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/881).
* #885
* #883
* #882
* __->__ #881
* #879


saqadri:
Yes, important to be reverted

saqadri (approved):

PR #882, opened 2024-01-11
[HF][fix] Allow `kwargs` in `ModelParser.run()`

rossdanlm:
[HF][fix] Allow `kwargs` in `ModelParser.run()`

See https://github.com/lastmile-ai/aiconfig/pull/877 for context, and https://github.com/lastmile-ai/aiconfig/pull/880 for why that original fix needed to be reverted

Just a quick hack to get unblocked. I tried originally to make the ASR and Image2Text ParameterizedModel but that caused other errors.

## Test Plan

Before

https://github.com/lastmile-ai/aiconfig/assets/151060367/a23a5d5c-d9a2-415b-8a6e-9826da56e985

After

https://github.com/lastmile-ai/aiconfig/assets/151060367/f29580e9-5cf6-43c5-b848-bb525eb368f2

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/882).
* #885
* #883
* __->__ #882
* #881
* #879

### DIFF
--- a/python/src/aiconfig/default_parsers/parameterized_model_parser.py
+++ b/python/src/aiconfig/default_parsers/parameterized_model_parser.py
@@ -45,7 +45,7 @@ class ParameterizedModelParser(ModelParser):
         aiconfig: AIConfig,
         options: Optional[InferenceOptions] = None,
         parameters: Dict = {},
-        **kwargs,
+        **kwargs, #TODO: We should remove and make arguments explicit
     ) -> List[Output]:
         # maybe use prompt metadata instead of kwargs?
         if kwargs.get("run_with_dependencies", False):
--- a/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/automatic_speech_recognition.py
+++ b/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/automatic_speech_recognition.py
@@ -85,7 +85,7 @@ class HuggingFaceAutomaticSpeechRecognitionTransformer(ModelParser):
         await aiconfig.callback_manager.run_callbacks(CallbackEvent("on_deserialize_complete", __name__, {"output": completion_data}))
         return completion_data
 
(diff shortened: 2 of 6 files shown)


saqadri (commented):
We can fix on top but this is a critical API we are adding this to.

rossdanlm:
> We can fix on top but this is a critical API we are adding this to.

The problem is this currently breaks though so we need this

saqadri:
I looked at the code and I understand what the issue has been -- until now **all** our model parsers have been ParameterizedModelParser instances -- that one accepts kwargs in `run`. It stands to reason that the base class should also accept `kwargs`. Your change makes things more consistent, so thanks for the fix! And we can improve on this further. `Config.run` also accepts kwargs, and we should ensure every derived model parser accepts them too.

saqadri (approved):

python/src/aiconfig/default_parsers/parameterized_model_parser.py
```diff
@@ -45,7 +45,7 @@ async def run(
         aiconfig: AIConfig,
         options: Optional[InferenceOptions] = None,
         parameters: Dict = {},
-        **kwargs,
+        **kwargs, #TODO: We should remove and make arguments explicit
```
rossdanlm, line 48: We need to add `run_with_dependencies` arg here otherwise it will fail. Ex: https://github.com/lastmile-ai/aiconfig/blob/b7b9910cc7abbd9b08dbdfe634e25f98396afcef/python/src/aiconfig/editor/server/server.py#L381

It's weird that your test plan showed this working, this should've failed. Can you do `pip list | grep aiconfig` and make sure that your aiconfig is linked to local and not the production package? If it's not seeing `python-aiconfig` to a local path for you, you need to go to `aiconfig/python` and run `pip install -e .`

Issue #822, opened 2024-01-09
Fix serialize in all HF model parsers to take in model id from data param


rossdanlm:
Ankush pointed out that we're passing in the model parser id instead of the model id, which will cause `serialize()` to crash: https://github.com/lastmile-ai/aiconfig/blob/main/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/text_2_image.py#L229

Instead we should be passing in the model from the data param and using that

We should also be building automated testing for serialize (similar to Dall-E) so that we don't run into this issue again

rossdanlm:
Similarly, should support the "model" param arg for deserialize(): https://github.com/lastmile-ai/aiconfig/pull/883#discussion_r1448874623

rossdanlm:
Note that ASR __call__ supports `model` param: https://github.com/lastmile-ai/aiconfig/pull/883#discussion_r1449043026

PR #883, opened 2024-01-11
[HF][fix] Small nit fixes to ASR: Remove "model" from completion params, get_output_text in right format

rossdanlm:
[HF][fix] Small nit fixes to ASR: Remove "model" from completion params, get_output_text in right format

Getting an error saying that "model" param is defined twice, which it is because we're passing it through completion params and again through the explicit model param. This along with the change in  https://github.com/lastmile-ai/aiconfig/pull/879 will ensure that the right model of `None` is passed if the value is empty string.

I also fixed typo in the aiconfig.json file

## Test Plan

I also had to change the data field from `"./hi.mp3"` to `"cookbooks/Gradio/hi.mp3"` becuase I can only run the config from the high level `aiconfig` directory due to this line when running the editor: https://github.com/lastmile-ai/aiconfig/blob/b632094618c8f96b186b1849d0036e1313e02c5a/python/src/aiconfig/scripts/aiconfig_cli.py#L126

After

https://github.com/lastmile-ai/aiconfig/assets/151060367/7387706f-deea-414b-b429-732c78322615

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/883).
* #885
* __->__ #883
* #882
* #881
* #879

### DIFF
--- a/cookbooks/Gradio/huggingface.aiconfig.json
+++ b/cookbooks/Gradio/huggingface.aiconfig.json
@@ -214,14 +214,13 @@
       },
       "metadata": {
         "model": {
-          "name": "HuggingFaceAutomaticSpeechRecognition",
+          "name": "HuggingFaceAutomaticSpeechRecognitionTransformer",
           "settings": {
             "model": "openai/whisper-small"
           }
         },
         "parameters": {}
-      },
-      "outputs": []
+      }
     }
   ],
   "$schema": "https://json.schemastore.org/aiconfig-1.0"
--- a/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/automatic_speech_recognition.py
+++ b/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/automatic_speech_recognition.py
@@ -85,7 +85,7 @@ class HuggingFaceAutomaticSpeechRecognitionTransformer(ModelParser):
         await aiconfig.callback_manager.run_callbacks(CallbackEvent("on_deserialize_complete", __name__, {"output": completion_data}))
         return completion_data
 
-    async def run(self, prompt: Prompt, aiconfig: "AIConfigRuntime", options: InferenceOptions, parameters: Dict[str, Any]) -> list[Output]:
+    async def run(self, prompt: Prompt, aiconfig: "AIConfigRuntime", options: InferenceOptions, parameters: Dict[str, Any], **kwargs) -> list[Output]:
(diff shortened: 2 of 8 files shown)


cookbooks/Gradio/huggingface.aiconfig.json
```diff
@@ -207,21 +207,20 @@
       "input": {
         "attachments": [
           {
-            "data": "./hi.mp3",
+            "data": "cookbooks/Gradio/hi.mp3",
```
rossdanlm, line 210: For testing I had to change this line becuase I could only run from top level `aiconfig` dir

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/automatic_speech_recognition.py
```diff
@@ -198,7 +196,6 @@ def refine_pipeline_creation_params(model_settings: Dict[str, Any]) -> List[Dict
     """
 
     supported_keys = {
-        "model",
```
saqadri, line 201: Not blocking for this diff but removing this would now mean deserialize won't return "model" back, so in the case where it's called by itself without calling "run", no model setting would be present.

I think the alternative is to also call get_hf_model in deserialize and pass it in to this function.

But that can be done at top of stack

rossdanlm: I don't think any of the other HF model parsers have "model" as a supported key, because that's technically not one of the completion params passed into the HF `__call__` methods. Added a note of this in https://github.com/lastmile-ai/aiconfig/issues/822

saqadri: Yep, understood. However this specific HF __call__ method does accept `model`, so `refine_completion_params` for this model parser should have it as a supported key, and set it properly based on the model name/settings

saqadri (approved):
Approving to unblock, let's fix comments on top!

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/automatic_speech_recognition.py
rossdanlm: Got it, made a note of that in the issue above

PR #885, opened 2024-01-11
Use facebook/mms-tts-eng instead of suno/bark in Gradio cookbook for Text2Speech

rossdanlm:
Use facebook/mms-tts-eng instead of suno/bark in Gradio cookbook for Text2Speech

Bark is ~5GB and took my laptop 30mins to download for the first time. The Facebook MMS (https://huggingface.co/facebook/mms-tts-eng?fbclid=IwAR22U5VSaZ_V7QW4TJe77CM7f4fsTHjqCMjgLKV31wUl66ww1vXYSrEl-MA) is only ~100MB which is way better and at least for cookbook I feel much more worth it to quickly test things out. Yes the output quality is way more memey and "voice AI" stereotype, but I feel the tradeoff is worth it

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/885).
* __->__ #885
* #883
* #882
* #881
* #879

### DIFF
--- a/cookbooks/Gradio/huggingface.aiconfig.json
+++ b/cookbooks/Gradio/huggingface.aiconfig.json
@@ -16,7 +16,7 @@
     "default_model": "stevhliu/my_awesome_billsum_model",
     "model_parsers": {
       "Salesforce/blip-image-captioning-base": "HuggingFaceImage2TextTransformer",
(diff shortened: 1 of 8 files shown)


saqadri (approved):
Excellent idea

PR #888, opened 2024-01-11
Updated CHANGELOG for v1.1.11

rossdanlm:
Updated CHANGELOG for v1.1.11

TSIA, doing this so we can package a new release!

I messed up v1.1.4 for typescript sorry. I also messed up and didn't realize I needed to run `yarn build` from within the editor dir itself, so need to repackage python to v1.1.11


rholinshead:
Do we need to do major/minor due to some of the bigger changes like output type, etc?

PR #890, opened 2024-01-11
[editor] Run Build on Main Before Publishing

rholinshead:
# [editor] Run Build on Main Before Publishing

When we publish the editor we need to make sure it's built with the latest code otherwise we're getting a blank screen.

Ran `yarn build` in `/aiconfig/python/src/aiconfig/editor/client`
then `aiconfig edit --aiconfig-path python/src/aiconfig/editor/travel.aiconfig.json --server-mode='prod'` and see the editor UI at localhost:8080 now


rossdanlm (approved):
Oh thanks, did not know this

Let's also update the OSS release note procedure! https://www.notion.so/lastmile-ai/OSS-Release-process-Publishing-Guides-1260b4199f3946ff8abf5713f7c3fffc

rossdanlm:
Updated the OSS release note procedure, but for future would be good to add to automated script too

Screenshot 2024-01-11 at 12 39 00

PR #891, opened 2024-01-11
[editor][server] fix save

Ankush-lastmile:
[editor][server] fix save

on creating a new config, server does not persist file path into server state, nor does it update the AIConfigRuntime with file_path

Quick fix: set aiconfig.file_path to the path in server, when creating the New AIconfig/AIConfigRuntime

## Testplan
1. aiconfig edit on a config that doesn't exist

https://github.com/lastmile-ai/aiconfig/assets/141073967/99008b27-0ff3-4891-b9eb-c1b826c6ecab

### DIFF
--- a/python/src/aiconfig/editor/server/server_utils.py
+++ b/python/src/aiconfig/editor/server/server_utils.py
@@ -221,6 +221,7 @@ def init_server_state(app: Flask, edit_config: EditServerConfig) -> Result[None,
     else:
         LOGGER.info(f"Creating new AIConfig at {edit_config.aiconfig_path}")
         aiconfig_runtime = AIConfigRuntime.create()  # type: ignore
+        aiconfig_runtime.file_path = edit_config.aiconfig_path # type: ignore
         model_ids = ModelParserRegistry.parser_ids()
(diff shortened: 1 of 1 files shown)


python/src/aiconfig/editor/server/server_utils.py
```diff
@@ -77,6 +77,7 @@ def convert_to_mode(cls, value: Any) -> ServerMode:  # pylint: disable=no-self-a
 class ServerState:
     aiconfig: AIConfigRuntime | None = None
     events: dict[str, Event] = field(default_factory=dict)
+    file_path: str | None = None
```
jonathanlastmileai, line 80: We don't need to store two copies of this, especially if one can be None

jonathanlastmileai (changes requested):

python/src/aiconfig/editor/server/server_utils.py
Ankush-lastmile: sorry added that in while trying to understand. Removed

rossdanlm (approved):
nice

Ankush-lastmile:
#892

PR #896, opened 2024-01-11
[editor] Only stream output, not entire AIConfig

rossdanlm:
[editor] Only stream output, not entire AIConfig

This will be a bug for anything that streams multiple return outputs, but for now we don't have that and all our models which support streaming only output a single output

## Test Plan
Ok wow yea, even on Chrome this is *noticeably* faster!

https://github.com/lastmile-ai/aiconfig/assets/151060367/3613c2e5-8a8c-43c8-994d-d7ae62469594

### DIFF
--- a/python/src/aiconfig/schema.py
+++ b/python/src/aiconfig/schema.py
@@ -1,3 +1,4 @@
+import json
 import warnings
 from typing import Any, Dict, List, Literal, Optional, Union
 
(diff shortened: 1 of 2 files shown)


python/src/aiconfig/schema.py
```diff
@@ -89,6 +90,13 @@ class ExecuteResult(BaseModel):
     mime_type: Optional[str] = None
     # Output metadata
     metadata: Dict[str, Any]
+    
+    def to_json(self) -> JSONObject:
+        """
+        Helper method used to ensure this is formatted to a valid JSON object
+        so that it can be used for streaming
+        """
+        return self.__dict__
```
saqadri, line 99: should this be model.dumps?

saqadri (approved):

python/src/aiconfig/schema.py
rossdanlm: Good suggestion!

PR #864, opened 2024-01-10
[editor] Starting Point for Telemetry

Ankush-lastmile:
[editor] Starting Point for Telemetry

rebased version of #804 onto main

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/864).
* #899
* __->__ #864

PR #899, opened 2024-01-11
[editor][client] Enable Telemetry based on User Config settings

Ankush-lastmile:
[editor][client] Enable Telemetry based on User Config settings

This diff builds on top of @rholinshead's starting point for telemetry diff. Most of that looked good to me so I didn't touch it beside @rossdanlm 's .

- Building off of #869, if `allow_usage_data_sharing` is set to True, initialize Datadog Browser logging with a session ID.
- Disabled telemetry in dev mode as per @rholinshead 's comment that the hot reload spams the logs with something like "datadog logging already initialized"
- Moved the initialization logic away from`index.tsx` into `editor.tsx` so that it can be configurable.

## Testplan:

1. yarn build
2. run in "prod" mode
3. Edit AIconfig
4. Hit Save button -> validate datadog log sent

repeat with aiconfig.rc set with false logging

|Screenshot 2024-01-12 at 7 24 33 PM  | Screenshot 2024-01-12 at 7 23 51 PM |
|
Screenshot 2024-01-12 at 7 26 35 PM  |  -> No logs |

I tried taking a video but datadog doesn't immediately update with logs which made the video too long to upload

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/899).
* __->__ #899
* #864

### DIFF
--- a/python/src/aiconfig/editor/client/src/LocalEditor.tsx
+++ b/python/src/aiconfig/editor/client/src/LocalEditor.tsx
@@ -5,13 +5,7 @@ import AIConfigEditor, {
   RunPromptStreamErrorEvent,
 } from "./components/AIConfigEditor";
 import { Flex, Loader, MantineProvider, Image } from "@mantine/core";
-import {
-  AIConfig,
-  InferenceSettings,
-  JSONObject,
-  Output,
-  Prompt,
-} from "aiconfig";
+import { AIConfig, InferenceSettings, JSONObject, Output, Prompt } from "aiconfig";
 import { useCallback, useEffect, useMemo, useState } from "react";
 import { ufetch } from "ufetch";
 import { ROUTE_TABLE } from "./utils/api";
@@ -30,6 +24,34 @@ export default function Editor() {
   useEffect(() => {
     loadConfig();
   }, [loadConfig]);
+
+  const setupTelemetryIfAllowed = useCallback(async () => {
+    const isDev = (process.env.NODE_ENV ?? "development") === "development";
+  
+    // Don't enable telemetry in dev mode because hot reload will spam the logs. 
+    if (isDev) {
+      return;
+    }
+  
+    const res = await ufetch.get(ROUTE_TABLE.GET_AICONFIGRC, {});
+
+    const enableTelemetry = res.allow_usage_data_sharing;
+
+    if (enableTelemetry) {
+      datadogLogs.init({
+        clientToken: "pub356987caf022337989e492681d1944a8",
+        env: process.env.NODE_ENV ?? "development",
+        service: "aiconfig-editor",
+        site: "us5.datadoghq.com",
+        forwardErrorsToLogs: true,
+        sessionSampleRate: 100,
+      });
+    }
+  }, []);
+
+  useEffect(() => {
+    setupTelemetryIfAllowed();
+  }, [setupTelemetryIfAllowed]);
 
   const save = useCallback(async (aiconfig: AIConfig) => {
     const res = await ufetch.post(ROUTE_TABLE.SAVE, {
--- a/python/src/aiconfig/editor/client/src/LocalEditor.tsx
+++ b/python/src/aiconfig/editor/client/src/LocalEditor.tsx
@@ -16,6 +16,8 @@ import { useCallback, useEffect, useMemo, useState } from "react";
 import { ufetch } from "ufetch";
 import { ROUTE_TABLE } from "./utils/api";
 import { streamingApiChain } from "./utils/oboeHelpers";
+import { datadogLogs } from "@datadog/browser-logs";
+import { LogEvent, LogEventData } from "./shared/types";
 
 export default function Editor() {
   const [aiconfig, setAiConfig] = useState<AIConfig | undefined>();
@@ -171,6 +173,14 @@ export default function Editor() {
     return await ufetch.get(ROUTE_TABLE.SERVER_STATUS);
   }, []);
 
+  const logEvent = useCallback((event: LogEvent, data?: LogEventData) => {
+    try {
+      datadogLogs.logger.info(event, data);
+    } catch (e) {
+      // Ignore logger errors for now
+    }
+  }, []);
+
   const callbacks: AIConfigCallbacks = useMemo(
     () => ({
       addPrompt,
@@ -179,6 +189,7 @@ export default function Editor() {
       deletePrompt,
       getModels,
       getServerStatus,
+      logEvent,
       runPrompt,
       save,
       setConfigDescription,
@@ -194,6 +205,7 @@ export default function Editor() {
       deletePrompt,
       getModels,
       getServerStatus,
+      logEvent,
       runPrompt,
       save,
       setConfigDescription,
--- a/python/src/aiconfig/editor/client/src/components/AIConfigContext.tsx
+++ b/python/src/aiconfig/editor/client/src/components/AIConfigContext.tsx
@@ -1,5 +1,5 @@
 import { createContext } from "react";
-import { ClientAIConfig } from "../shared/types";
+import { ClientAIConfig, LogEvent, LogEventData } from "../shared/types";
 
 /**
  * Context for overall editor config state. This context should
@@ -7,6 +7,7 @@ import { ClientAIConfig } from "../shared/types";
  */
 const AIConfigContext = createContext<{
   getState: () => ClientAIConfig;
+  logEvent?: (event: LogEvent, data?: LogEventData) => void;
 }>({
   getState: () => ({ prompts: [], _ui: { isDirty: false } }),
 });
--- a/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
+++ b/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
@@ -170,7 +170,7 @@ export default function EditorContainer({
   const stateRef = useRef(aiconfigState);
   stateRef.current = aiconfigState;
 
-  const logEvent = callbacks.logEvent;
+  const logEventCallback = callbacks.logEvent;
 
   const saveCallback = callbacks.save;
   const onSave = useCallback(async () => {
@@ -520,7 +520,7 @@ export default function EditorContainer({
       };
 
       dispatch(action);
-      logEvent?.("ADD_PROMPT", { model, promptIndex });
+      logEventCallback?.("ADD_PROMPT", { model, promptIndex });
 
       try {
         const serverConfigRes = await addPromptCallback(
@@ -542,7 +542,7 @@ export default function EditorContainer({
         });
       }
     },
-    [addPromptCallback, logEvent]
+    [addPromptCallback, logEventCallback]
   );
 
   const deletePromptCallback = callbacks.deletePrompt;
@@ -767,9 +767,9 @@ export default function EditorContainer({
   const contextValue = useMemo(
     () => ({
       getState,
-      logEvent,
+      logEvent: logEventCallback,
     }),
-    [getState, logEvent]
+    [getState, logEventCallback]
   );
 
   const isDirty = aiconfigState._ui.isDirty !== false;
@@ -879,7 +879,7 @@ export default function EditorContainer({
                 loading={isSaving}
                 onClick={() => {
                   onSave();
-                  logEvent?.("SAVE_BUTTON_CLICKED");
+                  logEventCallback?.("SAVE_BUTTON_CLICKED");
                 }}
                 disabled={!isDirty}
                 size="xs"
--- a/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
+++ b/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
@@ -31,6 +31,8 @@ import { v4 as uuidv4 } from "uuid";
 import aiconfigReducer, { AIConfigReducerAction } from "./aiconfigReducer";
 import {
   ClientPrompt,
+  LogEvent,
+  LogEventData,
   aiConfigToClientConfig,
   clientConfigToAIConfig,
   clientPromptToAIConfigPrompt,
@@ -98,6 +100,7 @@ export type AIConfigCallbacks = {
   deletePrompt: (promptName: string) => Promise<void>;
   getModels: (search: string) => Promise<string[]>;
   getServerStatus?: () => Promise<{ status: "OK" | "ERROR" }>;
+  logEvent?: (event: LogEvent, data?: LogEventData) => void;
   runPrompt: (
     promptName: string,
     onStream: RunPromptStreamCallback,
@@ -166,6 +169,8 @@ export default function EditorContainer({
 
   const stateRef = useRef(aiconfigState);
   stateRef.current = aiconfigState;
+
+  const logEvent = callbacks.logEvent;
 
   const saveCallback = callbacks.save;
   const onSave = useCallback(async () => {
@@ -515,6 +520,7 @@ export default function EditorContainer({
       };
 
       dispatch(action);
+      logEvent?.("ADD_PROMPT", { model, promptIndex });
 
       try {
         const serverConfigRes = await addPromptCallback(
@@ -536,7 +542,7 @@ export default function EditorContainer({
         });
       }
     },
-    [addPromptCallback, dispatch]
+    [addPromptCallback, logEvent]
   );
 
   const deletePromptCallback = callbacks.deletePrompt;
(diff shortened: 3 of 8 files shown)


python/src/aiconfig/editor/client/src/LocalEditor.tsx
```diff
@@ -31,6 +26,31 @@ export default function Editor() {
...
+        site: "us5.datadoghq.com",
+        forwardErrorsToLogs: true,
+        sessionSampleRate: 100,
+      });
+
+      const sessionID = generateUniqueSessionId();
+
+      datadogLogs.setGlobalContext({ session_id: sessionID });
+    }
+  }, []);
```
Ankush-lastmile, line 48: setGlobalContext will add this as an attribute to every log sent

Ankush-lastmile: also should clientToken be a constant somewhere? was unsure if that should be the case

rossdanlm:
> This diff builds on top of @rholinshead's starting point for telemetry diff.

Can you link diff?

python/src/aiconfig/editor/client/src/LocalEditor.tsx
```diff
@@ -29,6 +26,31 @@ export default function Editor() {
...
   }, [loadConfig]);
 
+  const enableTelemetryIfAllowed = useCallback(async () => {
+    const res = await ufetch.post(ROUTE_TABLE.LOAD, {}); // change this to the endpoint used to get the config
+
+    const enableTelemetry = res.allow_usage_data_sharing;
+
+    if (enableTelemetry) {
+      datadogLogs.init({
+        clientToken: "pub356987caf022337989e492681d1944a8",
```
rossdanlm, line 36: What's this?

Ankush-lastmile: datadog browser logging uses this token for authentication

python/src/aiconfig/editor/client/src/LocalEditor.tsx
```diff
@@ -29,6 +26,31 @@ export default function Editor() {
...
+        env: process.env.NODE_ENV ?? "development",
+        service: "aiconfig-editor",
+        site: "us5.datadoghq.com",
+        forwardErrorsToLogs: true,
+        sessionSampleRate: 100,
```
rossdanlm, line 41: In diff summary, can you link to website for API docs so we can read more on what these mean?

python/src/aiconfig/editor/client/src/LocalEditor.tsx
Ankush-lastmile: its generated in datadog (done in the datadog ui probably by ryan)

python/src/aiconfig/editor/client/src/utils/logging.ts
```diff
@@ -0,0 +1,6 @@
+import * as uuid from "uuid";
+
+
+export function generateUniqueSessionId(): string {
+    return uuid.v4();
```
rossdanlm, line 5: Oh nice, just like what fb users

python/src/aiconfig/editor/client/src/utils/logging.ts
```diff
@@ -0,0 +1,6 @@
...
+}
```
rossdanlm, line 6: nit: add new line

python/src/aiconfig/editor/client/src/utils/logging.ts
```diff
@@ -0,0 +1,6 @@
+import * as uuid from "uuid";
+
```
rossdanlm, line 2: nit: is it normal to have two empty lines or just 1?

python/src/aiconfig/editor/client/src/shared/types.ts
```diff
@@ -61,3 +61,6 @@ export function aiConfigToClientConfig(aiconfig: AIConfig): ClientAIConfig {
     },
   };
 }
+
+export type LogEvent = "ADD_PROMPT" | "SAVE_BUTTON_CLICKED";
```
rossdanlm, line 65: Nice, good start

python/src/aiconfig/editor/client/src/LocalEditor.tsx
```diff
@@ -174,6 +196,14 @@ export default function Editor() {
     return await ufetch.get(ROUTE_TABLE.SERVER_STATUS);
   }, []);
 
+  const logEvent = useCallback((event: LogEvent, data?: LogEventData) => {
+    try {
+      datadogLogs.logger.info(event, data);
+    } catch (e) {
+      // Ignore logger errors for now
+    }
```
rossdanlm, line 204: Does this get handled automatically if we remove try-catch by `forwardErrorsToLogs` field?

python/src/aiconfig/editor/client/src/LocalEditor.tsx
```diff
@@ -29,6 +26,31 @@ export default function Editor() {
     loadConfig();
   }, [loadConfig]);
 
+  const enableTelemetryIfAllowed = useCallback(async () => {
```
rossdanlm, line 29: nit: `enable` implies that we're turning on a boolean flag. Let's change this to `setupTelemetryIfAllowed`

python/src/aiconfig/editor/client/src/shared/types.ts
```diff
@@ -61,3 +61,6 @@ export function aiConfigToClientConfig(aiconfig: AIConfig): ClientAIConfig {
...
+export type LogEventData = JSONObject;
```
rossdanlm, line 66: not a blocker, but let's add an issue to schematize the logging payload

python/src/aiconfig/editor/client/src/LocalEditor.tsx
```diff
@@ -174,6 +196,14 @@ export default function Editor() {
     return await ufetch.get(ROUTE_TABLE.SERVER_STATUS);
   }, []);
 
+  const logEvent = useCallback((event: LogEvent, data?: LogEventData) => {
+    try {
+      datadogLogs.logger.info(event, data);
```
rossdanlm, line 201: So if telemetry is false, this is going to error? Instead of try catch I think just checking for non-null and no-op is better? We can stil wrap the actual `datadogLogs.logger.info(event, data)` call in a try-catch

python/src/aiconfig/editor/client/src/components/AIConfigContext.tsx
```diff
@@ -1,12 +1,13 @@
...
-import { ClientAIConfig } from "../shared/types";
+import { ClientAIConfig, LogEvent, LogEventData } from "../shared/types";
 
 /**
  * Context for overall editor config state. This context should
  * be memoized to prevent unnecessary re-renders
  */
 const AIConfigContext = createContext<{
   getState: () => ClientAIConfig;
+  logEvent?: (event: LogEvent, data?: LogEventData) => void;
```
rossdanlm, line 10: Why do we need to pass this in? Does it get called anywhere in this PR?

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
```diff
@@ -771,8 +777,9 @@ export default function EditorContainer({
   const contextValue = useMemo(
     () => ({
       getState,
+      logEvent,
```
rossdanlm, line 780: Oh I see, it's passed into here from the editor wrapper

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
```diff
@@ -171,6 +174,8 @@ export default function EditorContainer({
   const stateRef = useRef(aiconfigState);
   stateRef.current = aiconfigState;
 
+  const logEvent = callbacks.logEvent;
```
rossdanlm, line 177: nit: a bit confusing since this isn't the `logEvent`, it's a `logEventCallback`, can we rename it to that (as well as prop name within `AIConfigContext`)?

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
```diff
@@ -878,7 +885,10 @@ export default function EditorContainer({
               <Button
                 leftIcon={<IconDeviceFloppy />}
                 loading={isSaving}
-                onClick={onSave}
+                onClick={() => {
+                  onSave();
+                  logEvent?.("SAVE_BUTTON_CLICKED");
```
rossdanlm, line 890: Edit: Oh nm I see that the `data` field is optional: `data?: LogEventData`. Makes sense

** Old **
Do we need to pass in empty event data here?

```
logEvent?.("SAVE_BUTTON_CLICKED", {})
```

rossdanlm (approved):
Generally lgtm, just some comments and nits

python/src/aiconfig/editor/client/src/LocalEditor.tsx
Ankush-lastmile: I manually tested this and this doesn't error, it just becomes a no-op

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
rossdanlm: Ok so in general how FB did logging, we'd have a field for UI component/surface, and then another field for EVENT (generated, clicked, etc).

But yea not a blocker, just FYI, this is good for now

python/src/aiconfig/editor/client/src/LocalEditor.tsx
```diff
@@ -31,6 +26,31 @@ export default function Editor() {
     loadConfig();
   }, [loadConfig]);
 
+  const enableTelemetryIfAllowed = useCallback(async () => {
+    const res = await ufetch.post(ROUTE_TABLE.LOAD, {}); // change this to the endpoint used to get the config
```
rholinshead, line 30: I think we probably will want to skip logging in dev (should check before sending the request) now that it's being initialized within the component mount -- just looked and it tries to initialize a ton, seems to be from hot reloading:

Screenshot 2024-01-11 at 10 43 17 PM

Ankush-lastmile:
- removed custom sessionID because @rholinshead pointed out that datadog already passes those in by default.
- added in check for aiconfigrc
- addressed some nits

python/src/aiconfig/editor/client/src/utils/logging.ts
Ankush-lastmile: removed custom id

python/src/aiconfig/editor/client/src/LocalEditor.tsx
Ankush-lastmile: Two seperate things.

This try catch ignores any errors by datadog logging module.

`forwardErrorsToLogs` forwards any `console.warn` to logs

python/src/aiconfig/editor/client/src/LocalEditor.tsx
Ankush-lastmile: updated, thanks for the suggestion

python/src/aiconfig/editor/client/src/shared/types.ts
Ankush-lastmile: added a todo

Ankush-lastmile: are we tracking editor issues with github issues?

python/src/aiconfig/editor/client/src/components/AIConfigContext.tsx
Ankush-lastmile: I think this was answered by yourself so resolving

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
Ankush-lastmile: updated per your suggestion

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
Ankush-lastmile: thanks for the insight!

python/src/aiconfig/editor/client/src/LocalEditor.tsx
Ankush-lastmile: updated to only initialize when in "prod" mode as per your suggestion

Ankush-lastmile:
> Can you link diff?

Its the diff below this one, sapling created a new pr

python/src/aiconfig/editor/client/src/LocalEditor.tsx
Ankush-lastmile: whoops forgot to link docs. Generally, datadog docs are easily googleable and easily navigate able. Can just google "datadog <search term>" 

"datadog browser logging" ->  https://docs.datadoghq.com/logs/log_collection/javascript/

python/src/aiconfig/editor/client/src/LocalEditor.tsx
```diff
@@ -29,6 +25,29 @@ export default function Editor() {
     loadConfig();
   }, [loadConfig]);
 
+  const setupTelemetryIfAllowed = useCallback(async () => {
+    const res = await ufetch.get(ROUTE_TABLE.GET_AICONFIGRC, {}); // change this to the endpoint used to get the config
```
rholinshead, line 29:
```suggestion
    const res = await ufetch.get(ROUTE_TABLE.GET_AICONFIGRC, {});
```
can remote comment now. nit, don't think {} is needed, either

python/src/aiconfig/editor/client/src/LocalEditor.tsx
```diff
@@ -29,6 +25,29 @@ export default function Editor() {
...
+
+    const enableTelemetry = res.allow_usage_data_sharing;
+    const isDev = (process.env.NODE_ENV ?? "development") === "development";
```
rholinshead, line 32: Let's do the dev check at the top so we don't need to make a request unnecessarily

python/src/aiconfig/editor/client/src/shared/types.ts
rholinshead: We should, TODOs were mainly from when it was pre-release

python/src/aiconfig/editor/client/src/utils/api.ts
```diff
@@ -14,6 +14,7 @@ export const ROUTE_TABLE = {
   CANCEL: urlJoin(API_ENDPOINT, "/cancel"),
   CLEAR_OUTPUTS: urlJoin(API_ENDPOINT, "/clear_outputs"),
   DELETE_PROMPT: urlJoin(API_ENDPOINT, "/delete_prompt"),
+  GET_AICONFIGRC: urlJoin(API_ENDPOINT, "/get_aiconfigrc"),
```
rholinshead, line 17: What does rc mean?

rholinshead (approved):
LGTM, but let's do the dev check & return before sending the request unnecessarily

python/src/aiconfig/editor/client/src/utils/api.ts
Ankush-lastmile: fairly certain that `rc` stands for run commands. Most people will have multiple of these on their system already, you can see which rc files you have on your own computer by running this in terminal `ls -a $HOME | grep rc`.

To define if telemetry is to be disabled, user will need to define their own `.aiconfigrc`. Will need to add this to docs as well

python/src/aiconfig/editor/client/src/LocalEditor.tsx
Ankush-lastmile: Checked for isDev before making the request, and if so return early.

Manually tested the flows and verified telemetry is only sent in prod and not dev

python/src/aiconfig/editor/client/src/LocalEditor.tsx
Ankush-lastmile: removed comment

python/src/aiconfig/editor/client/src/shared/types.ts
Ankush-lastmile: sounds good, created #915

Ankush-lastmile:
- moved isDev check to occur before making the request to get the aiconfigrc data. This way the request is only made if needed
- - Manually tested the flows and verified telemetry is only sent in prod and not dev

PR #900, opened 2024-01-12
Connect local HF model parser prompt input schemas to the right classnames

rossdanlm:
Connect local HF model parser prompt input schemas to the right classnames

We did a bit of changing of classnames, so just updating here, as well as re-ordering some of the imports

Also important to note is that we currently don't have this for the remote inference text generation class:

https://github.com/lastmile-ai/aiconfig/blob/2364ee32d340a105b4485fa134086f9093d42384/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/remote_inference_client/text_generation.py#L124

I'll handle that next diff, as well as rename that class (and test that it works without model name specified). For now I just copy pasted the remote_inference/text_generation.py --> hf.py (core parser). Just to show that they're indeed the same functionality and that core parser can be deleted

## Test Plan
Before

https://github.com/lastmile-ai/aiconfig/assets/151060367/bf10e6f2-56db-423d-8ff0-a636a12f7f96

After

https://github.com/lastmile-ai/aiconfig/assets/151060367/56ddc0fc-3416-4542-81ee-27c1e4c338fb

### DIFF
--- a/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/remote_inference_client/text_generation.py
+++ b/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/remote_inference_client/text_generation.py
@@ -315,7 +315,14 @@ class HuggingFaceTextGenerationParser(ParameterizedModelParser):
             output_data = output.data
             if isinstance(output_data, str):
                 return output_data
-            
+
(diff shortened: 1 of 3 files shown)


saqadri (approved):
Ship it

PR #901, opened 2024-01-12
Do not run prompt if any of required model settings are missing

rossdanlm:
Do not run prompt if any of required model settings are missing

There's several UI ways to do this, but for now I just did:

1. User clicks RunPromptButton
2. onRunPrompt checks if any of the fields are undefined or "". This "" I'm not sure if it's right and that some string fields have valid empty strings, but we can fix forward if needed by explicitly removing the setting key in `SettingPropertyRenderer` for certain fields. We can make another field in the schema called `allowEmptyValue: boolean = False` or something
3. Open the model settings drawer with red text showing which fields are missing

We can extend this later to make the rendering of missing fields in red border or something, but will let Ryan do that

It's kind of a big mess and I passed `onUpdateMissingRequirementsFields` throughout the renderers unti lwe get to final leve,

## Testing
For some reason whenever I press the run prompt button, even if the required fields are not empty or null or undefined, they're being reset so I can never actually run prompt

https://github.com/lastmile-ai/aiconfig/assets/151060367/b97eb9f8-c896-4bcb-b987-191d1815fba7

## Appendix
Here's how the code is all connected to make it easier to review this diff

7891bfd4-971d-45b0-8f28-d3a6efd5d507


python/src/aiconfig/editor/client/src/components/SettingsPropertyRenderer.tsx
```diff
@@ -329,6 +346,68 @@ export default function SettingsPropertyRenderer({
     }
     case "object": {
       const requiredFields = new Set<string>(property.required ?? []);
+      // useEffect(() => {
+      //   requiredFields.forEach((fieldName) => {
+      //     onUpdateMissingRequiredFields(fieldName, propertyValue);
+      //   });
+      // }, [onUpdateMissingRequiredFields, propertyValue, requiredFields]);
+      /* 
+Warning: Cannot update a component (`PromptContainer`) while rendering a different component (`SettingsPropertyRenderer`). To locate the bad setState() call inside `SettingsPropertyRenderer`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render
```
rossdanlm, line 355: See L76 for more context on what I initially tried to do here and why I had to move it there instead

python/src/aiconfig/editor/client/src/components/prompt/PromptContainer.tsx
```diff
@@ -66,10 +70,54 @@ export default memo(function PromptContainer({
...
+  const onUpdateMissingRequiredFields = useCallback(
+    (fieldName: string, fieldValue: JSONValue) => {
+      setMissingRequiredFields((prevState) => {
+        const newState = new Set<string>(prevState);
+        const isPrimitiveValue =
+          typeof fieldValue === "string" ||
+          typeof fieldValue === "number" ||
+          typeof fieldValue === "boolean" ||
+          typeof fieldValue === "undefined" ||
+          typeof fieldValue === "bigint";
```
rossdanlm, line 85: I realized a lot of the time property from `SettingsPropertyRenderer` was an object, and I finally realized it's basically just recursively checking it's own JSON file at each nested level. But yea added this check here so that we don't do any operations on values that don't make sense

rholinshead:
Hmm, I think this is overcomplicating things quite a bit by maintaining this separate state, and is also going to have false positives for settings that are globally set for the model, right?

I think the best starting point is ensuring the parsers surface an error if required fields aren't provided so that run fails. Then next step would be to surface the 'required' asterisk for the fields in the model settings UI.

If we want to do things like surface missing required fields in the UI or disable run prompt button if required fields are missing then we will probably want:
- a helper function that takes in an object and a schema and determines what required fields in the schema are missing in the object
- pass in global model settings for the prompt's model to the prompt container
- in PromptContainer, have a memoized `const missingRequiredModelSettings` which calls this helper on an object constructed from global model settings overwritten with local model settings. Note this is kind of a pain for two reasons: 1) need to recursively spread in the local values for different nested keys; and 2) this couples client-side with the aiconfig logic of settings resolution (client needs to know how aiconfig resolves model settings). Number 2 is why I was just leaving the required settings logic up to the parsers to throw an error for now.
- can pass this to the input renderer for disabling the run button and to the settings to show the error text (nit on that -- probably better to actually show about the input in case they don't have the drawer open, or potentially show like an error notification badge on the drawer / settings tab in the future)

rossdanlm:
This diff was mostly draft and Proof of Concept to get feedback from Ryan, closing

PR #902, opened 2024-01-12
[editor][client] render array objects in settings properly

Ankush-lastmile:
[editor][client] render array objects in settings properly

Upon loading the OpenAI function calling aiconfig, the editor client does not render the function data. This is because the functions are an array, and the client does not initialize array data.

In this diff, if the Settings data is an array with an initial value, properly set/intiailize and render the array data (inside SettingsPropertyRenderer)

Also updated the object renderer/settings to render json editor for an unspecified Json

## Testplan

Load the function calling aiconfig and open the model settings drawer. Notice the functions get rendered.

Modified function call information inside the json editor. Tested the save was successful.

https://github.com/lastmile-ai/aiconfig/assets/141073967/f477f970-3de1-4c7b-adb4-eba1a8ca0bce

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/SettingsPropertyRenderer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/SettingsPropertyRenderer.tsx
@@ -20,6 +20,7 @@ import UnionPropertyControl, {
   UnionProperty,
 } from "./property_controls/UnionPropertyControl";
 import { JSONObject, JSONValue } from "aiconfig";
+import JSONEditor from "./JSONEditor";
 
 export type StateSetFromPrevFn = (prev: JSONValue) => void;
 export type SetStateFn = (val: StateSetFromPrevFn | JSONValue) => void;
@@ -87,10 +88,28 @@ export default function SettingsPropertyRenderer({
     [propertyName, propertyValue, setValue]
   );
 
+  const itemValues = useRef(
+    Array.isArray(propertyValue) ? new Map(propertyValue.map(val => [uniqueId(), val])) : 
+    new Map<string, JSONValue>());
+
   // Used in the case the property is an array
-  // TODO: Should initialize with values from settings if available
-  const [itemControls, setItemControls] = useState<JSX.Element[]>([]);
-  const itemValues = useRef(new Map<string, JSONValue>());
+  const [itemControls, setItemControls] = useState<JSX.Element[]>(() => Array.from(itemValues.current, ([key, value]) => (
+    <Group key={key}>
+      <SettingsPropertyRenderer
+        propertyName=""
+        property={property.items}
+        initialValue={value}
+        setValue={newItem => {
+          itemValues.current.set(key, newItem);
+          setAndPropagateValue(Array.from(itemValues.current.values()));
+        }}
+     />
+     <ActionIcon onClick={() => removeItemFromList(key)}>
+       <IconTrash size={16} />
+     </ActionIcon>
+   </Group>
+  )));
+
 
   const removeItemFromList = useCallback(
     async (key: string) => {
@@ -367,8 +386,22 @@ export default function SettingsPropertyRenderer({
             <Stack>{subpropertyControls}</Stack>
           </>
         );
+      } else {
+        propertyControl = (
+          <Stack>
+            <PropertyLabel
+             propertyName={propertyName}
+             propertyDescription={propertyDescription}
+           />
+            <div style={{minWidth: "350px"}}>
+              <JSONEditor
+                content={initialValue as JSONObject}
+                onChangeContent={setAndPropagateValue}
+              />
+            </div>
+          </Stack>
+        );
       }
-
       break;
     }
     case "select": {


python/src/aiconfig/editor/client/src/components/SettingsPropertyRenderer.tsx
```diff
@@ -124,6 +125,33 @@ export default function SettingsPropertyRenderer({
     ]);
   }, [property.items, removeItemFromList, setAndPropagateValue]);
 
+ 
+
+  useEffect(() => {
```
rholinshead, line 130: We don't need an effect for this. This is just setting the initial state for itemControls so we should instead just initialize it inline above using the init function for useState:

```
const [itemControls, setItemControls] = useState<JSX.Element[]>(() => {
  if (!Array.isArray(initialValue)) {
    return [];
  }

  return initialValue.map((arrayItem,index) => {
    const key = uniqueId();
    return (
      <Group key={key}>
        <SettingsPropertyRenderer
          propertyName=""
          property={property.items}
          initialValue={arrayItem}
          setValue={newItem => {
            itemValues.current.set(key, newItem);
            setAndPropagateValue(Array.from(itemValues.current.values()));
          }}
       />
       <ActionIcon onClick={() => removeItemFromList(key)}>
         <IconTrash size={16} />
       </ActionIcon>
     </Group>
    );
  });
});
```

python/src/aiconfig/editor/client/src/components/SettingsPropertyRenderer.tsx
```diff
@@ -367,8 +395,14 @@ export default function SettingsPropertyRenderer({
             <Stack>{subpropertyControls}</Stack>
           </>
         );
+      } else {
+        propertyControl = (
+          <JSONEditor
+            content={initialValue as JSONObject}
+            onChangeContent={setAndPropagateValue}
+          />
```
rholinshead, line 403:
```suggestion
        <Stack>
             <PropertyLabel
                propertyName={propertyName}
                propertyDescription={propertyDescription}
              />
            <JSONEditor
            content={initialValue as JSONObject}
            onChangeContent={setAndPropagateValue}
          />
        </Stack>
```

Let's add the label

rholinshead: Also can you play around with it to see if we can get the width to look nicer? Maybe wrap the editor here with a div with 100% width or something?

python/src/aiconfig/editor/client/src/components/SettingsPropertyRenderer.tsx
```diff
@@ -90,7 +91,7 @@ export default function SettingsPropertyRenderer({
   // Used in the case the property is an array
   // TODO: Should initialize with values from settings if available
   const [itemControls, setItemControls] = useState<JSX.Element[]>([]);
-  const itemValues = useRef(new Map<string, JSONValue>());
+  const itemValues = useRef(propertyValue ?? new Map<string, JSONValue>());
```
rholinshead, line 94: We shouldn't use propertyValue here since that's a state object that shouldn't be mutated directly (need to use setState for that). Instead, we should construct this map with the propertyValue. Maybe something like:

```
const itemValues = useRef(
  Array.isArray(propertyValue) ? new Map(propertyValue.map(val => [uniqueId(), val])) : 
  new Map<string, JSONValue>());
```

rholinshead (changes requested):

python/src/aiconfig/editor/client/src/components/SettingsPropertyRenderer.tsx
Ankush-lastmile: updated as suggested. Is there anything bad that would happen if we were to continue to use `useEffect()`?

python/src/aiconfig/editor/client/src/components/SettingsPropertyRenderer.tsx
Ankush-lastmile: added label. Tried setting width to 100%, that did nothing. Anything > 100 would make it render nice but it would throw an exception on render

python/src/aiconfig/editor/client/src/components/SettingsPropertyRenderer.tsx
Ankush-lastmile: Updated. 

>that shouldn't be mutated directly (need to use setState for that

Makes sense, but then why is it okay to set it to a newMap with the same property value?

rholinshead: The code never mutates the map values, just the map itself. Before it would have been modifying the actual propertyValue object

python/src/aiconfig/editor/client/src/components/SettingsPropertyRenderer.tsx
rholinshead: Nothing critically bad, it's just the scenario I told you about --> doing it in a useEffect would first render with default [] in state, then hit the effect and then re-render again with new state from the effect. It's an extra full render cycle. This way the state is set initially before anything is ever added to the DOM and no re-render is needed

python/src/aiconfig/editor/client/src/components/SettingsPropertyRenderer.tsx
```diff
@@ -367,8 +398,22 @@ export default function SettingsPropertyRenderer({
...
+            <PropertyLabel
+             propertyName={propertyName}
+             propertyDescription={propertyDescription}
+           />
+            <JSONEditor
+              content={initialValue as JSONObject}
+              onChangeContent={setAndPropagateValue}
+            />
+          </Stack>
+          </div>
```
rholinshead, line 414:
```suggestion
          <Stack>
            <PropertyLabel
             propertyName={propertyName}
             propertyDescription={propertyDescription}
           />
           <div style={{width: "100%"}}>
              <JSONEditor
                content={initialValue as JSONObject}
                onChangeContent={setAndPropagateValue}
              />
            </div>
          </Stack>
```
Let's wrap the div around the JSONEditor only. If that still doesn't make it wider can you try {{minWidth: "350px}} or something like that? We have a min width of 400px for the settings container (https://github.com/lastmile-ai/aiconfig/blob/3b91a530a376faf13b2ae7ad3bf0ba1b20114548/python/src/aiconfig/editor/client/src/components/prompt/PromptActionBar.tsx#L48)

python/src/aiconfig/editor/client/src/components/SettingsPropertyRenderer.tsx
```diff
@@ -13,13 +13,14 @@ import {
   AutocompleteItem,
   Select,
 } from "@mantine/core";
-import { useState, useCallback, useRef } from "react";
+import { useState, useCallback, useRef, useEffect } from "react";
```
rholinshead, line 16:
```suggestion
import { useState, useCallback, useRef } from "react";
```
not needed

rholinshead (approved):
Accepting to unblock, but can you also test that changing a setting and saving the config properly persists these generic json settings?

And before landing let's try and get that width sorted out. Please ping me if you can't get it and I can play around with it a bit as well

python/src/aiconfig/editor/client/src/components/SettingsPropertyRenderer.tsx
rholinshead: Ya we won't want to go over 100% because there will be logic happening in monaco editor internals to handle resizing, probably based on parent element or something like that, and exceeding the parent width will cause troubles there

Ankush-lastmile:
rebase, updated json editor for model settings to be 350px wide.

Unfortunately getting a weird side effect where updating the json editor inside settings now adds to it than updating.

https://github.com/lastmile-ai/aiconfig/assets/141073967/47cf6275-711a-4ba7-a2fd-6a24ab48fb2f

python/src/aiconfig/editor/client/src/components/SettingsPropertyRenderer.tsx
Ankush-lastmile: removed

Ankush-lastmile:
fixed some spacing

Ankush-lastmile:
>Unfortunately getting a weird side effect where updating the json editor inside settings now adds to it than updating.

Fixed this by
1. Initialize `itemValues` first with propertyValue
2. Then Initialize `itemControls` with the resultant `itemValues`

This way the JSX keys lineup. 

Also removed the TODO comment

rholinshead (approved):

PR #905, opened 2024-01-12
[1/n] Disable Run Button: UI Display

rossdanlm:
[1/n] Disable Run Button: UI Display

For now no functionality, just setting up the visual UI display, will connect next PR. Doing this to make PRs smaller and easier to review + land

## Test Plan
When setting `disabled` to True
Screenshot 2024-01-12 at 14 57 38

When setting `disabled` to False, still works as before

https://github.com/lastmile-ai/aiconfig/assets/151060367/1c6ad683-8c50-46aa-9760-8eea8bd40fed

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/905).
* #907
* __->__ #905

PR #907, opened 2024-01-12
[2/n] Disable Run Button: Create runningPromptId field to determine if button should be disabled

rossdanlm:
[2n] Disable Run Button: Create runningPromptId field to determine if button should be disabled

I did a few things:

1. Create a new state id `runningPromptId` to check which prompt is running, making sure to set these correctly on run prompt actions
2. Read from this state in the AIConfigEditor and passed it as a prop down to the `RunPromptButton`

Question: Is there a better way of doing step #2 so that we don't need to keep piping it?

## Test Plan
Make sure that:
1. When you are running a prompt, you can still cancel it
2. When another prompt is running, you can't run any other prompt
3. When prompt is complete (or cancelled or errored), all the prompts can run again

https://github.com/lastmile-ai/aiconfig/assets/151060367/6e9a9cbf-6469-4ecb-ba74-1ede2ed1a292

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/907).
* __->__ #907
* #905

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
+++ b/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
@@ -828,6 +828,8 @@ export default function EditorContainer({
     return () => clearInterval(interval);
   }, [getServerStatusCallback, serverStatus]);
 
+  const runningPromptId: string | undefined = aiconfigState._ui.runningPromptId;
+
   return (
     <AIConfigContext.Provider value={contextValue}>
       <Notifications />
@@ -907,6 +909,8 @@ export default function EditorContainer({
           />
         </div>
         {aiconfigState.prompts.map((prompt: ClientPrompt, i: number) => {
+          const isAnotherPromptRunning =
+            runningPromptId !== undefined && runningPromptId !== prompt._ui.id;
           return (
             <Stack key={prompt._ui.id}>
               <Flex mt="md">
@@ -925,6 +929,7 @@ export default function EditorContainer({
                   onUpdateModelSettings={onUpdatePromptModelSettings}
                   onUpdateParameters={onUpdatePromptParameters}
                   defaultConfigModelName={aiconfigState.metadata.default_model}
+                  isRunButtonDisabled={isAnotherPromptRunning}
                 />
               </Flex>
               <div className={classes.addPromptRow}>
--- a/python/src/aiconfig/editor/client/src/components/aiconfigReducer.ts
+++ b/python/src/aiconfig/editor/client/src/components/aiconfigReducer.ts
@@ -181,22 +181,41 @@ function reduceConsolidateAIConfig(
       );
     }
     case "RUN_PROMPT": {
-      return reduceReplacePrompt(state, action.id, (prompt) => {
-        const responsePrompt = responseConfig.prompts.find(
-          (resPrompt) => resPrompt.name === prompt.name
-        );
+      // Note: If we are calling "RUN_PROMPT" directly as a dispatched event
+      // type, we automatically set the state there to `isRunning` for that
+      // prompt. That logic does not happen here, it happens in
+      // `aiconfigReducer`.
+      // If we are calling "RUN_PROMPT" indirectly via the action of a
+      // "CONSOLIDATE_AICONFIG" dispatch, we end up here. We need to check
+      // if we actually want to set the prompt state to `isRunning`
+      const isRunning = action.isRunning ?? false;
+      const stateWithUpdatedRunningPromptId = {
+        ...state,
+        _ui: {
+          ...state._ui,
+          runningPromptId: isRunning ? action.id : undefined,
+        },
+      };
+      return reduceReplacePrompt(
+        stateWithUpdatedRunningPromptId,
+        action.id,
+        (prompt) => {
+          const responsePrompt = responseConfig.prompts.find(
+            (resPrompt) => resPrompt.name === prompt.name
+          );
 
-        const outputs = responsePrompt?.outputs ?? prompt.outputs;
+          const outputs = responsePrompt?.outputs ?? prompt.outputs;
 
-        return {
-          ...prompt,
-          _ui: {
-            ...prompt._ui,
-            isRunning: action.isRunning ?? false,
-          },
-          outputs,
-        };
-      });
+          return {
+            ...prompt,
+            _ui: {
+              ...prompt._ui,
+              isRunning,
+            },
+            outputs,
+          };
+        }
+      );
     }
     case "UPDATE_PROMPT_INPUT": {
       return reduceReplacePrompt(state, action.id, consolidatePrompt);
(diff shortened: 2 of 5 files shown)


python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
```diff
@@ -828,6 +828,57 @@ export default function EditorContainer({
...
+                onChangePromptName={onChangePromptName}
+                cancel={callbacks.cancel}
+                onRunPrompt={onRunPrompt}
+                onUpdateModel={onUpdatePromptModel}
+                onUpdateModelSettings={onUpdatePromptModelSettings}
+                onUpdateParameters={onUpdatePromptParameters}
+                defaultConfigModelName={aiconfigState.metadata.default_model}
+                isRunButtonDisabled={
+                  runningPromptId !== undefined &&
+                  runningPromptId !== prompt._ui.id
```
rossdanlm, line 861: cc @rholinshead I feel confident I don't need to do the explicit check for `runningPromptId !== undefined`, but was just worried in case we can have situations where any of the `prompt._ui.id` fields could be undefined. Is that possible?

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
```diff
@@ -828,6 +828,57 @@ export default function EditorContainer({
     return () => clearInterval(interval);
   }, [getServerStatusCallback, serverStatus]);
 
+  const runningPromptId: string | undefined = aiconfigState._ui.runningPromptId;
+  const promptListContainer = (
```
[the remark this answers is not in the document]
rholinshead: Also I know this is a very big file and we can fix that down the line:
- pulling our callback definitions into their own hooks
- using context instead of passing a bunch of props through here (but I want us to align on whether we need to support custom react components before doing that)

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
rholinshead: No need for the check -- _ui.id is required to exist in ClientPrompt

python/src/aiconfig/editor/client/src/components/aiconfigReducer.ts
```diff
@@ -181,22 +181,38 @@ function reduceConsolidateAIConfig(
...
     }
     case "RUN_PROMPT": {
-      return reduceReplacePrompt(state, action.id, (prompt) => {
-        const responsePrompt = responseConfig.prompts.find(
-          (resPrompt) => resPrompt.name === prompt.name
-        );
+      // Note: We are relying on the callsite to explicitly set `isRunning`
+      // to true when it's needed. If isRunning is not explicitly defined
+      // in the action, and we have called consolidate aiconfig, then we
+      // assume that we are finished running the prompt.
```
rholinshead, line 187: Side note, I left a comment on the PR that added this to the action but we should revisit it. Actions generally should not need to be passing state to set. We should determine if a prompt is running based on the action itself (i.e. maybe we add a different action RUN_PROMPT_STREAM_CHUNK to dispatch when a chunk is returned instead of using CONSOLIDATE_AICONFIG with RUN_PROMPT action).

We can revisit after gradio integration is complete, though

rholinshead (approved):
LGTM, just a comment on using a variable for the jsx and some other small things

Also just noting this could be done without all the state stuff by just checking
```
const isConfigRunning = aiconfigState.some(prompt => prompt._ui.isRunning)
```
and checking within the PromptInputRenderer `isConfigRunning && !isRunning`.

Downside to that is it would run the check over the prompt elements on every state change/re-render. So, using state for this avoids that extra work.

rholinshead:
> Question: Is there a better way of doing step https://github.com/lastmile-ai/aiconfig/pull/2 so that we don't need to keep piping it?

Yes, technically. We could use context to make the value accessible to any child below it (e.g. wrap around the promptlist). But if we do end up supporting custom PromptContainers/InputRenderers then we'd probably do that by supporting render callbacks which would not necessarily be able to access the context and would require the value passed by prop instead.

python/src/aiconfig/editor/client/src/components/aiconfigReducer.ts
rossdanlm: > Actions generally should not need to be passing state to set.

Is this a general best practice, or more of your personalized opinion? 

I've also seen cases where the dispatch event types are more granular and therefore the event data passed along with it doesn't need to be as detailed. Ex:

```
dispatch.("RUN_PROMPT_START", {id: prompt.id})
...
dispatch("RUN_PROMPT_FINISH", {id: prompt.id})
```

^Do you have any preference?

rholinshead: It's general best practice, just separates the concerns. Generally actions shouldn't need to worry about the underlying state, that logic should live within the reducer / state management.

More granular is fine as well; we could do the same and even pull the consolidate stuff out to explicit actions if desired. But not something to do anytime soon

rossdanlm:
> Also just noting this could be done without all the state stuff by just checking
> 
> ```
> const isConfigRunning = aiconfigState.some(prompt => prompt._ui.isRunning)
> ```
> 
> and checking within the PromptInputRenderer `isConfigRunning && !isRunning`.
> 
> Downside to that is it would run the check over the prompt elements on every state change/re-render. So, using state for this avoids that extra work.

Yea this is why I wanted to use config-level state. It feels a bit strange perform logic checks within a specific prompt that the prompt itself doesn't have control over and is only affected by it's sibling prompts logic. 

This is my understanding on how react components are rendered, please let me know if I'm wrong:
By making this disabled logic passed as props instead of re-checking logic directly within `promptInputRenderer`, since props are read-only, we actually aren't rendering the entire promptInputRenderer, only the `runPromptButton` where the changed prop affects the rendering logic of that component?

rossdanlm:
> if we do end up supporting custom PromptContainers/InputRenderers then we'd probably do that by supporting render callbacks

Can you give me an example/speudocode on how this would work?

rholinshead:
> > Also just noting this could be done without all the state stuff by just checking
> > ```
> > const isConfigRunning = aiconfigState.some(prompt => prompt._ui.isRunning)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and checking within the PromptInputRenderer `isConfigRunning && !isRunning`.
> > Downside to that is it would run the check over the prompt elements on every state change/re-render. So, using state for this avoids that extra work.
> 
> Yea this is why I wanted to use config-level state. It feels a bit strange perform logic checks within a specific prompt that the prompt itself doesn't have control over and is only affected by it's sibling prompts logic.
> 
> This is my understanding on how react components are rendered, please let me know if I'm wrong: By making this disabled logic passed as props instead of re-checking logic directly within `promptInputRenderer`, since props are read-only, we actually aren't rendering the entire promptInputRenderer, only the `runPromptButton` where the changed prop affects the rendering logic of that component?

Not exactly -- if a component's prop changes that component will re-render, regardless of rendering logic

rholinshead:
> > if we do end up supporting custom PromptContainers/InputRenderers then we'd probably do that by supporting render callbacks
> 
> Can you give me an example/speudocode on how this would work?

There's a bit of detail here: https://docs.google.com/document/d/1YwUlIQZX8CiuvtbVR4OvpFQ2nphFITRD_XZR3CDNtoE/edit#heading=h.3hy566aov149

PR #911, opened 2024-01-13
Delete `aiconfig_complete` stream response, replace with `aiconfig`

rossdanlm:
Delete `aiconfig_complete` stream response, replace with `aiconfig`

Before we used to not support streaming, so when we would return `aiconfig` it would be from a blocking hanging operation. This meant that we needed to set `isRunning` prompt state to be true while we were waiting, but now we don't need to do that anymore after we migrated all run events to return in streaming response format, even for non-streaming models: https://github.com/lastmile-ai/aiconfig/pull/806

Also we are now no longer using the `streamApi` helper since we added and are now using `streamingApiChain`, which was added in https://github.com/lastmile-ai/aiconfig/pull/789

Finally, if you want more resources on how streaming is connected, you can check out https://github.com/lastmile-ai/aiconfig/pull/910 which is a teaching guide I built for adding explaining how the code is connected

## Test Plan
Both streaming and non-streaming models work as before

https://github.com/lastmile-ai/aiconfig/assets/151060367/b62e7887-20af-4c0c-ab85-eeaacaab64e0

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/911).
* #912
* __->__ #911

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
+++ b/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
@@ -71,10 +71,6 @@ export type RunPromptStreamEvent =
   | {
       type: "aiconfig";
       data: AIConfig;
-    }
-  | {
-      type: "aiconfig_complete";
-      data: AIConfig;
     };
 
 export type RunPromptStreamErrorEvent = {
@@ -641,15 +637,7 @@ export default function EditorContainer({
                 type: "CONSOLIDATE_AICONFIG",
                 action: {
                   ...action,
-                  // Ensure we keep the prompt in a running state since this is an in-progress update
-                  isRunning: true,
                 },
-                config: event.data,
-              });
-            } else if (event.type === "aiconfig_complete") {
-              dispatch({
-                type: "CONSOLIDATE_AICONFIG",
-                action,
                 config: event.data,
               });
             }
@@ -686,6 +674,8 @@ export default function EditorContainer({
           cancellationToken
         );
 
+        // Keep this here in case any server implementations don't return
+        // aiconfig as a streaming format
         if (serverConfigResponse?.aiconfig) {
           dispatch({
             type: "CONSOLIDATE_AICONFIG",
--- a/python/src/aiconfig/editor/client/src/LocalEditor.tsx
+++ b/python/src/aiconfig/editor/client/src/LocalEditor.tsx
@@ -101,9 +101,6 @@ export default function Editor() {
           aiconfig: (data) => {
             onStream({ type: "aiconfig", data: data as AIConfig });
           },
-          aiconfig_complete: (data) => {
-            onStream({ type: "aiconfig_complete", data: data as AIConfig });
-          },
           error: (data) => {
             onError({
               type: "error",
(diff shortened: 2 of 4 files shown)


python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
```diff
@@ -641,17 +637,9 @@ export default function EditorContainer({
                 type: "CONSOLIDATE_AICONFIG",
                 action: {
                   ...action,
-                  // Ensure we keep the prompt in a running state since this is an in-progress update
-                  isRunning: true,
```
saqadri, line 645: Maybe also remove this property from the CONSOLIDATE_AICONFIG action in aiConfigReducer

saqadri (approved):

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
rholinshead: Wasn't this added specifically to support streaming? It wasn't needed for the single awaited request run calls. Also, just noting that we still probably want to support non-streaming execution since external deployments might not have streaming implemented

rholinshead: https://github.com/lastmile-ai/aiconfig/commit/010f499a7cea9b383292b0ff67315079510338de

> Fixed some bugs on the frontend along the way -- specifically, because we return the full aiconfig on every chunk update, we were resetting isRunning to false after the first chunk was received (by dispatching CONSOLIDATE_AICONFIG). Ideally, we update just the output chunk, but to fix this I've added a separate aiconfig_complete event to distinguish the finalized aiconfig from the in-progress one

I guess my main concern is will this be needed for gradio streaming?

saqadri: Ah that's a good point. Gradio would need the full updated value

rossdanlm: > Also, just noting that we still probably want to support non-streaming execution since external deployments might not have streaming implemented

We still have another CONSOLIDATE_AICONFIG call at the end after the stream outputs. 

> Wasn't this added specifically to support streaming? It wasn't needed for the single awaited request run calls.

I think you're right about this, if we return the entire AIConfig as a streamed response this will incorrectly make the `isRunning` param to false. However if we keep it here, we'll never terminate the `isRunning` state to false, because everything we return from the server is in stream format. 


https://github.com/lastmile-ai/aiconfig/assets/151060367/a6db0ff2-0390-4b2c-80e9-4c1364d79ae3



What I will do in a followup PR is add another streaming event called `aiconfig_stream` and have the `isRunning` there set to true (kind of the equivalent of `aiconfig` stream event today, while `aiconfig_complete` is now the same functionality as `aiconfig`). I just feel the names for that are a bit simpler and it's apparent that one is for streaming, while the other is for "regular" output as a "non-stream" (even if it's being implemented by our particular backend as a streaming event, other people can implement it as a non-stream event and know what it is supposed to do)

PR #913, opened 2024-01-14
[ez] Move AIConfigContext into a `contexts` folder

rossdanlm:
[ez] Move AIConfigContext into a `contexts` folder

No functional changes. I think this will be a bit cleaner when we make a READ-ONLY mode because I'm going to use a context for that too and better to keep these files together

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/913).
* #916
* __->__ #913


rholinshead:
We should just add isReadonly to AIConfigContext, no need for a separate one

rossdanlm:
> We should just add isReadonly to AIConfigContext, no need for a separate one

Alright sure, but why not move into it's separate folder anyways since it's not really a "component"?

rholinshead (approved):

rholinshead:
LGTM, just looks like there's a merge conflict to resolve before landing

PR #914, opened 2024-01-14
Add explicit "stop_streaming" event to signal streaming is finished

rossdanlm:
Add explicit "stop_streaming" event to signal streaming is finished

I think this is cleaner and better for us to deal with. It also breaks down the aiconfig streaming logic to be more explicitly handled for next PR.

See comments about this in

- https://github.com/lastmile-ai/aiconfig/pull/907#discussion_r1451013546
- https://github.com/lastmile-ai/aiconfig/pull/911#discussion_r1451629515

## Test Plan
Both the streaming and non-streaming models work the same as before

https://github.com/lastmile-ai/aiconfig/assets/151060367/2c3fd802-e95e-417c-ac80-7b6e18983520

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
+++ b/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
@@ -71,6 +71,10 @@ export type RunPromptStreamEvent =
   | {
       type: "aiconfig";
       data: AIConfig;
+    }
+  | {
+      type: "stop_streaming";
+      data: null;
     };
 
 export type RunPromptStreamErrorEvent = {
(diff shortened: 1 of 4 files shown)


python/src/aiconfig/editor/server/server.py
```diff
@@ -350,6 +350,10 @@ def kill_thread(thread_id: int | None):
             yield "["
             yield json.dumps({"aiconfig": aiconfig_json})
             yield "]"
+        
+        yield "["
+        yield json.dumps({"stop_streaming": None})
+        yield "]"
```
rholinshead, line 356: Oh, interesting. I didn't realize each streamed message was being wrapped in `[` and `]` separately -- just noting the original intention was for the full streamed response to be a single array of JSON objects ([ marking the start, ] marking the end). I guess if it works ¯\_(ツ)_/¯

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
```diff
@@ -633,13 +637,26 @@ export default function EditorContainer({
...
                   ...action,
+                  // Keep the prompt running state until the end of streaming
+                  isRunning: true,
                 },
                 config: event.data,
               });
+            } else if (event.type === "stop_streaming") {
+              // Pass this event at the end of streaming to signal
+              // that the prompt is done running and we're ready
+              // to consolidate the entire AIConfig result
```
rholinshead, line 655: > and we're ready to consolidate the entire AIConfig result

nit, consolidation can't really happen on this action since there's no aiconfig data to use to consolidate with, right? We could technically return the aiconfig as part of the stop_streaming response and then do consolidation with it, then this comment would be accurate -- and I guess that could be done in place of the final aiconfig response instead of sending an extra response

rholinshead (approved):
just a nit with the comment

python/src/aiconfig/editor/server/server.py
rossdanlm: Yea I actually tried doing the entire thing, and that works for single instance, but when we're iterating through the stream queue, we need to reset every yield around individual `[` and `]` otherwise it will not update on client

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
rossdanlm: You are correct, let me adjust this comment to say "ready to reset the AIConfig to a non-running state"

PR #916, opened 2024-01-14
2/n: Read-only mode: Connect context to Run Prompt button

rossdanlm:
2/n: Read-only mode: Connect context to Run Prompt button

Created a new value that gets read from props in `AIConfigEditor` and passed into AIConfigContext to be read by downstream components. Going to do this diff stack each component 1 at a time for easier review +  unblocking

I decided not to use a tooltip for read-only mode. Instead planning to add a big text at top saying that you're in read-only mode, and you won't be able to make any changes

## Test Plan
Still works as before

Read-only shows no tooltip
Screenshot 2024-01-13 at 23 32 35

Not in read-only and running a prompt shows a tooltip
Screenshot 2024-01-13 at 23 33 08

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/916).
* __->__ #916
* #913

### DIFF
--- a/python/src/aiconfig/editor/client/src/LocalEditor.tsx
+++ b/python/src/aiconfig/editor/client/src/LocalEditor.tsx
@@ -5,7 +5,13 @@ import AIConfigEditor, {
   RunPromptStreamErrorEvent,
 } from "./components/AIConfigEditor";
 import { Flex, Loader, MantineProvider, Image } from "@mantine/core";
-import { AIConfig, InferenceSettings, JSONObject, Output, Prompt } from "aiconfig";
+import {
+  AIConfig,
+  InferenceSettings,
+  JSONObject,
+  Output,
+  Prompt,
+} from "aiconfig";
 import { useCallback, useEffect, useMemo, useState } from "react";
 import { ufetch } from "ufetch";
 import { ROUTE_TABLE } from "./utils/api";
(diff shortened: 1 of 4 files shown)


rholinshead:
> Instead planning to add a big text at top saying that you're in read-only mode, and you won't be able to make any changes

Let's hold off on that for now; readonly will only be used in the lastmile shared aiconfig view for now and the messaging is probably something that should be managed by the application rendering the AIConfigEditor instead of by AIConfigEditor itself

python/src/aiconfig/editor/client/src/components/prompt/RunPromptButton.tsx
```diff
@@ -15,39 +16,49 @@ export default memo(function RunPromptButton({
...
+        ) : (
+          <>
+            <IconPlayerPlayFilled size="16" />
+          </>
+        )}
+      </Button>
+    );
+  }, [disabledOrReadOnly, onClick, isRunning]);
+
+  const disabledButton = useMemo(() => {
```
rholinshead, line 53: Let's not memoize these or use `useCallback` since that is actually going to result in extra unnecessary processing. This `RunPromptButton` is already memoized, so it will not re-render unless the props change. Then if that happens, note that the deps chain will result in button's useMemo always creating a new button anyway. So, this just results in extra processing to compute/compare the dependency changes within the RunPromptButton every time it re-renders

python/src/aiconfig/editor/client/src/LocalEditor.tsx
```diff
@@ -17,7 +17,11 @@ import { ufetch } from "ufetch";
 import { ROUTE_TABLE } from "./utils/api";
 import { streamingApiChain } from "./utils/oboeHelpers";
 
-export default function Editor() {
+type Props = {
+  readOnly?: boolean;
+};
+
+export default function Editor({ readOnly = false }: Props) {
```
rholinshead, line 24:
```suggestion

export default function Editor() {
```

LocalEditor shouldn't have `readOnly` as a prop since the component that is being packaged and reused is the `AIConfigEditor` component. LocalEditor will never be readonly

rholinshead (approved):
Accepting to unblock, but some minor things to change:
- LocalEditor won't need readonly prop
- No need for all the memoization within RunPromptButton since it will actually negatively affect render perf

rholinshead:
Also please address the merge conflicts before landing

PR #918, opened 2024-01-14
[ez] Rename `aiconfig` stream response to `aiconfig_chunk` response

rossdanlm:
[ez] Rename `aiconfig` stream response to `aiconfig_chunk` response

I was having some trouble converting everything to `STREAM_AICONFIG_CHUNK` so I'm just doing really easy steps to keep diffs easy to review and test. You'll be able to see final result at end of diff stack

## This diff
Just doing very simple name coy and replace, no functional changes

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/918).
* #928
* #926
* #925
* #924
* #922
* #921
* #920
* #919
* __->__ #918
* #917

### DIFF
--- a/python/src/aiconfig/editor/client/src/LocalEditor.tsx
+++ b/python/src/aiconfig/editor/client/src/LocalEditor.tsx
@@ -98,8 +98,8 @@ export default function Editor() {
           output_chunk: (data) => {
             onStream({ type: "output_chunk", data: data as Output });
           },
-          aiconfig: (data) => {
-            onStream({ type: "aiconfig", data: data as AIConfig });
+          aiconfig_chunk: (data) => {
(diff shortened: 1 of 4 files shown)


rholinshead (approved):
Accepting to unblock but not entirely sure I understand the reasoning. Based on server code comment

> Ensure that the run process is complete to yield final output

Previous 'aiconfig' stream response is the last response after all the outputs are sent. So there will only be one single 'aiconfig_chunk' right? Or is gradio going to have streaming send multiple `aiconfig_chunks`? If so, then I agree makes sense to have it the same instead of supporting different action names depending on the server

PR #922, opened 2024-01-14
Create new action RUN_PROMPT_CANCEL

rossdanlm:
Create new action RUN_PROMPT_CANCEL

Splitting the RUN_PROMPT action into 3 sub-categories:

1. RunPrompt
2. RunPromptCancel
3. RunPromptSuccess
4. RunPromptError

Having 4 separate events means that we can call these directly without having to use RUN_PROMPT within the CONSOLIDATE_AICONFIG

Another benefit of this is that it fixes a bug where the state doesn't get properly refreshed, so the output is shown with the existing cancelled output and not the previous one

## Test Plan
Before (the output state lingers instead of being cleared --> If this is a feature not a bug, that's easy to keep just don't reset the client prompt output to what the server responds back with)

Everything works as previously (can update input + metadata on client and that remains unchanged, only output gets restored)


rholinshead:
> Another benefit of this is that it fixes a bug where the state doesn't get properly refreshed, so the output is shown with the existing cancelled output and not the previous one

FWIW I do think it's a bit strange to revert back to previous output (I left a comment on that in the original PR as well), since I can imagine a scenario where I'm using a model parser that takes a while to execute a long prompt and I'm just iterating on it and want to see the response format itself. I would cancel if the format is wrong right away but then probably expect it to stay there to reference instead of having it reverted.

I think it's ok to keep as-is for now since that was the original intention and the fix ensures consistency with that. We can revisit later to update based on user feedback if applicable

rholinshead (approved):

Issue #664, opened 2023-12-29
[python] Add run_with_dependencies to API


rholinshead:
In typescript, the AIConfig API has `runWithDependencies`. In python, the `run_with_dependencies` is supported by passing kwargs to the run method. Instead of supporting this arbitrary kwargs data we should add the `run_with_dependencies` implementation as a method on the config to be consistent with typescript

rossdanlm:
See comment thread in https://github.com/lastmile-ai/aiconfig/pull/661#discussion_r1437945315 for details

There are two parts to this:
1. Instead of add `run_with_dependencies` to API, we should just re-use `run()` method but pass in an explicit boolean argument (`run_with_dependencies`) instead of relying on `kwarg`
2. We should also delete `kwargs` argument from the `run()` API method

rholinshead:
> Instead of adding run_with_dependencies to API, we should just re-use run() method but pass in an explicit boolean argument (run_with_dependencies) instead of relying on kwarg

Should this be part of inference options instead?

Also, note that this requires the same change to typescript side as well to remain consistent

rossdanlm:
> Should this be part of inference options instead?

I don't think. Conceptually, inference options (things like setting streaming to true) are more like global functions you want to set for your entire session, not just 1 single action. I don't think automatically making every prompt required to re-run all it's dependencies is desirable

sp6370:
I would like to pick this up! 🚀

rossdanlm:
cc @sp6370 please follow instructions for this in https://github.com/lastmile-ai/aiconfig/issues/664#issuecomment-1871676370

Right now this function exists in typescript in https://github.com/lastmile-ai/aiconfig/blob/e47ce1ff59bd383cdda2c2682528d858adc376cd/typescript/lib/config.ts#L456, but I think it's better to remove it and simply make the `run_with_dependencies` an explicit method arg in the `run()` method

sp6370:
@rholinshead Clarification: Both Python and Type Script implementations need to be updated?

rholinshead:
> @rholinshead Clarification: Both Python and Type Script implementations need to be updated?

@sp6370 ideally both python and typescript implementations will match but it's ok to update the python implementation first and create an issue for the typescript side to match the python implementation afterwards, if you are primarily interested in the python side.

@rossdanlm's comment above outlines the python side changes and that's all that should be needed to close out this issue

sp6370:
[...]

PR #923, opened 2024-01-14
Updated ParameterizedModelParser run method for handing run_with_dependencies parameter

sp6370:
Context: Currently `ParameterizedModelParser` run method utilized `kwargs` to decide if to execute prompt with dependencies or not. This pull request eliminates the usage of `kwargs` and introduces a new optional prameter for `run` method to pass information about running the prompt with dependencies.

Closes: #664 

Test Plan:
- [x] Add a new test that invokes `run` with `run_with_dependencies=True` for a prompt which requires output from another prompt.
- [x] Ensure that all the prompts in the hugging face cookbook work.


sp6370:
Hi @rossdanlm!

Could I get some feedback on the test for the PR?

python/tests/test_run_config.py
```diff
@@ -42,3 +44,22 @@ async def test_load_parametrized_data_config(set_temporary_env_vars):
...
+        config_absolute_path = get_absolute_file_path_from_relative(__file__, config_relative_path)
+        config = AIConfigRuntime.load(config_absolute_path)
+
+        combined_prompt_parameters = {
+            "city": "NYC",
+            "order_by": "geographic location",
+        }
+        await config.run("gen_itinerary", combined_prompt_parameters, run_with_dependencies=True)
+
+    assert mock_openai.call_count == 2
```
Ankush-lastmile, line 65: Just checking that you're rebased onto main, specifically after #999?

sp6370: Haven't rebased yet.

sp6370: I did notice tests failing after a merge. I guess #999 was the case.
I reverted the merge later.

Ankush-lastmile: Yep following the changes in PR #999, its necessary to revise our approach to mocking the OpenAI client. Your proposed code looks solid to me. My suggestions are:

1. Rebase your diff to main
2. Temporarily mark this test as an expected failure. Please add the @pytest.mark.xfail decorator above the existing @pytest.mark.asyncio decorator."

python/tests/aiconfigs/tarvel_gpt_prompts_with_dependency.json
```diff
@@ -0,0 +1,35 @@
+{
```
Ankush-lastmile, line 1: nit: `tarvel_gpt_prompts_with_dependency.json` -> `travel_gpt_prompts_with_dependency.json`

sp6370: 🤣 Nice find!

python/tests/test_run_config.py
sp6370: Will do so!

python/src/aiconfig/default_parsers/parameterized_model_parser.py
```diff
@@ -45,13 +45,10 @@ async def run(
         aiconfig: AIConfig,
         options: Optional[InferenceOptions] = None,
         parameters: Dict = {},
-        **kwargs,  # TODO: We should remove and make arguments explicit
+        run_with_dependencies: Optional[bool] = False,
```
jonathanlastmileai, line 48: Is this backwards compatible? It might be de facto based on call sites, but it is not by virtue of the signature.

python/tests/aiconfigs/tarvel_gpt_prompts_with_dependency.json
jonathanlastmileai: What is this new example showing?

python/tests/test_run_config.py
```diff
@@ -48,3 +50,23 @@ async def test_load_parametrized_data_config(set_temporary_env_vars):
                 },
             ],
         }
+
+@pytest.mark.xfail
```
jonathanlastmileai, line 54: Why are we adding a test we expect to fail?

jonathanlastmileai (changes requested):
Thanks for the PR! Back to your queue for now

python/tests/test_run_config.py
sp6370: Done!

python/tests/aiconfigs/tarvel_gpt_prompts_with_dependency.json
sp6370: It's a aiconfig which uses output from a previous prompt as input to another prompt.

python/tests/test_run_config.py
sp6370: As mentioned by @Ankush-lastmile [here](https://github.com/lastmile-ai/aiconfig/pull/923#discussion_r1472205356). There have been changes to OpenAI model paeser as such new way to for Mocking OpenAI API is required. 

The new test introduced works fines based on the earlier mocking approach. Unless a mocking strategy is implemented we would have to use the work around as used here as well https://github.com/sp6370/aiconfig/blob/9617bd9d54e3bff3bb7eef2889af9f30adf39602/python/tests/test_run_config.py#L11.

python/src/aiconfig/default_parsers/parameterized_model_parser.py
sp6370: Yes, it based on the call sites.

rossdanlm: It's not backwards compatible, but we explicitly want to remove kwargs. I will list this as a potential breaking change in release notes, but this is intentional

python/tests/test_run_config.py
rossdanlm: I think it's good to add, we can edit it later

rossdanlm (approved):
I think this is good, thanks for the fix, pls ping me when this lands!

python/src/aiconfig/default_parsers/parameterized_model_parser.py
sp6370: Correction! @rossdanlm is correct here!

I meant yes it is based on call sites.

rossdanlm (changes requested):
Actually wait sorry, can you pls also remove kwargs from https://github.com/lastmile-ai/aiconfig/blob/9617bd9d54e3bff3bb7eef2889af9f30adf39602/python/src/aiconfig/model_parser.py#L69 ? 

See more comments and details in https://github.com/lastmile-ai/aiconfig/pull/882

sp6370:
> Actually wait sorry, can you pls also remove kwargs from
> 
> https://github.com/lastmile-ai/aiconfig/blob/9617bd9d54e3bff3bb7eef2889af9f30adf39602/python/src/aiconfig/model_parser.py#L69
> 
> ?
> See more comments and details in #882

Hey Ross! 
I have addressed this. Also, here's a short video for testing Hugging Face Functionality.
https://www.loom.com/share/a7e09d403dcd4d30a7fd93f8a9e8b7c5

python/src/aiconfig/model_parser.py
```diff
@@ -66,7 +66,6 @@ async def run(
         aiconfig: AIConfig,
         options: Optional["InferenceOptions"] = None,
         parameters: Dict = {},
-        **kwargs,  # TODO: Remove this, just a hack for now to ensure that it doesn't break
```
rossdanlm, line 69: We need to add `run_with_dependencies` arg here otherwise it will fail. Ex: https://github.com/lastmile-ai/aiconfig/blob/b7b9910cc7abbd9b08dbdfe634e25f98396afcef/python/src/aiconfig/editor/server/server.py#L381

It's weird that your test plan showed this working, this should've failed. Can you do `pip list | grep aiconfig` and make sure that your aiconfig is linked to local and not the production package? If it's not seeing `python-aiconfig` to a local path for you, you need to go to `aiconfig/python` and run `pip install -e .`

rossdanlm (changes requested):
Pls ensure that when testing it's linked to the local pacakge

python/src/aiconfig/model_parser.py
sp6370: @rossdanlm Yes! I am using the local package. I resintall the source package at 0.30 in the test demo video. 

As for why if doesn't fail is because `HuggingFaceTextSummarizationRemoteInference` inherits from `ParameterizedModelParser` and run with dependencies is taken care of for Parameterized Model Parser.

As for `ModelParser`, we don't have `run_with_dependencies` method implemented.

sp6370: https://github.com/lastmile-ai/aiconfig/blob/b7b9910cc7abbd9b08dbdfe634e25f98396afcef/python/src/aiconfig/editor/server/server.py#L381

I'm unsure about the mechanism behind aiconfig.run in determining which version of run to employ. Based on my understanding, it might be contingent upon the behavior defined in the model registry. Specifically, for HuggingFace models, my speculation is that it defaults to utilizing the ParameterizedModelParser.

rossdancraig: > As for why if doesn't fail is because HuggingFaceTextSummarizationRemoteInference inherits from ParameterizedModelParser and run with dependencies is taken care of for Parameterized Model Parser.

Sorry can you please test on models like `HuggingFaceImage2TextRemoteInference` (ex: image_caption prompt) which does not depend on `ParameterizedModelParser` and inherits directly from `ModelParser`. This is what I meant that I suspect this these should fail 

> I'm unsure about the mechanism behind aiconfig.run in determining which version of run to employ

All the model parsers that are text-based inputs, will inherit from `ParameterizedModelParser` which inherits from `ModelParser`. All the models that are attachment-based inputs (ex: `HuggingFaceImage2TextRemoteInference`) will inherit from `ModelParser` directly instead

python/src/aiconfig/Config.py
```diff
@@ -309,7 +309,7 @@ async def run(
                 "prompt_name": prompt_name,
                 "params": params,
                 "options": options,
-                "kwargs": kwargs,
+                "run_with_dependencies": run_with_dependencies,
```
rossdanlm, line 312: Pls update other callsites to be consistent

python/tests/conftest.py
```diff
@@ -99,6 +99,99 @@ def mock_openai_chat_completion(**kwargs):
...
+                    },
+                }
+            ),
+        ],
+    ]
+
+    for input_params, response in response_map_list:
+        if kwargs == input_params:
+            return response
+    raise Exception("Unexpected arguments:\n {}".format(kwargs))
```
rossdanlm, line 192: Probably shoudl remove now to show that it works now that kwargs is removed

python/tests/conftest.py
```diff
@@ -99,6 +99,99 @@ def mock_openai_chat_completion(**kwargs):
     raise Exception("Unexpected arguments:\n {}".format(kwargs))
 
 
+def mock_openai_chat_completion_with_dependencies(**kwargs):
```
rossdanlm, line 102: run_with_dependencies =....

```
self,
        prompt_name: str,
        params: Optional[dict] = None,
        options: Optional[InferenceOptions] = None,
        run_with_dependencies: Optional[bool] = False,
```

rossdanlm (commented):
Generally lgtm, just make sure that we've chcked all the default model parsers too:

`aiconfig/python/src/aiconfig/default_parsers/.*`

python/src/aiconfig/Config.py
sp6370: Done! Updated all the call sites.

python/src/aiconfig/model_parser.py
sp6370: Updated the `ModelParser` as requested.

sp6370:
Cross checked the default model parsers `aiconfig/python/src/aiconfig/default_parsers/.*`. They don't have any implementations of `run` which requires update.

python/tests/conftest.py
sp6370: This doesn't reference run with dependencies.

python/tests/test_run_config.py
sp6370: I have e-enabled the test.

python/tests/test_run_config.py
```diff
@@ -48,3 +53,37 @@ async def test_load_parametrized_data_config(set_temporary_env_vars):
...
+        side_effect=mock_openai_chat_completion_with_dependencies
+    )
+
+    with patch.object(
+        openai.resources.chat.Completions,
+        "create",
+        new=mock_openai,
+    ):
+        config_relative_path = (
+            "aiconfigs/tarvel_gpt_prompts_with_dependency.json"
```
rossdanlm, line 72: nit: travel typo

rossdanlm (approved):
Nice! Thanks for the fixes! I'll follow up later with extensions + publishing requirements!

sp6370:
@rossdancraig I share the info on the test plan later today.

sp6370:
Tests: 
- Hugging Face Local Inference: https://www.loom.com/share/256142de11e147d19640ce5fd075383b?sid=546a14da-5907-4b19-a9ae-5b97de16e543

- Hugging Face Remote Inference: https://www.loom.com/share/cb99fab92e7a47f78aae09906ef3f44a?sid=7c71e5d7-28da-4e0f-aa01-065db9085a04

rossdanlm:
> Tests:
> 
> * Hugging Face Local Inference: https://www.loom.com/share/256142de11e147d19640ce5fd075383b?sid=546a14da-5907-4b19-a9ae-5b97de16e543
> * Hugging Face Remote Inference: https://www.loom.com/share/cb99fab92e7a47f78aae09906ef3f44a?sid=7c71e5d7-28da-4e0f-aa01-065db9085a04

Thank you for the super detailed test plans!!!

PR #919, opened 2024-01-14
Replace RUN_PROMPT with STREAM_AICONFIG_CHUNK inside of streaming implementation

rossdanlm:
Replace RUN_PROMPT with STREAM_AICONFIG_CHUNK inside of streaming implementation

This is the exact same logic, but I'm just going to do this as intermediate step to keep things clear and easy to review

## Test Plan
No functional changes, streaming and non streaming models still work

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/919).
* #928
* #926
* #925
* #924
* #922
* #921
* #920
* __->__ #919
* #918
* #917

PR #926, opened 2024-01-15
Split Action Types and AIConfigReducer into it's own folder

rossdanlm:
Split Action Types and AIConfigReducer into it's own folder

I just feel having the action type definitions vs. the reducer implementation make this a lot easier to read. You can now have file side by side instead of having to jump back and forth across the same one. I also merged the run commands together under the `RunPromptAction` type to make things cleaner as well. This will make it easier next diff once I merge logic together for the run prompts

I also cleaned up the `UPDATE_PROMPT_INPUT` implementation inside of the aiConfigReducer because it wasn't passing in the correct input type anyways, so now that's cleaner too.

## Test Plan
Everything works the same as before

https://github.com/lastmile-ai/aiconfig/assets/151060367/46c11941-c055-4307-99b2-3cd6202b78cb

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/926).
* #928
* __->__ #926
* #925
* #924
* #922
* #921
* #920
* #919
* #918
* #917

### DIFF
--- a/python/src/aiconfig/editor/client/src/LocalEditor.tsx
+++ b/python/src/aiconfig/editor/client/src/LocalEditor.tsx
@@ -98,8 +98,8 @@ export default function Editor() {
           output_chunk: (data) => {
             onStream({ type: "output_chunk", data: data as Output });
           },
-          aiconfig: (data) => {
-            onStream({ type: "aiconfig", data: data as AIConfig });
+          aiconfig_chunk: (data) => {
+            onStream({ type: "aiconfig_chunk", data: data as AIConfig });
           },
           stop_streaming: (_data) => {
             onStream({ type: "stop_streaming", data: null });
--- a/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
+++ b/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
@@ -670,8 +670,8 @@ export default function EditorContainer({
 
                 showNotification({
                   title: `Execution interrupted for prompt${
-                    promptName ? ` ${promptName}` : ""
-                  }. Resetting to previous state.`,
+                    promptName ? ` '${promptName}'` : ""
+                  }. Resetting output to previous state.`,
                   message: event.data.message,
                   color: "yellow",
                 });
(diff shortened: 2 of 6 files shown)


rholinshead (approved):
> I also cleaned up the `UPDATE_PROMPT_INPUT` implementation inside of the aiConfigReducer because it wasn't passing in the correct input type anyways, so now that's cleaner too.

Can you explain what you mean? As far as I can tell the implementation here is the same except inline instead of a helper function, which is fine - we can add it back if we have other actions that update input later.

rossdanlm:
> > I also cleaned up the `UPDATE_PROMPT_INPUT` implementation inside of the aiConfigReducer because it wasn't passing in the correct input type anyways, so now that's cleaner too.
> 
> Can you explain what you mean? As far as I can tell the implementation here is the same except inline instead of a helper function, which is fine - we can add it back if we have other actions that update input later.

So the method `reduceReplacePrompt()` expects a replace function callback of type `(prompt) -> ClientPrompt`. Before this replace function was of type `() -> ClientPrompt`

Screenshot 2024-01-15 at 15 16 03

For some reason this error was only flagged to me once I moved around logic in this diff. Let me know if I'm missing anything

rholinshead:
> > > I also cleaned up the `UPDATE_PROMPT_INPUT` implementation inside of the aiConfigReducer because it wasn't passing in the correct input type anyways, so now that's cleaner too.
> > 
> > 
> > Can you explain what you mean? As far as I can tell the implementation here is the same except inline instead of a helper function, which is fine - we can add it back if we have other actions that update input later.
> 
> So the method `reduceReplacePrompt()` expects a replace function callback of type `(prompt) -> ClientPrompt`. Before this replace function was of type `() -> ClientPrompt`
> 
> Screenshot 2024-01-15 at 15 16 03
> For some reason this error was only flagged to me once I moved around logic in this diff. Let me know if I'm missing anything

In `reduceReplaceInput` we define that `(prompt) => ClientPrompt` function inline as:
```
(prompt) => ({
    ...prompt,
    input: replacerFn(prompt.input),
  })
```
so it's not an error

rossdanlm:
Ohhhh, thanks that makes sense! I think I just removed it to keep it cleaner. It felt a bit weird to me we were using this as a helper function when we already had the more generic `reduceReplacePrompt()`!

PR #927, opened 2024-01-15
[ez] Rename logEvent handler to logEventHandler

rossdanlm:
[ez] Rename logEvent handler to logEventHandler

There's the type `logEvent` and the handler `logEvent` which can be confusing. Having this as it's own separate variable name names it clear that this is a handler, not a type.

## Test Plan
no functional changes

### DIFF
--- a/python/src/aiconfig/editor/client/src/contexts/AIConfigContext.tsx
+++ b/python/src/aiconfig/editor/client/src/contexts/AIConfigContext.tsx
@@ -7,7 +7,7 @@ import { ClientAIConfig, LogEvent, LogEventData } from "../shared/types";
  */
 const AIConfigContext = createContext<{
   getState: () => ClientAIConfig;
-  logEvent?: (event: LogEvent, data?: LogEventData) => void;
+  logEventHandler?: (event: LogEvent, data?: LogEventData) => void;
   readOnly?: boolean;
 }>({
   getState: () => ({ prompts: [], _ui: { isDirty: false } }),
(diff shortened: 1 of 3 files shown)


python/src/aiconfig/editor/client/src/contexts/AIConfigContext.tsx
```diff
@@ -7,7 +7,7 @@ import { ClientAIConfig, LogEvent, LogEventData } from "../shared/types";
  */
 const AIConfigContext = createContext<{
   getState: () => ClientAIConfig;
-  logEvent?: (event: LogEvent, data?: LogEventData) => void;
+  logEventHandler?: (event: LogEvent, data?: LogEventData) => void;
```
rholinshead, line 10: nit, I kind of like the context one just being `logEvent` for succinctness since I usually think of 'handler' suffix referring to a callback prop vs a function from context

rholinshead (approved):

python/src/aiconfig/editor/client/src/contexts/AIConfigContext.tsx
rossdanlm: Oh ok, I really want the function itself to have a different name though than the type event param that it takes in. Otherwise it's not obvious what `logEvent` really means (I know one has capitalized L but still not super clear). Do you have any other suggestions? If so I'll change. If not, I'll just land and we can change later, not a P0

PR #928, opened 2024-01-15
Merge STOP_STREAMING event into RUN_PROMPT_SUCCESS

rossdanlm:
Merge STOP_STREAMING event into RUN_PROMPT_SUCCESS

I just feel it makes a bit of sense to merge these two events together. I think this will be the final refacotring change I do

I also:

1. Changed the `id` fields to `promptId` to be more explicit
2. Centralized logic for setting the ClientAIConfig into a `setRunningPromptId` function
3. Re-arranged the action cases in the aiconfigReducer switch to have the run actions together, as well as save success at the end

## Test Plan
Everything still works as usual for streaming and non-streaming. Same video as the PRs earlier in this stack

### DIFF
--- a/python/src/aiconfig/editor/client/src/reducers/aiconfigReducer.ts
+++ b/python/src/aiconfig/editor/client/src/reducers/aiconfigReducer.ts
@@ -34,6 +34,19 @@ function reduceInsertPromptAtIndex(
   };
 }
 
+function setRunningPromptId(
+  state: ClientAIConfig,
+  runningPromptId: string | undefined
+): ClientAIConfig {
+  return {
+    ...state,
+    _ui: {
+      ...state._ui,
+      runningPromptId,
+    },
+  };
+}
+
 function reduceConsolidateAIConfig(
   state: ClientAIConfig,
   action: ConsolidateAIConfigSubAction,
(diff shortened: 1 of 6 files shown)


rholinshead (approved):

python/src/aiconfig/editor/client/src/reducers/aiconfigReducer.ts
```diff
@@ -365,6 +217,129 @@ export default function aiconfigReducer(
         },
       };
     }
+    case "RUN_PROMPT_START": {
+      const runningState = setRunningPromptId(dirtyState, action.promptId);
```
rholinshead, line 221: I don't think it will be significant, but just noting this technically results in an extra object creation/destructured copy (vs just doing a single one that sets dirty and running prompt id simultaneously). Can revisit later if needed

rossdanlm: Is this an extra creation/destruction from the previous implementation because it's now a function instead of inlined?

Screenshot 2024-01-15 at 15 21 41

If so, I'd probably still prefer to do it this way because it's re-used in many different action types and keeps logic centralized and easier to understand

PR #929, opened 2024-01-15
[rfc][sdk] 1/n Standardize Input Schema - Input Attachment Data

Ankush-lastmile:
[rfc][sdk] 1/n Standardize Input Schema - Input Attachment Data

Starting point of standardizing Input Schema for prompts

This diff introduces the type `AttachmentDataWithStringValue` to python and typescript sdk which will allow attachments to have structure as well as validation in python (pydantic validation).

## testplan

Screenshot 2024-01-15 at 5 52 11 PM

### DIFF
--- a/python/src/aiconfig/schema.py
+++ b/python/src/aiconfig/schema.py
@@ -79,6 +79,15 @@ OutputDataWithValue = Union[
     OutputDataWithToolCallsValue,
 ]
 
+class AttachmentDataWithStringValue(BaseModel):
+    """
+    This represents the attachment data that is stored as a string, but we use
+    both the `kind` field here and the `mime_type` in Attachment to convert
+    the string into the input format we want.
+    """
+
+    kind: Literal["file_uri", "base64"]
+    value: str
 
 class ExecuteResult(BaseModel):
     """
@@ -152,7 +161,7 @@ class Attachment(BaseModel):
     """
 
     # The data representing the attachment
-    data: Any
+    data: Union[AttachmentDataWithStringValue, str, Any]
     # The MIME type of the result. If not specified, the MIME type will be assumed to be text/plain
     mime_type: Optional[str] = None
     # Output metadata
--- a/typescript/types.ts
+++ b/typescript/types.ts
@@ -72,11 +72,22 @@ export type SchemaVersion =
   | "v1"
   | "latest";
 
+/**
+ * This represents the attachment data that is stored as a string, but we use
+ * both the `kind` field here and the `mime_type` in Attachment to convert
+ * the string into the input format we want.
+ */
+type AttachmentDataWithStringValue = {
+  kind: "file_uri" | "base64";
+  value: string;
+};
+
+
 export type Attachment = {
   /**
    * The data representing the attachment
    */
-  data: JSONValue;
+  data: JSONValue | AttachmentDataWithStringValue;
 
   /**
    * The MIME type of the result. If not specified, the MIME type will be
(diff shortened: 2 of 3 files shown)


python/src/aiconfig/schema.py
```diff
@@ -79,6 +79,15 @@ class OutputDataWithToolCallsValue(BaseModel):
...
 
+class AttachmentDataWithStringValue(BaseModel):
+    """
+    This represents the input data that is storied as a string, but we use
+    both the `kind` field here and the `mime_type` to convert
+    the string into the input format we want.
+    """
+
+    kind: Literal["file_uri", "base64"]
+    value: str
```
Ankush-lastmile, line 90: This type and code is pretty much a duplicate of 
```
class OutputDataWithStringValue(BaseModel):
    """
    This represents the output content that is storied as a string, but we use
    both the `kind` field here and the `mime_type` in ExecuteResult to convert
    the string into the output format we want.
    """

    kind: Literal["file_uri", "base64"]
    value: str
```

Should this be merged into one type and abstracted? That would reduce the explicitness and readability of this addition

python/src/aiconfig/schema.py
```diff
@@ -79,6 +79,15 @@ class OutputDataWithToolCallsValue(BaseModel):
     OutputDataWithToolCallsValue,
 ]
 
+class AttachmentDataWithStringValue(BaseModel):
+    """
+    This represents the input data that is storied as a string, but we use
```
rholinshead, line 84: nit, `input data that is storied` -> `attachment data that is stored` and should say 'the `mime_type` in Attachment'

typescript/types.ts
```diff
@@ -72,11 +72,22 @@ export type SchemaVersion =
   | "v1"
   | "latest";
 
+/**
+ * This represents the input data that is storied as a string, but we use
```
rholinshead, line 76: same here

rholinshead (approved):

rholinshead:
Nice, LGTM, just some comments. Would you be able to hook this up to the editor in a subsequent PR so that we are using the correct types there? I wonder if we can just remove the attachments stuff from PromptInputObjectSchema now with this? There's technically support for attachment metadata there but it could also just be set with the JSON toggle

Ankush-lastmile:
>Would you be able to hook this up to the editor in a subsequent PR so that we are using the correct types there? I wonder if we can just remove the attachments stuff from PromptInputObjectSchema now with this? There's technically support for attachment metadata there but it could also just be set with the JSON toggle

Looking into this, will sync offline

python/src/aiconfig/schema.py
```diff
@@ -152,7 +161,7 @@ class Attachment(BaseModel):
     """
 
     # The data representing the attachment
-    data: Any
+    data: Union[AttachmentDataWithStringValue, Any]
```
saqadri, line 164:
```suggestion
    data: Union[AttachmentDataWithStringValue, str, Any]
```

Technically, an attachment can also be a string, just like `ExecuteResult.data`: `data: Union[OutputDataWithValue, str, Any]`

saqadri (approved):
One change required, otherwise looks good. This doesn't actually update the model parsers it seems to actually use this new type yet, coming on top of this I assume?

python/src/aiconfig/schema.py
Ankush-lastmile: discussed offline: updating to include `str` to be explicit, since we know some models like gemini take in multiple parts or stablediffusion has a 'negative prompt` which is just as much as input. We can iterate as we go about how this looks.

Typescript is defined as JSONValue so no need to add explicit string; Already supported.

Ankush-lastmile:
> This doesn't actually update the model parsers it seems to actually use this new type yet, coming on top of this I assume?

Yes; those diffs will come on top

PR #932, opened 2024-01-15
[extensions] 2/n update asr model parser with input schema

Ankush-lastmile:
[extensions] 2/n update asr model parser with input schema

With a new input attachment type `AttachmentDataWithStringValue` defined, This diff updates the ASR Model parser to validate input is in the expected form/type.

Also made some updates to to make input attachment validation clearer.

## Testplan
Screenshot 2024-01-16 at 3 29 02 PM

### Dependencies
Sapling removed the dependency pr. Depends on
#929

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/932).
* #944
* #940
* __->__ #932

### DIFF
--- a/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/automatic_speech_recognition.py
+++ b/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/automatic_speech_recognition.py
@@ -5,7 +5,7 @@ from transformers import pipeline, Pipeline
 from aiconfig_extension_hugging_face.local_inference.util import get_hf_model
 from aiconfig import ModelParser, InferenceOptions
 from aiconfig.callback import CallbackEvent
-from aiconfig.schema import Prompt, Output, ExecuteResult, Attachment
+from aiconfig.schema import AttachmentDataWithStringValue, Prompt, Output, ExecuteResult, Attachment, PromptInput
 
 if TYPE_CHECKING:
     from aiconfig import AIConfigRuntime
@@ -154,7 +154,7 @@ def validate_attachment_type_is_audio(attachment: Attachment):
 
 def validate_and_retrieve_audio_from_attachments(prompt: Prompt) -> list[str]:
     """
-    Retrieves the audio uri's from each attachment in the prompt input.
+    Retrieves the audio uri's or base64 from each attachment in the prompt input.
 
     Throws an exception if
     - attachment is not audio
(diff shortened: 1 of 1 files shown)


extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/automatic_speech_recognition.py
```diff
@@ -163,21 +163,24 @@ def validate_and_retrieve_audio_from_attachments(prompt: Prompt) -> list[str]:
...
-    audio_uris: list[str] = []
+    audio_inputs: list[str] = []
 
     for i, attachment in enumerate(prompt.input.attachments):
         validate_attachment_type_is_audio(attachment)
 
-        if not isinstance(attachment.data, str):
-            # See todo above, but for now only support uri's
-            raise ValueError(f"Attachment #{i} data is not a uri. Please specify a uri for the audio attachment in prompt {prompt.name}.")
+        if not isinstance(attachment.data, AttachmentDataWithStringValue):
```
rholinshead, line 177: Do we need to have backwards-compatibility here for potential existing configs with the older format?

Ankush-lastmile: > Do we need to have backwards-compatibility here for potential existing configs with the older format?

I believe gradio example was the only config that used it. I think also what @saqadri meant by backwards-compatibility was that the AIConfig itself should be able to be loaded with `AIConfigRuntime.load()` if it doesn't use `AttachmentDataWithStringValue`

Ankush-lastmile: If that's not the case though we can allow arbitrary attachment data and let the model error as needed, but not sure if thats what we want

Ankush-lastmile: discussed offline with @saqadri, lets ship this, mark this as breaking change in changelog to keep anyone aware because this is technically a breaking change but doubt on who is using it but we want to inform anyone who is using it just in case.

rossdanlm (approved):
Alright cool, I'll make note of it in changelog

PR #935, opened 2024-01-16
[3/n] Read-only mode: Connect context to Add Prompt button

rossdanlm:
[3/n] Read-only mode: Connect context to Add Prompt button

For now I'm just returning empty component. It looks a bit weird to me right now, but I think that's becuase the background color of the prompt cells should be darker than the background color of the notebook itself. Once that's fixed it'll look better I think. From the Figma: https://www.figma.com/file/7YVa6KpEDgptj8DiAc5BaU/AiConfig-Editor?type=design&node-id=271-4785&mode=design

Screenshot 2024-01-16 at 08 52 01

## Test Plan
Change `readOnly = false` to `readOnly = true`

Before
Screenshot 2024-01-16 at 10 29 00

After
Screenshot 2024-01-16 at 10 28 34

## Old
Before
Screenshot 2024-01-16 at 08 55 58

After

Screenshot 2024-01-16 at 08 52 46

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
+++ b/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
@@ -918,12 +918,14 @@ export default function EditorContainer({
         onUpdateParameters={onUpdateGlobalParameters}
       />
       <Container maw="80rem" className={classes.promptsContainer}>
-        <div className={classes.addPromptRow}>
-          <AddPromptButton
-            getModels={callbacks.getModels}
-            addPrompt={(model: string) => onAddPrompt(0, model)}
-          />
-        </div>
+        {!readOnly && (
+          <div className={classes.addPromptRow}>
+            <AddPromptButton
+              getModels={callbacks.getModels}
+              addPrompt={(model: string) => onAddPrompt(0, model)}
+            />
+          </div>
+        )}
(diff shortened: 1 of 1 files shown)


rholinshead (changes requested):
In the screenshot it looks like the prompt component's width is reduced in readonly? I think we should fix that to make sure it stays full width.

A couple other things:
- should return null instead of <></> per recommended approach https://legacy.reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering; my understanding is that <></> will require extra work for react to construct the fragment each time
- in this case, I think we actually don't want to do this logic within the button itself. Generally if you aren't going to render a component at all based on logic available to the parent, that logic should be done by the parent of that component instead of within the component itself. It's especially useful to do that here because doing so will also skip the unnecessary attempt to load models (although readonly shouldn't have any loadModels implementation anyway, it's best to be cautious in case external implementations do that)

rossdanlm:
> like the prompt component's width is reduced in readonly? 

I don't think it is? Can you show where in screenshot it looks like this?

rholinshead:
> > like the prompt component's width is reduced in readonly?
> 
> I don't think it is? Can you show where in screenshot it looks like this?

Oh wait lol. It was the figma image I was looking at

rholinshead (approved):

PR #936, opened 2024-01-16
[4/n] Read-only mode:  Remove Clear Output and Save buttons

rossdanlm:
[4/n] Read-only mode:  Remove Clear Output and Save buttons

TSIA

## Test Plan
Change `readOnly = false` to `readOnly = true`
Before
Screenshot 2024-01-16 at 08 52 46

After
Screenshot 2024-01-16 at 11 14 05

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/936).
* __->__ #936
* #935

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
+++ b/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
@@ -878,32 +878,37 @@ export default function EditorContainer({
       <Container maw="80rem">
         <Flex justify="flex-end" mt="md" mb="xs">
           <Group>
-            <Button
-              loading={undefined}
-              onClick={onClearOutputs}
-              size="xs"
-              variant="gradient"
-            >
-              Clear Outputs
-            </Button>
-
-            <Tooltip
-              label={isDirty ? "Save changes to config" : "No unsaved changes"}
-            >
+            {!readOnly && (
               <Button
-                leftIcon={<IconDeviceFloppy />}
-                loading={isSaving}
-                onClick={() => {
-                  onSave();
-                  logEventHandler?.("SAVE_BUTTON_CLICKED");
-                }}
-                disabled={!isDirty}
+                loading={undefined}
+                onClick={onClearOutputs}
                 size="xs"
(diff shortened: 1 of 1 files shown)


rholinshead (approved):
Accepting to unblock but I think we could just not render the <ClearOutputsButton/> in readonly instead, especially since we've taken that stance with the AddPromptButton in previous PR instead of just making it disabled. Same with save button

rossdanlm:
> Accepting to unblock but I think we could just not render the in readonly instead

I think there's a typo, what do you not want to render? 

> especially since we've taken that stance with the RunPromptButton in previous PR instead of just making it disabled

We we did this for the run button in https://github.com/lastmile-ai/aiconfig/pull/916, do you mean pass in the disabled as a prop for the Button component instead or using `disabled` field directly? https://github.com/lastmile-ai/aiconfig/blob/17b54b5ffccf53d9429b77e9ab4d41e167524b52/python/src/aiconfig/editor/client/src/components/prompt/RunPromptButton.tsx#L33. Sure I can do that

rholinshead:
> > Accepting to unblock but I think we could just not render the in readonly instead
> 
> I think there's a typo, what do you not want to render?
> 
> > especially since we've taken that stance with the RunPromptButton in previous PR instead of just making it disabled
> 
> We we did this for the run button in #916, do you mean pass in the disabled as a prop for the Button component instead or using `disabled` field directly?
> 
> https://github.com/lastmile-ai/aiconfig/blob/17b54b5ffccf53d9429b77e9ab4d41e167524b52/python/src/aiconfig/editor/client/src/components/prompt/RunPromptButton.tsx#L33
> 
> . Sure I can do that

Sorry, realized I mistyped and edited too late. I meant the AddPromptButton in the previous PR, which we removed. I think we should also not render save/clear outputs/ in readonly.
I think we can render disabled runbutton in readonly just to give an idea that it was "run" and is now readonly. But the other buttons don't really mean anything from a presentational standpoint

PR #941, opened 2024-01-16
[editor] Make AIConfigEditor callbacks Prop Optional

rholinshead:
# [editor] Make AIConfigEditor callbacks Prop Optional

Make the `callbacks` prop optional so that we don't need to pass a huge callbacks object with no-ops in readonly mode (can just do <AIConfig config={config} readOnly/>)

## Testing:
With callbacks prop, make sure all actions work

Without callbacks prop, ensure all callback related options are just no-op'd and don't send any requests

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/prompt/PromptContainer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/prompt/PromptContainer.tsx
@@ -12,8 +12,8 @@ import ModelSelector from "./ModelSelector";
 
 type Props = {
   prompt: ClientPrompt;
-  cancel: (cancellationToken: string) => Promise<void>;
-  getModels: (search: string) => Promise<string[]>;
+  cancel?: (cancellationToken: string) => Promise<void>;
+  getModels?: (search: string) => Promise<string[]>;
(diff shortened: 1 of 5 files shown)


python/src/aiconfig/editor/client/src/components/prompt/PromptContainer.tsx
```diff
@@ -12,8 +12,8 @@ import ModelSelector from "./ModelSelector";
 
 type Props = {
   prompt: ClientPrompt;
-  cancel: (cancellationToken: string) => Promise<void>;
-  getModels: (search: string) => Promise<string[]>;
+  cancel?: (cancellationToken: string) => Promise<void>;
```
rossdanlm, line 15: nit: for future diff, can we call this `onCancel` so it's more obvious this is a callback? Not a blocker

rossdanlm: Oh nm, I see that this is just the method itself, not a callback that invokes other functionality. Interesting

rossdanlm (approved):
(y)

PR #942, opened 2024-01-16
yarn build

rossdanlm:
yarn build

Oh whoops, keep forgetting to do this when making client changes


saqadri (approved):

Ankush-lastmile:
> yarn build

do we need to do this on every pr?

rossdanlm:
> do we need to do this on every pr?

We need to do it on anything with `editor/client` changes

rholinshead:
We really only need to do it before publishing (and check pros server mode works)

PR #940, opened 2024-01-16
[extensions] 3\n Image 2 text

Ankush-lastmile:
[extensions] 3\n Image 2 text

With a new input attachment type `AttachmentDataWithStringValue` defined, This diff updates the HF image to text Model parser to validate input is in the expected form/type.

Also made some updates to to make input attachment validation clearer.

## Testplan
Screenshot 2024-01-16 at 3 32 52 PM

### Dependencies
Sapling removed the dependency pr. Depends on
#929

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/940).
* #944
* __->__ #940
* #932

PR #944, opened 2024-01-16
[editor] 4/n refactor input attachments to use new input data type

Ankush-lastmile:
[editor] 4/n refactor input attachments to use new input data type

- update renderer to acknowledge new type
- On New attachment, update aiconfig to use new values to input data when new file is updated
- updated Gradio AIConfig by removing and adding files.

Note: This diff doesn't test or change base64 functionality but should techncially be possible. <audio> HTML taglets you pass base64 source. Unsure about image

## Testplan:
1. edit config
2. reupload files & run. note: the incorrect rendering is due to the old aiconfig with the now incorrect input attachment data type

https://github.com/lastmile-ai/aiconfig/assets/141073967/eff42748-e3e6-48df-b90b-34bac4e997c4

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/944).
* __->__ #944
* #940
* #932

### DIFF
--- a/cookbooks/Gradio/huggingface.aiconfig.json
+++ b/cookbooks/Gradio/huggingface.aiconfig.json
@@ -26,11 +26,17 @@
       "input": {
         "attachments": [
           {
-            "data": "https://s3.amazonaws.com/lastmileai.aiconfig.public/uploads/2024_1_10_18_41_31/1742/bear_with_honey.png",
+            "data": {
+              "kind": "file_uri",
+              "value": "https://s3.amazonaws.com/lastmileai.aiconfig.public/uploads/2024_1_16_20_22_54/7748/fox_in_forest.png"
+            },
             "mime_type": "image/png"
           },
           {
-            "data": "https://s3.amazonaws.com/lastmileai.aiconfig.public/uploads/2024_1_10_18_41_31/7275/fox_in_forest.png",
+            "data": {
+              "kind": "file_uri",
+              "value": "https://s3.amazonaws.com/lastmileai.aiconfig.public/uploads/2024_1_16_20_20_57/508/bear_with_honey.png"
+            },
             "mime_type": "image/png"
           }
         ]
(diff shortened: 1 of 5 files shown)


cookbooks/Gradio/huggingface.aiconfig.json
```diff
@@ -221,7 +230,14 @@
...
       },
-      "outputs": []
+      "outputs": [
+        {
+          "output_type": "execute_result",
+          "execution_count": 0,
+          "data": " Hi.",
+          "metadata": {}
+        }
+      ]
```
Ankush-lastmile, line 240: noticed the asr prompt didn't have an output persisted but all the other ones did

python/src/aiconfig/editor/client/src/components/prompt/prompt_input/attachments/AttachmentContainer.tsx
```diff
@@ -41,7 +41,7 @@ export default memo(function AttachmentContainer({
       </Flex>
       <MimeTypeRenderer
         mimeType={attachment.mime_type}
-        content={attachment.data as string}
+        content={(attachment.data as AttachmentDataWithStringValue).value as string}
```
rholinshead, line 44: Nice. In a followup PR, I think it would be good to add 'kind' as a prop (can default to "file_uri") to MimeTypeRenderer and pass it in here and in PromptOutputsRenderer then handle the `data:${output.mime_type};base64,` prefix within MimeTypeRenderer; for PromptOutputsRenderer, we'd be able to get rid of a bit of code to make it
```
      case "file_uri":
      case "base64":
        return (
          <PromptOutputWrapper
            copyContent={(output.data as OutputDataWithValue).value as string}
            output={output}
            withRawJSONToggle
          >
            <MimeTypeRenderer
              mimeType={output.mime_type}
              content={(output.data as OutputDataWithValue).value as string}
              kind={output.data.kind as "file_uri" | "base64"}
            />
          </PromptOutputWrapper>
        );
```
instead of 2 separate cases

PR #946, opened 2024-01-16
[editor] add telemetry for more events

Ankush-lastmile:
[editor] add telemetry for more events

As discussed offline, we want to enable telemetry in the local editor for the following events:

- RUN_PROMPT_START
- RUN_PROMPT_CANCEL
- RUN_PROMPT_ERROR
- RUN_PROMPT_SUCCESS

This diff implements the logging telemetry for those events.
- Add mode "local" to all telemetry sent from local editor in prod mode
- Send telemetry log events after dispatching rather than before, so that the client render action gets called first.
- - This technically shouldn't matter since these are all async calls.
- Note: This only implements logging for Local Editor. Next up is to add the datadog initialization to Gradio Workbook

## Testplan

1. yarn build -> run editor in 'prod' mode
2. run prompt, cancel, rerun to success. Check telemetry sent to datadog.

https://github.com/lastmile-ai/aiconfig/assets/141073967/52043456-3682-4dd7-b95b-846ad7480019

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
+++ b/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
@@ -681,6 +681,7 @@ export default function AIConfigEditor({
         promptId,
         cancellationToken,
       });
+      logEventHandler?.("RUN_PROMPT_START");
 
       const onPromptError = (message: string | null) => {
         dispatch({
@@ -688,6 +689,7 @@ export default function AIConfigEditor({
           promptId,
           message: message ?? undefined,
         });
+        logEventHandler?.("RUN_PROMPT_ERROR");
 
         const promptName = getPrompt(stateRef.current, promptId)?.name;
 
@@ -732,6 +734,7 @@ export default function AIConfigEditor({
                 type: "RUN_PROMPT_SUCCESS",
                 promptId,
               });
+              logEventHandler?.("RUN_PROMPT_SUCCESS");
             }
           },
           (event) => {
@@ -748,6 +751,7 @@ export default function AIConfigEditor({
                   // Returned config output is reset to before running RUN_PROMPT
                   config: event.data.data,
                 });
+                logEventHandler?.("RUN_PROMPT_CANCELED");
 
                 const promptName = getPrompt(stateRef.current, promptId)?.name;
 
@@ -770,11 +774,13 @@ export default function AIConfigEditor({
         // Keep this here in case any server implementations don't return
         // aiconfig as a streaming format
         if (serverConfigResponse?.aiconfig) {
+          // Do we need to log here
           dispatch({
             type: "RUN_PROMPT_SUCCESS",
             promptId,
             config: serverConfigResponse.aiconfig,
           });
+          logEventHandler?.("RUN_PROMPT_SUCCESS");
         }
       } catch (err: unknown) {
         const message = (err as RequestCallbackError).message ?? null;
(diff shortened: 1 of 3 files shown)


python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
```diff
@@ -681,13 +681,15 @@ export default function AIConfigEditor({
...
       });
+      logEventHandler?.("RUN_PROMPT_START");
 
       const onPromptError = (message: string | null) => {
         dispatch({
           type: "RUN_PROMPT_ERROR",
           promptId,
           message: message ?? undefined,
         });
+        logEventHandler?.("RUN_PROMPT_ERROR");
```
rholinshead, line 692: Should we log the error message in the data as well? And maybe the prompt model

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
```diff
@@ -681,13 +681,15 @@ export default function AIConfigEditor({
         promptId,
         cancellationToken,
       });
+      logEventHandler?.("RUN_PROMPT_START");
```
rholinshead, line 684: Would probably be good to add the prompt model to the logged event data

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
```diff
@@ -748,6 +751,7 @@ export default function AIConfigEditor({
                   // Returned config output is reset to before running RUN_PROMPT
                   config: event.data.data,
                 });
+                logEventHandler?.("RUN_PROMPT_CANCELED");
```
rholinshead, line 754: Same here r.e. model

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
```diff
@@ -770,11 +774,13 @@ export default function AIConfigEditor({
         // Keep this here in case any server implementations don't return
         // aiconfig as a streaming format
         if (serverConfigResponse?.aiconfig) {
+          // Do we need to log here
```
rholinshead, line 777: Yes, this is for external implementations that might not have streaming implemented for the run callback

rholinshead (approved):
Accepting to unblock, but would be good to add additional data to the events, mainly so that we can understand which models/parsers are being run most (and which ones are errored/cancelled).


Long-term we can also consider logging a unique id for each distinct run -> success/error/cancel flow so that we can add monitoring for when models/parsers are errored/cancelled at a higher-than-normal rate

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
Ankush-lastmile: I was thinking so too but I think righ now the error logs pass the entire aiconfig. wasn't sure if we want to be logging that since we said we don't want to log the config. Thoughts? 

Should we pass in the message?

rholinshead: If the `message` includes the entire aiconfig, we shouldn't log that (and should probably fix to not do that). I  was pretty sure it just returns an error message without config, though

Ankush-lastmile:
- added model to Run events. It wasn't straight forward to find the model so did some logic using stateaiconfig -> getPrompt -> getModelName
- rebase

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
Ankush-lastmile: added and removed comment

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
Ankush-lastmile: going to ship this and then investigate this (and make change as necessary ontop)

PR #950, opened 2024-01-17
Add 'mode' Prop to AIConfigEditor & Map to Associated Theme

rholinshead:
# Add 'mode' Prop to AIConfigEditor & Map to Associated Theme

Add an optional 'mode' prop to the AIConfigEditor to use for telemetry and predefined styling (for 'local' editor, 'gradio' or 'vscode') which can be used when we package the editor for external use. I decided to make this optional so that we can support custom styles for external deployments (we can document the classes to use), rather than defaulting to 'local', but we can revisit if we don't want to support that level of customizability.

## Local
Screenshot 2024-01-16 at 7 14 41 PM

## Gradio
Screenshot 2024-01-16 at 7 14 18 PM

## No Mode
Screenshot 2024-01-16 at 7 22 56 PM

### DIFF
--- a/python/src/aiconfig/editor/client/src/LocalEditor.tsx
+++ b/python/src/aiconfig/editor/client/src/LocalEditor.tsx
@@ -4,7 +4,7 @@ import AIConfigEditor, {
   RunPromptStreamErrorCallback,
   RunPromptStreamErrorEvent,
 } from "./components/AIConfigEditor";
-import { Flex, Loader, MantineProvider, Image } from "@mantine/core";
+import { Flex, Loader, Image, createStyles } from "@mantine/core";
 import {
   AIConfig,
   InferenceSettings,
@@ -19,8 +19,27 @@ import { streamingApiChain } from "./utils/oboeHelpers";
 import { datadogLogs } from "@datadog/browser-logs";
 import { LogEvent, LogEventData } from "./shared/types";
 
-export default function Editor() {
+const useStyles = createStyles(() => ({
+  editorBackground: {
+    background:
+      "radial-gradient(ellipse at top,#08122d,#030712),radial-gradient(ellipse at bottom,#030712,#030712)",
+    margin: "0 auto",
+    minHeight: "100vh",
+  },
+
+  logo: {
+    maxWidth: "80rem",
+    margin: "0 auto",
+    padding: "32px 0 0 32px",
+    display: "flex",
+    justifyContent: "space-between",
+    alignItems: "center",
+  },
+}));
+
+export default function LocalEditor() {
   const [aiconfig, setAiConfig] = useState<AIConfig | undefined>();
+  const { classes } = useStyles();
 
   const loadConfig = useCallback(async () => {
     const res = await ufetch.post(ROUTE_TABLE.LOAD, {});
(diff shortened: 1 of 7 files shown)


python/src/aiconfig/editor/client/src/LocalEditor.tsx
```diff
@@ -19,8 +19,27 @@ import { streamingApiChain } from "./utils/oboeHelpers";
...
+  logo: {
+    maxWidth: "80rem",
+    margin: "0 auto",
+    padding: "32px 0 0 32px",
+    display: "flex",
+    justifyContent: "space-between",
+    alignItems: "center",
+  },
+}));
+
```
Ankush-lastmile, line 39: This sets the logo/top bar to local editor theme for every mode right? Is this intended?

Screenshot 2024-01-17 at 12 14 22 PM

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
```diff
@@ -931,134 +935,143 @@ export default function EditorContainer({
...
-            <Flex align="center">
-              <CopyButton
-                value={JSON.stringify(
-                  clientConfigToAIConfig(aiconfigState),
-                  null,
-                  2
+    <AIConfigEditorThemeProvider mode={mode}>
+      <AIConfigContext.Provider value={contextValue}>
+        <Notifications />
+        <div className="editorBackground">
```
Ankush-lastmile, line 941: is this className=editorBackground automatically available when `AIConfigEditorThemeProvider` is used?

python/src/aiconfig/editor/client/src/themes/AIConfigEditorThemeProvider.tsx
```diff
@@ -0,0 +1,36 @@
...
+type Props = {
+  children: React.ReactNode;
+  mode?: AIConfigEditorMode;
+};
+
+const THEMES = {
+  local: LOCAL_THEME,
+  gradio: GRADIO_THEME,
+  // TODO: Implement VSCODE_THEME
+  vscode: LOCAL_THEME,
```
Ankush-lastmile, line 16: is there a figma or already-made design for the vscode theme?

python/src/aiconfig/editor/client/src/themes/LocalTheme.ts
```diff
@@ -0,0 +1,167 @@
+import { MantineThemeOverride } from "@mantine/core";
```
Ankush-lastmile, line 1: What does override mean here? I tried looking at the docs but wasn't sure what exactly it would be "overriding"

python/src/aiconfig/editor/client/src/LocalEditor.tsx
rholinshead: Ya, it's intended in order to show the image in local editor. It's only in our own local editor component, where mode will always be 'local', so shouldn't be a problem

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
rholinshead: Yes, as long as the themes define it. The mantine provider will put all the theme-defined classes into global css

python/src/aiconfig/editor/client/src/themes/LocalTheme.ts
rholinshead: Mantine has a default theme if no MantineProvider is used to override it; the default has primary color of blue for buttons, hover outlines, etc., as well as specific fonts for text

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
```diff
@@ -931,134 +935,143 @@ export default function EditorContainer({
...
+                      />
+                    </div>
+                  )}
+                </Stack>
+              );
+            })}
+          </Container>
+        </div>
+      </AIConfigContext.Provider>
+    </AIConfigEditorThemeProvider>
```
Ankush-lastmile, line 1075: question for myself to understand react/mantine: I'm looking at `RunPromptButton.tsx`, how does it pull the styling/theme from the relevant theme.ts for itself?

Ankush-lastmile (approved):
lgtm

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
Ankush-lastmile: nvm I see that the button is defined as 
```
    <Button
      onClick={onClick}
      disabled={disabledOrReadOnly}
      p="xs"
      size="xs"
      className="runPromptButton"
    >
    ```
    
    and the theme.ts has runPromptButton defined

rholinshead: The theme/styles would be applied by:
- any element rendered by RunPromptButton which has a class name will have the style applied from the global styles set by theme (e.g. runPromptButton class)
- within mantine core code, the Button, Flex, Loader, Tooltip components used within RunPromptButton will themselves have styles applied based on applicable values set in context from the closest MantineThemeProvider

PR #951, opened 2024-01-17
[AIC-py][eval][1/n] generalize to take params dict input

jonathanlastmileai:
[AIC-py][eval][1/n] generalize to take params dict input

We should be able to take an arbitrary params dict, not just a single input
associated with "{{the_query}}".

This generalizes UserTestSuiteWithInputs to allow a dict, and makes the necessary
internal changes to respect that (while being near-perfectly type safe :D)

lib: unit tests. The tests are pretty thorough, checking dynamic types that are
not checkable statically, and fuzzy testing the possible inputs using
the hypothesis library.

This also affects this script which is not e2e tested, so I did that manually.

run_aiconfig script:

`python -m 'aiconfig.scripts.run_aiconfig' travel_aiconfig_test_suite_settings.json food`

...
```
1. Savour culinary delights at Chelsea Market.
2. Sample diverse food at Smorgasburg weekend market.
3. Explore immersive exhibits at the Museum of Ice Cream.
```

### DIFF
--- a/python/src/aiconfig/eval/lib.py
+++ b/python/src/aiconfig/eval/lib.py
@@ -1,6 +1,8 @@
 import asyncio
+import json
 import logging
 from dataclasses import dataclass
+from frozendict import frozendict
 from functools import partial
 from typing import Any, Generic, NewType, Sequence, Tuple, TypeVar
 
@@ -17,10 +19,20 @@ LOGGER = logging.getLogger(__name__)
 
 # TODO: figure out a way to do heterogenous list without Any
 # Each test is a (input_datum, Metric) pair
-UserTestSuiteWithInputs = Sequence[Tuple[str, Metric[str, Any]]]
+UserTestSuiteWithInputs = Sequence[Tuple[str | dict[str, str], Metric[str, Any]]]
 
 # Each test is a (output_datum, Metric) pair
 UserTestSuiteOutputsOnly = Sequence[Tuple[str, Metric[str, Any]]]
+
+
+# NOTE: it's probably better to avoid NewType in the future, because it doesn't
+# ... create a ... new type. For example, you can't pattern match against it.
+TextOutput = NewType("TextOutput", str)
+
+
+@dataclass(frozen=True)
+class TextBasedInputDatum:
+    value: str | frozendict[str, str]
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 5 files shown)


python/src/aiconfig/eval/lib.py
```diff
@@ -223,23 +264,23 @@ def _zip_inputs_outputs(outputs: list[TextOutput]):
...
-async def run_aiconfig_helper(runtime: AIConfigRuntime, prompt_name: str, question: str) -> Result[str, str]:
-    params = {
-        "the_query": question,
-    }
+async def run_aiconfig_on_text_based_input(runtime: AIConfigRuntime, prompt_name: str, params: TextBasedInputDatum) -> Result[str, str]:
+    def _get_params_for_aiconfig(params: TextBasedInputDatum) -> dict[str, str]:
+        match params.value:
+            case str(input_text):
+                return {"the_query": input_text}
+            case frozendict():
```
rholinshead, line 279: Is this checking if it's an empty frozendict right now (i.e. should this be instanceof)?

rholinshead (approved):
Just a question on the python match. Also, would the alternative to this be to just eval multiple prompt with the same input but dynamic parameters defined in the prompt itself?

Accepting to unblock, can probably just fix forward after folks with python experience have taken a look

jonathanlastmileai:
> Also, would the alternative to this be to just eval multiple prompt with the same input but dynamic parameters defined in the prompt itself

Explain more?

python/src/aiconfig/eval/lib.py
jonathanlastmileai: Very good question - I struggled with some of the subtleties of py3 pattern matching. One thing I learned is it doesn't really work with NewType, which doesn't, um, actually really make a new type :D.


To answer your question, this matches all frozendict instances. I don't fully understand how this works for primitives and non-record container types like this, but I have confirmed this works. 

The main thing about match...case is that pyright can check for exhaustive matching. So if this were only matching empty frozendicts, pyright would tell me I havent matched the entire input type.

rholinshead:
> > Also, would the alternative to this be to just eval multiple prompt with the same input but dynamic parameters defined in the prompt itself
> 
> Explain more?

Oh, I just meant from an eval perspective if you are running/evaluating different parameter values on the same prompt, an alternative would be to construct a new prompt per set of parameters to evaluate. Then you could have a single AIConfigRuntime which has each of the evaluations as prompts/outputs within it

PR #957, opened 2024-01-18
[easy][editor][client] 6/n Read-only mode: Global Parameters

Ankush-lastmile:
[easy][editor][client] 6/n Read-only mode: Global Parameters

- refactor `isReadonly` -> `readOnly` inside `ParametersRenderer.tsx`
- useState() to get readOnly bool

## Testplan

1. Pass readOnly=True prop in `AIConfigEditor`

| readOnly={false}  | readOnly={true} |
| Screenshot 2024-01-18 at 11 29 08 AM  | Screenshot 2024-01-18 at 11 32 28 AM |

## Dependencies
Built ontop of #939

#961 is next on the stack

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/ParametersRenderer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/ParametersRenderer.tsx
@@ -10,8 +10,9 @@ import {
 } from "@mantine/core";
 import { IconTrash, IconPlus } from "@tabler/icons-react";
 import { debounce, uniqueId } from "lodash";
-import { useState, useCallback, memo, useMemo } from "react";
+import { memo, useCallback, useContext, useMemo, useState } from 'react';
 import { JSONValue, JSONObject } from "aiconfig";
+import AIConfigContext from "../contexts/AIConfigContext";
 
 type Parameter = { parameterName: string; parameterValue: JSONValue };
 
@@ -36,7 +37,7 @@ const ParameterInput = memo(function ParameterInput(props: {
 }) {
   const { initialItemValue, removeParameter, onUpdateParameter } = props;
   // TODO: saqadri - update this once we have a readonly mode
-  const { isReadonly } = { isReadonly: false };
+  const { readOnly } = useContext(AIConfigContext);
 
   const [parameterName, setParameterName] = useState(
     initialItemValue?.parameterName ?? ""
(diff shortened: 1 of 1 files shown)


python/src/aiconfig/editor/client/src/components/ParametersRenderer.tsx
```diff
@@ -10,8 +10,9 @@ import {
 } from "@mantine/core";
 import { IconTrash, IconPlus } from "@tabler/icons-react";
 import { debounce, uniqueId } from "lodash";
-import { useState, useCallback, memo, useMemo } from "react";
+import { useState, useCallback, memo, useMemo, useContext } from "react";
```
rossdanlm, line 13: nit: let's order these alphabetically

python/src/aiconfig/editor/client/src/components/ParametersRenderer.tsx
```diff
@@ -100,7 +101,7 @@ const ParameterInput = memo(function ParameterInput(props: {
         />
         <Textarea
           placeholder="Enter parameter value"
```
rossdanlm, line 103: Not related to this diff, but cc @saqadri why is paramater name `TextInput` while parameter value is `TextArea`?

python/src/aiconfig/editor/client/src/components/ParametersRenderer.tsx
```diff
@@ -100,7 +101,7 @@ const ParameterInput = memo(function ParameterInput(props: {
...
-          disabled={isReadonly}
+          disabled={readOnly}
```
rossdanlm, line 104: I personally don't really like the greyed out style, would rather change theming (probably globally is better since we'll probably do this for text components in prompt input, prompt output, model settings, prompt params as well

cc @zakariassoul what do you think about the greyed out styling?

Not a blocker, but think it should be P1
cc @rholinshead can you give Ankush code pointers on this after bug-bash/MVP is released?

rossdanlm (approved):
Good to unblock, think it looks ugly though

rossdanlm (commented):
I also think we should remove the `Use parameters in your prompt or system prompt with` text for read-only mode

python/src/aiconfig/editor/client/src/components/ParametersRenderer.tsx
rholinshead: TextInput is intended for shorter text (like one line), text area is for larger text (multiple lines, expands)

python/src/aiconfig/editor/client/src/components/ParametersRenderer.tsx
rholinshead: +1, I do think the style is generally pretty dark and a bit hard to see things

Ankush-lastmile:
sorted a line of imports

Ankush-lastmile:
Sapling created #967 on rebase

PR #960, opened 2024-01-18
[editor] Support Updating AIConfig State when aiconfig Prop Changes

rholinshead:
# Support Updating AIConfig State when aiconfig Prop Changes

For gradio, events get dispatched and result in propagating aiconfig updates through a `value` which is passed as a prop to the top-level gradio component. From there, we will need to update the AIConfigEditor state with the updated value (e.g. for outputs after running prompts).

To support that, add a new action, `PROVIDED_AICONFIG_UPDATE` and dispatch it whenever the `aiconfig` prop to AIConfigEditor changes.

## Testing:
Updated LocalEditor to leverage state instead of callbacks for streaming:
```
          output_chunk: (data) => {
            //onStream({ type: "output_chunk", data: data as Output });
            const output = data as Output;
            setAiConfig(
              (prev) =>
                ({
                  ...prev,
                  prompts: prev!.prompts.map((prompt) => {
                    if (prompt.name === "get_activities") {
                      return { ...prompt, outputs: [output] };
                    }
                    return prompt;
                  }),
                } as AIConfig)
            );
          },
          aiconfig_chunk: (data) => {
            //onStream({ type: "aiconfig_chunk", data: data as AIConfig });
            setAiConfig(data as AIConfig);
          },
```

Then ensure streaming works:

https://github.com/lastmile-ai/aiconfig/assets/5060851/f7cfc694-2815-4901-a76c-c0ce5458c462

We'll probably need to handle runningPromptId in a similar way so that the spinners don't show indefinitely in gradio

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/960).
* __->__ #960
* #959

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
+++ b/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
@@ -750,7 +750,7 @@ export default function AIConfigEditor({
         onPromptError(message);
       }
     },
-    [runPromptCallback]
+    [logEventHandler, runPromptCallback]
   );
 
   const setNameCallback = callbacks?.setConfigName;
--- a/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
+++ b/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
@@ -131,7 +131,7 @@ export type AIConfigCallbacks = {
 type RequestCallbackError = { message?: string };
 
 export default function AIConfigEditor({
-  aiconfig: initialAIConfig,
+  aiconfig: providedAIConfig,
   callbacks,
   mode,
   readOnly = false,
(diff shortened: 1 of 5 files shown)


rholinshead:
May not be needed for sure; need to test out whether we can fake the callback req/resp format for 'run' in gradio via promises

python/src/aiconfig/editor/client/src/reducers/aiconfigReducer.ts
```diff
@@ -47,25 +47,46 @@ function setRunningPromptId(
...
-      ...statePrompt,
-      metadata: {
-        ...responsePrompt?.metadata,
-        ...statePrompt.metadata,
-      },
-    } as ClientPrompt;
+
+    if (!responsePrompt) {
+      return statePrompt;
+    }
```
rossdanlm, line 87: Nice this is cleaner!

python/src/aiconfig/editor/client/src/reducers/aiconfigReducer.ts
```diff
@@ -47,25 +47,46 @@ function setRunningPromptId(
...
+): ClientPrompt {
+  return {
+    ...externalPrompt,
+    ...statePrompt,
+    metadata: {
+      ...externalPrompt.metadata,
+      ...statePrompt.metadata,
+    },
+    outputs: externalPrompt.outputs,
+    _ui: statePrompt._ui,
```
rossdanlm, line 71: Why was this new logic needed?

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
```diff
@@ -140,9 +140,21 @@ export default function AIConfigEditor({
...
   const [aiconfigState, dispatch] = useReducer(
     aiconfigReducer,
-    aiConfigToClientConfig(initialAIConfig)
+    aiConfigToClientConfig(providedAIConfig)
   );
 
+  const [prevProvidedConfig, setPrevProvidedConfig] =
+    useState<AIConfig>(providedAIConfig);
+  // After initializing the aiconfigState, we should also support updating the
+  // state if the provided AIConfig changes externally
```
rossdanlm, line 149: Nice, I like this solutino because it doesn't change localEditor implementation either since that only sets aiconfig initial state on load api call

rossdanlm (approved):
Good change!

rholinshead:
Closing for now since Rossdan was able to get updates working in gradio without this

PR #964, opened 2024-01-18
Fix UPDATE_MODEL action to retain settings

rholinshead:
# Fix UPDATE_MODEL action to retain settings

Currently, if you update the model for a prompt all settings are wiped out on the client. On the server (and in sdk), we actually retain all of the settings when the model name changes.

This PR updates the client to match that logic, retaining the existing prompt model settings.

https://github.com/lastmile-ai/aiconfig/assets/5060851/2c0f9c51-1f30-4247-98d4-0caaa59f6de9

In the future, both server/sdk and client should leverage the prompt schemas for the models to retain only the supported settings in order to prevent errors (example client code below for reference)
```
    case "UPDATE_PROMPT_MODEL": {
      return reduceReplacePrompt(dirtyState, action.id, (prompt) => {
        // When updating the model for a prompt, retain existing settings that
        // are supported by the new model (if a schema exists for the new model)
        // or retain all existing settings if the schema is not known
        let modelSettings;
        const existingPromptModel = prompt.metadata?.model;
        if (
          action.modelName &&
          typeof existingPromptModel !== "string" &&
          existingPromptModel?.settings
        ) {
          modelSettings = existingPromptModel.settings;
          const schema = PROMPT_SCHEMAS[action.modelName];

          if (schema) {
            modelSettings = Object.fromEntries(
              Object.entries(modelSettings).filter(
                ([key]) => schema.model_settings?.properties[key] !== undefined
              )
            );
          }
        }

        return {
          ...prompt,
          metadata: {
            ...prompt.metadata,
            model: action.modelName
              ? {
                  name: action.modelName,
                  settings: modelSettings,
                }
              : undefined,
          },
        };
      });
    }
```

### DIFF
--- a/python/src/aiconfig/editor/client/src/reducers/aiconfigReducer.ts
+++ b/python/src/aiconfig/editor/client/src/reducers/aiconfigReducer.ts
@@ -167,18 +167,27 @@ export default function aiconfigReducer(
       }));
     }
     case "UPDATE_PROMPT_MODEL": {
-      return reduceReplacePrompt(dirtyState, action.id, (prompt) => ({
-        ...prompt,
-        metadata: {
-          ...prompt.metadata,
-          model: action.modelName
-            ? {
-                name: action.modelName,
-                // TODO: Consolidate settings based on schema union
-              }
-            : undefined,
-        },
-      }));
+      return reduceReplacePrompt(dirtyState, action.id, (prompt) => {
+        // TODO: Consolidate settings based on schema union, server-side
+        // For now, just keep all the settings to match the server-side implementation
+        let modelSettings;
+        const promptModel = prompt.metadata?.model;
+        if (promptModel && typeof promptModel !== "string") {
+          modelSettings = promptModel.settings;
+        }
+        return {
+          ...prompt,
+          metadata: {
+            ...prompt.metadata,
+            model: action.modelName
+              ? {
+                  name: action.modelName,
+                  settings: modelSettings,
+                }
+              : undefined,
+          },
+        };
+      });
     }
     case "UPDATE_PROMPT_MODEL_SETTINGS": {
       return reduceReplacePrompt(dirtyState, action.id, (prompt) => ({


python/src/aiconfig/editor/client/src/reducers/aiconfigReducer.ts
```diff
@@ -167,18 +167,27 @@ export default function aiconfigReducer(
...
-            ? {
-                name: action.modelName,
-                // TODO: Consolidate settings based on schema union
-              }
-            : undefined,
-        },
-      }));
+      return reduceReplacePrompt(dirtyState, action.id, (prompt) => {
+        // TODO: Consolidate settings based on schema union, server-side
+        // For now, just keep all the settings to match the server-side implementation
```
rossdanlm, line 172: I thought it was intentional to keep priority for client side over server (https://github.com/lastmile-ai/aiconfig/blob/05b10de06062876e8c91bbf60db38dcb16cbafa2/python/src/aiconfig/editor/client/src/reducers/aiconfigReducer.ts#L55-L56), but yea if we're updating the model action itself this makes sense to have server override:

python/src/aiconfig/editor/client/src/reducers/aiconfigReducer.ts
```diff
@@ -167,18 +167,27 @@ export default function aiconfigReducer(
...
+          modelSettings = promptModel.settings;
+        }
+        return {
+          ...prompt,
+          metadata: {
+            ...prompt.metadata,
+            model: action.modelName
+              ? {
+                  name: action.modelName,
+                  settings: modelSettings,
```
rossdanlm, line 185: we can just do `const` on L176 and this wil be undefined if never set right?

rossdanlm (approved):

python/src/aiconfig/editor/client/src/reducers/aiconfigReducer.ts
rholinshead: I did let with the if instead of using a const with huge ternary. I feel pretty strongly about not initializing as undefined since the above is pretty standard in js and explicitly defining it to undefined is really weird (you would in fact be defining it ... as undefined lol)

python/src/aiconfig/editor/client/src/reducers/aiconfigReducer.ts
rholinshead: Ya, for the actions where we get the updated value on the server we want to have client values take precedent if they are values that can easily change quickly (input, settings). Probably part of this TODO will be implementing the logic on server and client, we just can't do it solely on the client right now because then we'd be in a weird state where the server has really different values in some cases

PR #966, opened 2024-01-18
[editor] Default Theme ColorScheme to Match User System Settings

rholinshead:
# [editor] Default Theme ColorSchema to Match User System Settings

The theme is always light by default unless passing in a colorScheme. We need to use this mantine hook `useColorScheme` which does the media query for system (browser) settings, same as in lastmile.

## Testing:
- Ensure local editor remains the same for branding, regardless of system color theme:
Screenshot 2024-01-18 at 5 31 54 PM
Screenshot 2024-01-18 at 5 32 51 PM

- Ensure gradio properly reflects system theme now:
Screenshot 2024-01-18 at 5 31 04 PM
Screenshot 2024-01-18 at 5 32 13 PM

### DIFF
--- a/python/src/aiconfig/editor/client/src/themes/AIConfigEditorThemeProvider.tsx
+++ b/python/src/aiconfig/editor/client/src/themes/AIConfigEditorThemeProvider.tsx
@@ -3,6 +3,8 @@ import { AIConfigEditorMode } from "../shared/types";
 import { LOCAL_THEME } from "./LocalTheme";
 import { GRADIO_THEME } from "./GradioTheme";
 import ConditionalWrapper from "../components/ConditionalWrapper";
+import { useColorScheme } from "@mantine/hooks";
+import { useMemo } from "react";
 
 type Props = {
   children: React.ReactNode;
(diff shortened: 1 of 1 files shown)


python/src/aiconfig/editor/client/src/themes/AIConfigEditorThemeProvider.tsx
```diff
@@ -17,15 +19,20 @@ const THEMES = {
 };
 
 export default function AIConfigEditorThemeProvider({ children, mode }: Props) {
+  const preferredColorScheme = useColorScheme();
+  const theme = useMemo(
+    () => ({
+      colorScheme: preferredColorScheme,
+      ...THEMES[mode!],
```
rossdanlm, line 26: nit: let's do `mode.toLowerCase()` to be safe

rossdanlm (approved):
Very dark and edgy

python/src/aiconfig/editor/client/src/themes/AIConfigEditorThemeProvider.tsx
rholinshead: Mode is a union of string enums so can’t be different case unless we do a union of capitalized values as well (which is pretty gross lol)

PR #969, opened 2024-01-19
[ui] Reset outputs when first clicking run button

rossdanlm:
[ui] Reset outputs when first clicking run button

I think it's kind of weird sometimes becuase it's not obvious or you don't get clear user feedback right away that the model is running. Reset kind of tells the user it's in a waiting state.

Also weird for some models that simply return back the same output (ex: image2text captioning)

Also ran `yarn build`

## Test Plan
Check that output resets when you click run button

Before

https://github.com/lastmile-ai/aiconfig/assets/151060367/7fb50cec-5309-451a-9f95-b76bc18ffb72

After

https://github.com/lastmile-ai/aiconfig/assets/151060367/b4f8aee8-c520-41ce-85a8-eedc9fc1860b


rossdanlm:
I think this is a blocker for Gradio (don't know why but it's able to re-run again as long as the output is cleared, but if the text is still there the component doesn't update with streaming)


https://github.com/lastmile-ai/aiconfig/assets/151060367/0af62405-4453-4665-a496-0ffcc141c5fa

rholinshead (approved):
Can you check that cancelling in-progress running prompt still has the correct behaviour of resetting to prev output (not empty)? Otherwise LGTM after you resolve the merge conflicts in the build files

python/src/aiconfig/editor/client/src/reducers/aiconfigReducer.ts
```diff
@@ -221,6 +221,7 @@ export default function aiconfigReducer(
       const runningState = setRunningPromptId(dirtyState, action.promptId);
       return reduceReplacePrompt(runningState, action.promptId, (prompt) => ({
         ...prompt,
+        outputs: [],
```
rholinshead, line 224: Can you add a comment, maybe
```
// clear the outputs before running so new outputs are noticeable, even if they are identical
```

PR #987, opened 2024-01-19
3/n add to model parser registry on init

Ankush-lastmile:
3/n add to model parser registry on init

- add Gemini Parser to Model Parser Registry
- reorganize imports to avoid circular dependency imports

## Testplan

Screenshot 2024-01-25 at 1 06 03 AM

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/987).
* #1023
* __->__ #987
* #986
* #973

### DIFF
--- a/python/src/aiconfig/Config.py
+++ b/python/src/aiconfig/Config.py
@@ -9,6 +9,7 @@ from aiconfig.default_parsers.anyscale_endpoint import (
     DefaultAnyscaleEndpointParser,
 )
 from aiconfig.default_parsers.openai import DefaultOpenAIParser
+from aiconfig.default_parsers.gemini import GeminiModelParser
 from aiconfig.default_parsers.palm import PaLMChatParser, PaLMTextParser
 from aiconfig.model_parser import InferenceOptions, ModelParser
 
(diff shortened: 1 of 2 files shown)


python/src/aiconfig/Config.py
```diff
@@ -51,12 +52,12 @@
...
 )
-ModelParserRegistry.register_model_parser(PaLMChatParser())
-ModelParserRegistry.register_model_parser(PaLMTextParser())
+ModelParserRegistry.register_model_parser(GeminiModelParser("gemini-pro"), ["gemini-pro"])
 ModelParserRegistry.register_model_parser(HuggingFaceTextGenerationParser())
 for model in gpt_models_extra:
     ModelParserRegistry.register_model_parser(DefaultOpenAIParser(model))
-
+ModelParserRegistry.register_model_parser(PaLMChatParser())
+ModelParserRegistry.register_model_parser(PaLMTextParser())
```
rholinshead, line 60: Maybe we can axe PaLM registry by default soon?

rholinshead (approved):

Ankush-lastmile:
> Maybe we can axe PaLM registry by default soon?

Yea, we need to figure out what our gameplan is for having a model as core or not. 

It does use the api dependency as Gemini though so something to think about.

PR #993, opened 2024-01-23
[hf inference] Add HuggingFaceTextSummarizationParser

rholinshead:
[hf inference] Add HuggingFaceTextSummarizationParser

### DIFF
--- a/extensions/HuggingFace/python/README.md
+++ b/extensions/HuggingFace/python/README.md
@@ -5,13 +5,13 @@ This extension contains AIConfig model parsers with two main subfolders:
 
 ## Usage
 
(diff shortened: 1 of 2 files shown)


Ankush-lastmile (approved):
lgtm

rholinshead:
> > Added support for user-provided api_token as part of InferenceOptions
> 
> Sorry I remember this being mentioned and something having to do with editor client passing api_token. Could you explain why this change is needed again?

In Gradio, the server for the space will initialize parser instances in the registry and those parsers will be shared by anyone using the space. But, we want users to be able to provide their own HF api token to use for the actual inference requests, so it needs to be done at the run layer

PR #999, opened 2024-01-23
[sdk][python] Refactor OpenAI model paser to use client abstraction

Ankush-lastmile:
[sdk][python] Refactor OpenAI model paser to use client abstraction

- Refactor OpenAI Model parser to use an OpenAI Client construct instead of the weird default base client in the OpenAI Library
- - This change will make it easier to reuse this model parser with other model providers that are compatible with the OpenAI Library/client (such as OpenAI azure)
- Marked 2 tests as xfail because they currently break with this change and need to be updated to not break with client changes

TODO: investigate how to update mock tests to handle a client.

## Testplan

1. Run streaming in local editor with gpt-4

https://github.com/lastmile-ai/aiconfig/assets/141073967/2133a8cd-1c6f-4816-9286-1556272363a6

2. Validate Function Calling Notebook
Screenshot 2024-01-24 at 3 59 22 PM

### DIFF
--- a/python/src/aiconfig/default_parsers/openai.py
+++ b/python/src/aiconfig/default_parsers/openai.py
@@ -35,6 +35,8 @@ if TYPE_CHECKING:
 class OpenAIInference(ParameterizedModelParser):
     def __init__(self):
         super().__init__()
+        # Will be set in the run method. This is to avoid having to set the api key in the constructor
+        self.client = None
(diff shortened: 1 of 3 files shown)


saqadri (approved):

Ankush-lastmile:
created #1022

Issue #1000, opened 2024-01-23
Need to pass in token for llama guard cookbook


rossdanlm:
Getting this error:
```
OSError: You are trying to access a gated repo.
Make sure to request access at https://huggingface.co/meta-llama/LlamaGuard-7b and pass a token having permission to this repo either by logging in with `huggingface-cli login` or by passing `token=<your_token>`.
```

I think this might be related to https://github.com/lastmile-ai/aiconfig/pull/772

PR #1001, opened 2024-01-23
Some small fixes to the llama_guard_cookbook.ipynb

rossdanlm:
Some small fixes to the llama_guard_cookbook.ipynb

Ran into some issues that tracked in https://github.com/lastmile-ai/aiconfig/issues/1000

Made the setting OPENAI_API_KEY easier (can do it locally not just through `google-colab`)
Screenshot 2024-01-23 at 16 23 02


Ankush-lastmile (approved):
Screenshot 2024-01-23 at 5 52 42 PM

lgtm, though you can use try catch to handle which platform is executing (colab vs local)

PR #1003, opened 2024-01-23
Set huggingface-hub version >= 0.20.3

rholinshead:
Set huggingface-hub version >= 0.20.3


Ankush-lastmile (approved):
lgtm - curious what version you had installed locally for this to not work

PR #1004, opened 2024-01-23
[hf inference] Add HuggingFaceTextTranslationParser

rholinshead:
[hf inference] Add HuggingFaceTextTranslationParser

### DIFF
--- a/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/remote_inference_client/text_translation.py
+++ b/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/remote_inference_client/text_translation.py
@@ -1,0 +1,291 @@
+import copy
+import json
+from typing import TYPE_CHECKING, Any, List, Optional
+
+# HuggingFace API imports
+from huggingface_hub import InferenceClient
+
+from aiconfig import CallbackEvent
+from aiconfig.default_parsers.parameterized_model_parser import (
+    ParameterizedModelParser,
+)
+from aiconfig.model_parser import InferenceOptions
+from aiconfig.schema import (
+    ExecuteResult,
+    Output,
+    Prompt,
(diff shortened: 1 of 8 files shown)


extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/remote_inference_client/text_translation.py
```diff
@@ -0,0 +1,291 @@
...
+                "HUGGING_FACE_API_TOKEN", required=False
+            ).unwrap()
+
+        self.client = InferenceClient(model_id, token=token)
+
+    def id(self) -> str:
+        """
+        Returns an identifier for the Model Parser
+        """
+        return "HuggingFaceTextTranslationParser"
```
Ankush-lastmile, line 107: Should we include "RemoteInference" in the parser IDs for remote inference models? I know we haven't figured out model vs model provider yet but in the meantime I think this would be helpful to distinguish in the UI

rholinshead: Hmm, I'm not sure -- we already have a bit of a distinction with the 'Transformer' and 'Diffuser'. If we want to use RemoteInference would we also drop the 'Parser' for consistency with the local ones?

Ankush-lastmile (approved):
lgtm, seems like pretty much a copy paste of the other one

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/remote_inference_client/text_translation.py
Ankush-lastmile: I think that would be fine. Just felt like it would be easier to navigate if the naming were more explicit

PR #1005, opened 2024-01-23
HuggingFaceTextTranslationParserPromptSchema

rholinshead:
HuggingFaceTextTranslationParserPromptSchema


Ankush-lastmile:
> Note that the client doesn't actually support the other parameters defined in the api docs here:

What happens when you set them? does it just raise/throw exception?

Ankush-lastmile (approved):
lgtm

rholinshead:
> > Note that the client doesn't actually support the other parameters defined in the api docs here:
> 
> What happens when you set them? does it just raise/throw exception?

Presumably, since they would be unsupported arguments to the function call itself

PR #997, opened 2024-01-23
HuggingFaceTextSummarizationParserPromptSchema

rholinshead:
HuggingFaceTextSummarizationParserPromptSchema

# HuggingFaceTextSummarizationParserPromptSchema

Define the prompt schema using https://huggingface.co/docs/api-inference/detailed_parameters#summarization-task

Screenshot 2024-01-23 at 2 58 52 PM

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/997).
* #1016
* #1015
* #1012
* #1010
* #1009
* #1006
* #1005
* #1004
* #1003
* __->__ #997
* #993
* #992

PR #1006, opened 2024-01-23
[hf inference] Update HuggingFaceTextGenerationParser to support api_token in InferenceOptions

rholinshead:
[hf inference] Update HuggingFaceTextGenerationParser to support api_token in InferenceOptions

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1006).
* __->__ #1006
* #1005
* #1004
* #1003
* #997
* #993
* #992

### DIFF
--- a/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/remote_inference_client/text_generation.py
+++ b/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/remote_inference_client/text_generation.py
@@ -285,13 +285,19 @@ class HuggingFaceTextGenerationParser(ParameterizedModelParser):
         Returns:
             InferenceResponse: The response from the model.
         """
+        sanitized_options = copy.deepcopy(options)
+        run_override_api_token = getattr(sanitized_options, "api_token", None)
+        # Redact api token from logs if it exists
+        if run_override_api_token:
+            setattr(sanitized_options, "api_token", "hf_********")
+
         await aiconfig.callback_manager.run_callbacks(
             CallbackEvent(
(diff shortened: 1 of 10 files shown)


extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/remote_inference_client/text_generation.py
```diff
@@ -301,14 +307,24 @@ async def run_inference(
 
         # if stream enabled in runtime options and config, then stream. Otherwise don't stream.
         stream = True  # Default value
-        if options is not None and options.stream is not None:
-            stream = options.stream
+        if (
+            sanitized_options is not None
+            and sanitized_options.stream is not None
+        ):
+            stream = sanitized_options.stream
```
Ankush-lastmile, line 314: does this need to use santized options? curious to hear the reasoning behind using sanitized here vs keeping this unchanged

Ankush-lastmile (approved):
lgtm. Ideally we should figure out how to do this better long term but can't think of a better way to handle this api key right now

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/remote_inference_client/text_generation.py
rholinshead: Technically no, but it's possible someone might add code in the future and do something with the options (which could involve logging), so IMO it's safer to consistently use the sanitized_options throughout

PR #1009, opened 2024-01-24
[hf inference] Add HuggingFaceText2ImageRemoteInference

rholinshead:
[hf inference] Add HuggingFaceText2ImageRemoteInference

### DIFF
--- a/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/remote_inference_client/text_2_image.py
+++ b/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/remote_inference_client/text_2_image.py
@@ -1,0 +1,309 @@
+import base64
+import copy
+import io
+import json
+from typing import TYPE_CHECKING, Any, List, Optional
+
+# HuggingFace API imports
+from huggingface_hub import InferenceClient
+
+from aiconfig import CallbackEvent
+from aiconfig.default_parsers.parameterized_model_parser import (
+    ParameterizedModelParser,
+)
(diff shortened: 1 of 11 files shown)


extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/remote_inference_client/text_2_image.py
```diff
@@ -0,0 +1,309 @@
...
+        completion_data = await self.deserialize(prompt, aiconfig, parameters)
+
+        # If api token is provided in the options, use it for the client
+        client = self.client
+        if run_override_api_token:
+            client = InferenceClient(
+                self.client.model, token=run_override_api_token
+            )
+
+        response = client.text_to_image(**completion_data)
```
Ankush-lastmile, line 273: response is always going to be `PIL.Image` so I wonder why the method definition has the output type as a forward reference (type in quotes)
https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/inference/_client.py#L100

Ankush-lastmile:
The interesting red dog does not have a green hat

Ankush-lastmile (approved):
lgtm

extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/remote_inference_client/text_2_image.py
rholinshead: Ya, not really sure :s

PR #1017, opened 2024-01-24
vsCode multiTheme styles

zakariassoul:
vsCode multiTheme styles

Added my vsCode theme changes along with some other subtle changes to improve the UX (e.g. tooltips)

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/ConfigNameDescription.tsx
+++ b/python/src/aiconfig/editor/client/src/components/ConfigNameDescription.tsx
@@ -32,12 +32,9 @@ const useStyles = createStyles((theme) => ({
   },
   hoverContainer: {
     "&:hover": {
-      backgroundColor:
-        theme.colorScheme === "dark"
-          ? "rgba(255, 255, 255, 0.1)"
-          : theme.colors.gray[1],
+      backgroundColor: "rgba(226, 232, 255, 0.1)",
     },
-    borderRadius: theme.radius.sm,
+    borderRadius: "0px",
     width: "-webkit-fill-available",
   },
 }));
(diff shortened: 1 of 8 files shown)


python/src/aiconfig/editor/client/src/components/prompt/PromptName.tsx
```diff
@@ -21,20 +21,22 @@ export default memo(function PromptName({ promptId, name, onUpdate }: Props) {
...
-      onChange={onChange}
-      error={
-        getState().prompts.some(
-          (p) => p.name === nameInput && p._ui.id !== promptId
-        )
-          ? "Name already exists"
-          : null
-      }
-    />
+    <Tooltip label="Prompt Name" position="right">
```
rholinshead, line 24: Just confirming that we want this tooltip to show up in gradio and local editor as well? Same w/ the one in ModelSelector above

python/src/aiconfig/editor/client/src/themes/LocalTheme.ts
```diff
@@ -74,6 +74,7 @@ export const LOCAL_THEME: MantineThemeOverride = {
       borderLeft: "none",
       borderTopRightRadius: "4px",
       borderBottomRightRadius: "4px",
+      minWidth: "32px",
```
rholinshead, line 77: We should revisit later and move all the common styles to the component itself & just have overrides in the themes, otherwise if we add additional customization for theme props later we will force users of the component to need to duplicate a lot of the styles as well just to have the basic styles working (e.g. right now if there is not theme specified, the prompt cells are super narrow instead of expanding the full width because 'flex: 1' is defined in the theme styles instead of on the component

rholinshead (approved):
LGTM, just one question and minor comment for later

python/src/aiconfig/editor/client/src/components/ConfigNameDescription.tsx
```diff
@@ -32,12 +32,9 @@ const useStyles = createStyles((theme) => ({
   },
   hoverContainer: {
     "&:hover": {
-      backgroundColor:
-        theme.colorScheme === "dark"
-          ? "rgba(255, 255, 255, 0.1)"
-          : theme.colors.gray[1],
+      backgroundColor: "rgba(226, 232, 255, 0.1)",
```
rholinshead, line 35: Shouldn't this take into account the theme?

PR #1034, opened 2024-01-25
RFC: [python][sdk] Azure OpenAI Model Parser

Ankush-lastmile:
RFC: [python][sdk] Azure OpenAI Model Parser

## Testplan
Screenshot 2024-01-25 at 6 01 50 PM
Screenshot 2024-01-25 at 6 02 50 PM

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1034).
* __->__ #1034
* #1033

### DIFF
--- a/python/src/aiconfig/default_parsers/azure.py
+++ b/python/src/aiconfig/default_parsers/azure.py
@@ -1,0 +1,40 @@
+from aiconfig.default_parsers.openai import DefaultOpenAIParser
+from aiconfig.util.config_utils import get_api_key_from_environment
+
+from openai import AzureOpenAI
+
+
+class AzureOpenAIParser(DefaultOpenAIParser):
+    def __init__(self, deployment: str):
+        super().__init__(model_id = deployment)
+
+        # Deployment is a custom name that user defines in Azure portal. 
+        self.deployment = deployment 
+
+    async def deserialize(self, *args, **kwargs):
+        # Logic doesn't depend on input, pass it forward to the openai model parser deserialize 
+        openai_deserialized_params = await super().deserialize(*args, **kwargs)
+
+        # Should this just be the "model name" from config? 
+        # If so, Then the "model name" will have to be deployment name and be consistent 
+        # Which would mean we don't need to have this custom deserialization
+        openai_deserialized_params["model"] = self.deployment 
+        return openai_deserialized_params
+
+    def initialize_openai_client(self) -> None:
+        ## The Azure Client itself will check and retrieve these from environment 
+        # if its not explicitly passed in & will raise otherwise. Technically not needed
+        azure_openai_key = get_api_key_from_environment(
+            "AZURE_OPENAI_KEY"
+        ).unwrap()
+
+        azure_endpoint = get_api_key_from_environment(
+            "AZURE_OPENAI_ENDPOINT"
+        ).unwrap()
+
+        # Initialize Azure Client.
+        self.client = AzureOpenAI(
+            api_key=azure_openai_key,
+            api_version="2023-12-01-preview", # This will have to either be set here or passed in. Otherwise, azure client will check environment variables and throw error if not set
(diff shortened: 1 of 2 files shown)


python/src/aiconfig/default_parsers/azure.py
```diff
@@ -0,0 +1,40 @@
+from aiconfig.default_parsers.openai import DefaultOpenAIParser
+from aiconfig.util.config_utils import get_api_key_from_environment
+
+from openai import AzureOpenAI
+
+
+class AzureParser(DefaultOpenAIParser):
```
saqadri, line 7: AzureOpenAIParser

python/src/aiconfig/default_parsers/azure.py
```diff
@@ -0,0 +1,40 @@
...
+        return openai_deserialized_params
+
+    def initialize_openai_client(self) -> None:
+        ## The Azure Client itself will check and retrieve these from environment 
+        # if its not explicitly passed in & will raise otherwise. Technically not needed
+        azure_openai_key = get_api_key_from_environment(
+            "AZURE_OPENAI_KEY"
+        ).unwrap()
+
+        azure_endpoint = get_api_key_from_environment(
```
saqadri, line 31: Is there a default in case this isn't set? Or does this have to be set?

python/src/aiconfig/default_parsers/azure.py
```diff
@@ -0,0 +1,40 @@
...
+        self.deployment = deployment 
+
+    async def deserialize(self, *args, **kwargs):
+        # Logic doesn't depend on input, pass it forward to the openai model parser deserialize 
+        openai_deserialized_params = await super().deserialize(*args, **kwargs)
+
+        # Should this just be the "model name" from config? 
+        # If so, Then the "model name" will have to be deployment name and be consistent 
+        # Which would mean we don't need to have this custom deserialization
+        openai_deserialized_params["model"] = self.deployment
```
saqadri, line 21: Hmm this is tricky. So essentially in azure portal I need to deploy gpt4 and gpt3.5 separately, right? So I'd have to register those model parsers with different deployments and model id keys, right?

python/src/aiconfig/default_parsers/azure.py
```diff
@@ -0,0 +1,40 @@
+from aiconfig.default_parsers.openai import DefaultOpenAIParser
+from aiconfig.util.config_utils import get_api_key_from_environment
+
+from openai import AzureOpenAI
+
+
+class AzureParser(DefaultOpenAIParser):
+    def __init__(self, deployment: str):
```
saqadri, line 8: Alternatively this can take a mapping of deployment and model name, but I think this is better the way you've done it

saqadri (approved):
Looks good to me as a starting point. Can also try updating the anyscale endpoint model parser to use this structure

Some comments

python/src/aiconfig/default_parsers/azure.py
Ankush-lastmile: yep exactly.

python/src/aiconfig/default_parsers/azure.py
Ankush-lastmile: Nope there's no defaults. It has to be set otherwise it will throw an exception if one of the 3 required ones are missing

Ankush-lastmile: I think our error messages are a lot easier to understand but its the same error being thrown anyway

python/src/aiconfig/default_parsers/azure.py
Ankush-lastmile: updated

Ankush-lastmile:
merging as is, can iterate on top and remove comments as neccessary.

PR #1031, opened 2024-01-25
Share button [1/n]: Create button UI

rossdanlm:
Share button [1/n]: Create button UI

Doesn't do anything when you click it now, but just setting up the UI and tooltip. Will add upload feature next PR

## Test Plan
Run these commands from the `aiconfig` dir:
```bash
aiconfig_path  =./cookbooks/Gradio/huggingface.aiconfig.json
parsers_path=./cookbooks/Gradio/hf_model_parsers.py
aiconfig edit --aiconfig-path=$aiconfig_path --server-port=8080 --server-mode=debug_servers --parsers-module-path=$parsers_path
```

Screenshot 2024-01-25 at 15 58 42

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1031).
* #1037
* #1035
* #1032
* __->__ #1031

PR #1035, opened 2024-01-25
Share button [2/n]: Move uploadFile function into a utils file

rossdanlm:
Share button [2/n]: Move uploadFile function into a utils file

Going to re-use this next PR for uploading AIConfig files to S3

## Test Plan
No functional changes, make sure that uploading files still works for ASR and Image 2 text

```
```

```
aiconfig_path  =./cookbooks/Gradio/huggingface.aiconfig.json
```

https://github.com/lastmile-ai/aiconfig/assets/151060367/7a5141f1-bb1b-496c-b039-724a8eeb11dd

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1035).
* #1037
* __->__ #1035
* #1032
* #1031


rholinshead (approved):
Accepting to unblock but I don't think this is really needed, for 2 reasons:
1. We are intending to make the upload implementation for attachment files a callback passed from the root component (so companies can upload where they want, instead of to our s3 bucket)
2. For gradio share implementation, I'm pretty sure we should upload from the gradio backend (since that's where the source of truth for the config is, plus the python knows how to serialize the AIConfig runtime. Client only has an AIConfig object for the UI purposes). Also, we'll need to pass 'source' and 'sourceUrl' from gradio as well

rossdanlm:
> 1. We are intending to make the upload implementation for attachment files a callback passed from the root component (so companies can upload where they want, instead of to our s3 bucket)

Ok so even for `AttachmentUploader`, we're going to want to refactor that too for companies to upload to their own S3 bucket right? I think it's a bit unclear right now since there are two parts to sharing:

1. Upload file to S3 (which companies can change to their own)
2. Use the upload URI as the source for our `api/aiconfig/upload` post request. Are you saying that companies should also be able to override this too if they want? My feeling is that the answer should be yes

> we should upload from the gradio backend (since that's where the source of truth for the config is,

I thought client would be source of truth because it can be in the isDirty state and we'd want to export that right? Or are you saying that if `isDirty` is true, we should NOT share with those changes until it's been saved?

>  plus the python knows how to serialize the AIConfig runtime. Client only has an AIConfig object for the UI purposes)

Yea I made a helper function in https://github.com/lastmile-ai/aiconfig/pull/1037 to convert from ClientAIConfig --> AIConfig for this purpose, but sure we can define this ourselves

rholinshead:
> Ok so even for AttachmentUploader, we're going to want to refactor that too for companies to upload to their own S3 bucket right? 
Correct. Uploading attachment should be pulled out to its own handler and not necessarily even associated with s3


> 1. Upload file to S3 (which companies can change to their own)
> 2. Use the upload URI as the source for our api/aiconfig/upload post request. Are you saying that companies should also be able to override this too if they want? My feeling is that the answer should be yes

Yes. These shouldn't actually be two separate things -- it should just be one callback that gives you the share link for the config and that can be implemented in however many steps they desire

> I thought client would be source of truth because it can be in the isDirty state and we'd want to export that right? Or are you saying that if isDirty is true, we should NOT share with those changes until it's been saved?

isDirty means changes have not been saved to disk. The server should have the changes (within O(request time) seconds) since each action pushes to the server-side config state.

> Yea I made a helper function in https://github.com/lastmile-ai/aiconfig/pull/1037 to convert from ClientAIConfig --> AIConfig for this purpose, but sure we can define this ourselves

Ya, the client shouldn't need to know how to serialize the config. We could technically add any extra logic to the sdk for serialization (e.g. obfuscating PII) which the client would then also need to implement if we are saving from the client

PR #1038, opened 2024-01-26
Share Button [3/n]: Add share callback to callbacks prop

rossdanlm:
Share Button [3/n]: Add share callback to callbacks prop

Adding a share callback that takes the form of (AIConfig) -> string, where the string is a URL which the client will then use to display to the user.

## Test Plan
No functional changes yet


python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
```diff
@@ -148,6 +149,27 @@ export default function AIConfigEditor({
...
 
+  const shareCallback = callbacks?.share;
+  const onShare = useCallback(async () => {
+    if (!shareCallback) {
+      return;
+    }
+    try {
+      // TODO: While uploading, show a loader state for share button
+      const { share_url: shareUrl } = await shareCallback();
+      // clientConfigToAIConfig(stateRef.current)
```
rossdanlm, line 160: We will want to do this if we want to share the config in `isDirty` state, rather than server source of truth. I think for simplicitly, better just not to include for now since Gradio will always have server updated

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
```diff
@@ -148,6 +149,27 @@ export default function AIConfigEditor({
...
+      // TODO: display the shareUrl in a dialog
+      console.log("Share URL: ", shareUrl);
```
Ankush-lastmile, line 162: can remove?

Ankush-lastmile (approved):
lgtm

PR #1043, opened 2024-01-26
wip default model for hf remote inference parsers

Ankush-lastmile:
rfc wip default model for hf remote inference parsers

- added huggingface task to each remote inference model parser as a static variable
- defined a method that will add defeault model to completion params if "model" is not set.
- - Note: we need to think about how to set the model name if its specified in the aiconfig.
- - we should have something similar to the [local inference `get_model`](https://github.com/lastmile-ai/aiconfig/blob/main/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/local_inference/util.py#L11) but right now as a workaround we have the model parser id's set to a task in gradio.
- This diff is to improve the Gradio workbook ux

## Testplan:

Verified that the 7 tasks we are supporting have default models:

Screenshot 2024-01-26 at 4 16 10 PM


Ankush-lastmile:
this way doesn't update client side in the local editor.

PR #1046, opened 2024-01-26
Set Model Selector Label to 'HuggingFace Task' for 'gradio' Mode

rholinshead:
Set Model Selector Label to 'HuggingFace Task' for 'gradio' Mode

# Set Model Selector Label to 'HuggingFace Task' for 'gradio' Mode

Per dogfood feedback, it's confusing for the model selector label and placeholder to say 'Model' while the parser is registered based on task name. As a temporary improvement, this PR uses 'HuggingFace Task" terminology instead of model. Long-term, once the model -> model parser mapping UX is available, we should undo this and have the model copy back in the selector.

## Updated for 'gradio' Mode:
Screenshot 2024-01-26 at 4 35 27 PM

## Local / non-gradio is unchanged:
Screenshot 2024-01-26 at 4 35 09 PM

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1046).
* #1047
* __->__ #1046

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
+++ b/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
@@ -862,9 +862,10 @@ export default function AIConfigEditor({
     () => ({
       getState,
       logEventHandler,
+      mode,
       readOnly,
     }),
-    [getState, logEventHandler, readOnly]
+    [getState, logEventHandler, mode, readOnly]
   );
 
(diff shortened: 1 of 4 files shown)


Ankush-lastmile (approved):
lgtm. 

We have a couple PR's that are "temporary fixes" for model vs model provider, we should probably consolidate all of these somewhere, which I think it will help us figure out how to solve this broad problem. I'll make a google doc to start that

PR #1047, opened 2024-01-26
Default to 400 max_new_tokens for HF Text Generation Remote Inference Parser

rholinshead:
# Default to 400 max_new_tokens for HF Text Generation Remote Inference Parser

Per gradio dogfooding. The default value for max_new_tokens is set to 20 by the hugging face text_generation API. This isn't a good default limit for use in gradio workbooks, so set to a more reasonable 400. Note that we set the default in 2 places here:
1. In the PromptSchema, default is set to 400 to show in the UI what the default is. This default, if untouched, *does not* propagate to the parser since otherwise any default setting value in the schemas would be persisted to the config
2. In the parser implementation itself -- this is where the default is set/used for the run call. In the future once we have prompt schemas moved to the parser layer we can have a common process of grabbing all defaults from the schemas to use

## Testing:
- Ensure default is now 400 (and not persisted to config if not changed)
Screenshot 2024-01-26 at 5 09 14 PM

- Ensure changing the setting works as expected (and persists to config)

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1047).
* __->__ #1047
* #1046 

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1047).
* __->__ #1047
* #1046

### DIFF
--- a/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/remote_inference_client/text_generation.py
+++ b/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/remote_inference_client/text_generation.py
@@ -56,6 +56,13 @@ def refine_completion_params(model_settings: dict[Any, Any]) -> dict[str, Any]:
         if key.lower() in supported_keys:
             completion_data[key.lower()] = model_settings[key]
 
+    # Default max_new_tokens is set to 20 at the api layer if not specified. We
+    # increase that to 400 for an improved user experience.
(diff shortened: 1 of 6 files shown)


Ankush-lastmile (approved):
lgtm

PR #1048, opened 2024-01-26
[editor] Hardcode default models for HuggingFace Remote Inference schemas

Ankush-lastmile:
Hardcode default models for HuggingFace Remote Inference schemas

HuggingFace Remote Inference Model parsers utilized the Remote Inference endpoint.

When a model is not provided, it uses a default model. For now, hardcode the default model into the prompt schema, such that the user sees the "default model". This will help improve user experience

Found the models by calling the Inference Endpoint API:
Screenshot 2024-01-26 at 4 16 10 PM

There probably is a way to do something similar in Typescript and such implement it in the schema.

## Testplan:

Verify selecting the 7 tasks populates the model field

https://github.com/lastmile-ai/aiconfig/assets/141073967/69ed36f5-2259-41b9-a535-d0819bb91997

### DIFF
--- a/python/src/aiconfig/editor/client/src/shared/prompt_schemas/HuggingFaceText2SpeechRemoteInferencePromptSchema.ts
+++ b/python/src/aiconfig/editor/client/src/shared/prompt_schemas/HuggingFaceText2SpeechRemoteInferencePromptSchema.ts
@@ -13,6 +13,7 @@ export const HuggingFaceText2SpeechRemoteInferencePromptSchema: PromptSchema = {
         type: "string",
         description: `Hugging Face model to use. Can be a model ID hosted on the Hugging Face Hub or a URL 
         to a deployed Inference Endpoint`,
+        default: "suno/bark",
       },
     },
   },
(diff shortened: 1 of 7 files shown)


rholinshead (approved):
Nice, thanks!

python/src/aiconfig/editor/client/src/shared/prompt_schemas/HuggingFaceText2SpeechRemoteInferencePromptSchema.ts
```diff
@@ -13,6 +13,7 @@ export const HuggingFaceText2SpeechRemoteInferencePromptSchema: PromptSchema = {
         type: "string",
         description: `Hugging Face model to use. Can be a model ID hosted on the Hugging Face Hub or a URL 
         to a deployed Inference Endpoint`,
+        default: "suno/bark",
```
rholinshead, line 16: Is this the one that fails by default? Maybe we should choose a different one and default it in the parser?

Ankush-lastmile: Screenshot 2024-01-26 at 5 36 12 PM

yep suno/bark requires a pro subscription. We could choose a different one but a user could just change the model

PR #1049, opened 2024-01-26
Share Button [8/n] Show modal on UI click

rossdanlm:
Share Button [8/n] Show modal on UI click

Showing the modal after button  returns a URL. I'm pretty annoyed, wasn't able to get it:

1. Show up right beside the Share button
2. Have it look beautiful in the modal
3. Have the copy button be displayed on the right side (I was able to use position absolute, but now the label does not work and it's also not aligned with the padding for the X close button)

## Test plan

https://github.com/lastmile-ai/aiconfig/assets/151060367/9fba2b38-0602-4a92-b77a-2d8624d8fe34

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1049).
* __->__ #1049
* #1044

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/global/ShareButton.tsx
+++ b/python/src/aiconfig/editor/client/src/components/global/ShareButton.tsx
@@ -1,12 +1,17 @@
-import { Button, Tooltip } from "@mantine/core";
+import { Button, Container, Flex, Modal, Text, Tooltip } from "@mantine/core";
 import { memo, useState } from "react";
+import { useDisclosure } from "@mantine/hooks";
+import CopyButton from "../CopyButton";
 
 type Props = {
   onShare: () => Promise<string | void>;
 };
 
 export default memo(function ShareButton({ onShare }: Props) {
+  const [isModalOpened, { open: openModal, close: closeModal }] =
+    useDisclosure(false);
   const [isLoading, setIsLoading] = useState<boolean>(false);
+  const [shareUrl, setShareUrl] = useState<string>("");
 
   const onClick = async () => {
     if (isLoading) {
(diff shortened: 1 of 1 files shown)


python/src/aiconfig/editor/client/src/components/global/ShareButton.tsx
```diff
@@ -1,12 +1,43 @@
...
 
 type Props = {
   onShare: () => Promise<string | void>;
 };
 
+const useStyles = createStyles((theme) => ({
+  copyIcon: {
+    alignItems: "flex-end",
+    align: "right",
+    justifyContent: "flex-end",
```
rholinshead, line 14: Instead of doing this, you could just wrap the button with a <Flex justify="flex-end">. I assume most of these are just from trying which one works.

python/src/aiconfig/editor/client/src/components/global/ShareButton.tsx
```diff
@@ -20,29 +51,39 @@ export default memo(function ShareButton({ onShare }: Props) {
...
+        Share
+      </Button>
     </Tooltip>
   );
+
+  return (
+    <>
+      <Modal opened={isModalOpened} onClose={closeModal} title="AIConfig URL">
+        <div className={classes.urlContainer}>
+          {shareUrl}
```
rholinshead, line 80: Should wrap this in <Text> so that theme colours are applied to the text

python/src/aiconfig/editor/client/src/components/global/ShareButton.tsx
```diff
@@ -1,12 +1,43 @@
...
 
+const useStyles = createStyles((theme) => ({
+  copyIcon: {
+    alignItems: "flex-end",
+    align: "right",
+    justifyContent: "flex-end",
+    position: "absolute",
+    right: 0,
+  },
+  urlContainer: {
```
rholinshead, line 18: You could also try using <Container> instead of this wrapping div since it will have padding already

python/src/aiconfig/editor/client/src/components/global/ShareButton.tsx
```diff
@@ -1,12 +1,43 @@
...
+    align: "center",
+    [theme.fn.smallerThan("sm")]: {
+      marginLeft: "0",
+      display: "block",
+      position: "static",
+      bottom: -10,
+      left: 0,
+      height: 28,
+      margin: "10px 0",
+    },
```
rholinshead, line 31: What does this look like on small screens?

rholinshead (approved):

python/src/aiconfig/editor/client/src/components/global/ShareButton.tsx
rossdanlm: It doesn't actually apply to us since we can never go smaller than certain width for some reason:


https://github.com/lastmile-ai/aiconfig/assets/151060367/3401fb87-d224-498b-a296-b315f8ba1e4e

PR #1051, opened 2024-01-28
Dynamically Set Settings/Parameters Height to Match Prompt Input/Output Height

rholinshead:
Dynamically Set Settings/Parameters Height to Match Prompt Input/Output Height

# Dynamically Set Settings/Parameters Height to Match Prompt Input/Output Height

This PR implements dynamic max height for the model settings and parameters contents to be bounded by the height of the prompt input/output cell container. I wasn't able to get anything working with pure css so this implementation requires some javascript -- not ideal, due to potential perf hit, but it works.

The main idea is that:
1. In PromptContainer, use a ref for the element containing the input/outputs and calculate its height whenever input or outputs change (using debounce so we don't do this constantly). Set this height in an exported const map, mapped under the prompt id
2. In the PromptActionBar, set the max height of the contents to one of:
a) the calculated height minus height of surrounding action bar contents (84px, found with some trial/error logging)
b) 300px if a) is less than 300px since we want the minimum height to be reasonable enough to show sufficient settings

## Testing:
Min height of 300px for empty new prompt
Screenshot 2024-01-28 at 3 46 35 PM

Dynamic resizing:

https://github.com/lastmile-ai/aiconfig/assets/5060851/591fe137-edfb-4c7b-9b6d-be6d282efb15

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1051).
* #1084
* __->__ #1051

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/GlobalParametersContainer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/GlobalParametersContainer.tsx
@@ -37,6 +37,7 @@ export default memo(function GlobalParametersContainer({
               <ParametersRenderer
                 initialValue={initialValue}
                 onUpdateParameters={onUpdateParameters}
+                maxHeight="300px"
               />
             )}
           </Accordion.Panel>
--- a/python/src/aiconfig/editor/client/src/components/ParametersRenderer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/ParametersRenderer.tsx
@@ -10,7 +10,7 @@ import {
 } from "@mantine/core";
 import { IconTrash, IconPlus } from "@tabler/icons-react";
 import { debounce, uniqueId } from "lodash";
-import { memo, useCallback, useContext, useMemo, useState } from 'react';
+import { memo, useCallback, useContext, useMemo, useState } from "react";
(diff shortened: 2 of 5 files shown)


python/src/aiconfig/editor/client/src/components/GlobalParametersContainer.tsx
```diff
@@ -37,6 +37,7 @@ export default memo(function GlobalParametersContainer({
               <ParametersRenderer
                 initialValue={initialValue}
                 onUpdateParameters={onUpdateParameters}
+                maxHeight="300px"
```
rholinshead, line 40: Moving this here so we can remove the hard-coded default from ParametersRenderer

saqadri (approved):
Ship it! Looks great. Not a huge fan of the add parameter button but that's a minor nit

rholinshead:
> Ship it! Looks great. Not a huge fan of the add parameter button but that's a minor nit

Ah, yep. It's an existing thing, but fixed to a nicer UI in https://github.com/lastmile-ai/aiconfig/pull/1084

PR #1057, opened 2024-01-29
[1/n] Add load_json and load_yaml methods to AIConfigRuntime

saqadri:
[1/n] Add load_json and load_yaml methods to AIConfigRuntime

Allow loading an AIConfigRuntime from a string. There's already a `loadJSON` method in the TS SDK.

This is needed for VSCode extension (top of stack), which gets the state of the document from the editor (and which can contain unsaved changes that aren't present on the filesystem).

Test Plan:

Tested out getting started ipynb with:

load JSON:

```
config = AIConfigRuntime.load('travel.aiconfig.json')
```

load YAML:
```
config.save('travel.aiconfig.yaml', mode="yaml")
yaml_str = ""
with open('travel.aiconfig.yaml') as file:
    yaml_str = file.read()

config = AIConfigRuntime.load_yaml(yaml_str)
```

Made sure config runtime is valid and tried `config.run`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1057).
* #1076
* #1060
* #1059
* #1058
* __->__ #1057

PR #583, opened 2023-12-22
Add YAML serialization to aiconfig

saqadri:
Add YAML serialization to aiconfig

YAML is a superset of JSON, and allows for much better readability. We can easily support YAML for AIConfig with our existing JSON schema.

I have also created a PR to add AIConfig schema to schemastore: https://github.com/SchemaStore/schemastore/pull/3474. Once that gets approved, it will automatically hook up syntax validation for *.aiconfig.yml/*.aiconfig.yaml, *.aiconfig.json in VSCode and other IDEs.

Test Plan:

(Tried in both TS and Python)

Validated idempotence of the following:
```
from aiconfig import AIConfigRuntime
config = AIConfigRuntime.load('travel.aiconfig.json')
config.save("travel.aiconfig.yaml", mode="yaml")

config = AIConfigRuntime.load('travel.aiconfig.yaml')
config.save("travel2.aiconfig.json")

config.save()
```

`travel.aiconfig.json`:
```json
{
  "name": "NYC Trip Planner",
  "description": "Intrepid explorer with ChatGPT and AIConfig",
  "schema_version": "latest",
  "metadata": {
    "models": {
      "gpt-3.5-turbo": {
        "model": "gpt-3.5-turbo",
        "top_p": 1,
        "temperature": 1
      },
      "gpt-4": {
        "model": "gpt-4",
        "max_tokens": 3000,
        "system_prompt": "You are an expert travel coordinator with exquisite taste."
      }
    },
    "default_model": "gpt-3.5-turbo",
  },
  "prompts": [
    {
      "name": "get_activities",
      "input": "Tell me 10 fun attractions to do in NYC.",
      "metadata": {

      }
    },
    {
      "name": "gen_itinerary",
      "input": "Generate an itinerary ordered by {{order_by}} for these activities: {{get_activities.output}}.",
      "metadata": {
        "model": "gpt-4",
        "parameters": {
          "order_by": "geographic location"
        }
      }
    }
  ]
}
```

`travel.aiconfig.yaml`:
```yaml
description: Intrepid explorer with ChatGPT and AIConfig
metadata:
  default_model: gpt-3.5-turbo
  models:
    gpt-3.5-turbo:
      model: gpt-3.5-turbo
      temperature: 1
      top_p: 1
    gpt-4:
      max_tokens: 3000
      model: gpt-4
      system_prompt: You are an expert travel coordinator with exquisite taste.
  parameters: {}
name: NYC Trip Planner
prompts:
[...]

PR #1058, opened 2024-01-29
[2/n] Add to_string method to AIConfigRuntime

saqadri:
[2/n] Add to_string method to AIConfigRuntime

Allow saving an AIConfigRuntime to a string by creating a `to_string` method that encapsulates the core business logic of the `save` method.

This is needed for VSCode extension (top of stack), which sends the state of the AIConfigRuntime to the editor as a string for management (the IDE is responsible for managing the document, and for saving it to disk).

Test Plan:

Tested out getting started ipynb with:

save JSON:

```
config = AIConfigRuntime.load('travel.aiconfig.json')
config.save('travel2.aiconfig.json')
```

save YAML:
```
config.save('travel2.aiconfig.yaml', mode="yaml")
```

to JSON string:
```
config_str = config.to_string()
print(config_str)
```

to YAML string:
```
config_yaml = config.to_string(mode="yaml")
print(config_yaml)
```

For further sanity made sure that the generated strings could be loaded back as an AIConfig and `config.run` worked:

```
config = config.load_yaml(config_yaml)
config.run(...)
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1058).
* #1076
* #1060
* #1059
* __->__ #1058
* #1057

### DIFF
--- a/python/src/aiconfig/Config.py
+++ b/python/src/aiconfig/Config.py
@@ -455,17 +455,6 @@ class AIConfigRuntime(AIConfig):
             config_filepath (str, optional): The file path to the JSON or YAML configuration file.
                 Defaults to "aiconfig.json" or "aiconfig.yaml", depending on the mode.
         """
-        # Decide if we want to serialize as YAML or JSON
-
-        # AIConfig json should only contain the core data fields. These are auxiliary fields that should not be persisted
-        exclude_options = {
-            "prompt_index": True,
-            "file_path": True,
-            "callback_manager": True,
-        }
-
-        if not include_outputs:
-            exclude_options["prompts"] = {"__all__": {"outputs"}}
 
         default_filepath = (
             "aiconfig.yaml" if mode == "yaml" else "aiconfig.json"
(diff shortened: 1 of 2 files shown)


python/src/aiconfig/Config.py
```diff
@@ -479,30 +468,54 @@ def save(
...
+            exclude=exclude_options,
+            exclude_none=True,
+        )
+
+        if json_data.get("$schema", None) is None:
+            # Set the schema if it is not set
+            json_data["$schema"] = "https://json.schemastore.org/aiconfig-1.0"
+
+        if mode == "yaml":
+            # Save AIConfig JSON as YAML string
```
rholinshead, line 508: nit, Save -> Return

python/src/aiconfig/Config.py
```diff
@@ -479,30 +468,54 @@ def save(
...
+            return yaml.dump(
+                json_data,
+                indent=2,
+            )
+        else:
+            # Save AIConfig as JSON string, with the schema specified
```
rholinshead, line 514: Save -> Return

typescript/lib/config.ts
```diff
@@ -233,12 +233,40 @@ export class AIConfigRuntime implements AIConfig {
...
    * @param filePath The path to the file to save to.
    * @param saveOptions Options that determine how to save the AIConfig to the file.
+   * @param mode Whether to save the AIConfig as JSON or YAML. If unspecified, the file extension will be used to determine the mode.
    */
   public save(
     filePath?: string,
     saveOptions?: SaveOptions,
     mode?: "json" | "yaml"
   ) {
+    const defaultFilePath = mode === "yaml" ? "aiconfig.yaml" : "aiconfig.json";
```
rholinshead, line 243: Shouldn't this be `<something>.aiconfig.yaml` to include`.aiconfig.` as part of the extension? I guess it's fine for consistency with python side but I had brought it up in https://github.com/lastmile-ai/aiconfig/pull/583 and think it's still worth the discussion

typescript/lib/config.ts
```diff
@@ -233,12 +233,40 @@ export class AIConfigRuntime implements AIConfig {
...
+    if (!filePath) {
+      filePath = this.filePath ?? defaultFilePath;
+    }
+
+    if (mode == null) {
```
rholinshead, line 248: nit, could alternatively just default to "json" in the function param

typescript/lib/config.ts
```diff
@@ -233,12 +233,40 @@ export class AIConfigRuntime implements AIConfig {
...
+
+    const aiConfigString = this.toString(saveOptions, mode);
+    fs.writeFileSync(filePath, aiConfigString);
+  }
+
+  /**
+   * Returns this AIConfig as a string.
+   *
+   * Note that this doesn't return the full string representation of the AIConfig object,
+   * but rather the string representation of the AIConfig object that can be persisted to a file.
```
rholinshead, line 265: nit, "string representation" -> "serialized representation" and also probably worth explaining what the difference between these is

typescript/lib/config.ts
```diff
@@ -233,12 +233,40 @@ export class AIConfigRuntime implements AIConfig {
...
+   * @param saveOptions Options that determine how to serialize the AIConfig to string.
```
rholinshead, line 266: nit, `saveOptions` is pretty specific to one use case of this method. Would be better to name it `serializeOptions` or something

typescript/lib/config.ts
```diff
@@ -254,32 +282,20 @@ export class AIConfigRuntime implements AIConfig {
...
-
-      if (mode == null) {
-        if (isYamlExt(filePath)) {
-          mode = "yaml";
-        } else {
-          // Default to JSON
-          mode = "json";
-        }
+      if (aiConfigObj["$schema"] == null) {
+        // Add the $schema property to the JSON object before saving it. In the future this can respect the version specified in the AIConfig.
```
rholinshead, line 286: Just a side note, there was some discussion in slack about whether schema_version is even needed in the serialized config if we have $schema as well (cc @andrew-lastmile )

jonathanlastmileai (approved):

typescript/lib/config.ts
```diff
@@ -254,32 +282,20 @@ export class AIConfigRuntime implements AIConfig {
...
+        aiConfigObj["$schema"] = "https://json.schemastore.org/aiconfig-1.0";
       }
 
       // TODO: saqadri - make sure that the object satisfies the AIConfig schema
       let aiConfigString;
       if (mode === "yaml") {
         aiConfigString = yaml.dump(aiConfigObj, { indent: 2 });
       } else {
-        // Add the $schema property to the JSON object before saving it. In the future this can respect the version specified in the AIConfig.
```
rholinshead, line 263: Oh, interesting - this was only done in the JSON format before. Curious if it was intended to be left out in the YAML or just accidentally missed before (and fixed here)?

rholinshead (approved):
Accepting to unblock, just some small things

PR #1059, opened 2024-01-29
[3/n] Add /api/load_content and /api/to_string endpoints to aiconfig server

saqadri:
[3/n] Add /api/load_content and /api/to_string endpoints to aiconfig server

These endpoints allow loading an AIConfigRuntime from a string, and serializing it to a string. These will be used at the top of the stack by the VSCode extension.

Test Plan:
* At the top of the stack. Validated locally that these work.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1059).
* #1076
* #1060
* __->__ #1059
* #1058
* #1057

PR #1060, opened 2024-01-29
[4/n] Add 'start' command to aiconfig server and CLI

saqadri:
[4/n] Add 'start' command to aiconfig server and CLI

The current aiconfig CLI has an 'edit' command which starts both the frontend and backend servers, and also loads the aiconfig from disk.

In the same vein, this diff introduces a 'start' command that is used to just start the aiconfig server. This is used by the vscode extension to spawn a server when an aiconfig file is opened.

Unlike the 'edit' command, 'start':
* Doesn't load the aiconfig on server init. Instead, the /api/load_content endpoint is called to initialize.
* Doesn't search for open ports. The vscode client is responsible for finding an open port. This is because the client needs to be in control of the server lifecycle.

A small change that was also added to this diff: using `dotenv` to load an env file.

Test Plan:
* At the top of stack

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1060).
* #1076
* __->__ #1060
* #1059
* #1058
* #1057

### DIFF
--- a/python/src/aiconfig/editor/server/server.py
+++ b/python/src/aiconfig/editor/server/server.py
@@ -25,6 +25,7 @@ from aiconfig.editor.server.server_utils import (
     OpArgs,
     ServerMode,
     ServerState,
+    StartServerConfig,
     ValidatedPath,
     get_http_response_load_user_parser_module,
     get_server_state,
@@ -62,37 +63,49 @@ CORS(app, resources={r"/api/*": {"origins": "*"}})
 
 
 def run_backend_server(
-    edit_config: EditServerConfig, aiconfigrc_path: str
+    initialization_settings: StartServerConfig | EditServerConfig,
+    aiconfigrc_path: str,
 ) -> Result[str, str]:
-    LOGGER.setLevel(edit_config.log_level)
-    LOGGER.info("Edit config: %s", edit_config.model_dump_json())
+    LOGGER.setLevel(initialization_settings.log_level)
+    LOGGER.info("Edit config: %s", initialization_settings.model_dump_json())
     LOGGER.info(
-        f"Starting server on http://localhost:{edit_config.server_port}"
+        f"Starting server on http://localhost:{initialization_settings.server_port}"
     )
-    try:
-        LOGGER.info(
-            f"Opening browser at http://localhost:{edit_config.server_port}"
-        )
-        webbrowser.open(f"http://localhost:{edit_config.server_port}")
-    except Exception as e:
-        LOGGER.warning(
-            f"Failed to open browser: {e}. Please open http://localhost:{port} manually."
-        )
+
+    if isinstance(initialization_settings, EditServerConfig):
+        try:
+            LOGGER.info(
+                f"Opening browser at http://localhost:{initialization_settings.server_port}"
+            )
+            webbrowser.open(
+                f"http://localhost:{initialization_settings.server_port}"
+            )
+        except Exception as e:
+            LOGGER.warning(
+                f"Failed to open browser: {e}. Please open http://localhost:{initialization_settings.server_port} manually."
+            )
+    else:
+        # In the case of the 'start' command, just the webserver is started up, and there's no need to open the browser
+        pass
 
     app.server_state = ServerState()  # type: ignore
     res_server_state_init = init_server_state(
-        app, edit_config, aiconfigrc_path
+        app, initialization_settings, aiconfigrc_path
     )
     match res_server_state_init:
         case Ok(_):
             LOGGER.info("Initialized server state")
-            debug = edit_config.server_mode in [
+            debug = initialization_settings.server_mode in [
                 ServerMode.DEBUG_BACKEND,
                 ServerMode.DEBUG_SERVERS,
             ]
-            LOGGER.info(f"Running in {edit_config.server_mode} mode")
+            LOGGER.info(
+                f"Running in {initialization_settings.server_mode} mode"
+            )
             app.run(
-                port=edit_config.server_port, debug=debug, use_reloader=debug
+                port=initialization_settings.server_port,
+                debug=debug,
+                use_reloader=debug,
             )
             return Ok("Done")
         case Err(e):
--- a/python/src/aiconfig/editor/server/server.py
+++ b/python/src/aiconfig/editor/server/server.py
@@ -7,7 +7,7 @@ import threading
 import time
 import uuid
 import webbrowser
-from typing import Any, Dict, Type, Union
+from typing import Any, Dict, Literal, Type, Union
 
 import lastmile_utils.lib.core.api as core_utils
 import result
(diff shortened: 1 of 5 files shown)


python/src/aiconfig/scripts/aiconfig_cli.py
```diff
@@ -122,6 +152,30 @@ def is_port_in_use(port: int) -> bool:
...
+
+    LOGGER.warning(f"Using {port} to start aiconfig server.")
+
+    results: list[Result[str, str]] = []
+    backend_res = run_backend_server(start_config, aiconfigrc_path)
+    match backend_res:
+        case Ok(_):
+            pass
+        case Err(e):
+            return Err(e)
```
jonathanlastmileai, line 173: nit: I don't think you need this block. You'll return the natural Ok or Err on line 176

python/src/aiconfig/editor/server/server_utils.py
```diff
@@ -77,6 +79,24 @@ def convert_to_mode(
...
+    @field_validator("server_mode", mode="before")
+    def convert_to_mode(
+        cls, value: Any
+    ) -> ServerMode:  # pylint: disable=no-self-argument
+        if isinstance(value, str):
+            try:
+                return ServerMode[value.upper()]
+            except KeyError as e:
+                raise ValueError(f"Unexpected value for mode: {value}") from e
+        return value
```
jonathanlastmileai, line 97: nit: we can dedup with EditServerConfig either by making aiconfig_path optional or nesting the common fields in a shared sub-record type. Also I think we have enough of these enum field validators now to justify a util function.

python/src/aiconfig/editor/server/server_utils.py
```diff
@@ -268,13 +288,25 @@ def safe_load_from_disk(
...
+    # TODO: saqadri - load specific .env file if specified
+    dotenv.load_dotenv()
+    _load_user_parser_module_if_exists(
+        initialization_settings.parsers_module_path
+    )
     state = get_server_state(app)
     state.aiconfigrc_path = aiconfigrc_path
 
+    if isinstance(initialization_settings, StartServerConfig):
+        # The aiconfig will be loaded later, when the editor sends the payload.
```
jonathanlastmileai, line 305: logger.info ?

python/src/aiconfig/editor/server/server_utils.py
```diff
@@ -268,13 +288,25 @@ def safe_load_from_disk(
...
+        return Ok(None)
+    # else:
+    edit_config = initialization_settings
```
jonathanlastmileai, line 308: is this intended?

python/src/aiconfig/scripts/aiconfig_cli.py
```diff
@@ -43,6 +44,7 @@ def run_subcommand(argv: list[str]) -> Result[str, str]:
     subparser_record_types = {
         "edit": EditServerConfig,
         "rage": rage.RageConfig,
+        "start": StartServerConfig,
```
jonathanlastmileai, line 47: nit: Instead of a new subcommand, this can be `edit --daemon` or something, loosely following emacs

python/src/aiconfig/scripts/aiconfig_cli.py
```diff
@@ -72,6 +74,14 @@ def run_subcommand(argv: list[str]) -> Result[str, str]:
         LOGGER.debug(f"{edit_config.is_ok()=}")
         out = _run_editor_servers_with_configs(edit_config, cli_config)
         return out
+    elif subparser_name == "start":
+        LOGGER.debug("Running start subcommand")
+        start_config = core_utils.parse_args(
+            main_parser, argv[1:], StartServerConfig
+        )
```
jonathanlastmileai, line 81: I should probably abstract this out at this point. argv[1:] everywhere is ugly

python/src/aiconfig/scripts/aiconfig_cli.py
```diff
@@ -104,6 +114,26 @@ def _run_editor_servers_with_configs(
...
+
+    server_outcomes = _start_server(
+        start_config.unwrap(), cli_config.unwrap().aiconfigrc_path
+    )
+    if server_outcomes.is_err():
+        return Err(
+            f"Something went wrong with starting the aiconfig server: {server_outcomes=}"
+        )
+
+    return Ok(",".join(server_outcomes.unwrap()))
```
jonathanlastmileai, line 134: nit

`.unwrap()` is like a TODO - let's try to use the chaining API where possible (.map, .and_then, do()...)

This can be done more concisely:

```
return do(
  Ok(",".join(server_outcomes_ok)
  for start_config_ok in start_config
  for cli_config_ok in cli_config
  for server_outcomes_ok in _start_server(
      start_config_ok, cli_config_ok.aiconfigrc_path
  )
)
```

jonathanlastmileai (approved):
nits

python/src/aiconfig/editor/server/server_utils.py
jonathanlastmileai: https://github.com/lastmile-ai/lastmile-utils/issues/31

rholinshead:
> spawn a server when an aiconfig file is opened
Just curious why we'd want to spawn the server and not also load/initialize the config for the file that was opened?

python/src/aiconfig/editor/server/server.py
```diff
@@ -61,37 +63,48 @@
...
+            )
+            webbrowser.open(
+                f"http://localhost:{initialization_settings.server_port}"
+            )
+        except Exception as e:
+            LOGGER.warning(
+                f"Failed to open browser: {e}. Please open http://localhost:{initialization_settings.server_port} manually."
+            )
+    else:
+        # In the case of the 'start' command, just the webserver is started up, and there's no need to open the browser
```
Ankush-lastmile, line 88: Please add a `pass` statement here or get rid of the `else`. Without it, python will throw an Indentation error

Ankush-lastmile:
>Test Plan:

At the top of stack

Could you clarify what the test plan is

python/src/aiconfig/editor/server/server_utils.py
saqadri: I thought about that, but figured there may be other things to add to the `start` command in the future that the edit command doesn't need, so opted to keep them separate

python/src/aiconfig/editor/server/server.py
saqadri: Great catch! Thanks

saqadri:
> Just curious why we'd want to spawn the server and not also load/initialize the config for the file that was opened?

@rholinshead this is because VSCode is in charge of the document, so it's not always possible to read from disk. For example, a document could be opened with previous unsaved changes, or an Untitled document (i.e. document that doesn't yet exist on file system) may be opened. In those cases we still need aiconfig server spawned, but the content it uses to load the aiconfig will come directly from the IDE.

Ankush-lastmile (approved):

saqadri:
Heads up -- going to merge this in and address feedback at the top of the stack

PR #1061, opened 2024-01-29
Download Button [1/n]: Setup UI on the editor client

rossdanlm:
Download Button [1/n]: Setup UI on the editor client

If the download callback is defined (not for local editor or VSCode, but will be for Gradio), then we display it and use the loading state to disable you from clicking on the button.

I'm making the signature of the callback take in an AIConfig object parameter, because we have function `clientConfigToAIConfig` within the editor pacakge: https://github.com/lastmile-ai/aiconfig/blob/c838856f36941b0796c8f12edcf846cc18ebdaa1/python/src/aiconfig/editor/client/src/shared/types.ts#L37

## Test Plan
Rebase onto https://github.com/lastmile-ai/aiconfig/pull/1045

```
```

```
```

https://github.com/lastmile-ai/aiconfig/assets/151060367/123319cf-2e92-49e8-8f93-2d930a3b0ac7


python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
```diff
@@ -150,6 +152,26 @@ export default function AIConfigEditor({
...
   const logEventHandler = callbacks?.logEventHandler;
 
+  const downloadCallback = callbacks?.download;
+  const onDownload = useCallback(
+    async (config: AIConfig) => {
+      if (!downloadCallback) {
+        return;
+      }
+      try {
+        await downloadCallback(config);
```
rholinshead, line 162: Hmm, this is kind of strange -- here we are using the client config as the source of truth and I guess sending it to the server to initiate the download? Can't we just use the server-copy similar to share?

rholinshead: r.e. using server link. If we had an endpoint for the server to give us the download we would be able to just use the download property for an anchor element: https://developer.mozilla.org/en-US/docs/Web/API/HTMLAnchorElement/download

rossdanlm: > endpoint for the server to give us the download

You mean give us the AIConfig from server? This is similar to the `load()` endpoint which we could re-use if desired (but we'd have to refactor it to accept sessionId to get the right AIConfig instance)

rholinshead: Yes. I think this is fine for now. In gradio we can just fetch from the server and download. It won't actually be 2 hops since it's just one server request (to download the config) and then the browser -> user machine is a synchronous (local) operation, right?

PR #1063, opened 2024-01-29
[editor] Support ThemeMode Override

rholinshead:
# [editor] Support ThemeMode Override

In Gradio, the theme/colors for the UI can be specified as part of the user's HuggingFace account settings. That theme value is propagated through gradio custom components as a theme_mode='system' | 'dark' | 'light' value (system by default). Our aiconfig-editor code already dynamically changes colors for gradio theme based on system theme. However, it doesn't currently have a way to dynamically change based on HF theme settings. This PR adds a 'themeMode' prop to the AIConfigEditor with which root components can override the dark/light theme used by the editor.

With this landed, we'll then be able to pass the prop from the Gradio component and fix the HF space theme issues

https://github.com/lastmile-ai/aiconfig/assets/5060851/f6141e64-ea83-44e5-bc5c-60aa1997d16a

## Testing
- In LocalEditor, set mode="gradio"
- Set themeMode to light/dark and ensure it overrides system theme

### DIFF
--- a/python/src/aiconfig/editor/client/src/shared/types.ts
+++ b/python/src/aiconfig/editor/client/src/shared/types.ts
@@ -63,15 +63,17 @@ export function aiConfigToClientConfig(aiconfig: AIConfig): ClientAIConfig {
   };
 }
 
-export type LogEvent = 
+export type LogEvent =
   | "ADD_PROMPT"
-  | "SAVE_BUTTON_CLICKED" 
+  | "SAVE_BUTTON_CLICKED"
   | "RUN_PROMPT_START"
-  | "RUN_PROMPT_CANCELED" 
-  | "RUN_PROMPT_ERROR" 
+  | "RUN_PROMPT_CANCELED"
+  | "RUN_PROMPT_ERROR"
   | "RUN_PROMPT_SUCCESS";
 
 // TODO: schematize this
 export type LogEventData = JSONObject;
 
 export type AIConfigEditorMode = "local" | "gradio" | "vscode";
+
+export type ThemeMode = "light" | "dark";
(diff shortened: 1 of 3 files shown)


rossdanlm (approved):
Nice!

python/src/aiconfig/editor/client/src/shared/types.ts
```diff
@@ -63,15 +63,17 @@ export function aiConfigToClientConfig(aiconfig: AIConfig): ClientAIConfig {
...
+  | "RUN_PROMPT_CANCELED"
+  | "RUN_PROMPT_ERROR"
   | "RUN_PROMPT_SUCCESS";
 
 // TODO: schematize this
 export type LogEventData = JSONObject;
 
 export type AIConfigEditorMode = "local" | "gradio" | "vscode";
+
+export type ThemeMode = "light" | "dark";
```
rossdanlm, line 79: Shouldn't we also allow `"system"`?

python/src/aiconfig/editor/client/src/themes/AIConfigEditorThemeProvider.tsx
```diff
@@ -18,8 +19,18 @@ const THEMES = {
   vscode: LOCAL_THEME,
 };
 
-export default function AIConfigEditorThemeProvider({ children, mode }: Props) {
-  const preferredColorScheme = useColorScheme();
+export default function AIConfigEditorThemeProvider({
+  children,
+  mode,
+  themeMode,
```
rossdanlm, line 25: Then we can set "system" as default here if undefined

python/src/aiconfig/editor/client/src/shared/types.ts
rossdanlm: Ok nm, "system" is purely Gradio convention. Both RN and Mantine use dark, light only: https://mantine.dev/hooks/use-color-scheme/#definition

rholinshead: Ya, 'system' for us is equivalent to not passing anything since we already grab the system theme in the ThemeProvider

PR #1065, opened 2024-01-29
[python] Update Parameterization to pull outputs regardless of input

Ankush-lastmile:
[python] Update Parameterization to pull outputs regardless of input

Today parameterization will skip prompts if the input is not parameterizable. Instead, update to paramterize outputs even if the input is not parameterizable.

Typescript sdk already does something similar.

## Testplan
Run with HF Remote Inference models

Screenshot 2024-01-29 at 4 38 09 PM

### DIFF
--- a/python/src/aiconfig/util/params.py
+++ b/python/src/aiconfig/util/params.py
@@ -1,6 +1,6 @@
 import re
 from collections import defaultdict
-from typing import TYPE_CHECKING, Dict, List, Set
+from typing import TYPE_CHECKING, Any, Dict, List, Set
 
 from aiconfig.registry import ModelParserRegistry
 from pybars import Compiler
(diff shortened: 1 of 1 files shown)


Ankush-lastmile:
removed an unintended ipynb

python/src/aiconfig/util/params.py
```diff
@@ -262,18 +262,18 @@ def collect_prompt_references(
...
+        
         prompt_output = (
             ai_config.get_output_text(
                 previous_prompt, ai_config.get_latest_output(previous_prompt)
             )
             if previous_prompt.outputs
             else None
         )
+
+        # If some value is None, pybars will skip it.
```
rholinshead, line 276: What does this mean? Oh, I guess pybars will ignore None when resolving the template?

rholinshead (approved):

python/src/aiconfig/util/params.py
Ankush-lastmile: yes that is correct. I'll update the comment to be more clear

PR #1062, opened 2024-01-29
[5/n] Create vscode extension scaffolding

saqadri:
[5/n] Create vscode extension scaffolding

Just boilerplate code.

Test Plan:
* `yarn install && yarn compile`
* Run in Extension Development Host
* Open travel.aiconfig.json
* Make sure "Hello World from aiconfig-editor" is displayed

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1062).
* #1075
* #1073
* #1070
* #1069
* #1068
* #1066
* __->__ #1062
* #1060
* #1059
* #1058
* #1057

PR #1068, opened 2024-01-29
[7/n] Create VSCode Editor webview

saqadri:
[7/n] Create VSCode Editor webview

Create React app that instantiates an AIConfigEditor using the `"@lastmileai/aiconfig-editor` package.

The key differences compared to local editor:
1. Event listener to receive messages from extension host. Messages include the aiconfig content, and the aiconfig server URL

2. vscode-webview context to be able to post messages back to the extension host, and to save state (which is needed because webviews get hydrated and dehydrated by the IDE for resource management).

3. notifying the extension host when a document is in 'dirty' state.

4. Build process to bundle a single `main.js` file for all the changes.

Test Plan:
* Made sure bundle builds successfully with `yarn build:vscode`.
* Detailed test plan coming at the top of the stack

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1068).
* #1075
* #1073
* #1070
* #1069
* __->__ #1068
* #1066
* #1062
* #1060
* #1059
* #1058
* #1057


vscode-extension/editor/src/VSCodeEditor.tsx
```diff
@@ -0,0 +1,449 @@
...
+      {}
+    );
+
+    const enableTelemetry = res.allow_usage_data_sharing;
+
+    if (enableTelemetry) {
+      datadogLogs.init({
+        clientToken: "pub356987caf022337989e492681d1944a8",
+        env: process.env.NODE_ENV ?? "development",
+        service: "aiconfig-editor",
```
rholinshead, line 147: nit, we probably want to distinguish vs code from local and gradio here. Maybe "aiconfig-editor-vscode" or "vscode-aiconfig-editor"

vscode-extension/editor/src/VSCodeEditor.tsx
```diff
@@ -0,0 +1,449 @@
...
+    async (
+      promptName: string,
+      onStream: RunPromptStreamCallback,
+      onError: RunPromptStreamErrorCallback,
+      enableStreaming: boolean = true,
+      cancellationToken?: string
+    ) => {
+      if (vscode) {
+        notifyDocumentDirty(vscode);
+      }
```
rholinshead, line 237: Is this needed here? I see it done later on in this function as well. Seems like it should only be done later on to be consistent with other places (done after the request is sent)

vscode-extension/editor/src/VSCodeEditor.tsx
```diff
@@ -0,0 +1,449 @@
...
+    [aiConfigServerUrl, vscode]
+  );
+
+  const clearOutputs = useCallback(async () => {
+    const res = await ufetch.post(
+      ROUTE_TABLE.CLEAR_OUTPUTS(aiConfigServerUrl),
+      {}
+    );
+
+    if (vscode) {
```
rholinshead, line 220: Not needed here, but one nice way to reduce the duplication would be to implement a new hook, `useDocumentActionCallback` which calls the passed-in callback and then does the notifyDocumentDirty(vscode) call

```
export function useDocumentActionCallback(callbackFn, dependencies) {
  const { vscode } = useContext(WebviewContext);
  return useCallback(async (...args) => {
    const res = await callbackFn(...args);
    if (vscode) {
      notifyDocumentDirty(vscode);
    }
    return res;
  }, [dependencies]);
}
```

But no promise this code works, so definitely a follow-up action if at all

vscode-extension/editor/src/VSCodeEditor.tsx
rholinshead: Oh maybe to handle the streaming, it's done before instead of on stream events

vscode-extension/editor/src/utils/vscodeUtils.ts
```diff
@@ -0,0 +1,104 @@
...
+ * Notification message to extension host informing it that the document has been modified in the webview.
+ */
+export type NotifyDocumentDirtyMessage = {
+  type: "document_changed";
+};
+
+/**
+ * Message to extension host requesting that it execute the given prompt.
+ */
+export type ExecuteRunMessage = {
```
rholinshead, line 84: Is this still needed? Don't see it used here but maybe later in the stack?

vscode-extension/editor/src/utils/vscodeUtils.ts
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * State that gets serialized and restored when the webview is recreated.
+ * This is important because vscode controls the webview lifecycle and
+ * will dehydrate the webview when it is inactive.
+ * We can store anything we want in there, and we use this type to define what we store.
+ * For more information, see https://code.visualstudio.com/api/extension-guides/webview#getstate-and-setstate
+ */
+export type WebviewState = {
+  aiconfigState?: ClientAIConfig;
```
rholinshead, line 13: Assuming this will be used later in the stack?

vscode-extension/editor/src/VSCodeEditor.tsx
```diff
@@ -0,0 +1,449 @@
...
+  },
+}));
+
+const MODE = "vscode";
+
+export default function VSCodeEditor() {
+  const { vscode } = useContext(WebviewContext);
+
+  // TODO: saqadri - does this need to be wrapped in a memo?
+  const webviewState = getWebviewState(vscode);
```
rholinshead, line 44: In the current code it doesn't since it's only used in one place (initializing the state). It would actually be better to just call it within the state initialization instead of here (where it will be called unnecessarily each render). But maybe it's used elsewhere later in the stack.

saqadri:
Will address PR feedback at the top of stack

PR #1069, opened 2024-01-30
[8/n] Add init command to vscode extension

saqadri:
[8/n] Add init command to vscode extension

The way we package dependencies for the vscode extension is by using pip to install the `python-aiconfig` package, as well as any aiconfig extensions we want to package together by default.

This diff introduces the `check_requirements.py` script which knows how to parse a requirements.txt file and check if the dependencies and versions listed in it are already installed on the system.

The `init` command then does the following:
* Checks if python is installed (if not, it prompts the user to install).
* Checks if pip is installed on the system (this should always be installed if python has been installed, but is done as a sanity check)
* Checks if given the requirements.txt, any packages need to be installed or updated.
* If an update is required, it installs the dependencies using pip.

This is displayed to the user in the form of a vscode progress notification.

Finally, we create an AIConfig output channel that displays logs in the Output pane. This will be used by the aiconfig server as well to display detailed logs.

Test Plan:
* See video in comment.
* Also did some manual stress testing of `check_requirements.py` script -- made sure it handles ==, <=, >=, <, > operators, weird spaces, and comments in the requirements.txt file.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1069).
* #1075
* #1073
* #1070
* __->__ #1069
* #1068
* #1066
* #1062
* #1060
* #1059
* #1058
* #1057


saqadri:
https://github.com/lastmile-ai/aiconfig/assets/25641935/8db9c788-d3a3-4f6b-8afb-00a32ae886fd

vscode-extension/src/extension.ts
```diff
@@ -2,11 +2,31 @@
...
+    `Congratulations, your extension ${EXTENSION_NAME} is now active!`
+  );
+
+  // Create an output channel for the extension
+  const extensionOutputChannel = vscode.window.createOutputChannel("AIConfig", {
+    log: true,
+  });
+
+  let setupCommand = vscode.commands.registerCommand(COMMANDS.INIT, () => {
+    installDependencies(context, extensionOutputChannel);
```
rholinshead, line 24: If any of the required steps fail during initialization (e.g. no pip), doesn't this still activate the extension (albeit with messages from the initialization)? Would it be better to fail the activation if we know it will probably have issues due to missing dependencies?

PR #1070, opened 2024-01-30
[9/n] Add AIConfigEditor custom text editor

saqadri:
[9/n] Add AIConfigEditor custom text editor

This change introduces the AIConfig custom text editor. For more information on how custom text editors work in vscode, please see: https://code.visualstudio.com/api/extension-guides/custom-editors

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1070).
* #1075
* #1073
* __->__ #1070
* #1069
* #1068
* #1066
* #1062
* #1060
* #1059
* #1058
* #1057


saqadri:
I understand this is likely going to be hard to review without context. I owe you folks a quick design doc on the vscode editor @rholinshead @rossdanlm @Ankush-lastmile. Will do that on Friday.

Very high level:
There are 3 components that are interacting with each other:
1. the vscode extension host (this is where the extension actually runs in the IDE)
2. the aiconfig editor webview (which is more or less identical to the local editor)
3. the aiconfig server (which is identical to the local editor server)

One key point is that vscode (1) is responsible for the document, since the document can be in-memory with unsaved changes, untitled, or on disk. The extension host communicates directly with the aiconfig server (3) to initialize its state, notify it of changes, and manage the process lifecycle (starting up a new server on document open, closing it on document close, handling errors, etc.)

The webview (2) also communicates with the aiconfig server, pretty much exactly like it does with local editor. The main difference from local editor is that the webview notifies the extension host when the document is dirty, so that vscode can ask the aiconfig server for the updated state of the document.

The principle I've followed is to still treat the aiconfig server's state as the source of truth. Before saving the document, vscode asks the server for the latest state of the document it has got. Additionally, vscode is responsible for notifying the server of updates to the documents (which triggers a reload of the AIConfigRuntime). This way we still are dealing with a single source of truth for the state of the AIConfig instead of it being split across 1, 2 and 3.

vscode-extension/src/aiConfigEditor.ts
```diff
@@ -0,0 +1,413 @@
...
+   `;
+  }
+
+  /**
+   * Replaces the entire existing document with the provided new text.
+   */
+  private updateTextDocument(document: vscode.TextDocument, newText: string) {
+    const edit = new vscode.WorkspaceEdit();
+
+    // TODO: saqadri - figure out a way to update the document piecemeal
```
rholinshead, line 404: Would we ever want each prompt to be its own document? I vaguely remember that being the case in other applications, I think due to LSP integrations?

rholinshead (approved):
I think this looks good, from my limited vscode extension experience. We can follow-up as needed

vscode-extension/src/aiConfigEditor.ts
```diff
@@ -0,0 +1,413 @@
...
+
+    // Tip: Install the es6-string-html VS Code extension to enable code highlighting below
+    return /*html*/ `
+	 <!DOCTYPE html>
+	 <html lang="en">
+	   <head>
+		 <meta charset="utf-8">
+		 <meta name="viewport" content="width=device-width,initial-scale=1,shrink-to-fit=no">
+		 <meta http-equiv="Content-Security-Policy" content="default-src 'none'; img-src data: http: https: 'self'; connect-src vscode-webview: ${editorServer.url} http: https:; style-src 'unsafe-inline' ${webview.cspSource} https://cdn.jsdelivr.net/npm/monaco-editor@0.43.0 https://cdn.jsdelivr.net/npm/monaco-editor@0.43.0/min/vs/editor/editor.main.css; script-src 'nonce-${nonce}' vscode-resource: https: http: https://cdn.jsdelivr.net/npm/monaco-editor@0.43.0 https://cdn.jsdelivr.net; font-src https://cdn.jsdelivr.net/npm/monaco-editor@0.43.0/min/vs/base/browser/ui/codicons/codicon/codicon.ttf; worker-src blob:;">
+		 <title>Hello World</title>
```
rholinshead, line 387: Does this title actually matter? If so, should it be the filename?

saqadri: Doesn't really matter, I can remove it too

saqadri:
Merging, will fix feedback at top of stack

PR #1071, opened 2024-01-30
Download Button [4/n]: Make this enabled for read-only mode

rossdanlm:
Download Button [4/n]: Make this enabled for read-only mode

I realized that after I moved `readOnly` check on entire group in https://github.com/lastmile-ai/aiconfig/pull/1032, I need to be able to still have the Download button for read-only mode so user can still download the AIConfig. We just need to make sure to define this endpoint in the lastMile repo

## Test Plan
Rebase onto https://github.com/lastmile-ai/aiconfig/pull/1045

```
```

```
```

readOnly is true

Screenshot 2024-01-30 at 00 45 03

readOnly is false

Screenshot 2024-01-30 at 00 44 36


rholinshead (changes requested):
minor things but will save us from needing another PR in the future for the share button

rholinshead (approved):

PR #1083, opened 2024-01-30
[gradio] Update 'Share' button text to 'Share Workbook'

rholinshead:
[gradio] Update 'Share' button text to 'Share Workbook'

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1083).
* __->__ #1083
* #1082

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/global/ShareButton.tsx
+++ b/python/src/aiconfig/editor/client/src/components/global/ShareButton.tsx
@@ -31,7 +31,7 @@ export default memo(function ShareButton({ onShare }: Props) {
 
   const tooltipMessage: string = isLoading
     ? "Generating share link..."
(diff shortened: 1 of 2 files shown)


jonathanlastmileai (approved):

saqadri:
Hmm, this may be confusing too if people don't know what AIConfig is, right?

Should the messaging be more about getting a shareable URL?

rholinshead:
> Hmm, this may be confusing too if people don't know what AIConfig is, right?
> 
> Should the messaging be more about getting a shareable URL?

I think the confusion @tanya-rai was more worried about is that "Share" could be confused for sharing the space itself. Even "Get Share URL" or similar could be misconstrued that way.

What about "Share Workbook" ?

PR #1087, opened 2024-01-31
[vscode] Add share command

rossdanlm:
[vscode] Add share command

Same functionality as from Gradio where we first upload to S3, and then pass in that source Uri to our aiconfig endpoint

Some things I still need to do:
1. Copy pasting feature --> right now it's kind of bulky and it just shows up tin eh dialogue at bottom right. Would be better UX to have it be copy-paste functionality
2. Remove hardcoding to file

Another thing I noticed is that when we try to open the HF extensions (or any extension that relies on non-core model parsers), the VSCode extension will fail. I think this is fine, just need to think more about the UX and how to globally register other model parsers to render this (also without relying on the client code to display input schema settings)
3. Remove hardcode and only show active open window

## Test Plan
Error gets properly thrown if s3 bucket is wrong

Screenshot 2024-01-31 at 00 20 18

Share URI works

https://github.com/lastmile-ai/aiconfig/assets/151060367/d1791c5c-e450-4ffa-a61e-e7af8488073b

### DIFF
--- a/vscode-extension/src/extension.ts
+++ b/vscode-extension/src/extension.ts
@@ -1,11 +1,17 @@
+
 // The module 'vscode' contains the VS Code extensibility API
 // Import the module and reference it with the alias vscode in your code below
 import * as vscode from "vscode";
 
+import * as AWS from 'aws-sdk';
 import { exec, spawn } from "child_process";
+import {readFileSync} from 'fs';
 import path from "path";
+import { ufetch } from "ufetch";
 import { EXTENSION_NAME, COMMANDS } from "./util";
 import { AIConfigEditorProvider } from "./aiConfigEditor";
+
+const LASTMILE_BASE_URI: string = "https://lastmileai.dev/"
 
 // This method is called when your extension is activated
 // Your extension is activated the very first time the command is executed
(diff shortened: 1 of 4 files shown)


vscode-extension/src/extension.ts
```diff
@@ -26,6 +32,46 @@ export function activate(context: vscode.ExtensionContext) {
...
+    // TODO: use vscode.window.activeTextEditor instead of hardcoding
+    const fileName: string = "/Users/rossdancraig/Projects/aiconfig/cookbooks/Gemini/gemini_demo.aiconfig.json";
+    const sanitizedFileName: string = sanitizeFileName(fileName);
+
+    // Use empty accessKeyId and secretAccessKey so bucket can be accessed publicly
+    AWS.config.update({
+      accessKeyId: '',
+      secretAccessKey: '',
+    });
+    const s3 = new AWS.S3();
```
rholinshead, line 45: Does this need any handling to work without aws credentials (as in https://github.com/lastmile-ai/gradio-workbook/pull/153/files)?

vscode-extension/src/extension.ts
```diff
@@ -26,6 +32,46 @@ export function activate(context: vscode.ExtensionContext) {
   });
   context.subscriptions.push(setupCommand);
 
+  let shareCommand = vscode.commands.registerCommand(COMMANDS.SHARE, async () => {
+    // TODO: use vscode.window.activeTextEditor instead of hardcoding
+    const fileName: string = "/Users/rossdancraig/Projects/aiconfig/cookbooks/Gemini/gemini_demo.aiconfig.json";
```
rholinshead, line 37: Should do this before landing

vscode-extension/src/extension.ts
```diff
@@ -26,6 +32,46 @@ export function activate(context: vscode.ExtensionContext) {
...
+    };
+    await s3.putObject(input).promise();
+    
+    const s3Url: string = `https://s3.amazonaws.com/${bucket}/${uploadKey.replace(/[ ]/g, "%20")}`
+    
+    const lastmileUploadUrl: string = LASTMILE_BASE_URI + "api/aiconfig/upload";
+    const response = await ufetch.post(lastmileUploadUrl, {"url": s3Url, "source": "vscode"});
+    if (response?.id !== null && response?.id !== undefined) {
+      const renderingUrl: string = LASTMILE_BASE_URI + `aiconfig/${response?.id}`;
+      vscode.window.showInformationMessage("shareUri: " + renderingUrl);
```
rholinshead, line 70: Will the share button also show its own modal on top of this?

rholinshead: Oh, I guess this is unrelated to the button at all, just a command palette command. Nvm

vscode-extension/src/extension.ts
rossdanlm: cc @saqadri can you land final version where you had it updated?

vscode-extension/src/extension.ts
rholinshead: Oh, ok. Can you please update the comment to make that a bit clear (specifically noting 'without credentials')

rholinshead (approved):
Accepting to unblock, just the updated comment and the fix for the filename before landing

saqadri:
This PR can be closed, right?

PR #1094, opened 2024-01-31
Support Markdown Rendering in Config Description

rholinshead:
# Support Markdown Rendering in Config Description

One thing that would be nice for making configs more readable (especially in Gradio instructions -- https://huggingface.co/spaces/lastmileai/gradio-workbook-template) is if the description renderer supported markdown, similar to our output renderers. This PR swaps the non-editing view of the description to use our TextRenderer (which handles markdown, code, etc. rendering)

Screenshot 2024-01-31 at 3 49 26 PM
Screenshot 2024-01-31 at 3 49 38 PM
Screenshot 2024-01-31 at 3 50 09 PM

Added special handling for clicking on links and the prism code copy button:

https://github.com/lastmile-ai/aiconfig/assets/5060851/bf7a8a12-3901-46b2-80e8-05d22924a9b1

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/ConfigNameDescription.tsx
+++ b/python/src/aiconfig/editor/client/src/components/ConfigNameDescription.tsx
@@ -1,14 +1,8 @@
-import {
-  createStyles,
-  Stack,
-  Text,
-  Textarea,
-  TextInput,
-  Title,
-} from "@mantine/core";
+import { createStyles, Stack, Textarea, TextInput, Title } from "@mantine/core";
 import { useClickOutside } from "@mantine/hooks";
 import { memo, useCallback, useContext, useRef, useState } from "react";
 import AIConfigContext from "../contexts/AIConfigContext";
+import { TextRenderer } from "./prompt/TextRenderer";
 
 type Props = {
   name?: string;
@@ -42,6 +36,11 @@ const useStyles = createStyles((theme) => ({
   },
 }));
 
+function isClickableChildElement(element: EventTarget | null) {
+  // Specifically handle markdown links and prism code copy svg elements
+  return element instanceof HTMLAnchorElement || element instanceof SVGElement;
+}
+
 export default memo(function ConfigNameDescription({
   name,
   description,
@@ -67,7 +66,8 @@ export default memo(function ConfigNameDescription({
 
   const onHandleEnter = useCallback(
     (event: React.KeyboardEvent<HTMLInputElement | HTMLTextAreaElement>) => {
-      if (event.key === "Enter") {
+      if (event.key === "Enter" && !event.shiftKey) {
+        // Shift+Enter to add new line
         event.stopPropagation();
         setIsEditing(false);
       }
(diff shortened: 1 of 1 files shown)


python/src/aiconfig/editor/client/src/components/ConfigNameDescription.tsx
```diff
@@ -80,11 +80,16 @@ export default memo(function ConfigNameDescription({
       if (readOnly) {
         return;
       }
-      setIsEditing(true);
-      if (event.target === nameDisplayRef.current) {
+
+      if (event.currentTarget === nameDisplayRef.current) {
```
rossdanlm, line 84: Why change this from `.target` to `.current`?

python/src/aiconfig/editor/client/src/components/ConfigNameDescription.tsx
```diff
@@ -42,6 +36,11 @@ const useStyles = createStyles((theme) => ({
   },
 }));
 
+function isClickableChildElement(element: EventTarget | null) {
+  // Specifically handle markdown links and prism code copy svg elements
+  return element instanceof HTMLAnchorElement || element instanceof SVGElement;
```
rossdanlm, line 41: Why do we need to add this check when we didn't have it before?

python/src/aiconfig/editor/client/src/components/ConfigNameDescription.tsx
```diff
@@ -80,11 +80,16 @@ export default memo(function ConfigNameDescription({
...
-      setIsEditing(true);
-      if (event.target === nameDisplayRef.current) {
+
+      if (event.currentTarget === nameDisplayRef.current) {
         setInitialFocus("name");
-      } else if (event.target === descriptionDisplayRef.current) {
+        setIsEditing(true);
+      } else if (
+        event.currentTarget === descriptionDisplayRef.current &&
+        !isClickableChildElement(event.target)
```
rossdanlm, line 89: Why do we need this check now?

python/src/aiconfig/editor/client/src/components/ConfigNameDescription.tsx
```diff
@@ -80,11 +80,16 @@ export default memo(function ConfigNameDescription({
...
+      ) {
         setInitialFocus("description");
+        setIsEditing(true);
```
rossdanlm, line 92: Why do we re-set this to true when hitting enter? Don't we already exit out on L72?

python/src/aiconfig/editor/client/src/components/ConfigNameDescription.tsx
rholinshead: This is just checking if the target element passed in is either an anchor tag (link) or svg (what prism is using for the 'copy code' button)

python/src/aiconfig/editor/client/src/components/ConfigNameDescription.tsx
rholinshead: So for events, the event happens on the lowest element and bubbles up to each parent element. In our case, we wrap the TextRenderer with a div that has a ref. When we click on the div, we want to toggle it to "edit" mode, to allow editing the text. Except, we don't want that to happen if the TextRenderer has rendered clickable elements (anchor or copy code button) and the click is on those elements. Otherwise, clicking the link or copy code button would immediately show the edit view instead of doing normal click behaviour

PR #1095, opened 2024-02-01
[gradio] Move Styles into GradioTheme

rholinshead:
[gradio] Move Styles into GradioTheme

# [gradio] Move Styles into GradioTheme

Ok, this is a long stack of incremental changes to hopefully get gradio styles to look decent / match intended GradioTheme styles once and for all.

This PR moves the styles from gradio's styles.css into the theme so that we can hopefully remove the styles.css in gradio. Note that we nest them under the `div.editorBackground` (meaning, they apply to any elements with those classes and which are also descendants of the div with editorBackground class). This should ensure a high specificity on their selectors.

Check that the buttons have the correct styles:
Screenshot 2024-01-31 at 11 30 32 PM
Screenshot 2024-01-31 at 11 30 57 PM

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1095).
* #1109
* #1105
* #1104
* #1103
* #1102
* #1101
* #1100
* #1099
* #1098
* #1097
* #1096
* __->__ #1095

### DIFF
--- a/python/src/aiconfig/editor/client/src/themes/GradioTheme.ts
+++ b/python/src/aiconfig/editor/client/src/themes/GradioTheme.ts
@@ -29,8 +29,70 @@ export const GRADIO_THEME: MantineThemeOverride = {
       minHeight: "400px",
       paddingTop: "2px",
 
-      // Apply nested styles on mantine text components for higher specificity
-      // than gradio text styles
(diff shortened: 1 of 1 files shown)


rossdanlm (approved):
Cool, big global changes

PR #1096, opened 2024-02-01
[gradio] Use Function/Return Instead of Inline Theme Object

rholinshead:
[gradio] Use Function/Return Instead of Inline Theme Object

# [gradio] Use Function/Return Instead of Inline Theme Object

This is a no-op, just changing `globalStyles: (theme) => { }` into a function that explicitly returns an object, instead of inline returning the object. This will be used in the next PR where we create a variable for use in the styles, outside of the returned object

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1096).
* #1109
* #1105
* #1104
* #1103
* #1102
* #1101
* #1100
* #1099
* #1098
* #1097
* __->__ #1096
* #1095

### DIFF
--- a/python/src/aiconfig/editor/client/src/themes/GradioTheme.ts
+++ b/python/src/aiconfig/editor/client/src/themes/GradioTheme.ts
@@ -16,278 +16,281 @@ export const GRADIO_THEME: MantineThemeOverride = {
   },
 
   //gradio light theme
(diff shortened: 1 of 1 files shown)


rossdanlm (approved):
Thanks for separating this!

PR #1097, opened 2024-02-01
[gradio] Use variable for input border color

rholinshead:
[gradio] Use variable for input border color

# [gradio] Use variable for input border color

Use a variable, `inputBorderColor` for referencing the border color for different inputs so that it can be used without manually repeating the colors based on theme everywhere we want to have the same border color

## Testing
- Make sure the border for all the inputs are correct (note title is focused so it's orange)
Screenshot 2024-01-31 at 11 36 10 PM
Screenshot 2024-01-31 at 11 36 37 PM

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1097).
* #1109
* #1105
* #1104
* #1103
* #1102
* #1101
* #1100
* #1099
* #1098
* __->__ #1097
* #1096
* #1095

### DIFF
--- a/python/src/aiconfig/editor/client/src/themes/GradioTheme.ts
+++ b/python/src/aiconfig/editor/client/src/themes/GradioTheme.ts
@@ -16,278 +16,281 @@ export const GRADIO_THEME: MantineThemeOverride = {
   },
 
   //gradio light theme
-  globalStyles: (theme) => ({
-    "div.editorBackground": {
-      background: theme.colorScheme === "light" ? "white" : "#0b0f19",
-      borderRadius: "8px",
-      // Gradio component is iframed so height should be in relation to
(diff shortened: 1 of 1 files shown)


rossdanlm (approved):
Ok so basically, the 3rd arg for `border` is always colour hex?

rholinshead:
> Ok so basically, the 3rd arg for `border` is always colour hex?

Here it is, but it could be rbg or string color as well. The order actually doesn't matter, either -- `border` is just short form for 3 different styles set simultaneously (border-color, border-style, border-width) -- but this order is generally the most common (https://developer.mozilla.org/en-US/docs/Web/CSS/border)

PR #1098, opened 2024-02-01
[gradio] Fix Classes Specificity

rholinshead:
[gradio] Fix Classes Specificity

# [gradio] Fix Classes Specificity

Many of the intended styles within the GradioTheme weren't being applied because gradio itself has classes on things with higher specificity. To fix, we should be able to enforce a higher specificity by having the classes/styles moved to being underneath div.editorBackground as descendants.

This now fixes the input/text areas in most places to match the intended background color (e.g. compare to instruction input in https://huggingface.co/spaces/internlm/internlm2-math-7b in dark mode -- they are the same background color now).

Screenshot 2024-01-31 at 11 42 52 PM
Screenshot 2024-01-31 at 11 43 20 PM

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1098).
* #1109
* #1105
* #1104
* #1103
* #1102
* #1101
* #1100
* #1099
* __->__ #1098
* #1097
* #1096
* #1095

### DIFF
--- a/python/src/aiconfig/editor/client/src/themes/GradioTheme.ts
+++ b/python/src/aiconfig/editor/client/src/themes/GradioTheme.ts
@@ -16,278 +16,281 @@ export const GRADIO_THEME: MantineThemeOverride = {
   },
 
   //gradio light theme
-  globalStyles: (theme) => ({
-    "div.editorBackground": {
-      background: theme.colorScheme === "light" ? "white" : "#0b0f19",
-      borderRadius: "8px",
-      // Gradio component is iframed so height should be in relation to
-      // the height of the containing iframe, not the viewport
-      height: "100%",
-      // Add some margin & padding to better visually separate from surrounding
-      // gradio card
-      margin: "14px auto 0 auto",
-      minHeight: "400px",
-      paddingTop: "2px",
-
-      // Apply nested styles on mantine elements for higher specificity
-      // than gradio element styles
-
-      /*
-       * For all buttons, excluding menu dropdown and .ghost class buttons, apply the gradio color
-       * styles
-       */
-      ".mantine-Button-root.mantine-UnstyledButton-root:not(.ghost):not([aria-haspopup='menu'])":
-        {
-          background: "linear-gradient(to bottom right, #ffedd5, #fdba74 100%)",
-          border: "1px solid #fed7aa",
-          boxShadow: "0px 1px 4px 0px rgba(0, 0, 0, 0.05)",
-          color: "#ea580c !important",
-          fontSize: "16px" /* var(--button-large-text-size) */,
-          fontWeight: 600 /* var(--button-large-text-weight) */,
-          padding: "0 1.25rem",
-        },
-
-      /*
-       * Same as above, only on hover
-       */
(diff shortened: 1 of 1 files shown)


python/src/aiconfig/editor/client/src/themes/GradioTheme.ts
```diff
@@ -87,17 +87,6 @@ export const GRADIO_THEME: MantineThemeOverride = {
             color: "#374151",
           },
 
-        /*
-         * Specifically set the .runPromptButton to have the desired styles specified in our custom gradio theme
-         */
-        ".mantine-Button-root.mantine-UnstyledButton-root.runPromptButton": {
```
rholinshead: Shouldn't be needed if the runPromptButton class specificity is high enough now below

rossdanlm: Sorry, just to avoid confusion, when you say "high enough" you don't mean "higher up the component stack towrads ancestors", you mean "closer enough to the child component so that it overwrites the parent-defined style" right?

python/src/aiconfig/editor/client/src/themes/GradioTheme.ts
```diff
@@ -149,133 +138,141 @@ export const GRADIO_THEME: MantineThemeOverride = {
...
+              "linear-gradient(to bottom right, #ffedd5, #fdba74 100%)",
+            border: "1px solid #fed7aa",
+            boxShadow: "0px 1px 4px 0px rgba(0, 0, 0, 0.05)",
+            color: "#ea580c !important",
+            path: {
+              color: "#E85921",
+            },
+            ":hover": {
+              background: "linear-gradient(to bottom right, #ffedd5, #fed7aa)",
+            },
```
rossdanlm, line 275: In general, how do you know how to add this stuff? Do you find it from somewhere, or do you just play around and figure it out? If you find it from somewhere, how do you extra the values on webpages etc?

rossdanlm (approved):
Ngl, don't fully understand how you get these specific values but it looks visually good enough so stamping to unblock

python/src/aiconfig/editor/client/src/themes/GradioTheme.ts
rholinshead: Actually, I mean the specificity itself has the highest relevance for the matching elements based on the combination of selectors (descendant of div.editorBackground AND runPromptButton class) -- see https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity for more info

python/src/aiconfig/editor/client/src/themes/GradioTheme.ts
rholinshead: Most of this was from Zak's existing styles. But generally you can just go and 'inspect element' in existing gradio spaces to see what the styles are. Technically gradio uses style variables for a lot of things (e.g. `var(input-background-color-fill)` but those variables aren't declared for us locally so just using the explicit value of those variables where applicable

PR #1102, opened 2024-02-01
[gradio] Fix Local Parameters AddParameter Button Style

rholinshead:
[gradio] Fix Local Parameters AddParameter Button Style

# [gradio] Fix Local Parameters AddParameter Button Style

The 'add parameter' button in the prompt parameters section currently doesn't match the global one in gradio. This fixes to apply the same class to both:

## Global:
Screenshot 2024-01-31 at 11 55 12 PM

## Local:
Screenshot 2024-01-31 at 11 55 01 PM

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1102).
* #1109
* #1105
* #1104
* #1103
* __->__ #1102
* #1101
* #1100
* #1099
* #1098
* #1097
* #1096
* #1095

PR #1101, opened 2024-02-01
[editor][rfc] Only Show Outputs Section if There are Outputs

rholinshead:
[editor][rfc] Only Show Outputs Section if There are Outputs

# [editor][rfc] Only Show Outputs Section if There are Outputs

From some feedback, it's kind of weird / confusing / takes up valuable screen real estate to show an 'Outputs' section even if there are no outputs. Let's remove it:

## Before:
Screenshot 2024-01-31 at 11 52 26 PM

## After:
Screenshot 2024-01-31 at 11 52 38 PM

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1101).
* #1109
* #1105
* #1104
* #1103
* #1102
* __->__ #1101
* #1100
* #1099
* #1098
* #1097
* #1096
* #1095

PR #1100, opened 2024-02-01
[editor] Add Prompt Label to Config Renderer

rholinshead:
[editor] Add Prompt Label to Config Renderer

# [editor] Add Prompt Label to Config Renderer

Inputs without a schema are currently missing the 'Prompt' label, so add it:

Screenshot 2024-01-31 at 11 50 51 PM

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1100).
* #1109
* #1105
* #1104
* #1103
* #1102
* #1101
* __->__ #1100
* #1099
* #1098
* #1097
* #1096
* #1095

PR #1099, opened 2024-02-01
[gradio] Fix Border Color for textarea in Settings

rholinshead:
[gradio] Fix Border Color for textarea in Settings

# [gradio] Fix Border Color for textarea in Settings

Similar to previous PR - fix the settings textarea to have proper border/background color to match other gradio spaces in dark mode:

## Before
Screenshot 2024-01-31 at 11 47 44 PM

## After
Screenshot 2024-01-31 at 11 46 38 PM

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1099).
* #1109
* #1105
* #1104
* #1103
* #1102
* #1101
* #1100
* __->__ #1099
* #1098
* #1097
* #1096
* #1095

PR #1105, opened 2024-02-01
[gradio] Change Side Panel Color to Differ from Background

rholinshead:
[gradio] Change Side Panel Color to Differ from Background

# [gradio] Change Side Panel Color to Differ from Background

Just change the background color of the prompt right side panel to be a tiny bit lighter than the background.

## Before:
Screenshot 2024-02-01 at 12 01 21 AM

## After:
Screenshot 2024-02-01 at 12 01 11 AM

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1105).
* #1109
* __->__ #1105
* #1104
* #1103
* #1102
* #1101
* #1100
* #1099
* #1098
* #1097
* #1096
* #1095

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/prompt/PromptContainer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/prompt/PromptContainer.tsx
@@ -156,8 +156,13 @@ export default memo(function PromptContainer({
             isRunning={prompt._ui.isRunning}
             isRunButtonDisabled={isRunButtonDisabled}
           />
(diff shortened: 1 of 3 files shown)


rossdanlm (approved):
Honestly I don't really see any difference in the images you shared, but I think it's fine to just unblock this 1-liner and we can forward fix later

PR #1109, opened 2024-02-01
[gradio] Fix Margin of Title / Global Parameters

rholinshead:
[gradio] Fix Margin of Title / Global Parameters

# [gradio] Fix Margin of Title / Global Parameters

In gradio, the prompt menu button has styles overridden by gradio button styles and therefore has no padding. This messes up the alignment of the title/description and global parameters compared to the prompts:

Screenshot 2024-02-01 at 11 23 00 AM

Similarly, in readOnly mode there is no prompt menu button but we still had the left margin on the title/description and global parameters:
Screenshot 2024-02-01 at 11 26 04 AM

This PR fixes both situations by:
- Explicitly setting the padding/margin for the prompt menu button in gradio theme to have higher specificity than gradio button styles
- Dynamically handling the left margin of the title/description and global parameters based on readOnly or not

Notes:
- Not sure how to test vscode theme here really since the actual vs code package has different styles, etc. from what's in prod. As best-effort, I just made sure the global parametersContainer and parametersContainerReadonly styles override local style when set (so they can be used as needed in vscode theme after)

## Testing:
- Gradio theme:
Screenshot 2024-02-01 at 11 28 52 AM
Screenshot 2024-02-01 at 11 29 04 AM

- Local theme:
Screenshot 2024-02-01 at 11 29 13 AM
Screenshot 2024-02-01 at 11 29 21 AM

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1109).
* __->__ #1109
* #1105
* #1104
* #1103
* #1102
* #1101
* #1100
* #1099
* #1098
* #1097
* #1096
* #1095

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/GlobalParametersContainer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/GlobalParametersContainer.tsx
@@ -1,12 +1,22 @@
-import { Accordion, Text } from "@mantine/core";
+import { Accordion, Text, createStyles } from "@mantine/core";
 import { JSONObject } from "aiconfig";
-import { memo, useState } from "react";
+import { memo, useContext, useState } from "react";
 import ParametersRenderer from "./ParametersRenderer";
+import AIConfigContext from "../contexts/AIConfigContext";
 
 type Props = {
   initialValue: JSONObject;
   onUpdateParameters: (newParameters: JSONObject) => void;
 };
+
+const useStyles = createStyles(() => ({
+  parametersContainer: {
+    margin: "16px auto 16px 36px",
+  },
+  parametersContainerReadonly: {
+    margin: "16px auto",
+  },
+}));
 
 export default memo(function GlobalParametersContainer({
   initialValue,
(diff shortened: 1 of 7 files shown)


python/src/aiconfig/editor/client/src/components/GlobalParametersContainer.tsx
```diff
@@ -1,21 +1,43 @@
...
   return (
-    <div className="parametersContainer">
+    // Set local and global classname. Global will override if specified
+    // Local is readonly or not. Global will always have parametersContainer class
+    // and if readonly, will also have parametersContainerReadonly class (to allow overrides)
+    <div
+      className={`${
+        readOnly
+          ? classes.parametersContainerReadonly
+          : classes.parametersContainer
```
rossdanlm, line 38: Wow cool trick!

python/src/aiconfig/editor/client/src/components/GlobalParametersContainer.tsx
```diff
@@ -1,21 +1,43 @@
...
+      } parametersContainer ${readOnly ? "parametersContainerReadonly" : ""}`}
```
rossdanlm, line 39: K yo newbie mcscrewbie over here: what do the 2nd and 3rd args do for the `className` param: https://stackoverflow.com/questions/23663151/whats-the-point-of-the-second-argument-in-element-classlist-toggle

rossdanlm (approved):

python/src/aiconfig/editor/client/src/components/GlobalParametersContainer.tsx
rholinshead: Classname only takes in a string. When we use `useStyles`, the classes we define are actually serialized to class name and values used for the element styles. It's a bit of a trick, but we can actually apply multiple class names by just defining more as part of the string literal. e.g. `className="classA classB classC" would apply all 3 classes to the element.
In our case, we always want the literal `parametersContainer` class applied (so we can override it in the theme styles). We also want our local styles (from `classes.parametersContainer` or `parametersContainerReadonly`) applied. And lastly, we want the literal `parametersContainerReadonly` applied if readonly so that we can use it in the global theme styles (where 'readOnly' state is not always available)

rholinshead:
Rebase

python/src/aiconfig/editor/client/src/components/GlobalParametersContainer.tsx
rossdanlm: Ahhhhh I understand thanks for explaining. I am familiar with his from React `styles={[]}` but didn't realize `classname` also did that, cool!

PR #1112, opened 2024-02-02
[ez] Export missing default model parsers

rossdanlm:
[ez] Export missing default model parsers

Now we can do `from aiconfig import ClaudeBedrockModelParser` for example. I also deleted the `hf.py` file since we now have `aiconfig_extension_hugging_face` for this

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1112).
* #1114
* __->__ #1112

### DIFF
--- a/python/src/aiconfig/Config.py
+++ b/python/src/aiconfig/Config.py
@@ -1,28 +1,27 @@
+from typing import Any, Dict, List, Literal, Optional, Tuple
+
 import json
 import os
-from typing import Any, Dict, List, Literal, Optional, Tuple
-from aiconfig.default_parsers.claude import ClaudeBedrockModelParser
-
 import requests
 import yaml
-from aiconfig.callback import CallbackEvent, CallbackManager
-from aiconfig.default_parsers.anyscale_endpoint import (
+
+from .callback import CallbackEvent, CallbackManager
+from .default_parsers.anyscale_endpoint import (
     DefaultAnyscaleEndpointParser,
 )
-from aiconfig.default_parsers.openai import DefaultOpenAIParser
-from aiconfig.default_parsers.gemini import GeminiModelParser
-from aiconfig.default_parsers.palm import PaLMChatParser, PaLMTextParser
-from aiconfig.model_parser import InferenceOptions, ModelParser
-
-from aiconfig.schema import JSONObject
-
+from .default_parsers.claude import ClaudeBedrockModelParser
 from .default_parsers.dalle import DalleImageGenerationParser
+from .default_parsers.openai import DefaultOpenAIParser
+from .default_parsers.gemini import GeminiModelParser
 from .default_parsers.hf import HuggingFaceTextGenerationParser
+from .default_parsers.palm import PaLMChatParser, PaLMTextParser
+from .model_parser import InferenceOptions, ModelParser
+
 from .registry import (
     ModelParserRegistry,
     update_model_parser_registry_with_config_runtime,
 )
-from .schema import AIConfig, Prompt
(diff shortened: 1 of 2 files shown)


python/src/aiconfig/Config.py
```diff
@@ -53,11 +52,14 @@
 ModelParserRegistry.register_model_parser(
     DefaultAnyscaleEndpointParser("AnyscaleEndpoint")
 )
+ModelParserRegistry.register_model_parser(
+    # TODO: Get a list of default models for AzureOpenAIParser
+    AzureOpenAIParser(deployment="gpt-35-turbo")
```
rholinshead, line 57: nit, maybe a comment explaining 35 is intentional (not 3.5) based on https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/models#gpt-35

python/src/aiconfig/Config.py
```diff
@@ -53,11 +52,14 @@
...
+)
 ModelParserRegistry.register_model_parser(
     GeminiModelParser("gemini-pro"), ["gemini-pro"]
 )
 ModelParserRegistry.register_model_parser(ClaudeBedrockModelParser())
-ModelParserRegistry.register_model_parser(HuggingFaceTextGenerationParser())
```
rholinshead, line 60: I think this will break the current HuggingFace/Mistral-aiconfig.json cookbook. Can you check that? Looks like it has an hf.py there but defines the parser as HuggingFaceTextParser -- probably just need to update the config there to use 'HuggingFaceTextParser' instead.

python/src/aiconfig/default_parsers/hf.py
rholinshead, line 1: Can you create an issue to do the same in ts?

rholinshead (changes requested):
Just requesting changes for the HF mistral cookbook

python/src/aiconfig/Config.py
Ankush-lastmile: I don't think we should not add this as a default model parser because that deployment id is specific to LastMile

python/src/aiconfig/Config.py
rossdanlm: Yea good point, I'll just not remove the main HF parser from this PR and do that in another one in the future. Issue: https://github.com/lastmile-ai/aiconfig/issues/1116

python/src/aiconfig/default_parsers/hf.py
rossdanlm: Task created in https://github.com/lastmile-ai/aiconfig/issues/1125

rossdanlm:
Based on comments:

1. Not removing core model parser for HF. Task for removing it in https://github.com/lastmile-ai/aiconfig/issues/1116
2. Not defining model parser for `AzureOpenAIParser`

python/src/aiconfig/Config.py
```diff
@@ -53,13 +52,13 @@
 ModelParserRegistry.register_model_parser(
     DefaultAnyscaleEndpointParser("AnyscaleEndpoint")
 )
+ModelParserRegistry.register_model_parser(ClaudeBedrockModelParser())
+for model in gpt_models_extra:
```
rholinshead, line 56: nit, would be better to have this colocate with the main registered ones above. Do you know why these are handled in a separate 'extra' array?

rholinshead (approved):

python/src/aiconfig/Config.py
Ankush-lastmile: this was a quick hack implemented in #994

PR #1114, opened 2024-02-02
Check model and model parser ids strings for spaces, strip them

rossdanlm:
Check model and model parser ids strings for spaces, strip them

Tanya noticed a bug when we accidentally added "field " with a space instead of "field", so trimming the value.

I originally tried to fix this in the editor config itself in `SettingsPropertyRenderer`, but that was complicated since it meant that this would reset the string value state for the string, and you couldn't keep typing empty spaces at the begininng or end of strings otherwise it would keep trimming them. Therefore, I decided to make this change in all places where we set the model names or fields for main SDK. This may seem a bit controversial (ex: you now can't have two separate models called "  my_model  " vs. "my_model" but I feel this is fine

Note: I changed both the getter and setter methods, so this will be a breaking change for existing AIConfigs that have leading or trailing whitespace in the (model)/(model parser) names

## Test Plan
```
parsers_path=./cookbooks/Gradio/aiconfig_model_registry.py
```
It now works with leading or trailing whitespace

https://github.com/lastmile-ai/aiconfig/assets/151060367/116b4ef4-b1f9-4a15-9d07-e1fb9d995245

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1114).
* __->__ #1114
* #1112


rholinshead:
> I originally tried to fix this in the editor config itself in SettingsPropertyRenderer, but that was complicated since it meant that this would reset the string value state for the string, and you couldn't keep typing empty spaces at the begininng or end of strings otherwise it would keep trimming them. 

Couldn't it just be done in `onUpdatePromptModel` in `AIConfigEditor`? Just trim the newModel before dispatch/request


> This may seem a bit controversial (ex: you now can't have two separate models called " my_model " vs. "my_model" but I feel this is fine

I do agree, I don't think leading/trailing whitespace should be part of the model name since that's asking for trouble in any other place that trims things.

python/src/aiconfig/schema.py
```diff
@@ -664,6 +668,20 @@ def update_prompt(self, prompt_name: str, prompt_data: Prompt):
...
+        """
+        if (prompt_data.metadata is not None and prompt_data.metadata.model is not None):
+            model = prompt_data.metadata.model
+            if isinstance(model, str):
+                model = model.strip()
+            else:
+                model.name = model.name.strip()
+                
+                if model.settings is not None and "model" in model.settings:
+                    model.settings["model"] = model.settings["model"].strip()
```
rholinshead, line 683: This needs to check if it's a string first, right?

python/src/aiconfig/schema.py
```diff
@@ -856,13 +877,14 @@ def _update_model_settings_for_prompt(
                 name=model_name, settings=settings
             )
         else:
+            if "model" in settings:
+                settings["model"] = settings["model"].strip()
```
rholinshead, line 881: Same here, needs string check

python/src/aiconfig/schema.py
```diff
@@ -884,6 +906,7 @@ def _update_model_for_aiconfig(
 `update_model` function with a specified `prompt_name` argument.
 """
         warnings.warn(warning_message)
+        model_name = model_name.strip()
```
rholinshead, line 909: If a config already has model name with leading or trailing space here, it can no longer have settings set by this?

rholinshead (commented):
I think the reasoning is right, but this is a lot of places where we're stripping the whitespace. Is there some minimum that we can do without needing to do it everywhere?

python/src/aiconfig/registry.py
```diff
@@ -61,7 +61,7 @@ def get_model_parser(model_id: str) -> ModelParser:
         Returns:
             ModelParser: The retrieved model parser
         """
-        return ModelParserRegistry._parsers[model_id]
+        return ModelParserRegistry._parsers[model_id.strip()]
```
rholinshead, line 64: For example, here is a "get" instead of set, so we probably don't need to change this?

python/src/aiconfig/schema.py
rossdanlm: You're right, thanks for the check

rossdanlm:
> Couldn't it just be done in `onUpdatePromptModel` in `AIConfigEditor`? Just trim the newModel before dispatch/request

It's a bit more complicated than this, because we get the model_name from the prompt when calling `config.run()` in https://github.com/lastmile-ai/aiconfig/blob/c571904edfafd878e60c81962a85b3f9f7de5125/python/src/aiconfig/Config.py#L322-L333

~~So in order to prevent breaking changes to existing AIConfigs which may have spaces already defined in that config and not trimmed, we need to continue using those ones~~. 

Ok nm, I realize that since we already strip the model name in `register_model_parser()`, which is what we call every time we do `config.create()` or `config.load()` then we're fine. 

However I still feel it's good to save it in the SDK so that this is centralized and not reliant on the editor client logic. Ex: if someone manually calls `config.update_model("  my_model  ", None, "prompt_name")` it will still work

python/src/aiconfig/registry.py
rossdanlm: The problem is that this function `get_model_parser()` is an external public call, so it's possible for someone to call into it with unsanitized trailing and leading whitespace

rossdanlm:
> Is there some minimum that we can do without needing to do it everywhere?

I think that for future nit, we can call `sanitize_model_name` before every centralized and public SDK method involving model_name. Created task for it in https://github.com/lastmile-ai/aiconfig/issues/1126

python/src/aiconfig/schema.py
rossdanlm: Since aiconfig-level model settings can have more than a single model, if we change the model name, we have no idea what the "previous" model name used to be so will error if we're only trying to pass in settings. Therefore, if you change the model name and don't pass in settings argument, it will just create a new model with empty settings. I wrote a comment about this in https://github.com/lastmile-ai/aiconfig/blob/43c7f9ea1592f1e00f812f2c75111475d4100f39/python/src/aiconfig/schema.py#L784-L788

python/src/aiconfig/registry.py
rholinshead: Doesn't that contradict the PR summary, though:
> Because we could have existing configs that have model names already, I only changed the setter methods, not the getter ones like get_model(), so this is not a breaking change

python/src/aiconfig/registry.py
```diff
@@ -74,7 +74,7 @@ def get_model_parser_for_prompt(prompt: Prompt, config: "AIConfigRuntime"):
         Returns:
             ModelParser: The retrieved model parser
         """
-        model_name = config.get_model_name(prompt)
+        model_name = config.get_model_name(prompt).strip()
```
rholinshead, line 77: I think we should just leave this as-is. If it's in the config already then we should have either sanitized it or it was before we started sanitizing it and stripping is going to make it inaccessible.

> Because we could have existing configs that have model names already, I only changed the setter methods, not the getter ones like get_model(), so this is not a breaking change

python/src/aiconfig/registry.py
rossdanlm: Yea you're right, let me re-word the the summary.

python/src/aiconfig/registry.py
rossdanlm: If use does something like this:
```
model_name = "  my_model  "
config.update_model(model_name, None)
# does some stuff...

config.get_model_parser_for_prompt(model_name)
```

This will fail if we don't do the strip, so I think it's important to keep it. I'll re-word the summary

rossdanlm: Made an issue to surface unfound model/model parser names in https://github.com/lastmile-ai/aiconfig/issues/1136 similar to what we have for prompt error message

rholinshead: Can you explain why this will fail? With the other changes in this PR, config.get_model_name(prompt) should not have any extra whitespace already, so no need to strip it, right?

python/src/aiconfig/schema.py
```diff
@@ -664,6 +668,22 @@ def update_prompt(self, prompt_name: str, prompt_data: Prompt):
...
+            model = prompt_data.metadata.model
+            if isinstance(model, str):
+                model = model.strip()
+            else:
+                model.name = model.name.strip()
+
+                if model.settings is not None and "model" in model.settings:
+                    # TODO: Support cases where "model" field under model settings isn't just a str, since it technically can be anything depending on the model parser
+                    if isinstance(model.settings["model"], str):
+                        model.settings["model"] = model.settings["model"].strip()
```
rholinshead, line 685: Was thinking on this some more. Let's not do this specific change. My reasoning is:
- parser/schema level should has no idea what this 'model' means. It's an implementation detail of the parser
- similarly, it's the parser's responsibility to use the 'model' from the settings as the 'model' for the inference run (that logic is in the parser, not the schema/core sdk)

So, if we want to strip the settings model it should be done when pulling it from the settings in the relevant parsers

python/src/aiconfig/schema.py
```diff
@@ -856,13 +879,16 @@ def _update_model_settings_for_prompt(
                 name=model_name, settings=settings
             )
         else:
+            if "model" in settings:
+                # TODO: Support cases where "model" field under model settings isn't just a str, since it technically can be anything depending on the model parser
```
rholinshead, line 883: Same here -- this is parser-specific logic

rholinshead (changes requested):
Ok, I think most places look good. Requesting changes because I don't think we should be touching 'model' value in the inference settings from the core/sdk code (it's parser implementation detail).
Also still not sure the strip is needed in `get_model_parser_for_prompt` but might just not be understanding your example

PR #1115, opened 2024-02-02
[ez] Autoformatting

rossdanlm:
[ez] Autoformatting

Ran `fd --glob '*.py' python/src/aiconfig/ | xargs python -m 'scripts.lint' --mode=fix --files`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1115).
* __->__ #1115
* #1114
* #1112


rholinshead (approved):
Is there some extension or something we each need to have set up locally to make sure these formatting changes happen on save? Not sure why we need to keep manually running the formatter

PR #1117, opened 2024-02-02
[gradio] Fix readonly RunPromptButton Margin

rholinshead:
[gradio] Fix readonly RunPromptButton Margin

# [gradio] Fix readonly RunPromptButton Margin

Pretty annoying, but we have custom margin for the runPromptButton in gradio to account for the Prompt label being shown (by our gradio styles). But in readOnly we don't show the label, but kept the same margin. This ends up looking like this:

Screenshot 2024-02-02 at 3 03 45 PM

To fix, add .runPromptButtonReadOnly class to the button in readOnly and use it to set the margin in the theme:

Screenshot 2024-02-02 at 3 22 05 PM
Screenshot 2024-02-02 at 3 23 58 PM

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1117).
* #1118
* __->__ #1117

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/prompt/RunPromptButton.tsx
+++ b/python/src/aiconfig/editor/client/src/components/prompt/RunPromptButton.tsx
@@ -33,7 +33,7 @@ export default memo(function RunPromptButton({
       disabled={disabledOrReadOnly}
       p="xs"
       size="xs"
-      className="runPromptButton"
+      className={`runPromptButton ${readOnly ? "runPromptButtonReadOnly" : ""}`}
     >
       {isRunning ? (
         <Flex align="center" justify="center">
--- a/python/src/aiconfig/editor/client/src/themes/GradioTheme.ts
+++ b/python/src/aiconfig/editor/client/src/themes/GradioTheme.ts
@@ -236,6 +236,10 @@ export const GRADIO_THEME: MantineThemeOverride = {
           height: "auto",
         },
 
+        ".runPromptButton.runPromptButtonReadOnly": {
+          marginTop: "13px",
+        },
+
         ".actionTabsPanel": {
           width: "400px",
         },


python/src/aiconfig/editor/client/src/components/prompt/RunPromptButton.tsx
```diff
@@ -33,7 +33,7 @@ export default memo(function RunPromptButton({
       disabled={disabledOrReadOnly}
       p="xs"
       size="xs"
-      className="runPromptButton"
+      className={`runPromptButton ${readOnly ? "runPromptButtonReadOnly" : ""}`}
```
rossdanlm, line 36: Ok sorry just to be clear, it's specifically this line that causes the prompt button to be vertically centered instead of at the top right? https://github.com/lastmile-ai/aiconfig/blob/c571904edfafd878e60c81962a85b3f9f7de5125/python/src/aiconfig/editor/client/src/themes/GradioTheme.ts#L234

And similar to React where we can overwrite components with spreading ex: 
```
{
  ...stuff_here 
  margin: <overrideMarginValue>
}
```

By declaring `.runPromptButton.runPromptButtonReadOnly` second it will override the previous `marginTop` value defined in `runPromptButton`?

rossdanlm (approved):

python/src/aiconfig/editor/client/src/components/prompt/RunPromptButton.tsx
rholinshead: Yes, specifically the 33px.

> By declaring .runPromptButton.runPromptButtonReadOnly second it will override the previous marginTop value defined in runPromptButton

Yes and no -- it will override the value, but not due to the ordering but because of the specificity: it applies the styles to any element with both `.runPromptButton` AND `.runPromptButtonReadOnly` classes, whereas the `runPromptButton` selector styles are only applied to an element with that class. Since the button has both classes, the selector will override any conflicting styles

PR #1123, opened 2024-02-02
[editor] Don't Try to Save if No Save Callback

rholinshead:
# [editor] Don't Try to Save if No Save Callback

If no `save` callback is provided to the editor we shouldn't have the interval set up to autosave, and should not override the default CMD+S listener.

## Testing:
- Ensure CMD + S and autosave still work correctly
- Remove `save` from the callbacks passed to the editor and ensure CMD+S opens the default browser save again:
Screenshot 2024-02-02 at 5 42 33 PM
- Add console log in the effect to ensure no autosave is attempted


rossdanlm (approved):
I'm mad I didn't notice this earlier

PR #1124, opened 2024-02-02
[editor][ez] Fix AddPromptButton Width

rholinshead:
[editor][ez] Fix AddPromptButton Width

# [editor][ez] Fix AddPromptButton Width

Fix the AddPromptButton so that it no longer extends beyond the left of the prompts. Requires both adding left margin (to offset its left side to match start of the prompt cells) and removing the 100% width (to prevent the right side from extending outside of the container right padding). The width is still the full width of the prompts without explicitly setting width

## Before
Screenshot 2024-02-02 at 5 59 23 PM

## After
Screenshot 2024-02-02 at 5 57 53 PM

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/prompt/AddPromptButton.tsx
+++ b/python/src/aiconfig/editor/client/src/components/prompt/AddPromptButton.tsx
@@ -10,6 +10,7 @@ import {
 import { IconPlus, IconSearch, IconTextCaption } from "@tabler/icons-react";
 import { memo, useCallback, useState } from "react";
 import useLoadModels from "../../hooks/useLoadModels";
+import { PROMPT_CELL_LEFT_MARGIN_PX } from "../../utils/constants";
 
 type Props = {
   addPrompt: (prompt: string) => void;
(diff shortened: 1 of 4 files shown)


python/src/aiconfig/editor/client/src/components/prompt/AddPromptButton.tsx
```diff
@@ -21,8 +21,8 @@ const useStyles = createStyles((theme) => ({
     borderRadius: rem(4),
     display: "flex",
     justifyContent: "center",
+    marginLeft: "36px",
```
rossdanlm, line 24: nit: Btw why 36px in particular? If this is used multiple places, should this be defined as a STYLE constant somewhere?

rossdanlm (approved):

python/src/aiconfig/editor/client/src/components/prompt/AddPromptButton.tsx
rholinshead: I'm not sure exactly what is making it add up to that -- the cell menu button padding  icon is what pushes the prompt cell over to require that alignment from the other components. Good call on using a const - will do

rholinshead:
Updated to use a const, PROMPT_CELL_LEFT_MARGIN_PX

PR #1127, opened 2024-02-03
[editor][1/4] Set Up Notification Provider/Context

rholinshead:
[editor][1/4] Set Up Notification Provider/Context

[editor][1/4] Set Up Notification Provider/Context

# [editor][notifications][1/4] Set Up Notification Provider/Context

For the vs code extension, we want to leverage the vs code notification UX instead of mantine for showing notifications. To support that in a usable way, this PR stack sets up a Notification context which will propagate provided `showNotification` function to any components that request it from context. If there is no `showNotification` provided to the `AIConfigEditor` component we will default to the mantine `showNotification` function as usual.

This PR sets up the context, managing which `showNotification` function to call, and providing it through the context provider. The function takes in an `AIConfigEditorNotification` object, which includes the properties of mantine's `NotificationProps` type that we are currently using and which may be desired to use in the future (e.g. autoClose, onOpen, onClose). The remaining props are pretty UI-focused (e.g. color, radius, icon, etc.) and not really useful for external contexts.

## Testing:
- No-op here. Make sure notifications still work as usual

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1127).
* #1130
* #1129
* #1128
* __->__ #1127

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/notifications/NotificationContext.tsx
+++ b/python/src/aiconfig/editor/client/src/components/notifications/NotificationContext.tsx
@@ -1,0 +1,16 @@
+import { createContext } from "react";
+import { AIConfigEditorNotification } from "./NotificationProvider";
+
+/**
+ * Context for providing showNotification method throughout the editor. For example,
+ * for VS Code extension we want to re-use VS Code's notification framework instead of
+ * using the default (mantine) -- set in context with the `showNotification` callback
+ * provided to AIConfigEditor.
+ */
+const NotificationContext = createContext<{
+  showNotification: (notification: AIConfigEditorNotification) => void;
+}>({
+  showNotification: () => {},
+});
+
+export default NotificationContext;
--- a/python/src/aiconfig/editor/client/src/components/notifications/NotificationProvider.tsx
+++ b/python/src/aiconfig/editor/client/src/components/notifications/NotificationProvider.tsx
@@ -1,0 +1,61 @@
+import {
+  Notifications,
+  showNotification as mantineShowNotification,
+} from "@mantine/notifications";
+import { useCallback, useMemo } from "react";
+import NotificationContext from "./NotificationContext";
+
+export type AIConfigEditorNotificationType =
+  | "info"
+  | "success"
+  | "warning"
+  | "error";
+
+export type AIConfigEditorNotification = {
+  title: string;
+  message: string | null;
+  type?: AIConfigEditorNotificationType;
+  autoClose?: boolean | number;
+  onClose?: () => void;
+  onOpen?: () => void;
+};
+
+type Props = {
+  children: React.ReactNode;
+  showNotification?: (notification: AIConfigEditorNotification) => void;
+};
+
+const NOTIFICATION_TYPE_COLOR = {
+  info: "blue",
+  success: "green",
+  warning: "yellow",
+  error: "red",
+};
+
+export default function NotificationProvider({
+  children,
+  showNotification: overrideShowNotification,
+}: Props) {
+  const defaultShowNotification = useCallback(
+    (notification: AIConfigEditorNotification) =>
+      mantineShowNotification({
+        ...notification,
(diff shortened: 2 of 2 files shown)


python/src/aiconfig/editor/client/src/components/notifications/NotificationContext.tsx
```diff
@@ -0,0 +1,13 @@
+import { createContext } from "react";
+import { AIConfigEditorNotification } from "./NotificationProvider";
+
+/**
+ * Context for providing showNotification method throughout the editor
+ */
+const NotificationContext = createContext<{
```
rossdanlm, line 7: Nit: can we put these files in the `/context` folder since we're not really making 'components': https://github.com/lastmile-ai/aiconfig/tree/10da93839d85f4dee2ccd5428c9bc10104eb2246/python/src/aiconfig/editor/client/src/contexts

python/src/aiconfig/editor/client/src/components/notifications/NotificationContext.tsx
```diff
@@ -0,0 +1,13 @@
+import { createContext } from "react";
+import { AIConfigEditorNotification } from "./NotificationProvider";
+
+/**
+ * Context for providing showNotification method throughout the editor
```
rossdanlm, line 5: Let's add a bit more detail in code. Ex: "For example, for VS Code extension we want to re-use VS Code's notification framework instead of using the one provided by Mantine"

python/src/aiconfig/editor/client/src/components/notifications/NotificationProvider.tsx
```diff
@@ -0,0 +1,61 @@
...
+      showNotification: overrideShowNotification ?? defaultShowNotification,
+    }),
+    [defaultShowNotification, overrideShowNotification]
+  );
+
+  return (
+    <NotificationContext.Provider value={notificationContext}>
+      {!overrideShowNotification && <Notifications />}
+      {children}
+    </NotificationContext.Provider>
```
rossdanlm, line 59: This hasn't been connected to rest of the code yet, so it's not super clear to me in this diff that it'll work as is, but that's fine I'll look at the other PRs in stack

python/src/aiconfig/editor/client/src/components/notifications/NotificationContext.tsx
rholinshead: Hmm, I have it colocated here under /notifications folder so it's right beside the provider. I think that makes more sense than having it in a separate folder a few levels away. Also if it's not here, this folder just has one file, which is kinda lonely :(

python/src/aiconfig/editor/client/src/components/notifications/NotificationProvider.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const notificationContext = useMemo(
+    () => ({
+      showNotification: overrideShowNotification ?? defaultShowNotification,
+    }),
+    [defaultShowNotification, overrideShowNotification]
+  );
+
+  return (
+    <NotificationContext.Provider value={notificationContext}>
+      {!overrideShowNotification && <Notifications />}
```
rossdanlm, line 57: Why do we need this line? Is this mantine's way of initializing Notification framework? Can we add comment in code?

rossdanlm (approved):

python/src/aiconfig/editor/client/src/components/notifications/NotificationProvider.tsx
rholinshead: <Notifications /> is the mantine component for showing notifications (imported here, just moved from AIConfigEditor here). I can add the comment, but feels a bit verbose

rossdanlm: Yea no worries, once I see how it was connected in the next PR it made sense to me

PR #1128, opened 2024-02-03
[editor][2/4] Wrap AIConfigEditor with NotificationProvider

rholinshead:
[editor][2/4] Wrap AIConfigEditor with NotificationProvider

# [editor][notifications][2/4] Wrap AIConfigEditor with NotificationProvider

Wrap the `AIConfigEditor` component core with the `NotificationProvider`, wrapping both with the ThemeProvider to ensure notifications also have the proper theme applied. With the `NotificationProvider` in place, we also remove the mantine `Notifications` component from the AIConfigEditor (it's in the NotificationProvider now).

## Testing:
Ensure notifications still work

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1128).
* #1130
* #1129
* __->__ #1128
* #1127

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
+++ b/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
@@ -7,7 +7,7 @@ import {
   Alert,
   Group,
 } from "@mantine/core";
-import { Notifications, showNotification } from "@mantine/notifications";
+import { showNotification } from "@mantine/notifications";
 import {
   AIConfig,
   InferenceSettings,
@@ -60,6 +60,7 @@ import AIConfigEditorThemeProvider from "../themes/AIConfigEditorThemeProvider";
 import DownloadButton from "./global/DownloadButton";
 import ShareButton from "./global/ShareButton";
 import PromptsContainer from "./prompt/PromptsContainer";
+import NotificationProvider from "./notifications/NotificationProvider";
 
 type Props = {
   aiconfig: AIConfig;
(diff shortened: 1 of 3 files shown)


python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
```diff
@@ -954,113 +954,117 @@ export default function AIConfigEditor({
...
+        />
+      </Container>
+    </AIConfigContext.Provider>
+  );
+}
+
+// Wrap the AIConfigEditor in the NotificationProvider to provide NotificationContext
+// to the AIConfigEditor. Wrap NotificationProvider and AIConfigEditor with the
+// theme provider to ensure all components have the proper theme
+export default function AIConfigEditorWrapper(props: Props) {
```
rossdanlm, line 1062: Nit: Would prefer to rename this as `AIConfigEditor` so it's more clear this is what is referred to when doing default imports. I get that with default we can call it anything, but for tracing code I think it would be more clear to have this with that name. My proposal would be:
1. Remove `<AIConfigContext.Provider>` tag from current `AIConfigEditor`, put it within current `AIConfigEditorWrapper` component instead
2. Rename `AIConfigEditor` to something like `AIConfigEditorBase` or `AIConfigEditorWithoutContexts` 
3. Rename `AIConfigEditorWrapper` or to `AIConfigEditor`

rossdanlm (approved):

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
rholinshead: I can rename. I can't move the AIConfigContext.Provider since most of the context values (callbacks) use `showNotification` from context and need to be a child of the notification context provider, which is why the components are separated here

rossdanlm: Ah ok makes sense, but yea let's rename, even if it's a bit bulky

rholinshead: Will do. Long-term, we can move all of the context / callback code in the base component into their own hooks to reduce the size of this file drastically. But holding offf on that for now since it'll be a pain to deal with for the packaging stuff

PR #1129, opened 2024-02-03
[editor][3/4] Update showNotification Calls to Use Context

rholinshead:
[editor][3/4] Update showNotification Calls to Use Context

# [editor][3/4] Update showNotification Calls to Use Context

Update all existing `showNotification` calls to use the `showNotification` from our new `NotificationContext`. By default, this is the same mantine `showNotification` implementation, just with our own mapping of notification type to mantine notification color.

## Testing
- Ensure notifications still work
- Added onClose, onOpen, and autoClose: false to a notification to make sure they work as expected

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1129).
* #1130
* __->__ #1129
* #1128
* #1127

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
+++ b/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
@@ -7,7 +7,6 @@ import {
   Alert,
   Group,
 } from "@mantine/core";
-import { showNotification } from "@mantine/notifications";
 import {
   AIConfig,
   InferenceSettings,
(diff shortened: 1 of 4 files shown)


rossdanlm (approved):
We'll probably need to update the Gradio callback handlers too right? Actually I guess not because it'll just automatically use the default provided by mantine right?

rholinshead:
> We'll probably need to update the Gradio callback handlers too right? Actually I guess not because it'll just automatically use the default provided by mantine right?

Ya, it'll just use the default

PR #1131, opened 2024-02-05
Add hoverContainer class to ConfigNameDescription

rholinshead:
Add hoverContainer class to ConfigNameDescription

# Add hoverContainer class to ConfigNameDescription

In https://github.com/lastmile-ai/aiconfig/pull/1121/files, the local hoverContainer class is modified to have vs code-specific styles. Instead, we can add the literal `hoverContainer` class to the elements so that our vscode theme can override the styles with greater specificity

## Testing:
- Make sure existing styles still work
- See class added:
Screenshot 2024-02-05 at 11 25 09 AM

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1131).
* #1134
* #1133
* #1132
* __->__ #1131

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/ConfigNameDescription.tsx
+++ b/python/src/aiconfig/editor/client/src/components/ConfigNameDescription.tsx
@@ -128,7 +128,9 @@ export default memo(function ConfigNameDescription({
           <Title
             ref={nameDisplayRef}
             onClick={onClickEdit}
-            className={!readOnly ? classes.hoverContainer : undefined}
+            className={
+              !readOnly ? `${classes.hoverContainer} hoverContainer` : undefined
+            }
           >
             {name}
           </Title>
@@ -136,7 +138,11 @@ export default memo(function ConfigNameDescription({
             <div
               ref={descriptionDisplayRef}
               onClick={onClickEdit}
-              className={!readOnly ? classes.hoverContainer : undefined}
+              className={
+                !readOnly
+                  ? `${classes.hoverContainer} hoverContainer`
+                  : undefined
+              }
             >
               <TextRenderer content={description} />
             </div>


python/src/aiconfig/editor/client/src/components/ConfigNameDescription.tsx
```diff
@@ -128,15 +128,21 @@ export default memo(function ConfigNameDescription({
           <Title
             ref={nameDisplayRef}
             onClick={onClickEdit}
-            className={!readOnly ? classes.hoverContainer : undefined}
+            className={
+              !readOnly ? `${classes.hoverContainer} hoverContainer` : undefined
```
Ankush-lastmile, line 132: this line effectively adds two different class names right?

what happens if they both exist? how is the specificity chosen?

python/src/aiconfig/editor/client/src/components/ConfigNameDescription.tsx
```diff
@@ -128,15 +128,21 @@ export default memo(function ConfigNameDescription({
...
           {description && (
             <div
               ref={descriptionDisplayRef}
               onClick={onClickEdit}
-              className={!readOnly ? classes.hoverContainer : undefined}
+              className={
+                !readOnly
+                  ? `${classes.hoverContainer} hoverContainer`
+                  : undefined
+              }
```
Ankush-lastmile, line 145: not related to this diff but curious why we set the class name to undefined in readonly mode

rholinshead: Just from the original implementation, when in readonly mode it doesn't need any extra styling since it should just render as regular text without any hover styles or border shown.

Ankush-lastmile (approved):
lgtm, some questions

python/src/aiconfig/editor/client/src/components/ConfigNameDescription.tsx
rholinshead: Correct. The `classes.hoverContainer` is pre-processed by mantine to a dynamic class that serializes the above styles into something like
```
.mantine-<randomid> {
  ... hoverContainer styles above
}
```

And then we add `hoverContainer`as well.

Both styles will be applied but we can ensure global `hoverContainer` has higher specificity since it should be nested under the `div.editorBackground` style as well

PR #1135, opened 2024-02-05
[editor] Support Running Prompt with Shift or CTRL + Enter

rholinshead:
[editor] Support Running Prompt with Shift or CTRL + Enter

# [editor] Support Running Prompt with Shift or CTRL + Enter

From https://github.com/lastmile-ai/aiconfig/pull/1121/files (and also general editor feedback), add support for executing text-input prompts with Shift+Enter (or CTRL+Enter)

https://github.com/lastmile-ai/aiconfig/assets/5060851/e103bf62-3edf-4a03-b2a8-09a7c1b1c9ce

Update: Show notification if trying to run a prompt while a prompt is already running:
Screenshot 2024-02-05 at 6 50 56 PM
Screenshot 2024-02-05 at 6 51 11 PM

## Testing:
- Ensure execution via button (and cancelation) work
- Ensure newlines can be added with enter (no shift/ctrl modifier)
- Ensure shift + enter and ctrl + enter run the prompt; ensure they do not run it when it's already running or when readOnly is true (though can't even edit then anyway)
- Ensure the same for a prompt with no schema (hf facebook/mms-tts-eng)

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/prompt/prompt_input/PromptInputRenderer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/prompt/prompt_input/PromptInputRenderer.tsx
@@ -1,5 +1,5 @@
 import { PromptInput } from "aiconfig";
-import { memo, useState } from "react";
+import { memo, useCallback, useContext, useState } from "react";
 import { PromptInputSchema } from "../../../utils/promptUtils";
 import PromptInputSchemaRenderer from "./schema_renderer/PromptInputSchemaRenderer";
 import PromptInputConfigRenderer from "./PromptInputConfigRenderer";
@@ -10,6 +10,7 @@ import { Text } from "@mantine/core";
 import JSONRenderer from "../../JSONRenderer";
 import JSONEditorToggleButton from "../../JSONEditorToggleButton";
 import RunPromptButton from "../RunPromptButton";
+import NotificationContext from "../../notifications/NotificationContext";
 
 type Props = {
   input: PromptInput;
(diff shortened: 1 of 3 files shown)


python/src/aiconfig/editor/client/src/components/prompt/prompt_input/PromptInputRenderer.tsx
```diff
@@ -91,14 +91,20 @@ export default memo(function PromptInputRenderer({
     </Flex>
   );
 
+  const runPrompt = useCallback(async () => {
+    if (!(isRunButtonDisabled || isRunning)) {
```
rossdanlm, line 95: Ok, just to confirm, is the only case where this is possible to enter here is if the prompt (or another prompt) is running and as it's running, user decides to change the input? Otherwise for read-only mode they won't be able to change this anyways right

python/src/aiconfig/editor/client/src/components/prompt/prompt_input/schema_renderer/PromptInputSchemaRenderer.tsx
```diff
@@ -95,6 +96,12 @@ export default memo(function PromptInputSchemaRenderer(props: Props) {
         value={props.input}
         label="Prompt"
         onChange={(e) => props.onChangeInput(e.target.value)}
+        onKeyDown={(event) => {
+          if (event.key === "Enter" && (event.shiftKey || event.ctrlKey)) {
```
rossdanlm, line 100: So does the ordering here matter? Should we instead be doing:

```
(event.shiftKey || event.ctrlKey) && event.key === "Enter"
```

so that "Enter" is the 2nd keyboard button that is pressed?

python/src/aiconfig/editor/client/src/components/prompt/prompt_input/PromptInputRenderer.tsx
rossdanlm: Also nit: I think it's slightly easier to exit early on the true condition
```
if (isRunButtonDisabled || isRunning) {
  // Do not let us invoke run operation when we're supposed to
  return;
}
await onRunPrompt();
```

rossdanlm (approved):
lgtm, just a few nits

python/src/aiconfig/editor/client/src/components/prompt/prompt_input/schema_renderer/PromptInputSchemaRenderer.tsx
rholinshead: Ordering doesn't matter -- it's one event that has both buttons pressed simultaneously (the event with just shift/ctrl key already happened and is ignored/not handled)

python/src/aiconfig/editor/client/src/components/prompt/prompt_input/PromptInputRenderer.tsx
rholinshead: Correct

rholinshead:
Changes from review: updated to show notifications for the two scenarios -- if current prompt is currently running, or if another prompt is currently running

rholinshead:
Screenshot 2024-02-05 at 6 50 56 PM
Screenshot 2024-02-05 at 6 51 11 PM

PR #1138, opened 2024-02-05
[ez] Fix dall-e-2 and dall-e-3 Prompt Schemas

rholinshead:
# [ez] Fix dall-e-2 and dall-e-3 Prompt Schemas

Update the prompt schemas to reflect the differing options between dall-e-2 and dall-e-3, per https://platform.openai.com/docs/api-reference/images/create

Screenshot 2024-02-06 at 10 23 43 AM
Screenshot 2024-02-06 at 10 28 47 AM

Made sure both models work with each option set to different values and with no settings specified


Ankush-lastmile:
Joe really loves cats

Ankush-lastmile (approved):
lgtm

PR #1139, opened 2024-02-05
[easy] Set Gemini model parser id to the model id being used

Ankush-lastmile:
[easy] Set Gemini model parser id to the model id being used

title

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1139).
* __->__ #1139
* #1137

### DIFF
--- a/python/src/aiconfig/default_parsers/gemini.py
+++ b/python/src/aiconfig/default_parsers/gemini.py
@@ -104,6 +104,7 @@ class GeminiModelParser(ParameterizedModelParser):
         super().__init__()
         # TODO(?): @Ankush-lastmile Figure out how to make this model parser impl and init not depend on the model
         self.model = genai.GenerativeModel(id)
+        self.model_id = id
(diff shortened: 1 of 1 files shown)


python/src/aiconfig/default_parsers/gemini.py
```diff
@@ -329,6 +330,10 @@ async def run_inference(
 
         if not self.api_key:
             self.api_key = get_api_key_from_environment("GOOGLE_API_KEY", required=True).unwrap()
+
+        # Gemini api stores a reference to the currently executing async event loop.
```
saqadri, line 334: This causes issues
```suggestion
        # Gemini api stores a reference to the currently executing async event loop.
        # This causes issues on reruns for the model parser, so to alleviate this, we
```

saqadri (approved):

PR #1141, opened 2024-02-06
Add additional commands for vscode extension

saqadri:
Add additional commands for vscode extension

* Create New AIConfig (JSON and YAML)
* Share
* Create model parser registry file
* Register model parser registry file path

As part of this I had to create an AIConfigEditorManager class that keeps track of which editors are currently open. This allows us to use that context when a user selects a "Share" command (for e.g.), so we know which context they are running that command in.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1141).
* #1142
* __->__ #1141
* #1140


vscode-extension/src/aiConfigEditor.ts
```diff
@@ -253,50 +270,34 @@ export class AIConfigEditorProvider implements vscode.CustomTextEditorProvider {
...
   private async startEditorServer(
     document: vscode.TextDocument
   ): Promise<ServerInfo> {
     this.extensionOutputChannel.info(
       this.prependMessage("Starting editor server", document)
     );
 
+    // If there is a custom model registry path, pass it to the server
+    let config = vscode.workspace.getConfiguration(EXTENSION_NAME);
+    let modelRegistryPath = config.get<string>("modelRegistryPath");
```
rholinshead, line 288: nit, `const`

vscode-extension/src/aiConfigEditorManager.ts
```diff
@@ -0,0 +1,83 @@
...
+  private editorsByUri: Map<string, AIConfigEditorState> = new Map();
+
+  constructor() {
+    this.activeEditor = null;
+    this.activeEditorUri = null;
+  }
+
+  setActiveEditor(editor: AIConfigEditorState | null) {
+    this.activeEditor = editor;
+    this.activeEditorUri = editor?.document.uri.toString();
```
rholinshead, line 39: Can this variable become outdated if changing the filename while it's open? Just wondering if we should be getting the uri directly / at time-of-need to ensure it's always up-to-date instead of setting it in a variable

vscode-extension/src/aiConfigEditorManager.ts
```diff
@@ -0,0 +1,83 @@
...
+  addEditor(editor: AIConfigEditorState, uri?: string) {
+    this.editorsByUri.set(uri ?? editor.document.uri.toString(), editor);
+    this.setActiveEditor(editor);
+    if (uri) {
+      this.activeEditorUri = uri;
+    }
+  }
+
+  getEditor(document: vscode.TextDocument) {
+    return this.getEditorByUri(document.uri.toString());
```
rholinshead, line 59: Ah, I guess we'd also run into issues with this mapping if that's the case

vscode-extension/src/aiConfigEditorManager.ts
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  getEditorByUri(uri: string) {
+    return this.editorsByUri.get(uri);
+  }
+
+  removeEditor(editor: AIConfigEditorState) {
+    this.editorsByUri.delete(editor.document.uri.toString());
+    if (this.activeEditor === editor) {
+      this.activeEditor = null;
```
rholinshead, line 69: Shouldn't this set activeEditorUri to null as well?

vscode-extension/src/extension.ts
```diff
@@ -26,14 +39,82 @@ export function activate(context: vscode.ExtensionContext) {
   });
   context.subscriptions.push(setupCommand);
 
+  let createAIConfigJSONCommand = vscode.commands.registerCommand(
```
rholinshead, line 42: nit, const for all of these commands

vscode-extension/src/extension.ts
```diff
@@ -26,14 +39,82 @@ export function activate(context: vscode.ExtensionContext) {
...
+
+  // Also handle file renames/moves -- inform the EditorManager of the change
+  context.subscriptions.push(
+    vscode.workspace.onDidRenameFiles((e) => {
+      e.files.forEach(async (file) => {
+        const editor = aiconfigEditorManager.getEditorByUri(
+          file.oldUri.toString()
+        );
+        if (editor) {
+          aiconfigEditorManager.removeEditorByUri(file.oldUri.toString());
```
rholinshead, line 113: Oh, nice, already a step ahead :p

vscode-extension/src/extension.ts
```diff
@@ -26,14 +39,82 @@ export function activate(context: vscode.ExtensionContext) {
...
+          aiconfigEditorManager.addEditor(editor, file.newUri.toString());
```
rholinshead, line 114: If this is renaming a file whose editor is not the active one in view, won't this incorrectly make that new editor active?
Maybe we should move this into a `replaceEditor` method in the manager which also handles setting the proper active editor if the rename happens on the active one?

vscode-extension/src/extension.ts
```diff
@@ -42,6 +123,242 @@ export function deactivate() {
...
+) {
+  const workspaceUri = vscode.workspace.workspaceFolders
+    ? vscode.workspace.workspaceFolders[0].uri
+    : null;
+  const untitledUri = workspaceUri
+    ? workspaceUri.with({
+        scheme: "untitled",
+        path: `${workspaceUri.path}/untitled.aiconfig.${mode}`,
+      })
+    : vscode.Uri.parse(`untitled:untitled.aiconfig.${mode}`);
```
rholinshead, line 142: What's the `untitled:` prefix for?

vscode-extension/src/extension.ts
```diff
@@ -42,6 +123,242 @@ export function deactivate() {
...
+  aiconfigEditorManager: AIConfigEditorManager,
+  mode: "json" | "yaml" = "json"
+) {
+  const workspaceUri = vscode.workspace.workspaceFolders
+    ? vscode.workspace.workspaceFolders[0].uri
+    : null;
+  const untitledUri = workspaceUri
+    ? workspaceUri.with({
+        scheme: "untitled",
+        path: `${workspaceUri.path}/untitled.aiconfig.${mode}`,
```
rholinshead, line 140: Do we need to worry about handling duplicate names?

vscode-extension/src/extension.ts
```diff
@@ -42,6 +123,242 @@ export function deactivate() {
...
+      })
+    : vscode.Uri.parse(`untitled:untitled.aiconfig.${mode}`);
+
+  // Specify the initial content here
+  const newAIConfigJSON = vscode.Uri.joinPath(
```
rholinshead, line 145: Should this use untitledUri instead? I see the only other line using untitledUri below is commented out

vscode-extension/src/extension.ts
```diff
@@ -42,6 +123,242 @@ export function deactivate() {
...
+  const newAIConfigYAML = vscode.Uri.joinPath(
+    context.extensionUri,
+    "static",
+    "untitled.aiconfig.yaml"
+  );
+
+  const fileContentPath = mode === "json" ? newAIConfigJSON : newAIConfigYAML;
+
+  const fileContentBuffer = await vscode.workspace.fs.readFile(fileContentPath);
+  const initialContent = fileContentBuffer.toString();
```
rholinshead, line 160: Do we just let the server initialize with empty config json?

vscode-extension/src/extension.ts
```diff
@@ -42,6 +123,242 @@ export function deactivate() {
...
+
+/**
+ * Opens the currently registered custom model registry file in the editor.
+ * If none is found, prompts the user to create a new one or use a pre-existing one.
+ */
+async function openModelRegistry(
+  context: vscode.ExtensionContext,
+  aiconfigEditorManager: AIConfigEditorManager
+) {
+  let config = vscode.workspace.getConfiguration(EXTENSION_NAME);
```
rholinshead, line 188: nit, const for all the lets not redeclared here

vscode-extension/src/extension.ts
rholinshead: Ah, wait, I realize now that the above is loading default contents from untitled.aiconfig files, right? Could we rename those files to make it clearer they're the default templates for new config files?

vscode-extension/src/extension.ts
```diff
@@ -42,6 +123,242 @@ export function deactivate() {
...
+    },
+  });
+}
+
+async function handleCustomModelRegistryUpdate(
+  config: vscode.WorkspaceConfiguration,
+  aiconfigEditorManager: AIConfigEditorManager,
+  modelRegistryPath: string
+) {
+  // TODO: saqadri - ask the user if they want us to apply the setting globally
```
rholinshead, line 332: +1, this will be important since I'd expect my configs to use the registered parsers in the same directory as them instead of using some global one that is registered once for use in a config

vscode-extension/src/extension.ts
```diff
@@ -42,6 +123,242 @@ export function deactivate() {
...
+  for (const editor of aiconfigEditors) {
+    if (editor.editorServer) {
+      promises.push(
+        updateModelRegistryPath(editor.editorServer.url, modelRegistryPath)
+      );
+    }
+  }
+
+  // TODO: saqadri - this by itself isn't enough -- we also need to reload the aiconfigs
+  await Promise.allSettled(promises);
```
rholinshead, line 355: Also will need to handle errors if any of them weren't updated

vscode-extension/src/extension.ts
```diff
@@ -42,6 +123,242 @@ export function deactivate() {
...
+
+  vscode.window.showInformationMessage(
+    `Notified open aiconfigs of model registry ${modelRegistryPath}`
```
rholinshead, line 358: nit, `Updated model registry for open aiconfigs to ${modelRegistryPath}`

vscode-extension/src/extension.ts
```diff
@@ -325,3 +642,92 @@ async function checkPip() {
...
+    secretAccessKey: "",
+  });
+  const s3 = new AWS.S3();
+
+  const randomPath = Math.round(Math.random() * 10000);
+  const bucket: string = "lastmileai.aiconfig.public";
+  const uploadKey: string = `aiconfigs/${getTodayDateString()}/${randomPath}/${sanitizedFileName}`;
+  const configString = activeEditor.document.getText();
+
+  // TODO: Will also need to check for yaml files and change the contentType accordingly
```
rholinshead, line 674: For sharing YAML, we'll also need to update the readonly implementation to parse to JSON for rendering it

vscode-extension/static/untitled.aiconfig.json
```diff
@@ -0,0 +1,32 @@
...
+    "models": {
+      "gpt-3.5-turbo": {
+        "model": "gpt-3.5-turbo",
+        "top_p": 1,
+        "temperature": 1
+      }
+    },
+    "default_model": "gpt-3.5-turbo"
+  },
+  "description": "Fill our a description about your AIConfig here.",
```
rholinshead, line 14: nit, our -> out

rholinshead (approved):
Accepting to unblock, just nits and a few things/questions for understanding

vscode-extension/package.json
```diff
@@ -36,9 +59,20 @@
             "filenamePattern": "*.aiconfig"
           }
         ],
+        "enableForDiffView": false,
```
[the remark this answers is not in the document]
saqadri: Sadly this doesn't work :(. We'll need to figure this out still. There may be something else

vscode-extension/README.md
```diff
@@ -1,71 +1,51 @@
...
-- Toggle preview (`Shift+Cmd+V` on macOS or `Shift+Ctrl+V` on Windows and Linux).
-- Press `Ctrl+Space` (Windows, Linux, macOS) to see a list of Markdown snippets.
-
-## For more information
-
-- [Visual Studio Code's Markdown Support](http://code.visualstudio.com/docs/languages/markdown)
-- [Markdown Syntax Reference](https://help.github.com/articles/markdown-basics/)
-
-**Enjoy!**
+| Provider                          | Model                                                                                                        | Language | Support                                                                                                     |
```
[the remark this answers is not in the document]
saqadri: That's true -- the main question is this readme displays in the extension marketplace so it might be useful to still keep, but agreed that it's pretty annoying to maintain two places. Will ship as is but we can maybe reword this to make it more generic, with a link to specific models supported

saqadri:
Fixed all the actionable feedback, and the remaining leaving as open comments so we can refer to it as we iterate on the extension. Thanks for the awesome diff reviews as always, @rholinshead!

PR #1142, opened 2024-02-06
Add Open In TextEditor button

saqadri:
Add Open In TextEditor button

This button just allows you to switch back to the raw texteditor in vscode.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1142).
* __->__ #1142
* #1141
* #1140


rholinshead (commented):
Seems a bit weird to couple this vs code specific UI within the editor component. Thoughts on alternatives:
- rendering the button above the editor component in VSCodeEditor
- setting an id on the button group containing the top bottoms and using react portal to inject the button from VSCodeEditor into the button group

saqadri:
> Seems a bit weird to couple this vs code specific UI within the editor component. Thoughts on alternatives:
> 
> * rendering the button above the editor component in VSCodeEditor (similar to hf token input in gradio, or source/download buttons in lastmile readonly)
> * setting an id on the button group containing the top bottoms and using react portal to inject the button from VSCodeEditor into the button group

Hmmm, this isn't that different from the other callbacks which, if undefined, won't render for the other editors. For local editor implementation, "open in text editor" button could toggle to a `JSONEditor` component for the entire config.

rholinshead:
> > Seems a bit weird to couple this vs code specific UI within the editor component. Thoughts on alternatives:
> > 
> > * rendering the button above the editor component in VSCodeEditor (similar to hf token input in gradio, or source/download buttons in lastmile readonly)
> > * setting an id on the button group containing the top bottoms and using react portal to inject the button from VSCodeEditor into the button group
> 
> Hmmm, this isn't that different from the other callbacks which, if undefined, won't render for the other editors. For local editor implementation, "open in text editor" button could toggle to a `JSONEditor` component for the entire config.

Ah, ok. If we want to use it in local as well it makes a bit more sense to be here. My comment was more about unnecessarily adding vs code-specific code to the component. Seems it's not vscode-specific if we'll support that in local as well, so lgtm

rholinshead (approved):
LGTM. We can probably wordsmith it a bit later so the text makes a bit more sense for local as well

PR #1144, opened 2024-02-06
[vscode] Fix white bar at top of config

rholinshead:
# [vscode] Fix white bar at top of config

In vscode there's a white bar at the top of the config in some themes. The cause is a margin-top set on one of the Flex containers. 
Screenshot 2024-02-06 at 12 51 41 PM

We can achieve the same effect using padding-top instead, but will result in the background colour being correct since there's no empty space between the elements now.

Just make sure local editor UI looks the same with margin-top changed to padding-top:
Screenshot 2024-02-06 at 2 15 07 PM

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
+++ b/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
@@ -992,7 +992,7 @@ function AIConfigEditorBase({
           </>
         )}
         <div>
-          <Flex justify="flex-end" mt="md" mb="xs">
+          <Flex justify="flex-end" pt="md" mb="xs">
             {
               <Group>
                 {downloadCallback && <DownloadButton onDownload={onDownload} />}


python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
```diff
@@ -992,7 +992,7 @@ function AIConfigEditorBase({
           </>
         )}
         <div>
-          <Flex justify="flex-end" mt="md" mb="xs">
+          <Flex justify="flex-end" pt="md" mb="xs">
```
saqadri, line 995: I would never have been able to get this. Thank you! :D

saqadri (approved):

Ankush-lastmile:
> We can achieve the same effect using padding-top instead, but will result in the background colour being correct since there's no empty space between the elements now.

Is this because the style sheets are defined for padding and not margin? curious how you figured this out

PR #1145, opened 2024-02-06
[editor] Support Explicit themeOverride for AIConfigEditor Component

rholinshead:
[editor] Support Explicit themeOverride for AIConfigEditor Component

# [editor] Support Explicit themeOverride for AIConfigEditor Component

From discussion earlier, the packaging devX for cleaning up the vscode theme for the vs code extension is still not in a good enough shape for easy iteration. To unblock Zak on theme work, we can support a themeOverride prop for the AIConfigEditor component and use it for overriding the theme instead of just using the static vs code theme defined. We can propagate the desired theme changes back to the static vs code theme afterwards.

Long-term this may be a nice prop to support anyway for external implementations to easily set the theme without requiring a bunch of css overrides to do so.

## Testing:
- Tested all permutations of existing mode, themeMode and system theme to ensure they all work correctly
- Explicitly set themeOverride value to gradio theme override object in LocalEditor and ensure it has proper theme. Also ensured the colorScheme in themeOverride takes precedence (if specified), otherwise default of system theme light/dark mode is used as expected

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
+++ b/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
@@ -7,6 +7,7 @@ import {
   Alert,
   Group,
   ActionIcon,
+  MantineThemeOverride,
 } from "@mantine/core";
 import {
   AIConfig,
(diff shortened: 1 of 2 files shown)


saqadri (approved):
Amazing!

rholinshead:
Fix merge conflict

Issue #1143, opened 2024-02-06
python version >=3.10 is required, 3.9 receives error


erdemunal35:
Hi,
I receive the error below when the python version of the virtual environment I install aiconfig is 3.9
<img width="1160" alt="Screenshot 2024-02-06 at 08 46 50" src="https://github.com/lastmile-ai/aiconfig/assets/46455260/9a872d39-d47e-42f3-aca3-bc77fe64296c">

when I use 3.10 it works fine. Can we say/add python version >=3.10 is required to the README.md?

Best,
Erdem

saqadri:
Thanks for reporting @erdemunal35! @Ankush-lastmile or @jonathanlastmileai can you PTAL?

rossdanlm:
Hey Erdem, we updated the pyprojects to all require `>=3.10` in https://github.com/lastmile-ai/aiconfig/pull/1146! Also made a note of this in our changelog (https://github.com/lastmile-ai/aiconfig/releases/tag/v1.1.20) and this is not fixed in `v1.1.20` for the main `python-aiconfig` library!

PR #1146, opened 2024-02-06
Changing all pyproject `requires-python` to 3.10

rossdanlm:
Changing all pyproject `requires-python` to 3.10

We have some libraries like `TypeAlias`, `match-case`, and shorthand for Union types (`X | Y` instead of `Union[X, Y]` that require Python 3.10, as flagged by https://github.com/lastmile-ai/aiconfig/issues/1143. Full list of features added to 3.10 listed here: https://docs.python.org/3/whatsnew/3.10.html

We're just going to be consistent for all extensions (even the ones that don't use the 3.10 libraries).

I'm also going to re-publish and build the libraries (and includes this in the release notes!)

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1146).
* #1147
* __->__ #1146

### DIFF
--- a/extensions/HuggingFace/python/pyproject.toml
+++ b/extensions/HuggingFace/python/pyproject.toml
@@ -11,7 +11,7 @@ authors = [
 ]
 description = "An extension for using Hugging Face tasks to parse models for AIConfig."
 readme = "README.md"
-requires-python = ">=3.8"
+requires-python = ">=3.10"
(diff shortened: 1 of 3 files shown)


Ankush-lastmile (approved):
lgtm

PR #1158, opened 2024-02-07
[python]  serialize first assistant message as prompt

Ankush-lastmile:
[python]  serialize first assistant message as prompt

A user was building with AIConfig and was attempting to serialize a set of prompts with OpenAI.

Their messages array looked like this
```
{ "role": "assistant", "content": "Welcome, ask me anything." },
{ "role": "user", "content": "Say this is a test" },
...
```

Notice how the first message was an assistant message! Previously serializing this with config.serialize() was not supported.

This diff adds minimal support by allowing serialization of the first assistant message, and creates a prompt with an empty input, and the assistant message as the output.

## testplan
Execute the user's test code. Also tested deserialize() and run() and they work as expected. No functional changes on that side

| Screenshot 2024-02-07 at 11 42 31 AM|Screenshot 2024-02-07 at 11 41 53 AM|

### DIFF
--- a/python/src/aiconfig/default_parsers/openai.py
+++ b/python/src/aiconfig/default_parsers/openai.py
@@ -74,7 +74,7 @@ class OpenAIInference(ParameterizedModelParser):
             },
         )
         await ai_config.callback_manager.run_callbacks(event)
-        prompts = []
+        prompts: list[Prompt] = []
 
         # Combine conversation data with any extra keyword args
         conversation_data = {**data}
@@ -154,6 +154,29 @@ class OpenAIInference(ParameterizedModelParser):
                         }
                     ),
                     outputs=assistant_output,
+                )
+                prompts.append(prompt)
+            elif i == 0 and role == "assistant":
+                # If the first message is an assistant message,
+                # build a prompt with an empty input, 
+                # and the assistant response as the output
+
+                # Pull assistant response
+                assistant_output = build_output_data(conversation_data["messages"][i])
+                prompt = Prompt(
+                    name= f"{prompt_name}_{len(prompts) + 1}",
+                    input="",
+                    metadata=PromptMetadata(
+                        model= copy.deepcopy(model_metadata),
+                        parameters= parameters,
+                        remember_chat_context = True
+                    ),
+                    outputs=[                        ExecuteResult(
+                            output_type="execute_result",
+                            execution_count=None,
+                            data=assistant_output,
+                            metadata={},
+                        )],
                 )
                 prompts.append(prompt)
             i += 1


python/src/aiconfig/default_parsers/openai.py
```diff
@@ -156,6 +156,25 @@ async def serialize(
...
                 prompts.append(prompt)
+            else:
+                # If the first message is an assistant message, create an empty prompt for it.
+                if i == 0:
+                    # Pull assistant response
+                    assistant_output = build_output_data(conversation_data["messages"][i])
+                    # Build prompt with empty input, output as the assistant response
+                    prompt = Prompt(
+                        name=new_prompt_name,
+                        input="",
```
saqadri, line 167: Can this be set to null instead of empty string?

python/src/aiconfig/default_parsers/openai.py
```diff
@@ -156,6 +156,25 @@ async def serialize(
                     outputs=assistant_output,
                 )
                 prompts.append(prompt)
+            else:
```
saqadri, line 159: nit: do an explicit check -- elif i==0 and assistant=true

saqadri (approved):
Just some comments. Test with config.deserialize, see what it's like in editor, etc

saqadri:
Also once you're happy with it please update TS implementation too

python/src/aiconfig/default_parsers/openai.py
Ankush-lastmile: our schema doesn't allow inputs to be None/null type

rholinshead: Would it be a valid alternative to use the PromptInput type for the input and just have the data be the assistant message?

Ankush-lastmile: I think my answer to that is no.

The PromptInput type is geared towards input, so storing this as an input would either have the implications that this is now an "input", or, we would have to define some additional logic to parse this into the "assistant message"

Some additional context to this it that the openai model parser does support storing the messages array itself as an input to the prompt; but that is a little hacky

python/src/aiconfig/default_parsers/openai.py
Ankush-lastmile: updated, added a check `role == "assistant"`

side note: our automated tests also caught this, which is great.

Ankush-lastmile:
- added check for assistant
- updated pr with test plan

saqadri:
Looks good, ship it! 

Please update the AIConfig package and let Yohann know that he can try again.

PR #1159, opened 2024-02-07
[vscode] Use VS Code Notifications for showNotification

rholinshead:
# [vscode] Use VS Code Notifications for showNotification

Just familiarizing myself a bit with the vscode extension dev so picked up this quick task. We support custom notifications via `showNotification` callback in the `AIConfigEditor`. For VS Code, we can leverage this callback to use the vs code `showXMessage` api, for X in 'information', 'warning', 'error'.

A caveat with vscode notifications is that they are intended to be very succinct -- they only allow a message string (and they strip newlines). They allow showing additional 'details' string but ONLY for modal UX, which is a bad UX to use compared to the toasts. As a workaround, this PR adds a 'Details' button to the notification toast so that the user can click it and see the full details message in the output log.

https://github.com/lastmile-ai/aiconfig/assets/5060851/1b0536b4-536b-4c44-a40c-faab80182677

### DIFF
--- a/vscode-extension/src/aiConfigEditor.ts
+++ b/vscode-extension/src/aiConfigEditor.ts
@@ -251,6 +251,62 @@ export class AIConfigEditorProvider implements vscode.CustomTextEditorProvider {
         case "open_in_text_editor":
           vscode.commands.executeCommand("workbench.action.reopenTextEditor");
           return;
+
+        case "show_notification":
+          const notification = e.notification as {
+            // AIConfigEditorNotification
+            title: string;
+            message: string | null;
+            type?: "info" | "success" | "warning" | "error";
+            autoClose?: boolean | number;
+            onClose?: () => void;
+            onOpen?: () => void;
+          };
+
+          let notificationFn;
+          let outputChannelFn;
+          switch (notification.type) {
+            case "info":
+            case "success":
+              notificationFn = vscode.window.showInformationMessage;
+              outputChannelFn = this.extensionOutputChannel.info;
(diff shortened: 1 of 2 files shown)


vscode-extension/src/aiConfigEditor.ts
```diff
@@ -251,6 +251,62 @@ export class AIConfigEditorProvider implements vscode.CustomTextEditorProvider {
...
           return;
+
+        case "show_notification":
+          const notification = e.notification as {
+            // AIConfigEditorNotification
+            title: string;
+            message: string | null;
+            type?: "info" | "success" | "warning" | "error";
+            autoClose?: boolean | number;
+            onClose?: () => void;
```
saqadri, line 262: Do onOpen and onClose work in this context? These functions are supplied from the webview?

saqadri (approved):
Looks awesome! Agreed with the usability decisions you've made.

vscode-extension/src/aiConfigEditor.ts
rholinshead: Ya, the editor component supports specifying onOpen and onClose as part of the notification object (which is also handled by mantine `showNotification`). In vscode case, we kind of implicitly handle it -- onOpen is just before the notification is called and onClose happens once the notification is resolved (happens on explicit close or button press). `autoClose` isn't supported directly by the vs code notification api but I've seen some hacks to support it - just prefer not to implement it until it's needed since it's hard to test with the current package scenario

PR #1160, opened 2024-02-07
[vscode] Propagate themeMode for dark/light Theme Switching

rholinshead:
[vscode] Propagate themeMode for dark/light Theme Switching

# [vscode] Propagate themeMode for dark/light Theme Switching

VS Code themes have a `ColorThemeKind` which is one of Light, Dark, HighContrast (dark) or HighConstrastLight. We can leverage this to explicitly set the `themeMode` for the `AIConfigEditor`, which will be used for a few things:
- default mantine highlight and text backgrounds
- JSON editor dark/light theme colors
- as needed, can use the theme 'dark' or 'light' in our custom VSCode theme styles to override defaults

## Testing:
Make sure the JSON editor background as well as hover and title text background are correct when changing between themes

https://github.com/lastmile-ai/aiconfig/assets/5060851/195e3d30-e3db-479c-80f5-c3aaddaa6a35

### DIFF
--- a/vscode-extension/editor/src/VSCodeEditor.tsx
+++ b/vscode-extension/editor/src/VSCodeEditor.tsx
@@ -28,8 +28,9 @@ import {
   updateWebviewState,
 } from "./utils/vscodeUtils";
 import { VSCODE_THEME } from "./VSCodeTheme";
-// TODO: Update package to export AIConfigEditorNotification type
+// TODO: Update package to export AIConfigEditorNotification and ThemeMode types
 import { AIConfigEditorNotification } from "@lastmileai/aiconfig-editor/dist/components/notifications/NotificationProvider";
+import { ThemeMode } from "@lastmileai/aiconfig-editor/dist/shared/types";
 
 const useStyles = createStyles(() => ({
   editorBackground: {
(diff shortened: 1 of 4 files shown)


vscode-extension/editor/src/VSCodeEditor.tsx
```diff
@@ -101,6 +103,11 @@ export default function VSCodeEditor() {
         // /get endpoint so we get the latest content from the server
         return;
       }
+      case "set_theme": {
+        const theme = message.theme;
+        setThemeMode(theme);
```
saqadri, line 108: I think we should save this in webview state similar to how we save the server url (see line 100). Otherwise when the webview is refreshed on tab switch it will have the wrong theme.

saqadri (approved):
Looks good, but I think missing one step to preserve the theme setting as vscode rehydrates the webview. Approving to unblock but important to fix before shipping.

saqadri (approved):
Looks good, but I think missing one step to preserve the theme setting as vscode rehydrates the webview. Approving to unblock but important to fix before shipping.

vscode-extension/editor/src/VSCodeEditor.tsx
rholinshead: Ah, thanks, nice catch

rholinshead:
Updated to handle rehydration when changing tabs:
- save theme in webview state --> initial theme state will be set to whatever was in state, which should technically be an improvement when NOT changing theme between tabs, since theme will be correct even before set_theme is dispatched
- importantly: dispatch set_theme with the correct theme whenever the tab becomes active again, handling the case where theme was changed when the tab was not active and then the tab becomes active again

PR #1162, opened 2024-02-07
[vscode] Connect openInTextEditor Callback

rholinshead:
# [vscode] Connect openInTextEditor Callback

Connect the button & event added in https://github.com/lastmile-ai/aiconfig/pull/1142/files with the callback to reopen the aiconfig editor webview as a text editor to interact with the underlying file directly. 

https://github.com/lastmile-ai/aiconfig/assets/5060851/76a25aa7-566f-491d-9949-b0f2f9ec3b1a

Note, this requires closing and reopening the file in order to re-initialize the aiconfig editor webview for the file. We could potentially add another command to handle this scenario (something like, "AIConfig: Open in Editor" which defaults to current file but shows others in current directory?)

### DIFF
--- a/vscode-extension/editor/src/VSCodeEditor.tsx
+++ b/vscode-extension/editor/src/VSCodeEditor.tsx
@@ -415,6 +415,10 @@ export default function VSCodeEditor() {
     []
   );
 
+  const openInTextEditor = useCallback(async () => {
+    vscode?.postMessage({ type: "open_in_text_editor" });
+  }, [vscode]);
+
   const showNotification = useCallback(
     (notification: AIConfigEditorNotification) =>
       vscode?.postMessage({ type: "show_notification", notification }),
@@ -430,6 +434,7 @@ export default function VSCodeEditor() {
       getModels,
       getServerStatus,
       logEventHandler,
+      openInTextEditor,
       runPrompt,
       setConfigDescription,
       setConfigName,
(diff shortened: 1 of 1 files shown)


saqadri (approved):
Thanks! I missed this in my stack

rholinshead:
@saqadri thoughts on the caveat?

> Note, this requires closing and reopening the file in order to re-initialize the aiconfig editor webview for the file. We could potentially add another command to handle this scenario (something like, "AIConfig: Open in Editor" which defaults to current file but shows others in current directory?)

Not sure that it's a blocker for shipping this, or if it can be added as an Issue for later

saqadri:
@rholinshead to clarify you mean in order to reopen the file as an aiconfig editor webview after having opened it as text, you need to close and reopen the editor, right? IMO this is ok. I like your idea of adding a command but not P0 in my opinion

rholinshead:
> @rholinshead to clarify you mean in order to reopen the file as an aiconfig editor webview after having opened it as text, you need to close and reopen the editor, right? IMO this is ok. I like your idea of adding a command but not P0 in my opinion

Yes, correct - need to close and reopen. Sounds good, I'll ship this

saqadri:
Actually, come to think of it, that command could be pretty useful in other situations too -- for example one thing people struggled with creating a config was naming the file to match the expected extension *.aiconfig.json/yaml

Having a command could allow people to open arbitrary json/yaml files as aiconfig editors

PR #1163, opened 2024-02-07
[vscode][ez] Fix Default Gradient Color (for Top Buttons)

rholinshead:
# [vscode][ez] Fix Default Gradient Color (for Top Buttons)

Most of the top buttons in the editor are using 'gradient' as their variant. We have the defaultGradient set to pink in the theme so changing it to match what the vscode button color should be.

## Before:
Screenshot 2024-02-07 at 4 39 38 PM

## After:
Screenshot 2024-02-07 at 4 34 55 PM


saqadri (approved):
Thank you!

PR #1164, opened 2024-02-07
[editor] Add 'OpenAI Config File' Command

rholinshead:
[editor] Add 'OpenAI Config File' Command

# [editor] Add 'OpenAI Config File' Command

From discussion in https://github.com/lastmile-ai/aiconfig/pull/1162, it will be handy to have a command for opening an AIConfig file, especially in a couple cases:
1. If wanting to open a `.json` or `.yaml` config file, without the `.aiconfig.` sub-extension in the name

https://github.com/lastmile-ai/aiconfig/assets/5060851/9a851dce-ff45-4ef1-a627-5526fafb5f4b

2. After opening an AIConfig file in a text editor using the 'Open in Text Editor' button, re-opening the file with the AIConfig editor

https://github.com/lastmile-ai/aiconfig/assets/5060851/5c07d02a-233c-48cf-b8c9-234640bec81d

## Testing:
- Above scenarios work
- Also tested opening a json file that doesn't match config spec, the server errors and leaves an empty editor:
Screenshot 2024-02-07 at 5 48 03 PM
This is technically the case for existing configs as well, so not related to this PR. We should probably improve that UX in the future as well

### DIFF
--- a/vscode-extension/src/extension.ts
+++ b/vscode-extension/src/extension.ts
@@ -16,7 +16,9 @@ import {
   sanitizeFileName,
   getTodayDateString,
   LASTMILE_BASE_URI,
-  getPythonPath
+  getPythonPath,
+  isSupportedConfigExtension,
+  SUPPORTED_FILE_EXTENSIONS,
 } from "./util";
 import { AIConfigEditorProvider } from "./aiConfigEditor";
 import { AIConfigEditorManager } from "./aiConfigEditorManager";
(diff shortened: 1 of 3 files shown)


vscode-extension/src/extension.ts
```diff
@@ -177,6 +187,36 @@ async function createNewAIConfig(
...
+    defaultUri = activeDocument.uri;
+  }
+
+  const openUri = await vscode.window.showOpenDialog({
+    defaultUri,
+    filters: {
+      // Allow opening .json and .yaml files regardless of .aiconfig. sub-extension
+      "AIConfig Extension": SUPPORTED_FILE_EXTENSIONS,
+    },
+  });
```
saqadri, line 207: I'm wondering if we should update this command or add a separate command that does `Open Active File in AIConfig Editor` (wording TBD), which instead of opening the file dialog just opens the current active file as an aiconfig.

Reason being there could be untitled files (e.g. copy-pasting content into a text editor) and wanting to open that as an aiconfig.

saqadri (approved):
This approach is decent, but maybe we should also have a command (maybe a second command) which doesn't open the file dialog and just uses the `defaultUri`

vscode-extension/src/extension.ts
rholinshead: I think it might make sense to have as a separate command. Otherwise, if it's part of this command we need to surface that choice somehow (e.g. like a dialog w/ button "current" or "search") and that makes it an extra click to get through this (imo way more common) flow.

rholinshead: Going to land this and can follow up with the other command tomorrow

rholinshead:
rebase

PR #1179, opened 2024-02-08
Add `log.txt` to .gitignore

rossdanlm:
Add `log.txt` to .gitignore

This was grinding my gears!!!!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # Logs
 logs
+log.txt
 *.log
 npm-debug.log*
 yarn-debug.log*


rholinshead:
Do you know where the `log.txt` file is coming from? This is more of a bandaid and we should fix whatever's creating the file because it happens with the prod vscode extension and is going to show up for all users as well

saqadri:
What process generates the log.txt file? Does it need to be created?

rossdanlm:
> Do you know where the log.txt file is coming from?

I don't know, just annoying for now and don't want this being generated in our repo. I created an issue to investigate: https://github.com/lastmile-ai/aiconfig/issues/1183

Ankush-lastmile (approved):

Ankush-lastmile:
lets file an issue to keep track of this

PR #1178, opened 2024-02-08
[vs-code][1/n] Welcome Page: Scaffolding

rossdanlm:
[vs-code][1/n] Welcome Page: Scaffolding

Just basic scaffolding for now, nothing fancy, will make UI better next diff

Following instructions from https://lastmileai.dev/workbooks/cls5af4s700pkqrsegl64gx3i

## Test Plan

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1178).
* #1182
* __->__ #1178

PR #1182, opened 2024-02-08
[vs-code][2/n] Welcome Page: HTML content

rossdanlm:
[vs-code][2/n] Welcome Page: HTML content

Just some HTML content.

I tried having links to commands, but turns out that's not super straightforward, so whatever just listing commands and they can run it themselves. See how VSCode does it in here:
- command definition: https://github.com/microsoft/vscode/blob/c1021c9573c1a60c652be64556b8ac024e543c83/src/vs/workbench/contrib/welcomeGettingStarted/common/gettingStartedContent.ts#L64C15-L64C45
- command execution: https://github.com/microsoft/vscode/blob/c1021c9573c1a60c652be64556b8ac024e543c83/src/vs/workbench/contrib/welcomeGettingStarted/browser/gettingStarted.ts#L1232

Can add more pictures later if you want.

Added some FAQ to learn more + get unblocked.

## Test Plan

Screenshot 2024-02-08 at 16 38 10

Video showing full page (can pause to read stuff)

https://github.com/lastmile-ai/aiconfig/assets/151060367/451dac81-df10-48ad-93af-e7b4570e39e1

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1182).
* __->__ #1182
* #1178

### DIFF
--- a/vscode-extension/src/welcomePage.html
+++ b/vscode-extension/src/welcomePage.html
@@ -2,7 +2,149 @@
 <html lang="en">
 
 <body>
-    <h1>Welcome to my extension!</h1>
-</body>
+    <h1>Welcome to AIConfig Editor!</h1>
 
+    AIConfig Editor is a tool that allows you to <b>create, edit, and manage AIConfig files</b>. 
+    This acts as an open-source playground that supports local and remote models from 
+    OpenAI, Anthropic, Meta, Google, HuggingFace, Anyscale, and more.<br>
+    <b> Tip:</b> To learn more about AIConfigs and why they're useful, check out the FAQ 
+    section at the bottom of this page.
+    <br>
+    <p>
+        <img src="https://s3.amazonaws.com/lastmileai.aiconfig.public/playing_around_with_editor.png" alt="Playing around with the editor" width="700" height="250">
+    </p>
+
+    <h2>Commands</h2>
+
+    Get started by trying out some of these commands! Press ("CMD" + "SHIFT" + P) 
+    to open the command palette and type in "AIConfig":
+
+    <ul>
+        <li>"AIConfig: Create New" -> Creates a new file (in JSON or YAML format)</li>
+        <li>"AIConfig: Share (get permalink)" -> Creates a link to read-only state of your AIConfig file. Use this to share with others :D</li>
+        <li>"AIConfig: Welcome" -> Opens this window that you can refer to again!</li>
+    </ul>
+
+    You can also open existing files that end in ".aiconfig.json" or ".aiconfig.yaml" 
+    (note that there are two periods in the filename!). <br>
+    Example: "my_cool_file.aiconfig.json"
+
+    <h2>API Keys</h2>
+    Some models require API keys in order to run them. 
+    If you want to run any of the models below, please create a ".env" file in
+    the root of your project and add the following keys:
+    
+    <ul>
+        <li>
+            <a href="https://ai.google.dev/tutorials/setup">GOOGLE_API_KEY</a>=your_key_here
+            <ol style="list-style-type: lower-alpha; padding-bottom: 0;">
+                <li>Used for Gemini</li>
+            </ol>
+        </li>
+        <li>
+            <a href="https://platform.openai.com/api-keys">OPENAI_API_KEY</a>=your_key_here
+            <ol style="list-style-type: lower-alpha; padding-bottom: 0;">
+                <li>Used for OpenAI models like ChatGPT and Dall-E</li>
+            </ol>
+        </li>
+        <li>
+            <a href="https://huggingface.co/settings/tokens">HUGGING_FACE_API_TOKEN</a>=your_token_here
+            <ol style="list-style-type: lower-alpha; padding-bottom: 0;">
+                <li>Used for any models <a href="https://huggingface.co/models">hosted on Hugging Face</a></li>
+            </ol>
+        </li>
+    </ul>
+
+    For example, your .env file should look similar to this:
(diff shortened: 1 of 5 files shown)


vscode-extension/src/welcomePage.html
```diff
@@ -0,0 +1,124 @@
+<!DOCTYPE html>
```
saqadri, line 1: Why isn't this markdown?

saqadri:
This should literally be a markdown file that we open.

saqadri (changes requested):
This is overly complicated. Just make an md markdown file that does the exact same thing. It's much simpler

vscode-extension/src/welcomePage.html
```diff
@@ -2,7 +2,123 @@
...
+    generative AI models using AIConfig files with easily button clicks. Think of it like being able
+    to open up a Word document with Microsoft Word instead of having to write it in Notepad. -->
+
+
+    <h1> FAQ </h1>
+    <p>
+        <b>Q: What is AIConfig?</b> <br>
+
+        A: AIConfig files are files in JSON or YAML format where you can define the 
+        structure of generative AI workflows. Each AIConfig file contains a set
```
rholinshead, line 85: I think we should speak more to configuring prompt inputs, models and settings instead of calling out structure of workflows

vscode-extension/src/welcomePage.html
```diff
@@ -2,7 +2,123 @@
...
+        A: AIConfig files are files in JSON or YAML format where you can define the 
+        structure of generative AI workflows. Each AIConfig file contains a set 
+        of prompts which define what AI model to use and what data (ex: input 
+        settings, prompt parameters) to pass into the model when calling it. It 
+        also contains the output for that call if it exists.
+        
+        AIConfigs are pretty gnarley because they let you:
+
+        <ol>
+        <li>Use any model (that we support)
```
rholinshead: nit, should we remove "(that we support)"

vscode-extension/src/welcomePage.html
```diff
@@ -2,7 +2,123 @@
...
+    <p></p>Here's what the raw data looks like:
+
+    Pretty ugly amiright? This is why we have the AIConfig Editor. It lets you play around with 
+    generative AI models using AIConfig files with easily button clicks. Think of it like being able
+    to open up a Word document with Microsoft Word instead of having to write it in Notepad. -->
+
+
+    <h1> FAQ </h1>
+    <p>
+        <b>Q: What is AIConfig?</b> <br>
```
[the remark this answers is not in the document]
rholinshead: Also, do we need to use this space for explaining what an AIConfig is? That's probably better suited for the extension page itself. I wouldn't think the Welcome page needs to reiterate that (should already be "sold" before installing the instruction).

We should use the welcome page real estate for extension specifics (commands, getting started)

rholinshead (commented):
I think we can trim out a lot of stuff to make this page more useful. The meme and sales pitch IMO don't belong / just take up a lot of the prime real estate that we should be using to better surface extension functionality and "getting started" tips (also noting that some critical info is missing, e.g. output for debugging, the conda/interpreter steps)

rholinshead:
Is the copy/design from Tanya?

> This should literally be a markdown file that we open.

My understanding was it was HTML so that we can support the link clicks -> action dispatch

vscode-extension/src/welcomePage.html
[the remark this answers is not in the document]
rholinshead: Would recommend removing this Q and then changing FAQ to "Troubleshooting"

vscode-extension/src/welcomePage.html
rossdanlm: Added an image from Gradio workbooks showing general instructions

vscode-extension/src/welcomePage.html
saqadri: Yes you can open the readme in preview mode, we don't need to open it as a webview.

saqadri: You should be able to do something like this:

```
function activate(context) {
    let previousVersion = context.globalState.get('version');
    let currentVersion = vscode.extensions.getExtension(EXTENSION_NAME).packageJSON.version;

    if(previousVersion === undefined || previousVersion !== currentVersion) {
        let readmePath = path.join(context.extensionPath, 'README.md');
        let readme = fs.readFileSync(readmePath, 'utf8');
        // Open raw markdown (not needed)
        //let document = await vscode.workspace.openTextDocument({ content: readme, language: 'markdown' });
        //vscode.window.showTextDocument(document);

        // Open markdown in preview mode
        vscode.commands.executeCommand('markdown.showPreview', vscode.Uri.file(readmePath))
    
        // Update the version in the globalState
        context.globalState.update('version', currentVersion);
    }
}
```

The code inside the IF block can be refactored into the WELCOME command

saqadri: Also we should have an information message that pops up when the Welcome page is shown. Can also have a setting called 'showWelcome' which can be set to false if a user selects "Don't Show Again" in that information message. Can happen at the top of the stack.

saqadri:
The content is very verbose and complicated in this welcome page. I guess fine for right now, but needs to be dramatically cleaned up. You also should register the commands as hyperlinks so someone can just click a link to create a new AIConfig, to install, etc etc. I think this is possible in markdown but not in a webview

saqadri (approved):
Approving to unblock. We need to redo how this works tomorrow

PR #1184, opened 2024-02-08
[1/N][RAG Demo fixes] remove faithfulness metric

jonathanlastmileai:
[1/N][RAG Demo fixes] remove faithfulness metric

It turns out even GPT4 can't do this with long contexts.


saqadri (commented):
Let's keep the metric but not rely on it. Otherwise it's an even less impressive demo

PR #1185, opened 2024-02-08
draft install dependencies if not available when opening an aiconfig

Ankush-lastmile:
draft install dependencies if not available.


Ankush-lastmile:
implemented in #1262 stack

PR #1201, opened 2024-02-08
[rfc] Prevent 'Error starting aiconfig server' notification when closing w/ unsaved changes

rholinshead:
[rfc] Prevent 'Error starting aiconfig server' notification when closing w/ unsaved changes

# [rfc] Prevent 'Error starting aiconfig server' notification when closing w/ unsaved changes

Currently, whenever an aiconfig editor webview is closed with unsaved changes, a notifcation shows at the bottom right:
Screenshot 2024-02-08 at 6 29 06 PM

This is caused by the following sequence of events:
1. changeDocumentSubscription happens with the "unsaved" changes being undone; this results in updating webview & server
2. server /load request is sent as a result
3. Another changeDocumentSubscription happens with no changed content (can probably just skip -- in all cases?)
4. Webview disposed, kills editor process
5. /load Request (STILL IN FLIGHT) fails, causing the notification

I spent some time but wasn't able to figure out why changeDocumentSubscription was happening (note -- NOT happening from the notifying dirty). Seems like this could potentially be a KP - https://github.com/microsoft/vscode/issues/81902 but not concrete.

For now, RFC on this, we could just handle the server request error specially in this case when the webview is already disposed (and ignore it).

https://github.com/lastmile-ai/aiconfig/assets/5060851/4e986654-31d9-4192-bba7-d07fe0756448

Make sure the error still shows for legit cases (e.g. opening config without correct parsers installed):

https://github.com/lastmile-ai/aiconfig/assets/5060851/912b8123-137f-4e8e-ab64-b05438adb44f

### DIFF
--- a/vscode-extension/src/aiConfigEditor.ts
+++ b/vscode-extension/src/aiConfigEditor.ts
@@ -6,7 +6,7 @@ import {
   getCurrentWorkingDirectory,
   getDocumentFromServer,
   getPythonPath,
-  initializeServerState,
+  updateServerState,
   updateWebviewEditorThemeMode,
   waitUntilServerReady,
 } from "./util";
@@ -144,23 +144,59 @@ export class AIConfigEditorProvider implements vscode.CustomTextEditorProvider {
             return;
           }
 
+          // TODO: Should we skip events with no content changes?
+          // if (e.contentChanges.length === 0) {
+          //   console.log(
+          //     `changeDocumentSubscription ${e.document.uri} - skipping event because there are no content changes`
+          //   );
+          //   return;
+          // }
+
           // TODO: saqadri - instead of sending the entire document to the webview,
           // can ask it to reload the document from the server
 
           // Update webview immediately, and server state should be updated in the background.
+          console.log(
+            `changeDocumentSubscription ${e.document.uri} -- updating webview`
+          );
           updateWebview();
 
           // Notify server of updated document
           if (editorServer) {
-            console.log("changeDocumentSubscription -- updating server");
-
             // TODO: saqadri - decide if we want to await here or just fire and forget
-            await initializeServerState(editorServer.url, e.document);
-          }
+            try {
+              console.log("changeDocumentSubscription -- updating server");
+              await updateServerState(editorServer.url, e.document);
+            } catch (e) {
+              if (isWebviewDisposed) {
+                // Ignore errors caused by closing the webview while the server update
+                // request is in flight (e.g. closing config w/ unsaved changes will
+                // emit onDidChangeTextDocument with reverted unsaved changes)
+                // See #1201 for full details.
+                console.info(
+                  "Ignoring server update error due to webview disposal"
+                );
+                return;
+              }
 
-          console.log(
-            `changeDocumentSubscription ${e.document.uri} -- updating webview`
-          );
+              vscode.window
+                .showErrorMessage(
+                  "Failed to update aiconfig server. You can view the aiconfig but cannot modify it.",
+                  ...["Details", "Retry"]
+                )
+                .then((selection) => {
+                  if (selection === "Details") {
+                    this.extensionOutputChannel.error(
+                      e?.message ?? JSON.stringify(e)
+                    );
+                    this.extensionOutputChannel.show(/*preserveFocus*/ true);
+                  }
+                  if (selection === "Retry") {
+                    updateServerState(editorServer.url, e.document);
+                  }
+                });
+            }
+          }
         }
       }
     );
@@ -199,17 +235,17 @@ export class AIConfigEditorProvider implements vscode.CustomTextEditorProvider {
 
     // Make sure we get rid of the listener when our editor is closed.
     webviewPanel.onDidDispose(() => {
+      isWebviewDisposed = true;
       console.log(`${document.fileName}: Webview disposed`);
+
       changeDocumentSubscription.dispose();
       willSaveDocumentSubscription.dispose();
 
-      // TODO: saqadri -- terminate the editor server process.
       if (editorServer) {
         console.log("Killing editor server process");
         editorServer.proc.kill();
+        editorServer = null;
       }
-
-      isWebviewDisposed = true;
     });
 
     // Receive message from the webview.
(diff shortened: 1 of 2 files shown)


vscode-extension/src/aiConfigEditor.ts
```diff
@@ -338,7 +373,24 @@ export class AIConfigEditorProvider implements vscode.CustomTextEditorProvider {
...
 
     // Now set up the server with the latest document content
-    await initializeServerState(editorServer.url, document);
+    try {
+      await updateServerState(editorServer.url, document);
+    } catch (e) {
+      vscode.window
+        .showErrorMessage(
+          "Failed to start aiconfig server. You can view the aiconfig but cannot modify it.",
+          ...["Details", "Retry"]
```
rossdanlm, line 382: What does the "Details" here add? Looks the same in test plans

vscode-extension/src/aiConfigEditor.ts
```diff
@@ -199,17 +234,17 @@ export class AIConfigEditorProvider implements vscode.CustomTextEditorProvider {
...
       console.log(`${document.fileName}: Webview disposed`);
+
       changeDocumentSubscription.dispose();
       willSaveDocumentSubscription.dispose();
 
-      // TODO: saqadri -- terminate the editor server process.
       if (editorServer) {
         console.log("Killing editor server process");
         editorServer.proc.kill();
+        editorServer = null;
```
rossdanlm, line 246: Why do we need to explicitly set this to null? Is this related to the logic in the rest of this PR or just finishing the TODO from Sarmad on L206?

vscode-extension/src/aiConfigEditor.ts
```diff
@@ -199,17 +234,17 @@ export class AIConfigEditorProvider implements vscode.CustomTextEditorProvider {
 
     // Make sure we get rid of the listener when our editor is closed.
     webviewPanel.onDidDispose(() => {
+      isWebviewDisposed = true;
```
rossdanlm, line 237: Ok so just to clarify, this needs to be moved up top here becasue it's a flag that will get read on L171? And L171 flow occurs when we call `changeDocumentSubscription` on L240 below?

vscode-extension/src/aiConfigEditor.ts
```diff
@@ -144,23 +144,58 @@ export class AIConfigEditorProvider implements vscode.CustomTextEditorProvider {
...
-            console.log("changeDocumentSubscription -- updating server");
-
             // TODO: saqadri - decide if we want to await here or just fire and forget
-            await initializeServerState(editorServer.url, e.document);
-          }
+            try {
+              console.log("changeDocumentSubscription -- updating server");
+              await updateServerState(editorServer.url, e.document);
+            } catch (e) {
+              if (isWebviewDisposed || !editorServer) {
```
rossdanlm, line 171: What are the use cases where this is true because of the 2nd condition instead of the 1st (ie: `isWebviewDisposed` is false, but `editorServer` is null)? Is it when we first try to open the file in AIConfig editor before it's had time to initialize the server connection?

vscode-extension/src/aiConfigEditor.ts
```diff
@@ -144,23 +144,58 @@ export class AIConfigEditorProvider implements vscode.CustomTextEditorProvider {
             return;
           }
 
+          // TODO: Should we skip events with no content changes?
+          // if (e.contentChanges.length === 0) {
+          //   console.log(
+          //     `changeDocumentSubscription ${e.document.uri} - skipping event because there are no content changes`
```
rossdanlm, line 150: In general, would this case even be possible? Why would `changeDocumentSubscription` change if the content hasn't been changed: https://lastmileai.dev/workbooks/clsf1z185004bqpof6z9lbpvn

vscode-extension/src/aiConfigEditor.ts
```diff
@@ -144,23 +144,58 @@ export class AIConfigEditorProvider implements vscode.CustomTextEditorProvider {
...
-            await initializeServerState(editorServer.url, e.document);
-          }
+            try {
+              console.log("changeDocumentSubscription -- updating server");
+              await updateServerState(editorServer.url, e.document);
+            } catch (e) {
+              if (isWebviewDisposed || !editorServer) {
+                // Ignore errors caused by closing the webview while the server update
+                // request is in flight (e.g. closing config w/ unsaved changes will
+                // emit onDidChangeTextDocument with reverted unsaved changes)
```
rossdanlm, line 174: Nit: add something like "please refer to https://github.com/lastmile-ai/aiconfig/pull/1201 for full details" becuase the explanation you have in summary is really helpful

rossdanlm (approved):
I think this is super well investigated, thanks for the thorough explanation in summary

vscode-extension/src/aiConfigEditor.ts
rholinshead: Honestly, I'm not sure if it needs to be (seems same behaviour with or without) but the fact that the existing code has webViewDisposed check and editorServer check suggests we *should* be setting it to null when its process is killed. More for completeness than anything

vscode-extension/src/aiConfigEditor.ts
rholinshead: When testing it didn't actually seem to matter if this was at the top or bottom but it makes more sense to set it immediately, considering there are async events that are checking it

vscode-extension/src/aiConfigEditor.ts
rholinshead: Hmm, that's a good point. In theory I guess that scenario may be able to happen. I had just added this check for completeness but maybe we do actually want to surface the error if this scenario *does* happen. I'll remove the `!editorServer` check

vscode-extension/src/aiConfigEditor.ts
rholinshead: Yep, it's possible and 100% happens every time you close the editor with unsaved changes. It also seems to happen pretty regularly when just making normal changes in the editor

vscode-extension/src/aiConfigEditor.ts
rholinshead: It adds the 'Details' button in the notification, clicking it opens the Output tab with the exception message

rholinshead:
Changes from review:
- add comment to refer to this PR for full details
- remove check for null server when webview not disposed so that notification shows in that case

vscode-extension/src/aiConfigEditor.ts
rossdanlm: Ok then yea, think it would be good just to ignore if there are no contentChanges detected, as long as the "Are you sure you want to remove unsaved changes" dialog still pops up

PR #1205, opened 2024-02-09
[editor] Add 'Metadata' Section Under Settings Tab

rholinshead:
# [editor] Add 'Metadata' Section Under Settings Tab

Currently the editor does not provide a way to modify general prompt metadata, outside of model settings and parameters. For some of our chat parsers, this means the user cannot modify the 'remember_chat_context' setting. This PR implements the UX for modifying this metadata, adding it below the model settings in the settings tab. For now, only doing so if the prompt schema specifies `prompt_metadata` since I have some concerns about this UX for general metadata (without prompt schema it just renders as JSON editor below the settings and that's kind of confusing; it may be improved by having metadata under an accordion fold but still not sure just yet how we can have some metadata above that fold and some not).

https://github.com/lastmile-ai/aiconfig/assets/5060851/94d5e915-6bbd-4796-919c-e356aba0aa85

Screenshot 2024-02-15 at 11 21 26 AM

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1205).
* #1209
* #1208
* __->__ #1205


python/src/aiconfig/editor/server/static/asset-manifest.json
rholinshead, line 1: Ugh, sorry, just ignore this build stuff here and below. I had run yarn build accidentally and forgot to undo. It's ok to land the build changes since it needs to be done before publishing the aiconfig package anyway

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
```diff
@@ -242,19 +242,21 @@ function AIConfigEditorBase({
       async (
         promptName: string,
         newPrompt: Prompt,
-        onSuccess: (aiconfigRes: AIConfig) => void,
-        onError: (err: unknown) => void
+        callbacks: {
+          onSuccess?: (aiconfigRes: AIConfig) => void;
+          onError?: (err: unknown) => void;
+        }
```
Ankush-lastmile, line 248: what's the rationale to make this an object vs keep the callbacks defined?

python/src/aiconfig/editor/client/src/reducers/aiconfigReducer.ts
```diff
@@ -149,6 +149,17 @@ export default function aiconfigReducer(
...
       }));
     }
+    case "UPDATE_PROMPT_METADATA": {
+      return reduceReplacePrompt(dirtyState, action.id, (prompt) => ({
+        ...prompt,
+        metadata: {
+          ...action.metadata,
+          // Keep existing model and parameters metadata (managed by separate actions)
+          model: prompt.metadata?.model,
+          parameters: prompt.metadata?.parameters,
```
Ankush-lastmile, line 159: could this just be 

```
{
action.metadata?
...prompt.metadata
}
```

Ankush-lastmile (approved):
lgtm, I had 2 questions

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
rossdanlm: Using `onSuccess` and `onError` is a very common frontend pattern. I think Ryan is just following conventions?

rossdanlm:
[UX] cc @zakariassoul  Why not make other stuff like `stream` setting part of metadata as well, since that's not actually "model settings"?

python/src/aiconfig/editor/client/src/reducers/aiconfigReducer.ts
rossdanlm: Following up discussion offline w. Ankush so that others are aware.

He's ensuring that whatever is defined in `action.metadata` (ex: "test": "Test value here") gets properly set through the action and not overridden by what was previously in the `prompt.metadata` field. We also can't just do 

```typescript
metadata: { ...prompt.metadata action.metadata? }
```

Because action is a generic `JSONObject` and Ryan is safe-guarding to ensure we can't possibly override the `model` and `parameters` unsafely through this `UPDATE_PROMPT_METADATA` action

python/src/aiconfig/editor/client/src/reducers/actions.ts
```diff
@@ -15,9 +15,10 @@ export type MutateAIConfigAction =
   | SetDescriptionAction
   | SetNameAction
   | UpdatePromptInputAction
-  | UpdatePromptNameAction
+  | UpdatePromptMetadataAction
   | UpdatePromptModelAction
   | UpdatePromptModelSettingsAction
+  | UpdatePromptNameAction
```
rossdanlm, line 21: Nice, alphabetical!

python/src/aiconfig/editor/client/src/components/prompt/prompt_metadata/PromptMetadataSchemaRenderer.tsx
```diff
@@ -0,0 +1,46 @@
...
+export default memo(function ModelSettingsSchemaRenderer({
+  schema,
+  metadata,
+  onUpdatePromptMetadata,
+}: Props) {
+  const debouncedConfigUpdate = useMemo(
+    () =>
+      debounce(
+        (newMetadata: JSONObject) => onUpdatePromptMetadata(newMetadata),
+        250
```
rossdanlm, line 24: Don't we have a constant val for this in utils?

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
```diff
@@ -387,6 +393,57 @@ function AIConfigEditorBase({
...
+      dispatch({
+        type: "UPDATE_PROMPT_METADATA",
+        id: promptId,
+        metadata: newMetadata,
+      });
+
+      const onError = (err: unknown) => {
+        const message = (err as RequestCallbackError).message ?? null;
+        showNotification({
+          title: "Error updating prompt name",
```
rossdanlm, line 413: Typo: should be metadata instead of name

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
```diff
@@ -387,6 +393,57 @@ function AIConfigEditorBase({
     );
   }, [updateModelCallback]);
 
+  const onUpdatePromptMetadata = useCallback(
+    async (promptId: string, newMetadata: JSONObject) => {
+      if (!debouncedUpdatePrompt) {
+        // Just no-op if no callback specified. We could technically perform
+        // client-side updates but that might be confusing
```
rossdanlm, line 400: >We could technically perform client-side updates but that might be confusing

Nah this is a good call

rossdanlm (approved):
K yo next time pls split this into small er diffs, would've been much easier to review. Ex:

1. UI for showing tabs + JSON editor (no callback implementations)
2. Implementation (getting model settings + schema renderer)
3. Implementation (writing model settings, updating the reducer, etc)
4. Other refactoring work (ex: `onSuccess` + `onError` into single `callbacks` object

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
rholinshead: Ya, for js usually it's common to use an object when props can be undefined rather than having literal 'undefined' prop before a defined one

zakariassoul:
@rossdanlm my overall feedback: 

1- Displaying information/Actions needs to be contextual. Settings is global tab that enables a user to pick a model, and tweak it's behavior in one go. 

2- Let's preserve the integrity of the UI. Simplicity is critical for adoption,  and we need to say no to adding everything as a button or a tab even if we strongly feel it's useful.

rholinshead:
> [UX] cc @zakariassoul Why not make other stuff like `stream` setting part of metadata as well, since that's not actually "model settings"?

`stream` *IS* a model setting for some model parsers (e.g. openai)

zakariassoul:
@rossdanlm, @rholinshead, & @Ankush-lastmile i suggest the following: 

1- Put remember context back in the settings tab
2- If and only If function calling needs it's own tab to reduce the clutter when setting it up, then repurpose the 3rd tab for function calling.

python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
rholinshead: Thanks, good catch

python/src/aiconfig/editor/client/src/components/prompt/prompt_metadata/PromptMetadataSchemaRenderer.tsx
rholinshead: Whoops, yep

rholinshead:
> K yo next time pls split this into small er diffs, would've been much easier to review. Ex:
> 
> 1. UI for showing tabs + JSON editor (no callback implementations)
> 2. Implementation (getting model settings + schema renderer)
> 3. Implementation (writing model settings, updating the reducer, etc)
> 4. Other refactoring work (ex: `onSuccess` + `onError` into single `callbacks` object

Ya, sorry. I had crunched through this last night and meant to spend some time cleaning it up this morning but kinda decided to put it up as-is and move to other priorities. Will be sure to do that next time

rholinshead:
> @rossdanlm, @rholinshead, & @Ankush-lastmile i suggest the following:
> 
> 1- Put remember context back in the settings tab 2- If and only If function calling needs it's own tab to reduce the clutter when setting it up, then repurpose the 3rd tab for function calling.

remember_chat_context wasn't in settings. It's a top-level metadata. Settings is specifically prompt.metadata.model.settings in the config. The UI needs to dynamically render based on the metadata of the config, we don't want to make our UI be coupled to these generic keys right now because we have nothing at the SDK level that currently dictates what these keys are

rholinshead:
Synced with @zakariassoul offline. Instead of its own tab, we can have the metadata below model settings (starting with remember_chat_context). Long-term, we can integrate an accordian-type UX for different "categories", e.g model metadata, lower-pri settings separate from hi-pri ones (will need schema changes). Caveat is we would want remember_chat_context to be part of the hi-pri settings section and not nested under metadata section

rholinshead:
Ok, updated per discussion with Zak. Move the metadata below the settings (if schema exists). The only new change since last PR is this placement (see change in PromptActionBar component)

saqadri (approved):

PR #1208, opened 2024-02-09
[editor][ez] Only Render SidePanel Tab Content When Active Tab

rholinshead:
[editor][ez] Only Render SidePanel Tab Content When Active Tab

# [editor][ez] Only Render SidePanel Tab Content When Active Tab

I noticed each of the tab components was being rendered even when they're not shown. We can improve render perf by only rendering them when their tab is selected.

## Testing:
- Just ensure tabs still work
- Use console.log to ensure the components are only rendered when shown

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1208).
* #1209
* __->__ #1208
* #1205


Ankush-lastmile:
> I noticed each of the tab components was being rendered even when they're not shown. 

How do you know if was being rendered? did you observe it in console or something?

rholinshead:
> > I noticed each of the tab components was being rendered even when they're not shown.
> 
> How do you know if was being rendered? did you observe it in console or something?

Ya, I confirmed it with adding console.log in the component function and saw it logging when the tab wasn't shown. I kind of had a hunch that would happen based on how the tab hierarchy looks. Technically mantine has a `keepMounted` prop which seems like it should actually do this but I was still seeing the render logged when I had it set to false so decided to just explicitly do this

rossdanlm (approved):

saqadri (approved):

PR #1213, opened 2024-02-10
[fix] Directly use `fetch` and `formData` in Share VSCode command instead of S3 client

rossdanlm:
[fix] Directly use `fetch` and `formData` in Share VSCode command instead of S3 client

I was never able to figure out how to use unsigned/anonymous credentials on TS. We use UNSIGNED in Gradio for Python backend here (https://github.com/lastmile-ai/gradio-workbook/blob/6911294e91612c2b97d9f77debe0a6cf1039c377/gradioworkbook/backend/gradio_notebook/events.py#L536), but can't ever get that equivalent to work, so just directly going to use post to a website link instead.

## Test Plan
To test, I temporarily deleted my `.aws/credentials` file, and restarted my computer

Before

Screenshot 2024-02-09 at 17 54 08

After

Screenshot 2024-02-09 at 20 15 52

When you click the link, it opened this for me: https://lastmileai.dev/aiconfig/clsfea2nt0004pecrbk5rw5lp

### DIFF
--- a/vscode-extension/package.json
+++ b/vscode-extension/package.json
@@ -110,7 +110,6 @@
     "typescript": "^5.3.3"
   },
   "dependencies": {
-    "aws-sdk": "^2.1547.0",
     "oboe": "^2.1.5",
     "portfinder": "^1.0.32",
     "ufetch": "^1.6.0"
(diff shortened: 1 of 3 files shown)


Ankush-lastmile (approved):
lgtm

PR #1231, opened 2024-02-13
[gradio] Minor UI Tweaks for Clean Up

rholinshead:
# [gradio] Minor UI Tweaks for Clean Up

Applying the minor UI tweaks proposed by Zak:
gradioNotebook

Mainly:
- Group share/download and use icons w/ tooltips (note, applying grouping style only when both are shown)
- Set min-height on top buttons and prompt name/model inputs to 36px
- Set min width of action bar to 32px when closed

## Before:
Screenshot 2024-02-13 at 3 14 24 PM

## After:
Screenshot 2024-02-13 at 3 18 41 PM

### DIFF
--- a/python/src/aiconfig/editor/client/src/themes/GradioTheme.ts
+++ b/python/src/aiconfig/editor/client/src/themes/GradioTheme.ts
@@ -51,6 +51,7 @@ export const GRADIO_THEME: MantineThemeOverride = {
             color: "#ea580c !important",
             fontSize: "16px" /* var(--button-large-text-size) */,
             fontWeight: 600 /* var(--button-large-text-weight) */,
+            minHeight: "36px",
             padding: "0 1.25rem",
           },
 
@@ -150,7 +151,7 @@ export const GRADIO_THEME: MantineThemeOverride = {
         ".ghost": {
           input: {
             border: `1px solid ${inputBorderColor}`,
-            maxHeight: "16px",
+            minHeight: "36px",
             fontFamily:
               "sf mono, ui-monospace, Menlo, Monaco, Consolas, 'Liberation Mono', 'Courier New', monospace",
             borderRadius: "8px",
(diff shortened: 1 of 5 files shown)


saqadri:
One feedback -- if share is supposed to be prominent let's make sure its tooltip is super clear. Ideally it has text as well but will defer to @zakariassoul's judgement. Just want to call out IMO share is a killer feature

python/src/aiconfig/editor/client/src/themes/GradioTheme.ts
```diff
@@ -166,6 +167,16 @@ export const GRADIO_THEME: MantineThemeOverride = {
...
 
+        ".buttonGroupLeft": {
+          borderBottomRightRadius: 0,
+          borderTopRightRadius: 0,
+        },
+
+        ".buttonGroupRight": {
+          borderBottomLeftRadius: 0,
+          borderTopLeftRadius: 0,
+        },
```
rossdanlm, line 178: Why define this separately in Gradio vs. in the component itself?

rossdanlm (approved):
I like the `isGrouped` button, seems cool

python/src/aiconfig/editor/client/src/themes/GradioTheme.ts
rholinshead: This is just in case we need to explicitly override in gradio if gradio overrides the default

rholinshead:
> One feedback -- if share is supposed to be prominent let's make sure its tooltip is super clear. Ideally it has text as well but will defer to @zakariassoul's judgement. Just want to call out IMO share is a killer feature

Tooltip is:
Screenshot 2024-02-13 at 4 33 31 PM

zakariassoul:
@saqadri Affordance is clear by using the icons and supplementing with a tooltip. Share icon is so universally recognized that it can stand alone without a title. 

Let's keep the UI clean, and keep secondary actions muted.

PR #1245, opened 2024-02-14
Set model field to None when HF task changes

rossdanlm:
Set model field to None when HF task changes

TSIA, have to do this on both the client and backend so that they're not out of sync

## Test Plan
Make sure to run `yarn && yarn build` from the editor client file, then go to `aiconfig` and run this:
```
parsers_path=./cookbooks/Gradio/aiconfig_model_registry.py
```
To test, I hardcoded the input schema for HuggingFaceText2ImageDiffusor to be `        default: "runwayml/stable-diffusion-v1-5",`. Notice that the other model settings are still intact, just not the model field

https://github.com/lastmile-ai/aiconfig/assets/151060367/be832cf2-b07c-4b6a-89c0-9812a798b877


python/src/aiconfig/editor/client/src/reducers/aiconfigReducer.ts
```diff
@@ -168,21 +168,23 @@ export default function aiconfigReducer(
...
+        // }
         return {
           ...prompt,
           metadata: {
             ...prompt.metadata,
             model: action.modelName
               ? {
                   name: action.modelName,
-                  settings: modelSettings,
+                  // TODO (rossdanlm): For now just clearing the model settings whenever new model name is selected
```
rholinshead, line 186: Wait, why doesn't this just remove 'model' if it exists instead? We were persisting he settings intentionally due to product feedback that changing the model from say gpt-4 to gpt-3.5 was dropping all of the settings even though they're supported

python/src/aiconfig/editor/client/src/shared/prompt_schemas/HuggingFaceText2ImageDiffusorPromptSchema.ts
```diff
@@ -20,6 +20,7 @@ export const HuggingFaceText2ImageDiffusorPromptSchema: PromptSchema = {
       model: {
         type: "string",
         description: `Hugging Face model to use`,
+        default: "runwayml/stable-diffusion-v1-5",
```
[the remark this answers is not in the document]
rholinshead: This isn't for remote inference. This is for local parsers, which has no default applied at the parser or api (diffuser) layer

python/src/aiconfig/schema.py
```diff
@@ -819,6 +819,14 @@ def _update_model_name_for_prompt(self, model_name: str, prompt_name: str):
             model_settings: InferenceSettings = (
                 prompt.metadata.model.settings or {}
             )
+            # TODO (rossdanlm): This is a hack so that when Hugging Face task
+            # (defined as model_name) is switched, the "model" field in setting
+            # is also updated and set to None so that it resets to default model
+            # for that HF task. We're setting the entire settings just to empty
+            # (using defaults from input schema) to make UI client updating easier
+            if hasattr(model_settings, "model"):
+                # model_settings["model"] = None
```
rholinshead, line 828: Can you clarify why it's easier to clear the whole thing instead of just setting the model to None? It's just a few lines of code difference but I think clearing is going to be a worse UX in some case (we used to clear and then updated to not clear based on product feedback)

rholinshead (changes requested):

rholinshead:
I think it would be better to just handle removing the model instead of clearing everything based on past product request

python/src/aiconfig/editor/client/src/reducers/aiconfigReducer.ts
rossdanlm: Sure I can do that

python/src/aiconfig/editor/client/src/shared/prompt_schemas/HuggingFaceText2ImageDiffusorPromptSchema.ts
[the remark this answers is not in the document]
rossdanlm: Yea I just did local for proof of concept to test

python/src/aiconfig/schema.py
rossdanlm: I just wanted to get a solution out quickly and ensure I could quickly test. Pretty tired so haven't been thinking too deeply

python/src/aiconfig/editor/client/src/reducers/aiconfigReducer.ts
```diff
@@ -182,7 +183,11 @@ export default function aiconfigReducer(
             model: action.modelName
               ? {
                   name: action.modelName,
-                  settings: modelSettings,
+                  // TODO (rossdanlm): For now just clearing the model field whenever new model name is selected
+                  settings: {
+                    ...modelSettings,
+                    model: undefined,
```
rholinshead, line 189: If we explicitly set to undefined here this is actually going to have an explicit value in the settings in the config, I think, even for settings that don't have model. Instead, can we just do this at the top when defining `modelSettings`:

```
... modelSettings = promptModel.settings;
// TODO (rossdanlm): For now just clearing the model field whenever new model name is selected
  delete modelSettings.model;
}
```

rholinshead (approved):
Accepting to unblock, but let's just delete the 'model' key from the settings instead of setting it to undefined

PR #1246, opened 2024-02-14
[gradio] Set Default Models for Summarization and Translation

rholinshead:
# [gradio] Set Default Models for Summarization and Translation

The default translation and summarization models are broken for the remote inference client, raising a `KeyError` for `summary_text` and `translation_text`, respectively.

We can hardcode working defaults for these parsers for now, based on the top most "liked" models for the relevant tasks:

Screenshot 2024-02-14 at 10 54 24 AM
Screenshot 2024-02-14 at 10 55 32 AM

NOTE: This translation model requires specifying the target and source lang, otherwise it errors about them missing. I think this is fine for now, for a few reasons:
- it's better than current default of failing without clear error message and with no way to resolve (now, just add tgt and src lang)
- there are very few good translation models in the hub that actually work with inference API and those that do either require the src and tgt lang or are a poor UX/quality:
Screenshot 2024-02-14 at 10 59 12 AM
- if we use this translation model with correct settings in our quickstart example, most people will probably clone and have the settings already or can at least use our quickstart as a reference

### DIFF
--- a/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/remote_inference_client/text_summarization.py
+++ b/extensions/HuggingFace/python/src/aiconfig_extension_hugging_face/remote_inference_client/text_summarization.py
@@ -57,6 +57,14 @@ def refine_completion_params(model_settings: dict[Any, Any]) -> dict[str, Any]:
             parameters[key.lower()] = model_settings[key]
             completion_data["parameters"] = parameters
 
+    # The default model in HF is sshleifer/distilbart-cnn-12-6, which raises a
+    # KeyError, "summary_text", when calling the inference client.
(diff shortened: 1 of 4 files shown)


Ankush-lastmile:
> those that do either require the src and tgt lang

Could we specify defaults in the schema to resolve this?

Ankush-lastmile (approved):
lgtm

rholinshead:
> > those that do either require the src and tgt lang
> 
> Could we specify defaults for src & tgt lang in the schema to resolve this?

I don't think so, because then we will be setting that if people pick models that don't support it. Also it seems like the format is just a string and interpreted by the model so there's no guarantee it will work even for those that support the params

PR #1250, opened 2024-02-14
[gradio] Fix Spinner Colors

rholinshead:
[gradio] Fix Spinner Colors

# [gradio] Fix Spinner Colors

The loading spinner on the run button, download and share buttons is not clearly visible in gradio right now, especially in light mode:

https://github.com/lastmile-ai/aiconfig/assets/5060851/a7204248-1a8a-4f3a-8d12-e553dd66c03d

To fix, we need to override the stroke for the svg that's used. Also, for running prompt button, updated to make the square in the runPromptButton fill with the gradio color instead of white.

Test in light & dark themeMode:

https://github.com/lastmile-ai/aiconfig/assets/5060851/489745c0-f5bb-4f17-8cfd-d817adc78c9b

https://github.com/lastmile-ai/aiconfig/assets/5060851/e37e543c-c43d-43f6-bf22-4bce4ea08644

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1250).
* #1251
* __->__ #1250

PR #1251, opened 2024-02-14
[gradio] Fix Output List Padding

rholinshead:
[gradio] Fix Output List Padding

# [gradio] Fix Output List Padding

Gradio has specific output list padding and margin styles which override mantine's:
Screenshot 2024-02-14 at 3 29 31 PM

This causes ordered lists in outputs to extend outside the cell card padding:
Screenshot 2024-02-14 at 3 28 32 PM

We can fix by overriding with mantine's styles at higher specificity:
Screenshot 2024-02-14 at 3 29 58 PM

## Testing:
- Locally, editor output list looks the same on main and in this PR for local and gradio modes:
Screenshot 2024-02-14 at 3 26 34 PM
Screenshot 2024-02-14 at 3 26 48 PM

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1251).
* __->__ #1251
* #1250

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/prompt/prompt_outputs/PromptOutputsRenderer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/prompt/prompt_outputs/PromptOutputsRenderer.tsx
@@ -103,5 +103,11 @@ const OutputRenderer = memo(function Output({ output }: { output: Output }) {
 });
 
 export default memo(function PromptOutputsRenderer({ outputs }: Props) {
-  return outputs.map((output, i) => <OutputRenderer key={i} output={output} />);
+  return (
+    <Flex direction="column" className="outputContainer">
+      {outputs.map((output, i) => (
+        <OutputRenderer key={i} output={output} />
+      ))}
+    </Flex>
+  );
 });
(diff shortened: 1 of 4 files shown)


saqadri (approved):
Nice catch, this makes sense. I was pretty confused why padding for some cells was off but not others. Didn't think it was output dependent!

python/src/aiconfig/editor/client/src/components/prompt/prompt_outputs/PromptOutputsRenderer.tsx
```diff
@@ -103,5 +103,11 @@ const OutputRenderer = memo(function Output({ output }: { output: Output }) {
 });
 
 export default memo(function PromptOutputsRenderer({ outputs }: Props) {
-  return outputs.map((output, i) => <OutputRenderer key={i} output={output} />);
+  return (
+    <Flex direction="column" className="outputContainer">
```
rholinshead, line 107: This has no actual difference in its own, but adding this container element is good since we can use the className for modifying styles as needed (also seems 'correct' to have the outputs all under a root element containing them)

PR #1256, opened 2024-02-15
[editor] Fix Model Setting Default Values

rholinshead:
[editor] Fix Model Setting Default Values

# [editor] Fix Model Setting Default Values

Currently, the model settings state is retained when changing models in the selector instead of updating to the new defaults:

https://github.com/lastmile-ai/aiconfig/assets/5060851/1c38d6df-96d8-46c1-b164-65e9c16241b8

To fix, we can check the default value compared to the previous value and if they differ (i.e. new prompt schema default value) AND the initialValue (i.e. config value) is null/undefined then we reset the property value to the default:

https://github.com/lastmile-ai/aiconfig/assets/5060851/56837ac5-8d3a-4f52-8c4b-9e6ec0119c97

Also, make sure it works to reset value that is explicitly set:

https://github.com/lastmile-ai/aiconfig/assets/5060851/9a559ccd-32dc-4359-b016-befb27c84162

Some notes:
- General FYI of why we don't use an effect for this: https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes
- We are NOT doing the same when initialValue (i.e. from config) changes since that would put us in a weird loop where the property value is updating config value on debounced, then the updated initialValue would overwrite the property value

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/SettingsPropertyRenderer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/SettingsPropertyRenderer.tsx
@@ -68,6 +68,18 @@ export default function SettingsPropertyRenderer({
     initialValue ?? defaultValue
   );
 
+  // If the default value (i.e. default from schema) changes, update the property value if
+  // the config value is not set. This will handle cases such as changing the prompt model
+  // selector when the model settings tab is open
+  // NOTE: We can't do the same for initialValue vs prevInitialValue because initialValue
+  // (config value) is updated on debounced propertyValue changes and that would cause issues
+  // with overwriting user changes with config value as they type
+  const [prevDefaultValue, setPrevDefaultValue] = useState(defaultValue);
+  if (prevDefaultValue !== defaultValue && initialValue == null) {
+    setPrevDefaultValue(defaultValue);
+    setPropertyValue(defaultValue);
+  }
+
   let propertyControl;
 
   const setAndPropagateValue = useCallback(


python/src/aiconfig/editor/client/src/components/SettingsPropertyRenderer.tsx
```diff
@@ -68,6 +68,18 @@ export default function SettingsPropertyRenderer({
     initialValue ?? defaultValue
   );
 
+  // If the default value (i.e. default from schema) changes, update the property value if
+  // the config value is not set. This will handle cases such as changing the prompt model
+  // selector when the model settings tab is open
```
rossdanlm, line 73: Perhaps an alternative solution would be to call `getState` AIConfigContext to refresh the settings (or something similar), and connect that variable as a dependency in here: https://github.com/lastmile-ai/aiconfig/blob/e88309434bba9f5ad4e271a8afb2917e8cbeb846/python/src/aiconfig/editor/client/src/components/prompt/PromptsContainer.tsx#L81

^So here instead of directly calling the callback, we can do something like:

```typescript
const prompt = getPrompt(const statePrompt = getPrompt(stateRef.current, promptId);
const settingsCallback = useCallback(() => {
  // update settings with `prompt` data instead of props.updatePromptModelSettings
  
},
[prompt]
);

```

rholinshead: Sorry, I don't really follow

rholinshead: prompt data comes from the reducer so it would always be stale if we aren't using onUpdatePromptModelSettings (which itself sets the reducer value)

rossdanlm: I guess what I mean is, this won't change the `onUpdatePromptModelSettings`, which means that the settings component won't re-render. I agree this method is more performant since we don't need to re-render the entire settings component, only the field itself if default value has changed

rholinshead:
Ok, might have been some caching weirdness because the config value is properly being updated now when switching models (see screen recording in description)

rossdanlm (approved):
Thanks for the fix!

python/src/aiconfig/editor/client/src/components/SettingsPropertyRenderer.tsx
rholinshead: > I guess what I mean is, even though prompt is using updated state, this won't change the onUpdatePromptModelSettings callback itself, which means that the settings component won't re-render.

It does re-render, since `initialValue` is the value of what's in state for the prompt (so it changes, debounced, whenever you change a setting), just would have the same 'model' propertyValue persisted.

This particular scenario is pretty complicated, though, since it's really specific to the model right now. Long-term this would be fixed by having the schemas implemented in the parsers to properly handle how to update the config.

For now, we can do this or I think we can get creative and use 'key' prop to actually remount the component (and reset it entirely to match config state) -- sort of along the lines of how I suggested we could handle undo/redo for some components. Will put up an RFC for that

rholinshead: Ok, https://github.com/lastmile-ai/aiconfig/pull/1257 might be better?

rholinshead:
Closing in favour of #1257

PR #1257, opened 2024-02-15
[editor] Remount Model Settings Renderer on Model Change

rholinshead:
# [editor] Remount Model Settings Renderer on Model Change

Currently, the model settings state is retained when changing models in the selector instead of updating to the new defaults:

https://github.com/lastmile-ai/aiconfig/assets/5060851/1c38d6df-96d8-46c1-b164-65e9c16241b8

To fix, we can force the model settings renderer to remount when the prompt model value changes by passing the model name as the `key` prop to the component. Specifically, this handles the scenario outlined in discussion here: https://github.com/lastmile-ai/aiconfig/pull/1256#discussion_r1491743289

https://github.com/lastmile-ai/aiconfig/assets/5060851/60263416-dbb0-41d1-947b-be679e80bf75

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/prompt/PromptActionBar.tsx
+++ b/python/src/aiconfig/editor/client/src/components/prompt/PromptActionBar.tsx
@@ -4,6 +4,7 @@ import { ClientPrompt } from "../../shared/types";
 import {
   PromptSchema,
   checkParametersSupported,
+  getPromptModelName,
 } from "../../utils/promptUtils";
 import { ActionIcon, Container, Flex, ScrollArea, Tabs } from "@mantine/core";
 import { IconClearAll } from "@tabler/icons-react";
@@ -13,6 +14,7 @@ import { JSONObject } from "aiconfig";
 import { PROMPT_CONTAINER_HEIGHT_MAP } from "./PromptContainer";
 
 type Props = {
+  defaultConfigModelName?: string;
   prompt: ClientPrompt;
   promptSchema?: PromptSchema;
   onUpdateModelSettings: (settings: Record<string, unknown>) => void;
@@ -58,6 +60,7 @@ const MIN_ACTION_BAR_HEIGHT = 300;
 const ADD_PARAMETER_BOTTOM_HEIGHT = 46;
 
 export default memo(function PromptActionBar({
+  defaultConfigModelName,
   prompt,
   promptSchema,
   onUpdateModelSettings,
@@ -107,6 +110,7 @@ export default memo(function PromptActionBar({
                   style={{ overflowY: "auto" }}
                 >
                   <ModelSettingsRenderer
+                    model={getPromptModelName(prompt, defaultConfigModelName)}
                     settings={getModelSettings(prompt)}
                     schema={modelSettingsSchema}
                     onUpdateModelSettings={onUpdateModelSettings}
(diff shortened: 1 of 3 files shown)


rossdanlm (approved):

python/src/aiconfig/editor/client/src/components/prompt/model_settings/ModelSettingsRenderer.tsx
```diff
@@ -93,6 +97,8 @@ export default memo(function ModelSettingsRenderer({
         >
           {rawJSONToggleButton}
           <ModelSettingsSchemaRenderer
+            // Use the model name as the key to force re-mount / updated state when model changes
```
rossdanlm, line 100: Could this cause an issue if we have multiple props where we re-use the default model and hence have the same key for two separate `ModelSettingsSchemaRenderer` objects? Or is React smart enough to understand the hierarchy and know that even though we set the same keys, they're not the same objects?

python/src/aiconfig/editor/client/src/components/prompt/PromptActionBar.tsx
```diff
@@ -107,6 +110,7 @@ export default memo(function PromptActionBar({
                   style={{ overflowY: "auto" }}
                 >
                   <ModelSettingsRenderer
+                    model={getPromptModelName(prompt, defaultConfigModelName)}
```
rossdanlm, line 113: I looked at the `getPromptModelName` method in https://github.com/lastmile-ai/aiconfig/blob/1c68007b971896e74c31fb68917cff46d22bb769/python/src/aiconfig/editor/client/src/utils/promptUtils.ts#L34

Just to sanity check, the value of this doesn't quite "matter" (since it's the Hugging Face Task (`"Text Generation"`) and not the model itself (`"mistralai/Mistral-7B-Instruct-v0.1"`)). All that matters is that we have changed the HF task (which for most cases will be the actual model itself like `dall-e 2`) and that changed value will be detected by the key and cause that component to re-render the settings right?

rholinshead: Ya, it's technically the model id that's registered in the ModelParserRegistry. And yep, that's correct (just nit on terminology, it would re-mount instead of re-render)

rossdanlm: This is beyond scope of this PR, just curious, what's the difference between re-mount vs. re-render and why does changing the key cause it to re-mount? Is Chat GPT wrong here? https://lastmileai.dev/workbooks/clsoufolp001pqrcsg97aj5dr

python/src/aiconfig/editor/client/src/components/prompt/model_settings/ModelSettingsRenderer.tsx
rossdanlm: > multiple props where we re-use the default model

Sorry typo, meant to say "multiple prompts"

python/src/aiconfig/editor/client/src/components/prompt/PromptActionBar.tsx
rholinshead: Yep, it's wrong -- see https://react.dev/reference/react/useState#resetting-state-with-a-key

> You can reset a component’s state by passing a different key to a component. In this example, the Reset button changes the version state variable, which we pass as a key to the Form. When the key changes, React re-creates the Form component (and all of its children) from scratch, so its state gets reset.

rholinshead: Re-mount basically means the component is removed from the DOM, recreated and re-added (state, etc. is re-initialized). Re-render just means updating the existing component (state, etc. is persisted)

PR #1260, opened 2024-02-16
[editor][rfc] Default to readOnly Until Server Is Working

rholinshead:
[editor][rfc] Default to readOnly Until Server Is Working

[editor][rfc] Default to readOnly Until Server Is Working

# [editor][rfc] Default to readOnly Until Server Is Working

When opening an aiconfig file, we currently open the webview with editable editor component before the server has successfully loaded the config content into state, then show an error message, `"Failed to start aiconfig server. You can view the aiconfig but cannot modify it."` with Retry button. A couple issues with the current state:
- Retry button is only shown once and can't retry again afterwards if it fails again
- Having editable state while the server isn't ready means the user can (confusingly) modify the UI, causing error messages and the changes NOT being reflected in the actual document (because we push to the document from the server)

https://github.com/lastmile-ai/aiconfig/assets/5060851/8ffd0b21-918a-49cf-ab9d-4f825314b952

We can improve this UX by:
- Making sure 'Retry' button is shown again (recursively) if the user clicks it and it still errors
- Setting default editor state to readOnly and updating it to editable only after the server successfully loads the content (and is ready to be used)
- NOTE: To prevent erroneous readOnly state when a non-working server state is "fixed", we assume that successful /load request means the server state is working, and set readOnly to false in that case as well

https://github.com/lastmile-ai/aiconfig/assets/5060851/d29b8b29-d44e-41d4-95d8-f08d06929fe9

## Testing:
- Made an aiconfig with model parsers that aren't loaded (e.g. from HF extension without the extension installed)
- Open the config, see readonly and error message. Click retry a few times, still shown to retry again after erroring again
- Update the config JSON (in another tab) to remove the bad parsers and fix the error. Click 'Retry', see success and now editable editor state
- Toggle between working server editor webview (editable) and one with errors (readonly) and back, ensuring the webview state is properly resaturated

### DIFF
--- a/vscode-extension/editor/src/VSCodeEditor.tsx
+++ b/vscode-extension/editor/src/VSCodeEditor.tsx
@@ -57,6 +57,12 @@ export default function VSCodeEditor() {
     webviewState?.theme
   );
 
+  // Default to readOnly until we receive a message from the extension host
+  // confirming it is safe to edit the content
+  const [isReadOnly, setIsReadOnly] = useState<boolean>(
+    webviewState?.isReadOnly ?? true
+  );
+
   const { classes } = useStyles();
 
   const updateContent = useCallback(async (text: string) => {
@@ -96,6 +102,14 @@ export default function VSCodeEditor() {
         updateContent(text);
         return;
       }
+      case "set_readonly_state": {
+        const isReadOnlyState = message.isReadOnly;
+        if (isReadOnlyState != null && isReadOnlyState !== isReadOnly) {
+          setIsReadOnly(isReadOnlyState);
+          updateWebviewState(vscode, { isReadOnly: isReadOnlyState });
+        }
+        return;
+      }
       case "set_server_url": {
         console.log("onMessage, message=", JSON.stringify(message));
         const url = message.url;
@@ -124,7 +138,10 @@ export default function VSCodeEditor() {
     const res = await ufetch.post(route, {});
     // console.log(`IN LOAD: route=${route}, res=${JSON.stringify(res)}`);
     setAIConfig(res.aiconfig);
-  }, [aiConfigServerUrl]);
+    // If we can load the config from the server, we assume it's in a good state
+    setIsReadOnly(false);
+    updateWebviewState(vscode, { isReadOnly: false });
+  }, [aiConfigServerUrl, vscode]);
 
   useEffect(() => {
     if (aiConfigServerUrl !== "") {
@@ -475,6 +492,7 @@ export default function VSCodeEditor() {
           aiconfig={aiconfig}
           callbacks={callbacks}
           mode={MODE}
+          readOnly={isReadOnly}
           themeOverride={VSCODE_THEME}
           themeMode={themeMode}
         />
--- a/vscode-extension/src/aiConfigEditor.ts
+++ b/vscode-extension/src/aiConfigEditor.ts
@@ -124,6 +124,55 @@ export class AIConfigEditorProvider implements vscode.CustomTextEditorProvider {
 
     let isWebviewDisposed = false;
 
+    function updateServerWithRetry(
+      serverUrl: string,
+      document: vscode.TextDocument
+    ) {
+      updateServerState(serverUrl, document)
+        .then(() => {
+          // In case of previous failure, reset to editable state
+          webviewPanel.webview.postMessage({
+            type: "set_readonly_state",
+            isReadOnly: false,
+          });
+        })
+        .catch((e) => {
+          if (isWebviewDisposed) {
+            // Ignore errors caused by closing the webview while the server update
+            // request is in flight (e.g. closing config w/ unsaved changes will
+            // emit onDidChangeTextDocument with reverted unsaved changes)
+            // See #1201 for full details.
+            console.info(
+              "Ignoring server update error due to webview disposal"
+            );
+            return;
+          }
+
+          webviewPanel.webview.postMessage({
+            type: "set_readonly_state",
+            isReadOnly: true,
+          });
+
+          vscode.window
+            .showErrorMessage(
+              "Failed to update aiconfig server. You can view the aiconfig but cannot modify it.",
+              ...["Details", "Retry"]
+            )
+            .then((selection) => {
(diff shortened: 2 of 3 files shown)


saqadri (approved):
Love this approach. Thanks for thinking this through Ryan!

rholinshead:
@saqadri sorry, I had it in review but moved back to draft for now because I'm not sure what we want to do in terms of persisting readOnly in webview. Using the scenario in the recording where I have tab A with the raw config JSON and tab B for the editor webview:
- If we persist it in state as we're doing here, fixing the JSON in tab A correctly gets the server state working but toggling back to tab B we see the webview is still readOnly. This is because `resolveCustomTextEditor` is not called again when toggling back to the tab since the editor was already created there. In the video it just happens to be that the error message with 'retry' is still shown so clicking that is what sets off the 'readOnly' toggle. If I had actually dismissed that instead then toggling back to tab B would have the readOnly editor with no way to get it back.  I made also just made the same change in `changeDocumentSubscription` to toggle isReadonly state on success/failure of the `updateServerState` call (see latest update). But when I change tab A to fix the config I see it getting called but since tab B is not open I assume the webview message is ignored since the editor component isn't mounted to handle it and so it stays readonly

- If we don't persist it in state (and default isReadOnly to false, only setting to true on error) the UX seems to work as intended in the above scenario. HOWEVER, I'm almost certain this will result in a readOnly editor with bad server becoming editable if you toggle to another tab and back to it (can't test that since that flow currently results in loading spinner since it fails to load the config -- side note, do you know how tab B with bad server is loading the config in the first place right now if all the /load requests fail?)

rholinshead:
> @saqadri sorry, I had it in review but moved back to draft for now because I'm not sure what we want to do in terms of persisting readOnly in webview. Using the scenario in the recording where I have tab A with the raw config JSON and tab B for the editor webview:
> 
> * If we persist it in state as we're doing here, fixing the JSON in tab A correctly gets the server state working but toggling back to tab B we see the webview is still readOnly. This is because `resolveCustomTextEditor` is not called again when toggling back to the tab since the editor was already created there. In the video it just happens to be that the error message with 'retry' is still shown so clicking that is what sets off the 'readOnly' toggle. If I had actually dismissed that instead then toggling back to tab B would have the readOnly editor with no way to get it back.  I made also just made the same change in `changeDocumentSubscription` to toggle isReadonly state on success/failure of the `updateServerState` call (see latest update). But when I change tab A to fix the config I see it getting called but since tab B is not open I assume the webview message is ignored since the editor component isn't mounted to handle it and so it stays readonly
> * If we don't persist it in state (and default isReadOnly to false, only setting to true on error) the UX seems to work as intended in the above scenario. HOWEVER, I'm almost certain this will result in a readOnly editor with bad server becoming editable if you toggle to another tab and back to it (can't test that since that flow currently results in loading spinner since it fails to load the config -- I think we can persist the config in context to show on re-saturation, after implementing the update from aiconfig prop in [[editor] Support Updating AIConfig State when aiconfig Prop Changes #1261](https://github.com/lastmile-ai/aiconfig/pull/1261))
>   ~Side note, do you know how tab B with bad server is loading the config in the first place right now if all the /load requests fail?~ ah, from 'update'

Ok, so updated here to set readOnly to false in `loadConfig` with the assumption that loading the config from server state implies the server is in a working state

vscode-extension/editor/src/VSCodeEditor.tsx
```diff
@@ -124,6 +138,8 @@ export default function VSCodeEditor() {
     const res = await ufetch.post(route, {});
     // console.log(`IN LOAD: route=${route}, res=${JSON.stringify(res)}`);
     setAIConfig(res.aiconfig);
+    // If we can load the config from the server, we assume it's in a good state
+    setIsReadOnly(false);
```
saqadri, line 142: Should we also update the webview state in this case?
i.e. `updateWebviewState(vscode, { isReadOnly: false });`

vscode-extension/src/aiConfigEditor.ts
```diff
@@ -124,6 +124,55 @@ export class AIConfigEditorProvider implements vscode.CustomTextEditorProvider {
...
+            isReadOnly: false,
+          });
+        })
+        .catch((e) => {
+          if (isWebviewDisposed) {
+            // Ignore errors caused by closing the webview while the server update
+            // request is in flight (e.g. closing config w/ unsaved changes will
+            // emit onDidChangeTextDocument with reverted unsaved changes)
+            // See #1201 for full details.
+            console.info(
```
saqadri, line 145: Just checking if this should be piped to the output channel or if this is just a console log for debugging

vscode-extension/src/aiConfigEditor.ts
```diff
@@ -163,39 +212,8 @@ export class AIConfigEditorProvider implements vscode.CustomTextEditorProvider {
...
-                      e?.message ?? JSON.stringify(e)
-                    );
-                    this.extensionOutputChannel.show(/*preserveFocus*/ true);
-                  }
-                  if (selection === "Retry") {
-                    updateServerState(editorServer.url, e.document);
-                  }
-                });
-            }
+            console.log("changeDocumentSubscription -- updating server");
```
saqadri, line 215: nit: remove

saqadri (approved):
I think this looks good! The main concern I have is with saving the readonly state in the webview state, and not updating it properly. I think the way you have it should work, but have a comment about that to confirm.

Overall, ship it!

vscode-extension/editor/src/VSCodeEditor.tsx
rholinshead: Ah, good catch - yep!

vscode-extension/src/aiConfigEditor.ts
rholinshead: This is mainly for debugging, I don't think this would be useful for users to see

vscode-extension/editor/src/VSCodeEditor.tsx
rholinshead: updated

PR #1261, opened 2024-02-20
[editor] Support Updating AIConfig State when aiconfig Prop Changes

rholinshead:
[editor] Support Updating AIConfig State when aiconfig Prop Changes

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
+++ b/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
@@ -155,7 +155,7 @@ export type AIConfigCallbacks = {
 type RequestCallbackError = { message?: string };
 
 function AIConfigEditorBase({
-  aiconfig: initialAIConfig,
+  aiconfig: providedAIConfig,
   callbacks,
   mode,
   readOnly = false,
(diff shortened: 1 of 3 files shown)


saqadri (approved):
Incredible writeup of the various cases and options considered. This change looks good to me. Main question -- did you check this in the vscode editor? I tried this from #960 when I was building the VSCode editor, and it didn't seem to propagate updates properly. However, it's possible that some of the other fixes you made to re-render as state changes were the missing piece.

rholinshead:
> Incredible writeup of the various cases and options considered. This change looks good to me. Main question -- did you check this in the vscode editor? I tried this from #960 when I was building the VSCode editor, and it didn't seem to propagate updates properly. However, it's possible that some of the other fixes you made to re-render as state changes were the missing piece.

Ah, no I didn't try it out since I'm not sure the best way to do that besides pulling my aiconfig-editor package PR and rebasing this on top and linking it; how did you test out #960 before?
Fwiw, the change of overwriting the state (mainly, the fact that the ClientPrompts have new ids/keys) is what will 'fix' the issue you saw

saqadri:
> how did you test out https://github.com/lastmile-ai/aiconfig/pull/960 before?

With difficulty. I copied the VSCodeEditor.tsx (and related files) into the aiconfig-editor folder, built the main.js bundle from there, and copied it into the vscode-extension folder lol. Probably overkill and better to just publish a package update.

One thing I'd suggest is we should have a test-aiconfig-editor package on npm that we update to test stuff out as needed. But also feel free to push an update to the main package directly right now to test it out.

rholinshead:
> > how did you test out #960 before?
> 
> With difficulty. I copied the VSCodeEditor.tsx (and related files) into the aiconfig-editor folder, built the main.js bundle from there, and copied it into the vscode-extension folder lol. Probably overkill and better to just publish a package update.
> 
> One thing I'd suggest is we should have a test-aiconfig-editor package on npm that we update to test stuff out as needed. But also feel free to push an update to the main package directly right now to test it out.

Ah, ok, that's what I feared :s 

I was planning to publish and bump the minor version with this since it does add some new behaviour that changes how the editor behaves.

I can publish to a test package and update test plan for vscode, but will land first so it's easier to do that without dealing with additional merge issues

PR #1262, opened 2024-02-20
[vscode installation] 1/n Python version validation

Ankush-lastmile:
[vscode installation] 1/n Python version validation

Add a helper method to validate the python interpreter version. Simply throws an error on a bad version. Notification dialog in diff on top.

Align's with Python AIConfig's version requirement of python3.10 and higher

## Testplan

Select a python interpreter 3.9, which is less than aiconfig's requirement version of 3.10.

https://github.com/lastmile-ai/aiconfig/assets/141073967/4cfcb153-50aa-40fb-9764-54e964d4b1b8

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1262).
* #1265
* __->__ #1262

### DIFF
--- a/vscode-extension/src/extension.ts
+++ b/vscode-extension/src/extension.ts
@@ -18,6 +18,7 @@ import {
   getPythonPath,
   isSupportedConfigExtension,
   SUPPORTED_FILE_EXTENSIONS,
+  isPythonVersionAtLeast310,
 } from "./util";
 import { AIConfigEditorProvider } from "./aiConfigEditor";
 import { AIConfigEditorManager } from "./aiConfigEditorManager";
@@ -673,6 +674,12 @@ async function checkPython() {
             }
           });
         resolve(false);
+      } else if (!isPythonVersionAtLeast310(pythonPath)) {
+        console.error(
+          "Python version is not 3.10 or higher. Please upgrade to Python 3.10 or higher."
+        );
+        resolve(false);
+        // show guide for installation
       } else {
         resolve(true);
         console.log("Python is installed");
--- a/vscode-extension/src/util.ts
+++ b/vscode-extension/src/util.ts
@@ -1,5 +1,5 @@
 import * as vscode from "vscode";
-import type { ChildProcessWithoutNullStreams } from "child_process";
+import { type ChildProcessWithoutNullStreams, execSync } from "child_process";
 import { setTimeout } from "timers/promises";
 import { ufetch } from "ufetch";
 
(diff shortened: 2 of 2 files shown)


vscode-extension/src/extension.ts
```diff
@@ -673,7 +674,13 @@ async function checkPython() {
...
           });
         resolve(false);
-      } else {
+      }
+      else if(!isPythonVersionAtLeast310(pythonPath)){
+        console.error("Python version is not 3.10 or higher. Please upgrade to Python 3.10 or higher.");
+        resolve(false);
+        // show guide for installation
+      } 
+      else if(stdout) {
```
rholinshead, line 683: why the added stdout check here?

vscode-extension/src/extension.ts
```diff
@@ -812,4 +819,4 @@ async function shareAIConfig(
         }
       });
   }
-}
+}
```
rholinshead, line 822: nit, newline. Do you have prettier running?

vscode-extension/src/extension.ts
```diff
@@ -673,7 +674,13 @@ async function checkPython() {
             }
           });
         resolve(false);
-      } else {
+      }
+      else if(!isPythonVersionAtLeast310(pythonPath)){
```
rholinshead, line 678:
```suggestion
      else if (!isPythonVersionAtLeast310(pythonPath)) {
```

vscode-extension/src/util.ts
```diff
@@ -301,3 +301,22 @@ export async function getPythonPath(): Promise<string> {
...
   return pythonPath;
 }
+
+/**
+ * Checks if the specified Python interpreter is version 3.10 or higher.
+ * @param pythonPath The path to the Python interpreter.
+ * @returns A promise that resolves to a boolean indicating if the version is >= 3.10.
+ */
+
+export function isPythonVersionAtLeast310(pythonPath: string): boolean {
```
rholinshead, line 311: I think this is ok to unblock but ideally if there's some way we can determine the required python version dynamically via pip or something that would be best

vscode-extension/src/util.ts
```diff
@@ -301,3 +301,22 @@ export async function getPythonPath(): Promise<string> {
...
+  try {
+    // Use Python to check compatible version
+    const command = `${pythonPath} -c "import sys; print(sys.version_info >= (3, 10))"`;
```
rholinshead, line 314: Why this command vs just doing `--version` check? IMO that would be cleaner so that we can do the comparison in code ourselves

rholinshead: Actually, we already should have that version from stdout above, right? I don't think we need to run any python here at all

rholinshead (changes requested):
Can we just use the python version from stdout and compare it to a constant (with note to keep constant in sync with our aiconfig package and maybe a TODO to look into getting it dynamically from pip)

vscode-extension/src/util.ts
Ankush-lastmile: executing `python --version` returns a literal like `Python 3.11.7`. 

Felt like parsing this into a version number and then comparing was trickier than using python itself to validate version. 

I can update if that would be easier to follow

vscode-extension/src/util.ts
Ankush-lastmile: what do you mean by dynamically? also pip itself has versions, which is not the same as python's version

rholinshead: Oh, what I mean is our pyproject.toml defines `requires-python = ">=3.10"` and that shows up as metadata in pip. Ideally we could dynamically get that version in the code without having to hardcode it

vscode-extension/src/util.ts
rholinshead: Ah, ok. I guess there are some nuances with it as well (e.g. comparator of version numbers). I'm fine with keeping as-is using python to check

rholinshead:
Ok, so let's keep the check logic as-is but please address the other comments:
- prettier (newline, spacing)
- remove stdout check in that else if not needed

On top of that, please update the comment to give more context on using python to check version (namely complexity of different version values and how they would need to be parsed and compared)

rholinshead (approved):
Accepting to unblock once latest comment is addressed

vscode-extension/src/extension.ts
Ankush-lastmile: removed, not needed

vscode-extension/src/extension.ts
Ankush-lastmile: my prettier settings were changed, I reset it to default settings and it worked like a charm

Ankush-lastmile:
- rebase
- fixed formatting
- updated comment to be clearer

PR #1264, opened 2024-02-20
Update changelog

rossdanlm:
Update changelog

TSIA

Also ran yarn build on editor/client


CHANGELOG.md
```diff
@@ -1,14 +1,33 @@
 # Changelog
 
+## (2024-02-20) Python Version 1.1.22, NPM Version 1.1.9
```
rholinshead, line 3: Shouldn't this be `1.1.26` and `1.1.8`?

rholinshead (approved):
Accepting to unblock but please make sure the versions are correct

CHANGELOG.md
rossdanlm: Yea good catch, the version is actually 1.1.10 (I skipped 1.1.8 one week to 1.1.9 and surprisingly npm didn't error with monotonic)

PR #1265, opened 2024-02-20
[vscode installation] 2/n refactor error message for invalid python install

Ankush-lastmile:
[vscode installation] 2/n refactor error message for invalid python install

Added a helper method for showing user friendly error when python install doesn't go well.

## Testplan

Select Interpreter version 3.9.4. Open an editor and expect an error notification.

https://github.com/lastmile-ai/aiconfig/assets/141073967/5bda86e0-690f-4e43-9e94-f9279e3b2f34

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1265).
* #1294
* #1289
* #1290
* __->__ #1265

### DIFF
--- a/vscode-extension/src/extension.ts
+++ b/vscode-extension/src/extension.ts
@@ -19,6 +19,7 @@ import {
   isSupportedConfigExtension,
   SUPPORTED_FILE_EXTENSIONS,
   isPythonVersionAtLeast310,
+  showGuideForPythonInstallation,
 } from "./util";
 import { AIConfigEditorProvider } from "./aiConfigEditor";
 import { AIConfigEditorManager } from "./aiConfigEditorManager";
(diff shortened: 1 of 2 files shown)


vscode-extension/src/util.ts
```diff
@@ -319,4 +319,16 @@ export function isPythonVersionAtLeast310(pythonPath: string): boolean {
     console.debug("Error checking Python version:", error);
     return false;
   }
-}
\ No newline at end of file
+}
+
+export function showGuideForInstallation(message: string): void {
```
rholinshead, line 324: nit -- forPythonInstallation

vscode-extension/src/extension.ts
```diff
@@ -657,28 +658,16 @@ async function checkPython() {
...
-          .then((selection) => {
-            if (selection === "Install Python") {
-              vscode.env.openExternal(
-                vscode.Uri.parse("https://www.python.org/downloads/")
-              );
-            } else if (selection === "Retry") {
-              vscode.commands.executeCommand(COMMANDS.INIT);
-            }
-          });
+        showGuideForInstallation("Specified Python Interpreter is not valid");
```
rholinshead, line 664: What does it mean for the interpreter to not be valid? To me, this message doesn't really connect to the python installation guide. My thought process would be: "is the interpreter not valid because it's pointing to incorrect path or is python not installed, or what?"

vscode-extension/src/util.ts
Ankush-lastmile: updated

vscode-extension/src/extension.ts
Ankush-lastmile: > What does it mean for the interpreter to not be valid

It would mean something more along the lines of "it's pointing to an incorrect path". Since a python interpreter/path will be specified at this point, I don't think its valid to say python is not installed since this code doesn't check for existence of python.

What would you recommend as a better message here?

rholinshead: I think the message makes sense but then the guide for python installation does not. Instead, they should probably be informed of how to select or configure a working interpreter

Ankush-lastmile: I think that makes sense, This diff is just refactoring into into helper method using existing logic.

vscode-extension/src/util.ts
```diff
@@ -322,3 +322,18 @@ export function isPythonVersionAtLeast310(pythonPath: string): boolean {
     return false;
   }
 }
+
+export function showGuideForPythonInstallation(message: string): void {
+  // Guide for installation
```
rossdanlm, line 327: You don't need this lmao

rossdanlm (approved):
Lgtm

Ankush-lastmile:
merging this diff, requested changes can come on top

PR #1272, opened 2024-02-20
Add Lint check to ci/cd

Ankush-lastmile:
Add Lint check to ci/cd

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,25 @@
+name: Continuous Integration
+
+# This action works with pull requests and pushes
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+
+jobs:
+  prettier:
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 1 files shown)


rholinshead (commented):
This is just running prettier, not actually linting, right? I would expect lint with `yarn`:
```
      - name: Lint
        run: |
          yarn lint
```

Ankush-lastmile:
> This is just running prettier, not actually linting, right? I would expect lint with yarn:

Yep, was just playing around with the github action from the marketplaces to figure out how it works. Its great and its extensible enough to be used with python and typescript. Running the python linter bubbles up 150+ exceptions 😅

Ankush-lastmile:
> This is just running prettier, not actually linting, right? I would expect lint with yarn:

But I was trying to run linting with prettier rules, not sure if that should be done in addition to yarn lint

Ankush-lastmile:
closing for now, will revisit later

PR #1273, opened 2024-02-21
[RFC] Initial walkthrough content for VS Code extension

rossdanlm:
[RFC] Initial walkthrough content for VS Code extension

Following the guide from VSCode tutorials here: https://code.visualstudio.com/api/references/contribution-points#contributes.walkthroughs

This is now just extremely barebones, want to get some comments on what other commands we feel should be included.

Also note that I can also add in images and markdown to the right side of panel for each item that's selected

Let's just get this out really quickly so that Tanya + Zak can help out with content + graphics in the markdown files

## Test Plan

https://github.com/lastmile-ai/aiconfig/assets/151060367/5827b98c-ac1c-4454-a337-ce9eb3c44b1a

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1273).
* #1276
* __->__ #1273

PR #1274, opened 2024-02-21
Activate extension after VSCode startup

rossdanlm:
Activate extension after VSCode startup

This will trigger the python dependencies to be installed, saving time for when we first open an AIConfig file. This will call the `activate()` method in the `extension.ts` file, which contains this line:

```
  // Run the setup command on activation
  vscode.commands.executeCommand(COMMANDS.INIT);
```

## Test Plan
No functional changes, but activation now gets triggered right away instead of opening an AIConfig file. This still takes a bit of time to startup because we need to run the flask server, but if I didn't have dependencies installed, it would be huge time savings!

Before

https://github.com/lastmile-ai/aiconfig/assets/151060367/c164d180-c7ce-4062-bbe9-5452449e0ffe

After

https://github.com/lastmile-ai/aiconfig/assets/151060367/fb45a79a-7593-4745-bfa2-41fa0856470c

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1274).
* __->__ #1274
* #1273

### DIFF
--- a/vscode-extension/package.json
+++ b/vscode-extension/package.json
@@ -15,6 +15,9 @@
     "Other"
   ],
   "main": "./out/extension.js",
+  "activationEvents": [
+    "onStartupFinished"
+  ],
   "contributes": {
     "commands": [
       {


rossdanlm:
Synced w. Sarmad offline, decided not to do this because we can get flagged by VS Code for taking up startup time

Screenshot 2024-02-21 at 14 28 44

saqadri:
We can revisit this based on how we structure extension activation. Currently because it's heavyweight and involves installing a lot of dependencies, we should try to activate the extension only when necessary.

Some more context on this from official vscode docs:

image

image

`onStartupFinished` is better than `*`, but still may lead to a suboptimal user experience overall.

rossdanlm:
> image 

> It is best practice to activate only when a user needs your extension. If your extension works on a subset of languages, it is better for the user to list that subset than to activate on all languages.

This part was specifically talking about activating on language flag and advising against using a catch-all 
```
"activationEvents": [
    "onLanguage"
]
```

I don't think it's clear from the docs what they recommend as best practice for overall activation. I also feel this is a one-time cost of installing dependencies once (and margin install deltas whenever we update our `requirements.txt` files) so not actually doing anything expensive except for initial dependency installs

PR #1280, opened 2024-02-21
[vscode][2/n] Setup Env Variables: Define env path

rossdanlm:
[vscode][2/n] Setup Env Variables: Define env path

Opening a window for user to enter in an `.env` path. Using the home dir for now, but can change to workspace if you want, not a big deal.

Next diffs will implement a static file template and use that when env path doesn't exist. Diff after that I'll deal with existing env path situation

## Test Plan

https://github.com/lastmile-ai/aiconfig/assets/151060367/cceadc3a-cd20-4164-b8b2-c8d11b36b5bb

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1280).
* #1284
* #1283
* __->__ #1280

### DIFF
--- a/vscode-extension/src/extension.ts
+++ b/vscode-extension/src/extension.ts
@@ -59,7 +59,7 @@ export function activate(context: vscode.ExtensionContext) {
     vscode.commands.registerCommand(
       COMMANDS.SETUP_ENVIRONMENT_VARIABLES,
       () => {
-        vscode.window.showInformationMessage("Will implement next PR");
+        setupEnvironmentVariables();
       }
     )
   );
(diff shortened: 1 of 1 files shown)


vscode-extension/src/extension.ts
```diff
@@ -716,6 +716,95 @@ async function checkPip() {
...
+  //     "Please customize your new model registry."
+  //   );
+  // }
+
+  // let config = vscode.workspace.getConfiguration(EXTENSION_NAME);
+  // await handleCustomModelRegistryUpdate(
+  //   config,
+  //   aiconfigEditorManager,
+  //   modelRegistryPath
+  // );
```
rossdanlm, line 805: Will delete later in stack, just doing small PRs now to get more quick feedback

Ankush-lastmile (approved):
lgtm, approving to unblock

PR #1282, opened 2024-02-21
[editor][ez] Fix Resaturation of Editor with Bad Server State

rholinshead:
[editor][ez] Fix Resaturation of Editor with Bad Server State

# [editor][ez] Fix Resaturation of Editor with Bad Server State

Currently, when re-saturating the webview for an editor with bad server state (e.g. invalid model parsers in the config), the /load call will fail and leave the webview with the loading spinner:

https://github.com/lastmile-ai/aiconfig/assets/5060851/3949274d-fbb1-479f-a54a-9ac8870e617c

To fix, we can serialize the config in webview context and then use it to re-saturate the aiconfig state in the webview:

https://github.com/lastmile-ai/aiconfig/assets/5060851/5d67dc31-7b68-4782-af73-a8b6294a5452

For a webview with working state, note that /load is called successfully after the webview loads and will update the aiconfig state in the editor as needed (now that aiconfig prop updates the editor state)

### DIFF
--- a/vscode-extension/editor/src/VSCodeEditor.tsx
+++ b/vscode-extension/editor/src/VSCodeEditor.tsx
@@ -46,9 +46,10 @@ export default function VSCodeEditor() {
 
   // TODO: saqadri - does this need to be wrapped in a memo?
   const webviewState = getWebviewState(vscode);
-  console.log(`webviewState = ${JSON.stringify(webviewState)}`);
 
-  const [aiconfig, setAIConfig] = useState<AIConfig | undefined>();
+  const [aiconfig, setAIConfig] = useState<AIConfig | undefined>(
+    webviewState?.aiconfigState
+  );
   const [aiConfigServerUrl, setAIConfigServerUrl] = useState<string>(
     webviewState?.serverUrl ?? ""
   );
@@ -65,23 +66,25 @@ export default function VSCodeEditor() {
 
   const { classes } = useStyles();
 
-  const updateContent = useCallback(async (text: string) => {
-    // TODO: saqadri - this won't work for YAML -- the handling of the text needs to include the logic from AIConfig.load
+  const updateContent = useCallback(
+    async (text: string) => {
+      if (text != null) {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        let updatedConfig: any = {};
+        try {
+          // Try parsing the string as JSON first
+          updatedConfig = JSON.parse(text);
+        } catch (e) {
+          // If that fails, try parsing the string as YAML
+          updatedConfig = yaml.load(text);
+        }
 
-    if (text != null) {
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      let updatedConfig: any = {};
-      try {
-        // Try parsing the string as JSON first
-        updatedConfig = JSON.parse(text);
-      } catch (e) {
-        // If that fails, try parsing the string as YAML
-        updatedConfig = yaml.load(text);
+        setAIConfig(updatedConfig);
+        updateWebviewState(vscode, { aiconfigState: updatedConfig });
       }
-
-      setAIConfig(updatedConfig);
-    }
-  }, []);
+    },
+    [vscode]
+  );
 
   // Register an event listener to handle messages from the extension host
   // This is how we'll receive updates to the webview's content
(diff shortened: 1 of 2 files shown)


vscode-extension/editor/src/VSCodeEditor.tsx
```diff
@@ -65,23 +67,27 @@ export default function VSCodeEditor() {
...
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      let updatedConfig: any = {};
-      try {
-        // Try parsing the string as JSON first
-        updatedConfig = JSON.parse(text);
-      } catch (e) {
-        // If that fails, try parsing the string as YAML
-        updatedConfig = yaml.load(text);
+        setAIConfig(updatedConfig);
+        updateWebviewState(vscode, { aiconfigState: updatedConfig });
```
rholinshead, line 86: This and adding vscode to deps array is the only change to this callback. The rest is auto-formatting

vscode-extension/editor/src/utils/vscodeUtils.ts
```diff
@@ -10,7 +9,7 @@ import { JSONValue } from "aiconfig";
  * For more information, see https://code.visualstudio.com/api/extension-guides/webview#getstate-and-setstate
  */
 export type WebviewState = {
-  aiconfigState?: ClientAIConfig;
+  aiconfigState?: AIConfig;
```
rholinshead, line 12: This is the proper type since this comes directly from the aiconfig server. ClientAIConfig is exclusive to the editor component since it contains _ui fields

vscode-extension/editor/src/VSCodeEditor.tsx
```diff
@@ -65,23 +67,27 @@ export default function VSCodeEditor() {
 
   const { classes } = useStyles();
 
-  const updateContent = useCallback(async (text: string) => {
-    // TODO: saqadri - this won't work for YAML -- the handling of the text needs to include the logic from AIConfig.load
+  const updateContent = useCallback(
+    async (text: string) => {
+      // TODO: saqadri - this won't work for YAML -- the handling of the text needs to include the logic from AIConfig.load
```
saqadri, line 72: nit: outdated comment

vscode-extension/editor/src/VSCodeEditor.tsx
```diff
@@ -48,7 +48,9 @@ export default function VSCodeEditor() {
   const webviewState = getWebviewState(vscode);
   console.log(`webviewState = ${JSON.stringify(webviewState)}`);
```
saqadri, line 49: nit: remove -- with aiconfig stored in state it may be very noisy

saqadri (approved):
I think this is good. The only two concerns I have:

1: in the case you trigger an execution, then navigate away from the tab, and come back, there is likely a period of time during which webview state AIConfig is inconsistent with server state -- in that case I think /load will make it eventually consistent.

2: size of webview state. Would recommend testing with huggingface.aiconfig.json which is 8MB I think.

I don't see any other way to solve this, but just wanted to call out some considerations (and why I had avoided saving the aiconfig in webview state)

rholinshead:
> I think this is good. The only two concerns I have:
> 
> 1: in the case you trigger an execution, then navigate away from the tab, and come back, there is likely a period of time during which webview state AIConfig is inconsistent with server state -- in that case I think /load will make it eventually consistent.
> 
> 2: size of webview state. Would recommend testing with huggingface.aiconfig.json which is 8MB I think.
> 
> I don't see any other way to solve this, but just wanted to call out some considerations (and why I had avoided saving the aiconfig in webview state)

Ah, ok. I had assumed it wasn't storing the config in state because of the prop not updating the editor state, but size constraint also makes sense. `huggingface.aiconfig.json` loads/resaturates quickly enough. Maybe we can add logic in the future to not persist state over some threshold size?

For 1. I just tested and toggling back currently only has updated output if the execution completed while toggled away. Otherwise, the outputs are not updated if toggling back before they complete (obviously streaming callback doesn't connect editor to the server anymore, but going to look into why it's not being updated as a result of the server updating the config since I would expect that to update the document itself. Maybe `isInternalDocumentChange` is true in this case

saqadri:
> > I think this is good. The only two concerns I have:
> > 1: in the case you trigger an execution, then navigate away from the tab, and come back, there is likely a period of time during which webview state AIConfig is inconsistent with server state -- in that case I think /load will make it eventually consistent.
> > 2: size of webview state. Would recommend testing with huggingface.aiconfig.json which is 8MB I think.
> > I don't see any other way to solve this, but just wanted to call out some considerations (and why I had avoided saving the aiconfig in webview state)
> 
> Ah, ok. I had assumed it wasn't storing the config in state because of the prop not updating the editor state, but size constraint also makes sense. `huggingface.aiconfig.json` loads/resaturates quickly enough. Maybe we can add logic in the future to not persist state over some threshold size?
> 
> For 1. I just tested and toggling back currently only has updated output if the execution completed while toggled away. Otherwise, the outputs are not updated if toggling back before they complete (obviously streaming callback doesn't connect editor to the server anymore, but going to look into why it's not being updated as a result of the server updating the config since I would expect that to update the document itself. Maybe `isInternalDocumentChange` is true in this case. In any case, this is the current behaviour on main as well and not a result of this change

Sounds good @rholinshead -- please feel free to ship this and we can iterate further on size constraints in the future if we encounter them :D

PR #1283, opened 2024-02-21
[vscode][3/n] Setup Env Variables: Use env template file for new files

rossdanlm:
[vscode][3/n] Setup Env Variables: Use env template file for new files

If the file does not exist, created a template file with some helper text to help guide people on how to set their environment variables.

Next PR I will handle the use case for when the env file already exists

## Test Plan

https://github.com/lastmile-ai/aiconfig/assets/151060367/4608833c-d64f-4906-a7ed-94af12174c6d

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1283).
* #1284
* __->__ #1283
* #1280

### DIFF
--- a/vscode-extension/src/extension.ts
+++ b/vscode-extension/src/extension.ts
@@ -59,7 +59,7 @@ export function activate(context: vscode.ExtensionContext) {
     vscode.commands.registerCommand(
       COMMANDS.SETUP_ENVIRONMENT_VARIABLES,
       () => {
-        setupEnvironmentVariables();
+        setupEnvironmentVariables(context);
       }
     )
   );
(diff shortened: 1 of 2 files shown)


vscode-extension/static/env_template.env
```diff
@@ -0,0 +1,15 @@
+# Uncomment the following lines to set these common env variables for
```
rossdanlm, line 1: Tanya said this is too much text. I'm just going to land and we can fix later

vscode-extension/src/extension.ts
```diff
@@ -756,53 +758,42 @@ async function setupEnvironmentVariables() {
...
+        preview: false,
+        // Tried using vscode.ViewColumn.Active but that overrides existing
+        // walkthrough window
+        viewColumn: vscode.ViewColumn.Beside,
+      });
+      vscode.window.showInformationMessage(
+        "Please define your environment variables."
+      );
+    }
+  }
```
Ankush-lastmile, line 796: nit: what if it doesn't open?

Ankush-lastmile (approved):
lgtm, approving to unblock

vscode-extension/src/extension.ts
rossdanlm: Good catch, will follow up someday

rossdanlm: https://github.com/lastmile-ai/aiconfig/issues/1287

PR #1286, opened 2024-02-22
Updated the empty config state

tanya-rai:
Updated the empty config state

Made it cleaner - this should be intuitive to use as is.

Also user doesn't have to delete as much text from initial state to jump in and get started.

### DIFF
--- a/vscode-extension/static/untitled.aiconfig.json
+++ b/vscode-extension/static/untitled.aiconfig.json
@@ -1,7 +1,8 @@
 {
-  "name": "New AIConfig -- please rename me",
+  "name": "Untitled",
   "schema_version": "latest",
   "metadata": {
+    "parameters": {},
     "models": {
       "gpt-3.5-turbo": {
         "model": "gpt-3.5-turbo",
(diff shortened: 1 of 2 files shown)


tanya-rai:
Screenshot 2024-02-21 at 8 10 43 PM

vscode-extension/static/untitled.aiconfig.json
```diff
@@ -11,21 +12,21 @@
     },
     "default_model": "gpt-3.5-turbo"
   },
-  "description": "Fill out a description about your AIConfig here.",
+  "description": "Enter a description for this AIConfig",
```
rossdanlm, line 15: I think a good compromise between keeping it clean and informative would be to link it to our AICOnfig getting started docs

vscode-extension/static/untitled.aiconfig.json
```diff
@@ -11,21 +12,21 @@
     },
     "default_model": "gpt-3.5-turbo"
   },
-  "description": "Fill out a description about your AIConfig here.",
+  "description": "Enter a description for this AIConfig. Learn about features in our [docs](https://aiconfig.lastmileai.dev/docs/category/overview)!",
```
rossdanlm, line 15: "Enter a description for this AIConfig. You can also check out [our docs](https://aiconfig.lastmileai.dev/docs/category/overview) to learn more!

rossdanlm (approved):

saqadri:
It would be really cool to also have sample output saved in the config. Maybe we should have a separate "Create from template" command for that, but now the Untitled is truly just an untitled. My goal was to have an educational untitled because this is the first time someone is seeing an aiconfig

PR #1289, opened 2024-02-22
[vscode installation] 5/n show view only when dependencies are not installed

Ankush-lastmile:
[vscode installation] 5/n show view only when dependencies are not installed

This diff removes blocking on server being ready. Now a user can view the aiconfig in readonly mode if their dependencies are not installed. Consequently so, any errors on dependency installation will be shown

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1289).
* __->__ #1289
* #1294
* #1290

### DIFF
--- a/vscode-extension/package.json
+++ b/vscode-extension/package.json
@@ -181,6 +181,7 @@
     "typescript": "^5.3.3"
   },
   "dependencies": {
+    "@vscode/python-extension": "^1.0.5",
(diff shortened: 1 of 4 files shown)


Ankush-lastmile:
implemented in #1306

PR #1290, opened 2024-02-22
3/n Update Initialize to invoke select interpreter

Ankush-lastmile:
3/n Update Initialize to invoke select interpreter

This diff updates the initialize flow to include the Python interpreter selection step. It ensures dependencies are installed only after the interpreter selection is finalized, accommodating the no-op change where the selected interpreter remains unchanged.

Note: If the extension is not activated, invoking the Initialize AIConfig command for an unknown reason installs dependencies on the pre-selected interpreter path

## testplan

Extension initializes on Opening An AIConfig

Test select a new interpreter

Test reselecting the same interpreter

https://github.com/lastmile-ai/aiconfig/assets/141073967/1a51ff89-3889-4fd6-8964-c341453ddf68

## Dependencies
https://github.com/lastmile-ai/aiconfig/pull/1265

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1290).
* #1289
* #1294
* __->__ #1290

### DIFF
--- a/vscode-extension/src/extension.ts
+++ b/vscode-extension/src/extension.ts
@@ -23,6 +23,7 @@ import {
 } from "./util";
 import { AIConfigEditorProvider } from "./aiConfigEditor";
 import { AIConfigEditorManager } from "./aiConfigEditorManager";
+import { PythonExtension } from "@vscode/python-extension";
 
 // This method is called when your extension is activated
 // Your extension is activated the very first time the command is executed
@@ -39,7 +40,7 @@ export function activate(context: vscode.ExtensionContext) {
   });
 
   const setupCommand = vscode.commands.registerCommand(COMMANDS.INIT, () => {
-    installDependencies(context, extensionOutputChannel);
+    initialize(context, extensionOutputChannel);
   });
   context.subscriptions.push(setupCommand);
 
@@ -905,4 +906,19 @@ async function shareAIConfig(
         }
       });
   }
+}
+
+/**
+ * Selects a Python interpreter and initiates the dependency installation flow.
+ */
+async function initialize(
+  context: vscode.ExtensionContext,
+  outputChannel: vscode.LogOutputChannel
+) {
+  // Make sure Python API is activated
+  const pythonApi: PythonExtension = await PythonExtension.api();
+
+  await vscode.commands.executeCommand("python.setInterpreter");
+
+  installDependencies(context, outputChannel);
 }
--- a/vscode-extension/package.json
+++ b/vscode-extension/package.json
@@ -181,6 +181,7 @@
     "typescript": "^5.3.3"
   },
   "dependencies": {
+    "@vscode/python-extension": "^1.0.5",
     "oboe": "^2.1.5",
     "portfinder": "^1.0.32",
     "ufetch": "^1.6.0"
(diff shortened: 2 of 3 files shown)


rholinshead:
> for unknown reasons, we are unable to block on the user selecting an interpreter. For now let this silently fail

Is this still true? In the code snippet I sent you it seemed to properly resolve only after the quickpick was closed

Ankush-lastmile:
> Is this still true? In the code snippet I sent you it seemed to properly resolve only after the quickpick was closed

The code snippet worked, realized I didn't update this diff properly so I took it back to draft mode.

Ankush-lastmile:
Ready for review

vscode-extension/src/extension.ts
```diff
@@ -23,10 +23,13 @@ import {
...
 import { AIConfigEditorProvider } from "./aiConfigEditor";
 import { AIConfigEditorManager } from "./aiConfigEditorManager";
+import {
+  PythonExtension,
+} from "@vscode/python-extension";
 
 // This method is called when your extension is activated
 // Your extension is activated the very first time the command is executed
-export function activate(context: vscode.ExtensionContext) {
+export async function activate(context: vscode.ExtensionContext) {
```
rholinshead, line 32: I don't think this needs to be made `async` since the async/await is a callback to a (sync) function call below

vscode-extension/src/extension.ts
```diff
@@ -906,3 +912,27 @@ async function shareAIConfig(
...
+  const pythonApi: PythonExtension = await PythonExtension.api();
+
+  await new Promise((resolve, reject) => {
+    vscode.commands
+    .executeCommand("python.setInterpreter")
+    .then(async () => {
+    // .then triggers on user selection or exiting interpreter selection.
+      resolve(undefined);
+    });
+  });
```
rholinshead, line 933: `await vscode.commands.executeCommand("python.setInterpreter");` should do the same thing. Shouldn't need to do a new promise and seems weird to use both await and promise.then in an async function

```suggestion
  await vscode.commands.executeCommand("python.setInterpreter");
```

vscode-extension/src/extension.ts
```diff
@@ -906,3 +912,27 @@ async function shareAIConfig(
...
+
+  installDependencies(context, outputChannel);
+  
+  return;
```
rholinshead, line 937:
```suggestion
```

vscode-extension/src/extension.ts
```diff
@@ -38,9 +41,12 @@ export function activate(context: vscode.ExtensionContext) {
     log: true,
   });
 
-  const setupCommand = vscode.commands.registerCommand(COMMANDS.INIT, () => {
-    installDependencies(context, extensionOutputChannel);
-  });
+  const setupCommand = vscode.commands.registerCommand(
+    COMMANDS.INIT,
+    async () => {
+      await initialize(context, extensionOutputChannel);
```
rholinshead, line 47: Does this actually need do async/await? I think it can just fire off the call to initialize since nothing else is waiting for it

vscode-extension/src/extension.ts
Ankush-lastmile: thanks for the clarification, updated

vscode-extension/src/extension.ts
Ankush-lastmile: not needed, added that while trying things out, removed

Ankush-lastmile:
- updated pr description
- addressed comments

rholinshead (approved):

PR #1291, opened 2024-02-22
[vscode][6/n] Setup Env Variables: Add more validation to env path input

rossdanlm:
[vscode][6/n] Setup Env Variables: Add more validation to env path input

Added two noteworthy checks that we didn't have before:

1. The filename itself must be ".env" (before it only checked for ending so things like "invalid/filename.env" could be valid)
2. The path must be a subset of the workspace path, to ensure that when we call the `loadenv()` function from Python in AIConfig SDK, it's able to access this `.env` file

I also made this into a separate function to keep it cleaner

## Test Plan

https://github.com/lastmile-ai/aiconfig/assets/151060367/d77305c4-2cc6-4cb8-a918-b74826ff3714

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1291).
* #1300
* __->__ #1291

### DIFF
--- a/vscode-extension/src/extension.ts
+++ b/vscode-extension/src/extension.ts
@@ -736,22 +736,14 @@ async function setupEnvironmentVariables(context: vscode.ExtensionContext) {
   const homedir = require("os").homedir(); // This is cross-platform: https://stackoverflow.com/a/9081436
   const defaultEnvPath = path.join(homedir, ".env");
 
+  const workspacePath = vscode.workspace.workspaceFolders
+    ? vscode.workspace.workspaceFolders[0].uri.fsPath
+    : null;
+
   const envPath = await vscode.window.showInputBox({
     prompt: "Enter the path of your .env file",
     value: defaultEnvPath,
-    validateInput: (text) => {
-      if (!text) {
-        return "File path is required";
-      } else if (!text.endsWith(".env")) {
-        return "File path must end in .env file";
-      }
-      // TODO: Check that file path is a "/.env" file (linux) or "\.env" (Windows)
-
-      // TODO: Check that env path is contained within workspace hierarchy
-      // (Ex: can't have .env file in a sibling dir otherwise AIConfig
-      // loadenv can't read it)
-      return null;
-    },
+    validateInput: (input) => validateEnvPath(input, workspacePath),
   });
 
   if (!envPath) {
@@ -810,6 +802,35 @@ async function setupEnvironmentVariables(context: vscode.ExtensionContext) {
       "Please define your environment variables."
     );
   }
+}
+
+function validateEnvPath(
+  inputPath: string,
+  workspacePath: string | null
+): string | null {
+  if (!inputPath) {
+    return "File path is required";
+  } else if (path.basename(inputPath) !== ".env") {
+    return 'Filename must be ".env"';
+  } else if (workspacePath !== null) {
+    // Loadenv from Python checks each folder from the file/program where it's
+    // invoked for the presence of an `.env` file. Therefore, the `.env` file
+    // must be saved either at the top-level directory of the workspace
+    // directory, or one of it's parent directories. This will ensure that if
+    // two AIConfig files are contained in separate paths within the workspace
+    // they'll still be able to access the same `.env` file.
+
+    // Note: If the `inputPath` directory is equal to the `workspacePath`,
+    // `relativePathFromEnvToWorkspace` will be an empty string
+    const relativePathFromEnvToWorkspace = path.relative(
+      path.dirname(inputPath),
+      workspacePath
+    );
+    if (relativePathFromEnvToWorkspace.startsWith("..")) {
+      return `File path must either be contained within the VS Code workspace directory ('${workspacePath}') or within a one of it's parent folders`;
+    }
+  }
+  return null;
 }
 
 async function shareAIConfig(


vscode-extension/src/extension.ts
```diff
@@ -739,22 +739,14 @@ async function setupEnvironmentVariables(context: vscode.ExtensionContext) {
...
-        return "File path must end in .env file";
-      }
-      // TODO: Check that file path is a "/.env" file (linux) or "\.env" (Windows)
-
-      // TODO: Check that env path is contained within workspace hierarchy
-      // (Ex: can't have .env file in a sibling dir otherwise AIConfig
-      // loadenv can't read it)
-      return null;
-    },
+    validateInput: (input) => validateEnvPath(input, workspacePath),
```
saqadri, line 746: This won't work in the case where a user has multiple workspace folders open. It could loop through all the workspace folders to see if the file is under any of those paths.

vscode-extension/src/extension.ts
```diff
@@ -815,6 +807,30 @@ async function setupEnvironmentVariables(context: vscode.ExtensionContext) {
...
+  inputPath: string,
+  workspacePath: string | null
+): string | null {
+  if (!inputPath) {
+    return "File path is required";
+  } else if (!inputPath.endsWith(".env")) {
+    return 'File path must end in ".env"';
+  } else if (path.basename(inputPath) !== ".env") {
+    return 'Filename of the fully qualified path must be ".env"';
+  } else if (workspacePath !== null) {
```
saqadri, line 817:
```suggestion
  } else if (workspacePath != null) {
```

vscode-extension/src/extension.ts
```diff
@@ -812,6 +804,30 @@ async function setupEnvironmentVariables(context: vscode.ExtensionContext) {
...
 
+function validateEnvPath(
+  inputPath: string,
+  workspacePath: string | null
+): string | null {
+  if (!inputPath) {
+    return "File path is required";
+  } else if (!inputPath.endsWith(".env")) {
+    return 'File path must end in ".env"';
+  } else if (path.basename(inputPath) !== ".env") {
```
saqadri, line 815: I feel like it would be simpler to just be able to specify the path to the .env on the aiconfig server instead of all these checks. Fine for now, but let's fix. The proper flow would be:

1. User can specify the path to env file
2. We save the .env path in the extension settings
3. We add or update it with the env template
4. On aiconfig server start, we specify the path to the .env file
5. In the aiconfig server, when calling load dotenv, we specify the path to the env file.

Then we don't have to validate that it's in the folder hierarchy.

vscode-extension/src/extension.ts
```diff
@@ -736,22 +736,14 @@ async function setupEnvironmentVariables(context: vscode.ExtensionContext) {
   const homedir = require("os").homedir(); // This is cross-platform: https://stackoverflow.com/a/9081436
   const defaultEnvPath = path.join(homedir, ".env");
 
+  const workspacePath = vscode.workspace.workspaceFolders
+    ? vscode.workspace.workspaceFolders[0].uri.fsPath
+    : null;
```
rholinshead, line 741:
```suggestion
  const workspacePath = vscode.workspace.workspaceFolders?.[0]?.uri?.fsPath ?? null;
```
nit, can optionally chain array indexing

vscode-extension/src/extension.ts
```diff
@@ -812,6 +804,30 @@ async function setupEnvironmentVariables(context: vscode.ExtensionContext) {
   }
 }
 
+function validateEnvPath(
+  inputPath: string,
+  workspacePath: string | null
+): string | null {
+  if (!inputPath) {
+    return "File path is required";
+  } else if (!inputPath.endsWith(".env")) {
```
rholinshead, line 813: Is this check needed now that the basename check exists?

vscode-extension/src/extension.ts
```diff
@@ -812,6 +804,30 @@ async function setupEnvironmentVariables(context: vscode.ExtensionContext) {
...
+    return 'File path must end in ".env"';
+  } else if (path.basename(inputPath) !== ".env") {
+    return 'Filename of the fully qualified path must be ".env"';
```
rholinshead, line 816: nit, "Filename must be .env" is much clearer imo

vscode-extension/src/extension.ts
```diff
@@ -812,6 +804,30 @@ async function setupEnvironmentVariables(context: vscode.ExtensionContext) {
...
+  } else if (workspacePath !== null) {
+    // Note: If the workspacePath is equal to inputPath,
+    // relativePathFromEnvToWorkspace will be empty string
+    const relativePathFromEnvToWorkspace = path.relative(
+      path.dirname(inputPath),
+      workspacePath
+    );
+    if (relativePathFromEnvToWorkspace.startsWith("..")) {
```
rholinshead, line 824: If it can't be any direct level above the workspace, doesn't the relative path need to be exactly empty string (match workspace)?  Because setting it to something like `${workspacePath}/folderA/.env` won't work if the config is in `${workspacePath/folderB/.env` even though it would be valid based on this logic.

To be completely correct we probably need to check the relative path to the env file from whatever config is opened for it, but that might a bit of work

vscode-extension/src/extension.ts
```diff
@@ -812,6 +804,30 @@ async function setupEnvironmentVariables(context: vscode.ExtensionContext) {
...
+  inputPath: string,
+  workspacePath: string | null
+): string | null {
+  if (!inputPath) {
+    return "File path is required";
+  } else if (!inputPath.endsWith(".env")) {
+    return 'File path must end in ".env"';
+  } else if (path.basename(inputPath) !== ".env") {
+    return 'Filename of the fully qualified path must be ".env"';
+  } else if (workspacePath !== null) {
```
Ankush-lastmile, line 817: when would the workspacePath be Null?

vscode-extension/src/extension.ts
```diff
@@ -812,6 +804,30 @@ async function setupEnvironmentVariables(context: vscode.ExtensionContext) {
...
+    // Note: If the workspacePath is equal to inputPath,
+    // relativePathFromEnvToWorkspace will be empty string
+    const relativePathFromEnvToWorkspace = path.relative(
+      path.dirname(inputPath),
+      workspacePath
+    );
+    if (relativePathFromEnvToWorkspace.startsWith("..")) {
+      return `File path must either be contained within the VS Code workspace directory ('${workspacePath}') or within a one of it's parent folders`;
+    }
```
Ankush-lastmile, line 826: what if the file path is from root? Does the path.relative create a  url that starts with `..`?

ie `/users/rossdan/something/.env`

Ankush-lastmile (approved):
lgtm, 2 questions

vscode-extension/src/extension.ts
rossdanlm: It's not but I feel it's nice to have and makes it more piece-wise. Some people might not understand what "basename" means so having these two together makes it more clear

vscode-extension/src/extension.ts
rossdanlm: Yea in this case it'll say:

`../Projects/aiconfig/" because it has to get out of the `something` path

vscode-extension/src/extension.ts
rossdanlm: > `${workspacePath}/folderA/.env` won't work if the config is in `${workspacePath}/folderB/.env`

That's correct and intentional. When running the `loadenv()` function in Python, it checks recursively up the file until it reaches the top folder from where you're running it. So when running the config from `${workspacePath}/folderB/.env`, it will check:

- `${workspacePath}/folderB/.env`
- `${workspacePath}/.env`
- and every recursive parent folder inside of `${workspacePath}` + `.env`

vscode-extension/src/extension.ts
```diff
@@ -812,6 +804,30 @@ async function setupEnvironmentVariables(context: vscode.ExtensionContext) {
...
+  } else if (!inputPath.endsWith(".env")) {
+    return 'File path must end in ".env"';
+  } else if (path.basename(inputPath) !== ".env") {
+    return 'Filename of the fully qualified path must be ".env"';
+  } else if (workspacePath !== null) {
+    // Note: If the workspacePath is equal to inputPath,
+    // relativePathFromEnvToWorkspace will be empty string
+    const relativePathFromEnvToWorkspace = path.relative(
+      path.dirname(inputPath),
+      workspacePath
```
saqadri, line 822: Just checking -- I think this should be switched right? workspacePath is first, and inputPath is second. Don't think you need to do `path.dirname`, just 

```suggestion
       const relativePathFromEnvToWorkspace = path.relative(
         workspacePath,
         inputPath
```
```suggestion
    const relativePathFromEnvToWorkspace = path.relative(
      path.dirname(inputPath),
      workspacePath
```

I tested this with:

```
const foo = path.relative(
  "/Users/saqadri/lm/aiconfig2",
  "/Users/saqadri/lm/aiconfig2/vscode-extension/editor/src/VSCodeEditor.tsx"
); 
```

which returned `vscode-extension/editor/src/VSCodeEditor.tsx`

and switching it to:

```
const foo = path.relative(
  "/Users/saqadri/lm/aiconfig2/vscode-extension/editor/src/VSCodeEditor.tsx",
  "/Users/saqadri/lm/aiconfig2"
);
```

returned `../../../..`

saqadri (changes requested):

vscode-extension/src/extension.ts
rossdanlm: There are cases where VSCode has defined a workspace (ex: when you open a new VSCode window and havent' set the project directory yet)

vscode-extension/src/extension.ts
rossdanlm: > On aiconfig server start, we specify the path to the .env file

I don't like doing this because this now makes AIConfig dependent on the AIConfig extension working. If we try to use the cookbooks with Jupyter notebooks for instance and the `.env` file is not properly set to be within the workspace or a parent of the workspace, it will fail

rossdanlm: > In the aiconfig server, when calling load dotenv, we specify the path to the env file.

Also prefer not to do this because now we need to keep more information about the user's configuration for the SDK to work

vscode-extension/src/extension.ts
rossdanlm: I took this code from https://github.com/lastmile-ai/aiconfig/blob/7a4cafc7f1fac7d9f180fc7090cbb45ae5bc59d1/vscode-extension/src/extension.ts#L171 which also checks the 1st one.

I feel that most users will define a single folder within a VS Code session. VSCode multi-workspaces seem not very common (https://code.visualstudio.com/docs/editor/workspaces#_multiroot-workspaces). The fix I have in mind is to check for the common lowest ancestor path between all workspaces and check the relative path based on that. I think this followup as a P1, want to get the other parts of the install flow prioritized over this

vscode-extension/src/extension.ts
saqadri: > Jupyter notebooks for instance and the .env file is not properly set to be within the workspace or a parent of the workspace, it will fail

This is a very good point. Fair enough

vscode-extension/src/extension.ts
rossdanlm: It will have validation error because when doing this 
```
path.relative(
  ${workspacePath}
  ${workspacePath}/folderA
)
```

This will return `..`

saqadri (approved):
Approving to unblock, chatted offline

vscode-extension/src/extension.ts
rossdanlm: Let me confirm your example and screenshot just o verify + sanity check

rossdanlm: Just to be clear, I'm saying that when you set the env path, it must be at he LOWEST level the same folder as the workspace path, and can be valid if it's in any direct parent folder (no aunts or uncles) along the path tree

vscode-extension/src/extension.ts
rossdanlm: Added issue to keep track of this to the overall `env` followups: https://github.com/lastmile-ai/aiconfig/issues/1299

vscode-extension/src/extension.ts
rossdanlm: Ok reading your comment now I feel this is the part of confusion:

> If it can't be any direct level above the workspace,

The validation check I'm making is the opposite of this where env file MUST be the workspace or a direct level above workspace. If that's true, then it will never need to start with `..`

vscode-extension/src/extension.ts
rossdanlm: I'm going to todo to check for multiple workspaces: https://github.com/lastmile-ai/aiconfig/issues/1299

vscode-extension/src/extension.ts
rossdanlm: Actually just removed, yea feel it's simpler

vscode-extension/src/extension.ts
rholinshead: Oh wait, did you swap the order? I think that was the confusion because it would have returned the path from workspace to the env file instead of from env file to workspace (so no "..")

rossdanlm: No the path needs to be from env file dir --> workspace

rossdanlm:
Add code comment to make it more clear
Screenshot 2024-02-22 at 17 23 34

PR #1292, opened 2024-02-22
updated empty state config

tanya-rai:
updated empty state config

added more detail - still needs improvement but better than the blank state from before.

to follow-up
- replace gpt4 with gemini to showcase a non-openai model. having some issues with the output not saving..so not using it for now.
- also wanted to add image cell with dalle3 but not saving output right now


tanya-rai:
Screenshot 2024-02-22 at 12 25 34 PM

rossdanlm (approved):
lgtm, if Sarmad has comments we can fix forward

PR #1296, opened 2024-02-22
[vscode] Fix Outputs Propagating to Text Document

rholinshead:
# [vscode] Fix Outputs Propagating to Text Document

Currently, the webview only notifies the extension of dirty document (on the server) before/after the run request to the server. For long-running/streaming prompts, the request's response is actually returned well before the streaming responses are completed, meaning the document is not actually notified when the stream completes with the new output.

To fix, we can call `notifyDocumentDirty` to [update the text document in the extension with the server state](https://github.com/lastmile-ai/aiconfig/blob/main/vscode-extension/src/aiConfigEditor.ts#L282) whenever any of the stream responses are obtained (excluding `output_chunk` since that happens very frequently and would lead to the extension host loading the file from the server too frequently)

## Testing:
Running a prompt now shows the output in the text document:

https://github.com/lastmile-ai/aiconfig/assets/5060851/1211cc1e-676f-4415-8d01-4b176296f4b5

### DIFF
--- a/vscode-extension/editor/src/VSCodeEditor.tsx
+++ b/vscode-extension/editor/src/VSCodeEditor.tsx
@@ -283,18 +283,28 @@ export default function VSCodeEditor() {
         {
           output_chunk: (data) => {
             onStream({ type: "output_chunk", data: data as Output });
+            // Don't notify dirty since output chunks are very frequent
           },
           aiconfig_chunk: (data) => {
             onStream({ type: "aiconfig_chunk", data: data as AIConfig });
+            if (vscode) {
+              notifyDocumentDirty(vscode);
+            }
           },
           stop_streaming: (_data) => {
             onStream({ type: "stop_streaming", data: null });
+            if (vscode) {
+              notifyDocumentDirty(vscode);
+            }
           },
           error: (data) => {
             onError({
               type: "error",
               data: data as RunPromptStreamErrorEvent["data"],
             });
+            if (vscode) {
+              notifyDocumentDirty(vscode);
+            }
           },
         }
       );


vscode-extension/editor/src/VSCodeEditor.tsx
```diff
@@ -283,18 +283,28 @@ export default function VSCodeEditor() {
         {
           output_chunk: (data) => {
             onStream({ type: "output_chunk", data: data as Output });
+            // Don't notify dirty since output chunks are very frequent
           },
           aiconfig_chunk: (data) => {
             onStream({ type: "aiconfig_chunk", data: data as AIConfig });
+            if (vscode) {
+              notifyDocumentDirty(vscode);
+            }
```
rossdanlm, line 292: Why add it here if not adding to `output_chunk`? They have very similar functions (just that this is used for non-streaming models).

saqadri (approved):

vscode-extension/editor/src/VSCodeEditor.tsx
saqadri: This one gets called on completion of the response, right? I think that's why this is added here

rholinshead: The editor server only emits `aiconfig_chunk` once, at the end of the prompt run, right? I guess it's technically redundant when paired with `stop_streaming`, just feel it's "correct" for the sake of completion

rossdanlm: Yea I asked because we have `stop_streaming` there to handle cases explicitly for when streaming is over and no longer need to do changes like this within `aiconfig_chunk`

PR #1297, opened 2024-02-22
updated the env_tempate.env

tanya-rai:
updated the env_tempate.env

updating the template for environment variables

### DIFF
--- a/vscode-extension/static/env_template.env
+++ b/vscode-extension/static/env_template.env
@@ -1,15 +1,10 @@
-# Uncomment the following lines to set these common env variables for 
-# accessing generative AI models. If your model requires other env variables
-# you can also set those in this file
+# This is your .env file to specify keys from model providers. Obtain your keys:
+# OpenAI (ChatGPT, gpt-4, DALL-E 3): https://platform.openai.com/api-keys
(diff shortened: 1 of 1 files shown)


vscode-extension/static/env_template.env
```diff
@@ -1,15 +1,10 @@
-# Uncomment the following lines to set these common env variables for 
-# accessing generative AI models. If your model requires other env variables
-# you can also set those in this file
+# This is your .env file to specify keys from model providers. Obtain your keys:
```
rossdanlm, line 1: We should say something more like "common keys include" because they don't necessarily need these, it depends on the model they're using

rossdanlm (approved):
Pls update comments

vscode-extension/static/env_template.env
```diff
@@ -1,15 +1,10 @@
...
+# OpenAI (ChatGPT, gpt-4, DALL-E 3): https://platform.openai.com/api-keys
+# Google (Gemini, Gemma, PaLM): https://ai.google.dev/tutorials/setup
```
Ankush-lastmile, line 3: nit: Gemma isn't available on the api

PR #1302, opened 2024-02-22
[vscode] Set datadog service to vscode-aiconfig

rholinshead:
[vscode] Set datadog service to vscode-aiconfig


Ankush-lastmile (approved):
lgtm, I guess "mode" is now deprecated

saqadri (approved):
Good catch

rholinshead:
> lgtm, I guess "mode" is now deprecated

still used for the editor component, just not sure if we may have other vscode extensions at some point so figured we may as well match the package name

PR #1301, opened 2024-02-22
[ez][vscode] Remove Initialization Flow from Extension Activation

Ankush-lastmile:
[ez][vscode] Remove Initialization Flow from Extension Activation

This diff removes the Initialization flow from ocurring on Extension Activation.

The landing page directs the user to invoke this command.

The Opening an AIConfig may invoke this if dependencies are not installed (not implemented yet)

Invoking the Initialize Command will no longer automatically install dependencies on the pre-selection python interpreter.

## Testplan

https://github.com/lastmile-ai/aiconfig/assets/141073967/04031cdc-2c9d-47e8-b43d-a9d2f6aa769d

PR #1304, opened 2024-02-23
Cache Python interpreter to VS Code settings

rossdanlm:
Cache Python interpreter to VS Code settings

Do this after we run the `checkPython()` flow

It's just a simple write. We'll be using it later in two flows:

1. On open/create AIConfig file init to determine if interpter path is set (see Ankush PR in here for more details: https://github.com/lastmile-ai/aiconfig/pull/1301)
2. When updating interpter to loop over existing open AIConfig editor windows

Got this code to update from https://github.com/lastmile-ai/aiconfig/blob/826565f7079bc30e228a1477541ccbf17702d4f6/vscode-extension/src/extension.ts#L398-L402

## Test Plan
When AIConfig Editor is activated, check that it's saved to VS Code settings

https://github.com/lastmile-ai/aiconfig/assets/151060367/b7977c08-5a5a-4ca9-bff8-3bce7914e52e

### DIFF
--- a/vscode-extension/package.json
+++ b/vscode-extension/package.json
@@ -94,6 +94,11 @@
           "type": "string",
           "default": "",
           "description": "Path to the custom model registry file."
+        },
+        "vscode-aiconfig.pythonInterpreter": {
+          "type": "string",
+          "default": "",
+          "description": "Path to the python interpreter for this VS Code workspace"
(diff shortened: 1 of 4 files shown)


Ankush-lastmile (approved):
nice!

PR #1306, opened 2024-02-23
[rfc] Show readonly Editor Before Server Is Started

rholinshead:
[rfc] Show readonly Editor Before Server Is Started

# [rfc] Show readonly Editor Before Server Is Started

Currently when opening an aiconfig file, the file contents / editor webview are not rendered at all until the server is started (or attempted to start). This is because the webview isn't rendered until `resolveCustomTextEditor` returns, and we are awaiting a few server-initialization-related calls within the function.

To fix, we can move the `startEditorServer` into a regular `.then`, which will call the Promise without blocking. Within the `.then` we can handle the resolved promise as normal and do the other things that depend on the server process running.

Another thing worth mentioning is that we previously required the server url to set up the csp for the webview HTML. I'm not sure if we actually need to set the `connect-src vscode-webview` at all (didn't seem to make a difference when testing) but for now, just setting to accept any localhost port instead of being dependent on server.

We can immediately set the webview content and see it rendered (in readonly mode by default):

https://github.com/lastmile-ai/aiconfig/assets/5060851/2bf8cd4a-2b4d-4aea-ad1e-8f15f245ba65

NOTE: This currently results in a notification, 'Error loading models' since the (readonly) editor is attempting to load the models before the server is set up. There is a simple fix needed in the editor to address that, so I'll do that and update the package for use in the vscode extension before landing this

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1306).
* __->__ #1306
* #1309

### DIFF
--- a/vscode-extension/src/aiConfigEditor.ts
+++ b/vscode-extension/src/aiConfigEditor.ts
@@ -62,24 +62,13 @@ export class AIConfigEditorProvider implements vscode.CustomTextEditorProvider {
     _token: vscode.CancellationToken
   ): Promise<void> {
     let editorServer: ServerInfo | null = null;
+    let isWebviewDisposed = false;
 
     // TODO: saqadri - clean up console log
     //console.log(`${document.fileName}: resolveCustomTextEditor called`);
 
     // We show the extension output channel so users can see some of the server logs as they use the extension
     this.extensionOutputChannel.show(/*preserveFocus*/ true);
-
-    // Start the AIConfig editor server process.
-    editorServer = await this.startEditorServer(document);
-
-    this.aiconfigEditorManager.addEditor(
-      new AIConfigEditorState(
-        document,
-        webviewPanel,
-        editorServer,
-        this.aiconfigEditorManager
-      )
-    );
 
     // Setup initial content for the webview
     webviewPanel.webview.options = {
@@ -93,10 +82,7 @@ export class AIConfigEditorProvider implements vscode.CustomTextEditorProvider {
         vscode.Uri.joinPath(this.context.extensionUri, "editor", "static"),
       ],
     };
-    webviewPanel.webview.html = this.getHtmlForWebview(
-      webviewPanel.webview,
-      editorServer
-    );
+    webviewPanel.webview.html = this.getHtmlForWebview(webviewPanel.webview);
 
     function updateWebview() {
       if (isWebviewDisposed) {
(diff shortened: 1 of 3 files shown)


vscode-extension/src/aiConfigEditor.ts
```diff
@@ -110,6 +96,38 @@ export class AIConfigEditorProvider implements vscode.CustomTextEditorProvider {
...
 
+    // Update webview immediately so we unblock the render; server init will happen in the background.
+    updateWebview();
+
+    // Start the AIConfig editor server process. Don't await at the top level here since that blocks the
+    // webview render (which happens only when resolveCustomTextEditor returns)
+    this.startEditorServer(document).then(async (editorServer) => {
+      editorServer = editorServer;
+
+      this.aiconfigEditorManager.addEditor(
```
saqadri, line 107: This seems fine, but might be worth having an update when an editor server is created for aiConfigManager to update its state. Because technically aiConfigManager should keep track of all open editors irrespective of the server.

Not blocking for this diff since this is the same functionality as before

vscode-extension/src/aiConfigEditor.ts
```diff
@@ -110,6 +96,38 @@ export class AIConfigEditorProvider implements vscode.CustomTextEditorProvider {
       });
     }
 
+    // Update webview immediately so we unblock the render; server init will happen in the background.
+    updateWebview();
+
+    // Start the AIConfig editor server process. Don't await at the top level here since that blocks the
+    // webview render (which happens only when resolveCustomTextEditor returns)
+    this.startEditorServer(document).then(async (editorServer) => {
```
saqadri, line 104: Should we have a .error as well? Ideally with some kind of retry?

saqadri: Can also be future diffs

vscode-extension/src/aiConfigEditor.ts
```diff
@@ -579,7 +575,7 @@ export class AIConfigEditorProvider implements vscode.CustomTextEditorProvider {
...
+		 <meta http-equiv="Content-Security-Policy" content="default-src 'none'; img-src data: http: https: 'self'; media-src data: http: https: 'self'; connect-src vscode-webview: http://localhost:* http: https:; style-src 'unsafe-inline' ${webview.cspSource} https://cdn.jsdelivr.net/npm/monaco-editor@0.43.0 https://cdn.jsdelivr.net/npm/monaco-editor@0.43.0/min/vs/editor/editor.main.css; script-src 'nonce-${nonce}' vscode-resource: https: http: https://cdn.jsdelivr.net/npm/monaco-editor@0.43.0 https://cdn.jsdelivr.net; font-src https://cdn.jsdelivr.net/npm/monaco-editor@0.43.0/min/vs/base/browser/ui/codicons/codicon/codicon.ttf; worker-src blob:;">
```
saqadri, line 578: This is ok, I vaguely remember running into CSP issues for the server but maybe that was a different issue. Makes sense to leave it for localhost but also feel free to remove if we want to try that out

saqadri (approved):
Nice! The flow makes sense to me. We can test it out tomorrow.

rholinshead:
rebased onto updated package so the models error no longer happens:


https://github.com/lastmile-ai/aiconfig/assets/5060851/2acc1eed-1353-4862-83db-897fc5d5ea66


Going to land and can address feedback (existing behaviour improvements) in a follow-up tomorrow

PR #1307, opened 2024-02-23
Check interpreter (initalize if needed) and call install dependency flow before starting editor server

rossdanlm:
Check interpreter (initalize if needed) and call install dependency flow before starting editor server

Three main things we do here:

1. Check if interpreter is defined in workspace settings (set in https://github.com/lastmile-ai/aiconfig/pull/1304)
2. Call either `initialize()` or `installDependencies()` depending on that result
3. Implement the `initializePythonFlow()` inside of `resolveCustomTextEditor()`

## Test Plan

Before (does not invoke flow)

https://github.com/lastmile-ai/aiconfig/assets/151060367/8f1e308c-975d-4a79-8f9a-127c0322f065

After (does invoke flow)

https://github.com/lastmile-ai/aiconfig/assets/151060367/28f97461-3e7c-4340-af2c-fb4993562c24

### DIFF
--- a/vscode-extension/src/aiConfigEditor.ts
+++ b/vscode-extension/src/aiConfigEditor.ts
@@ -10,7 +10,10 @@ import {
   updateWebviewEditorThemeMode,
   waitUntilServerReady,
 } from "./util";
-import { getPythonPath } from "./utilities/pythonSetupUtils";
+import {
+  getPythonPath,
+  initializePythonFlow,
+} from "./utilities/pythonSetupUtils";
 import { getNonce } from "./utilities/getNonce";
 import { getUri } from "./utilities/getUri";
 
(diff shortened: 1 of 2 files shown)


vscode-extension/src/aiConfigEditor.ts
```diff
@@ -99,6 +102,9 @@ export class AIConfigEditorProvider implements vscode.CustomTextEditorProvider {
     // Update webview immediately so we unblock the render; server init will happen in the background.
     updateWebview();
 
+    // Do not start the server until we ensure the Python setup is ready
+    await initializePythonFlow(this.context, this.extensionOutputChannel);
```
rholinshead, line 106: Can this be done with .then before the startEditorServer so it doesn't block (should fix the issue in the description):
```
// Do not start the server until we ensure Python setup is ready
// Don't await at the top level here since that blocks the webview 
// render (which happens only when resolveCustomTextEditor returns)
initializePythonFlow(this.context, this.extensionOutputChannel).then(() =>
  // Start the AIConfig editor server process. Similarly, don't await / block webview
  this.startServer(document.then(async(editorServer => {...
   ....
```

rossdanlm: Ok turns out it's no longer an issue we just had a hanging `await` statement: https://github.com/lastmile-ai/aiconfig/compare/61df58e276bd18267de0dad3c7be4b3fb8cc6879..50aa6431a3db9a06dbe5b14ddb717830df3fa81b

Ankush-lastmile (approved):

PR #1313, opened 2024-02-23
Create listener to open info dialog for when python interpreter path changes

rossdanlm:
Create listener to open info dialog for when python interpreter path changes

Haven't added the actual implementation for restarting server. Ryan is working on that now so will connect to it next PR when it's ready

Some additional context that Ankush and I discussed:

If user selects an invalid interpreter, that's fine because:
1. They AIconfig flow will be invalid and fail (upon restarting server), but we have checks to pester them to update their interpreter to the correct one via the `installDependencies()` flow (defined in https://github.com/lastmile-ai/aiconfig/pull/1307)

## Test Plan

https://github.com/lastmile-ai/aiconfig/assets/151060367/b9219c15-5843-4bf7-888f-357b20faff6a

Also tested this does not trigger the button to be shown when selecting an interpreter gets triggered because we haven't selected it before. This is good, because it would otherwise be annoying to refresh the window when it's already "active". It's technically not active yet because we haven't yet done the python checks and started the server

https://github.com/lastmile-ai/aiconfig/assets/151060367/9a90dac2-b59e-4b1e-be87-ca2d2d550093

### DIFF
--- a/vscode-extension/src/extension.ts
+++ b/vscode-extension/src/extension.ts
@@ -19,13 +19,24 @@ import {
   SUPPORTED_FILE_EXTENSIONS,
   setupEnvironmentVariables,
 } from "./util";
-import { initialize, getPythonPath } from "./utilities/pythonSetupUtils";
+import {
+  getPythonPath,
+  initialize,
+  savePythonInterpreterToCache,
+} from "./utilities/pythonSetupUtils";
 import { AIConfigEditorProvider } from "./aiConfigEditor";
-import { AIConfigEditorManager } from "./aiConfigEditorManager";
+import {
+  AIConfigEditorManager,
+  AIConfigEditorState,
+} from "./aiConfigEditorManager";
+import {
+  ActiveEnvironmentPathChangeEvent,
+  PythonExtension,
+} from "@vscode/python-extension";
 
 // This method is called when your extension is activated
 // Your extension is activated the very first time the command is executed
-export function activate(context: vscode.ExtensionContext) {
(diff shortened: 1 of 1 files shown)


vscode-extension/src/extension.ts
```diff
@@ -147,6 +158,33 @@ export function activate(context: vscode.ExtensionContext) {
...
+    pythonapi.environments.onDidChangeActiveEnvironmentPath(
+      async (e: ActiveEnvironmentPathChangeEvent) => {
+        await savePythonInterpreterToCache();
+        const editors: Array<AIConfigEditorState> = Array.from(
+          aiconfigEditorManager.getRegisteredEditors()
+        );
+        if (editors.length > 0) {
+          vscode.window
+            .showInformationMessage(
+              "Python Interpreter Updated: Would you like to refresh active AIConfig files?",
```
saqadri, line 173: nit

```suggestion
              "Python Interpreter Updated: Would you like to restart active AIConfig file servers?",
```

saqadri (approved):

rossdanlm:
TODO: Add a followup to learn more to link to doc on why changing python interpreter needs to trigger refreshing active AIConfig files

vscode-extension/src/extension.ts
rossdanlm: Synced offline w. Sarmad "file servers" is a bit confusing for regular people to understand, but we should have a learn more button so people can understand why on interpreter update we need to refresh

PR #1314, opened 2024-02-23
[vscode] Fix Setting editorServer

rholinshead:
# [vscode] Fix Setting editorServer

Realized when looking into restarting the server that the `editorServer` global var was not being set properly (was null on dispose). The problem was using the same name for the resolved started server. This fixes by using a different name

## Testing:
- Ensure the editorServer value is right when disposing the webview
- Close webview and open again and ensure new server at port 8000
- Make changes to JSON for a config and toggle back to the editor webview and see the changes reflected there


Ankush-lastmile (approved):
sneaky bug

rossdanlm (approved):
Can you also add the code pointer where this global var is first set in code in PR description?

PR #1315, opened 2024-02-23
[vscode] Don't block on Python Initialization

Ankush-lastmile:
[vscode] Don't block on Python Initialization

This diff removes the await statement on the initialize python flow when resolving a custom editor, to ensure that it is not blocking. Now, only the aiconfig server is blocked on python initialization. The webview server is not blocked.

Previously this was blocking the webview and not letting the webview server start.

## Testplan

https://github.com/lastmile-ai/aiconfig/assets/141073967/ee180bb4-7fb7-40c6-a33e-b7dc3546ae70


rossdanlm (approved):
nice! we are following up with the server status error being shown when it's no longer needed

PR #1318, opened 2024-02-23
[1/n] Implement Editor Server as Class

rholinshead:
[1/n] Implement Editor Server as Class

# [1/n] Implement Editor Server as Class

Currently we maintain EditorServer state as an object for the extension, accessing the process and server url from the object. In subsequent PR, we will need to expose the ability to restart the editor server. Instead of creating and exposing a restart function on the object, let's just move the editor server into a class so that we have a proper instance with methods to leverage for starting, stopping and (next PR) restarting.

## Testing:
Open an aiconfig, ensure the server starts and I can run a prompt
Close the config, verify in logging that the server process is properly killed
Open config again. See server start up on 8000 (i.e. the port is open again after server was killed)

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1318).
* #1328
* #1319
* __->__ #1318

### DIFF
--- a/vscode-extension/src/editorServer.ts
+++ b/vscode-extension/src/editorServer.ts
@@ -1,0 +1,77 @@
+import * as vscode from "vscode";
+import { getPortPromise } from "portfinder";
+import { EXTENSION_NAME } from "./util";
+import { getPythonPath } from "./utilities/pythonSetupUtils";
+import { ChildProcessWithoutNullStreams, spawn } from "child_process";
+
+/**
+ * Provider for AIConfig editors.
+ *
+ * AIConfig editors are used for `.aiconfig`, `.aiconfig.json` and `.aiconfig.yaml` files.
+ * These files are backed by a JSON schema under the hood.
+ */
+export class EditorServer {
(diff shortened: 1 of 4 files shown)


vscode-extension/src/editorServer.ts
```diff
@@ -0,0 +1,74 @@
...
+  public serverProc: ChildProcessWithoutNullStreams | null = null;
+  public url: string | null = null;
+
+  constructor(workingDirectory: string) {
+    this.cwd = workingDirectory;
+  }
+
+  public async start(): Promise<EditorServer> {
+    if (this.serverProc) {
+      throw new Error("Server process already started");
```
saqadri, line 28: nit: throw or just `return this`?

saqadri (approved):
LGTM

vscode-extension/src/editorServer.ts
rholinshead: Oh good point, can just log and return this

PR #1319, opened 2024-02-23
[2/n] Support Restarting Editor Server

rholinshead:
[2/n] Support Restarting Editor Server

# [2/n] Support Restarting Editor Server

Add an "AIConfig: Restart Active Editor Server" command so that users can restart the server for the active (in-view) editor. As part of this, implement the restart functionality for the EditorServer so that the server process is killed and a new one is started, ensuring the associated extension/webview is properly updated to associate with the new server process. As best effort, also implement an `onDidChangeState` so that we can listen to the editor server status changing and show a progress notification that the server is starting.

This logic will be leveraged in subsequent PR to handle restarting all editor servers when the workspace python interpreter changes.

https://github.com/lastmile-ai/aiconfig/assets/5060851/39739008-abc7-4538-9087-42972c1400aa

Note: The server heartbeat banner shows when restarting because of the existing server status request failing. This should be fixed by https://github.com/lastmile-ai/aiconfig/pull/1325

## Testing:
- Restart server and ensure (via logging) that the old one is killed and new one is set up, ensuring prompts run in new one
- Close webview and ensure the server is killed
- Activate restart server command and toggle away then back to webview, ensure it's in readonly until it starts up and becomes editable again when toggle back

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1319).
* #1328
* __->__ #1319
* #1318

### DIFF
--- a/vscode-extension/package.json
+++ b/vscode-extension/package.json
@@ -58,6 +58,11 @@
         "category": "AIConfig"
       },
       {
+        "command": "vscode-aiconfig.restartActiveEditorServer",
+        "title": "Restart Active Editor Server",
+        "category": "AIConfig"
+      },
+      {
         "command": "vscode-aiconfig.setApiKeys",
         "title": "Set API Keys",
         "category": "AIConfig"
--- a/vscode-extension/src/aiConfigEditor.ts
+++ b/vscode-extension/src/aiConfigEditor.ts
@@ -1,8 +1,6 @@
 import * as vscode from "vscode";
 import {
-  COMMANDS,
   EXTENSION_NAME,
-  ServerInfo,
   getCurrentWorkingDirectory,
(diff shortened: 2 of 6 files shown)


saqadri (approved):

vscode-extension/package.json
```diff
@@ -57,6 +57,11 @@
         "title": "Open Custom Model Registry File",
         "category": "AIConfig"
       },
+      {
+        "command": "vscode-aiconfig.restartActiveEditorServer",
+        "title": "Restart Active Editor Server",
+        "category": "AIConfig"
+      },
```
rossdanlm, line 64: This is just for testing right? We're not actually going to land with it?

PR #1320, opened 2024-02-23
[ez][vscode] Don't set getServerStatus Callback in VSCode

Ankush-lastmile:
[vscode] Don't set getServerStatus Callback in VSCode

This diff removes the `getServerStatus` callback from being passed into the `AIConfigEditor`  since we handle the servers within the vscode extension.

## Testplan

https://github.com/lastmile-ai/aiconfig/assets/141073967/85f8a303-a3c3-440e-a372-301cb57c9bd8

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1320).
* __->__ #1320
* #1315

### DIFF
--- a/vscode-extension/editor/src/VSCodeEditor.tsx
+++ b/vscode-extension/editor/src/VSCodeEditor.tsx
@@ -432,9 +432,6 @@ export default function VSCodeEditor() {
     [aiConfigServerUrl, vscode]
   );
 
-  const getServerStatus = useCallback(async () => {
-    return await ufetch.get(ROUTE_TABLE.SERVER_STATUS(aiConfigServerUrl));
-  }, [aiConfigServerUrl]);
 
   const logEventHandler = useCallback(
     (event: LogEvent, data?: LogEventData) => {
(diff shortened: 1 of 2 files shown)


rholinshead:
This makes sense to me. We are rendering the AIConfigEditor immediately in the webview so the server might not be up. One other thing we can/should do is to not call getServerStatusCallback in AIConfigEditor if the state is readOnly. I'm not sure if that will completely resolve the issue but in theory should prevent the requests from happening until the server is working (assuming our readonly state works for those cases)

saqadri (approved):

Ankush-lastmile:
> One other thing we can/should do is to not call getServerStatusCallback in AIConfigEditor if the state is readOnly.

Created #1325. Not sure if we still want to merge this diff. 

We can republish the aiconfig editor package to test 1325 to validate if this diff is needed.

rholinshead:
@Ankush-lastmile seems like this is no longer needed, right?

Ankush-lastmile:
>@Ankush-lastmile seems like this is no longer needed, right?

Yep. I haven't really seen this in the past week while using the vscode extension

PR #1323, opened 2024-02-23
[ez] Rename setupEnvironmentVariables to setApiKeys

saqadri:
[ez] Rename setupEnvironmentVariables to setApiKeys

TSIA

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1323).
* __->__ #1323
* #1316

PR #1316, opened 2024-02-23
Clean up Welcome

saqadri:
Clean up Welcome

Just a few wording updates to clean up the welcome walkthrough.

Also bring back buttons that got removed in #1295

https://github.com/lastmile-ai/aiconfig/assets/25641935/6a155304-1cb6-46b1-b11e-b9231fabc542

Walkthrough command pallette:

image

PR #1324, opened 2024-02-23
Support .yml file extensions

saqadri:
Support .yml file extensions

This was something we noticed at customer offsite where many people tried creating aiconfigs with *.yml extension.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1324).
* __->__ #1324
* #1323
* #1316

### DIFF
--- a/vscode-extension/package.json
+++ b/vscode-extension/package.json
@@ -80,6 +80,9 @@
             "filenamePattern": "*.aiconfig.yaml"
           },
           {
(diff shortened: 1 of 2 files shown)


rholinshead (approved):
LGTM. I did a quick pass through aiconfig core as well and looks like is_yaml_extension checks already include .yml and .yml extension for existing file should be preserved on save

PR #1325, opened 2024-02-23
[local editor] Don't poll server status when in read only view

Ankush-lastmile:
[local editor] Don't poll server status when in read only view

In order to prevent showing status error message when in read only view, this diff makes sure that the aiconfig editor client does not poll the server status when in read only mode.

See #1320 for more context

## Testplan
1. Manually set readonly prop to True
2. Load local editor (Loaded in prod mode, I first built the editor with yarn build)
3. kill the python process that is hosting the aiconfig server.

| Before  | After |
| after  | before |

Note: the `killpython` command in my testplan is a custom alias I have defined that tries to kill all pid's that have the descriptor "python" in its name.

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
+++ b/python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
@@ -1033,7 +1033,7 @@ function AIConfigEditorBase({
   // Don't poll if server status is in an error state since it won't automatically recover
   const getServerStatusCallback = callbacks?.getServerStatus;
   useEffect(() => {
-    if (!getServerStatusCallback || serverStatus !== "OK") {
+    if (readOnly || !getServerStatusCallback || serverStatus !== "OK") {
       return;
     }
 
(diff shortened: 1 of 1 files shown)


python/src/aiconfig/editor/client/src/components/AIConfigEditor.tsx
```diff
@@ -1038,16 +1038,18 @@ function AIConfigEditorBase({
     }
 
     const interval = setInterval(async () => {
-      try {
-        const res = await getServerStatusCallback();
-        setServerStatus(res.status);
-      } catch (err: unknown) {
-        setServerStatus("ERROR");
+      if (!readOnly) {
```
rholinshead, line 1041: This can just be done above with the other checks to return, rather than setting the interval (otherwise it will unnecessarily run this interval every heartbeat_interval ms

rholinshead: If `readOnly` is in the dependency array for the effect then the effect will rerun and set a new interval (and dispose of the existing one in the case of toggling *to* readonly)

Ankush-lastmile: thanks for clarifying, updated as suggested.

Ankush-lastmile:
- Moved readOnly check to the beggining checks of the useEffect()
- followed steps in testplan -> works as intended. No changes in test plan

rossdanlm:
UI change for VS Code: Let's have follow up to say "if we're in read-only now, tell them why they can't do changes" -->

1. because we're currently initializing (VS Code server start up)
2. we got an error and need to display it now

rossdanlm (approved):
Stamping to unblock, we can follow up more later

PR #1326, opened 2024-02-23
[Swapping imageds between mac and windows]

zakariassoul:
[Swapping imageds between mac and windows]


zakariassoul:
@Ankush-lastmile can you test this on your windows VM?

saqadri (commented):
I don't think we should do this

saqadri (changes requested):
I don't think we should do this

zakariassoul:
Reverting changed. Will push another PR

PR #1328, opened 2024-02-23
[3/n] Restart Editor Servers on Python Interpreter Change

rholinshead:
# [3/n] Restart Editor Servers on Python Interpreter Change

When interpreter changes restart all editors if the user selects it. We need to leverage async locking mechanism for managing the server port otherwise each of the simultaneously-restarted servers could end up with duplicate port number.

Seems that even with just the lock it won't work, so need to explicitly manage the starting port in the search (to a reasonable max, before restarting search at 8000 again).

https://github.com/lastmile-ai/aiconfig/assets/5060851/7708c288-2878-420e-9130-4074ca9a11b5

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1328).
* __->__ #1328
* #1319
* #1318

### DIFF
--- a/vscode-extension/src/editor_server/serverPortManager.ts
+++ b/vscode-extension/src/editor_server/serverPortManager.ts
@@ -1,0 +1,25 @@
+import { Mutex } from "async-mutex";
+import { getPortPromise } from "portfinder";
+
+/**
+ * Super simple class to manage server port allocation using a lock mechanism. If
+ * we don't do this, restarting multiple servers simultaneously can result in
+ * obtaining the same port for multiple servers.
+ */
+class ServerPortManager {
+  private lock;
+
+  constructor() {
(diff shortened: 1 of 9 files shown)


vscode-extension/src/editor_server/serverPortManager.ts
```diff
@@ -0,0 +1,25 @@
+import { Mutex } from "async-mutex";
+import { getPortPromise } from "portfinder";
+
+/**
+ * Super simple class to manage server port allocation using a lock mechanism. If
```
saqadri, line 5: niiiice. using locks at work :D

saqadri (approved):

PR #1333, opened 2024-02-24
Fix error starting server on new aiconfig yaml

saqadri:
Fix error starting server on new aiconfig yaml

When you create a new aiconfig yaml, the server fails to start.

This is because of changes to the flow where we are calling `updateServerState` in the launch flow of the editor, which in turn calls `getModeForDocument`. This function uses the file extension to determine if the document is JSON or YAML, but in the case of Untitled documents, there is no file URI.

Since it is reasonable to expect this flow for Untitled documents, the fix here is to use the document's content to determine the mode (try parsing it as JSON or as YAML).

https://github.com/lastmile-ai/lastmile/assets/25641935/1c96d099-d970-4e5f-9b58-3768451e591c

### DIFF
--- a/vscode-extension/package.json
+++ b/vscode-extension/package.json
@@ -180,6 +180,7 @@
   "dependencies": {
     "@vscode/python-extension": "^1.0.5",
     "async-mutex": "^0.4.1",
+    "js-yaml": "^4.1.0",
     "oboe": "^2.1.5",
     "portfinder": "^1.0.32",
     "ufetch": "^1.6.0"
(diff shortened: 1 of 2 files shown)


rholinshead (approved):
Ah, makes sense. I think this ok and accepting to unblock. I read through some related things (e.g. https://github.com/microsoft/vscode/issues/93441#issuecomment-646238906) and not sure if there is some workaround of setting the untitle file extension/path explicitly (would probably also resolve the other P0 about the file extension being just `.json` when saving the new JSON file)

PR #1334, opened 2024-02-24
[RFC] Handle creating files as *.aiconfig.yaml or *.aiconfig.json files

saqadri:
[RFC] Handle creating files as *.aiconfig.yaml or *.aiconfig.json files

Untitled files have a lot of constraints on how they can be set up. Currently it defaults to saving the file as *.json or *.yaml.

This RFC is about saving the file as *.aiconfig.json or *.aiconfig.yaml.

Unfortunately this isn't perfect either, because it now saves the file as untitled-1.aiconfig.json (where `1` gets incremented). No file save as dialog is shown in the flow.

An alternative is to have the `Create New` command ask for a file name or path on creation (and enforce the file ending in the accepted extensions)

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1334).
* #1336
* #1335
* __->__ #1334
* #1333

### DIFF
--- a/vscode-extension/src/extension.ts
+++ b/vscode-extension/src/extension.ts
@@ -208,12 +208,6 @@ async function createNewAIConfig(
   const workspaceUri = vscode.workspace.workspaceFolders
     ? vscode.workspace.workspaceFolders[0].uri
     : null;
-  const untitledUri = workspaceUri
-    ? workspaceUri.with({
-        scheme: "untitled",
-        path: `${workspaceUri.path}/untitled.aiconfig.${mode}`,
-      })
-    : vscode.Uri.parse(`untitled:untitled.aiconfig.${mode}`);
 
   // Specify the initial content here
   const newAIConfigJSON = vscode.Uri.joinPath(
@@ -233,15 +227,27 @@ async function createNewAIConfig(
   const fileContentBuffer = await vscode.workspace.fs.readFile(fileContentPath);
   const initialContent = fileContentBuffer.toString();
 
-  const doc = await vscode.workspace.openTextDocument({
-    content: initialContent,
-    language: mode,
-  });
+  let doc: vscode.TextDocument = null;
+  let i = 0;
+  while (!doc) {
+    const fileName = `untitled${i === 0 ? "" : "-" + i}.aiconfig.${mode}`;
+    const filePath = workspaceUri
+      ? path.join(workspaceUri.path, fileName)
+      : fileName;
+    const untitledUri = vscode.Uri.file(filePath).with({ scheme: "untitled" });
+    try {
+      doc = await vscode.workspace.openTextDocument(untitledUri);
+      break;
+    } catch (err) {
+      i++;
+    }
+  }
 
-  //const doc = await vscode.workspace.openTextDocument(untitledUri);
-  await vscode.window.showTextDocument(doc, {
-    preview: false,
-    viewColumn: vscode.ViewColumn.One,
+  const editor = await vscode.window.showTextDocument(doc, { preview: false });
+
+  // TODO: saqadri - if the edits aren't applied, consider retrying or showing an error message
+  const editsApplied = await editor.edit((editBuilder) => {
+    editBuilder.insert(new vscode.Position(0, 0), initialContent);
   });
 
   await vscode.commands.executeCommand(


rholinshead:
I really like the idea of prompting for the name. We can default to the first available untitled name and let the user change that before showing the file. Have https://github.com/lastmile-ai/aiconfig/pull/1337 that I just need to test and put up (will do so tomorrow since I need to head out)

rossdanlm:
> I really like the idea of prompting for the name. We can default to the first available untitled name and let the user change that before showing the file. Have #1337 that I just need to test and put up (will do so tomorrow since I need to head out)

Personally for UI experience would prefer being able to create new, then only name on save (similar to Jupyter notebooks, MS Word, etc). I like the solution you found in https://github.com/microsoft/vscode/issues/93441#issuecomment-646238906 which does:
1. creates an untitled empty doc (with `schema: "untitled"`)
2. populates it with empty content for user to get started
3. when users saves it, it already exists and has the name prefilled

But yea not a blocker, we can fix forward to what UI experience we want best

vscode-extension/src/extension.ts
```diff
@@ -233,15 +227,27 @@ async function createNewAIConfig(
...
+    const untitledUri = vscode.Uri.file(filePath).with({ scheme: "untitled" });
+    try {
+      doc = await vscode.workspace.openTextDocument(untitledUri);
+      break;
+    } catch (err) {
+      i++;
+    }
+  }
+
+  const editor = await vscode.window.showTextDocument(doc, { preview: false });
```
rossdanlm, line 246: I like preview: false. I wish this was the default setting whenever opening docs so you don't have to double-click to keep file tabs sticky lol

vscode-extension/src/extension.ts
```diff
@@ -233,15 +227,27 @@ async function createNewAIConfig(
   const fileContentBuffer = await vscode.workspace.fs.readFile(fileContentPath);
   const initialContent = fileContentBuffer.toString();
 
-  const doc = await vscode.workspace.openTextDocument({
-    content: initialContent,
-    language: mode,
-  });
+  let doc: vscode.TextDocument = null;
+  let i = 0;
+  while (!doc) {
```
rossdanlm, line 232: nit: would like this to be clearer like "newConfigDocument"

vscode-extension/src/extension.ts
```diff
@@ -233,15 +227,27 @@ async function createNewAIConfig(
...
+  }
+
+  const editor = await vscode.window.showTextDocument(doc, { preview: false });
 
-  //const doc = await vscode.workspace.openTextDocument(untitledUri);
-  await vscode.window.showTextDocument(doc, {
-    preview: false,
-    viewColumn: vscode.ViewColumn.One,
+  // TODO: saqadri - if the edits aren't applied, consider retrying or showing an error message
+  const editsApplied = await editor.edit((editBuilder) => {
```
rossdanlm, line 249: What does this `editsApplied` variable do? I don't see it used anywhere else? Is it a global var?

vscode-extension/src/extension.ts
```diff
@@ -233,15 +227,27 @@ async function createNewAIConfig(
...
-    content: initialContent,
-    language: mode,
-  });
+  let doc: vscode.TextDocument = null;
+  let i = 0;
+  while (!doc) {
+    const fileName = `untitled${i === 0 ? "" : "-" + i}.aiconfig.${mode}`;
+    const filePath = workspaceUri
+      ? path.join(workspaceUri.path, fileName)
+      : fileName;
```
rossdanlm, line 236: Similar comment I left in https://github.com/lastmile-ai/aiconfig/pull/1337#discussion_r1501707203 for Ryan's diff, but yea this works

vscode-extension/src/extension.ts
```diff
@@ -233,15 +227,27 @@ async function createNewAIConfig(
...
+    const untitledUri = vscode.Uri.file(filePath).with({ scheme: "untitled" });
+    try {
+      doc = await vscode.workspace.openTextDocument(untitledUri);
+      break;
+    } catch (err) {
+      i++;
+    }
```
rossdanlm, line 243: Ryan's PR has some more checks like determining if the file already exists (yours will open an existing file if that already exists instead of new one) (or at least overwrite existing file contents from lines 246 --> 249). But that's ok, we can land this and Ryan can fix forward

rossdanlm (approved):
I personally prefer the UI experience of opening a file immediately instead of requiring user to input a name, but Ryan has a few more checks in https://github.com/lastmile-ai/aiconfig/pull/1337 like checking if the `untitled-${i}.aiconfig.${mode}` file already exists, so going to stamp to unblock this, and Ryan can add those checks in here later, not a big deal. Making progress is better than doing nothing imo

rholinshead:
> I personally prefer the UI experience of opening a file immediately instead of requiring user to input a name, but Ryan has a few more checks in #1337 like checking if the `untitled-${i}.aiconfig.${mode}` file already exists, so going to stamp to unblock this, and Ryan can add those checks in here later, not a big deal. Making progress is better than doing nothing imo

Would be great to have feedback from @tanya-rai, @zakariassoul, or @andrew-lastmile for which implementation we do end up going with.

Current state (before this PR or #1337) is:
- Create new opens Untitled file
- On save, user is prompted to save the file with a name. The extension is defaulted either .json or .yaml (NOT .aiconfig.json or .aiconfig.yaml)
- Major problem is that saving without the .aiconfig. part of the extension means this file won't open in our custom editor next time it's opened

Based on how vscode handles new/untitled files, it doesn't appear to be possible to specify the extension as .aiconfig.json or .aiconfig.yaml, so we are going to go with either the UX of this PR or #1337.

This PR:
- Creates the file as untitled.aiconfig.json immediately (incrementing as untitled-1, etc.)
- no 'save as prompt' on save

#1337:
- Immediately prompts user for the name of the file, defaulting to untitled.aiconfig.json (incrementing as untitled-1, etc.)
- no 'save as prompt' on save

So we can consider 2 paths:
1. User is fine with default untitled name
- This PR: no action needed after save
- #1337: user must hit 'Enter' to confirm the default name

2. User wants a different name
- This PR: user must open the file, then find the file in the file tree, then right click -> rename
- #1337: user enters a different name immediately

My assumption is that flow 2 is going to be vastly more common. My thoughts are that the flow for renaming an existing file (i.e. this PR) is much higher friction than just immediately typing the name of the file (#1337). Similarly, naming the file immediately is also an existing flow when clicking + New File in the left file tree whereas there doesn't appear to be any precedent for defaulting to a file name and requiring manually renaming it after the fact (aside from the usual Untitled flow which explicitly prompts to rename, which as noted above is not a possible flow)

rossdanlm:
Sounds good, no matter what we choose, I agree whatever we do is better than what we have currently so happy to accept whatever and get us moving forwards and we can re-evaluate in the future as P2 for best UI if we want

rholinshead:
> Sounds good, no matter what we choose, I agree whatever we do is better than what we have currently so happy to accept whatever and get us moving forwards and we can re-evaluate in the future as P2 for best UI if we want

I'll land #1337 for now so we have something for tomorrow. We can discuss and I can update/fix forward with this UX if that ends up being the desired one

rossdanlm:
Closing since we now have https://github.com/lastmile-ai/aiconfig/pull/1351

PR #1336, opened 2024-02-24
Add some more contribution points

saqadri:
Add some more contribution points

While investigating untitled naming, I came across some additional contribution points that VSCode enables. Adding some additional places to surface our commands.

https://github.com/lastmile-ai/aiconfig/assets/25641935/ca806614-f15a-4df7-9965-7e5a89867c21

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1336).
* __->__ #1336
* #1335


rholinshead (approved):
Nice!

vscode-extension/package.json
```diff
@@ -86,7 +90,6 @@
             "filenamePattern": "*.aiconfig"
           }
         ],
-        "enableForDiffView": false,
```
rossdanlm, line 89: Why remove this?

rossdanlm:
> While investigating untitled naming, I came across some additional contribution points that VSCode enables.

Nit: for future pls link resources so it's easier for others to see where you learned that info. Ex: https://code.visualstudio.com/api/references/contribution-points

vscode-extension/package.json
```diff
@@ -150,7 +153,31 @@
           }
         ]
       }
-    ]
+    ],
+    "menus": {
+      "editor/title": [
+        {
+          "command": "vscode-aiconfig.share",
+          "when": "resourcePath =~ /.*(\\.aiconfig\\.json|\\.aiconfig\\.yaml|\\.aiconfig\\.yml|\\.aiconfig)$/",
```
rossdanlm, line 161: Will the double backslashes work on both windows and mac?

vscode-extension/package.json
```diff
@@ -150,7 +153,31 @@
...
+      ],
+      "file/newFile": [
+        {
+          "command": "vscode-aiconfig.createAIConfigJSON"
+        },
+        {
+          "command": "vscode-aiconfig.createAIConfigYAML"
+        }
+      ],
+      "webview/context": [
```
rossdanlm, line 173: Ok so just to confirm, this is whenever you right click inside of the editor webview? https://code.visualstudio.com/api/extension-guides/webview#context-menus

If so, do you think it would also be good to have the "set API Keys" command here too?

rossdanlm (approved):
Have some questions but this generally looks cool

vscode-extension/package.json
rholinshead: It was added previously but actually doesn't do anything. The diff view is configured to use default diff view as a result of the code further down:
```
    "configurationDefaults": {
      "workbench.editorAssociations": {
        "{git,gitlens,sapling-diff}:/**/*.{aiconfig.json, aiconfig.yaml}": "default"
      }
    },
```

vscode-extension/package.json
saqadri: That's just an escape for the regex string, it shouldn't be specific to the OS, just the regex parser.

saqadri: We can validate this on windows

vscode-extension/package.json
saqadri: Yes, that would be a great command to add as well. Good idea!

saqadri: Added

PR #1337, opened 2024-02-24
[vscode] Prompt User for Name When Creating New AIConfig

rholinshead:
[vscode] Prompt User for Name When Creating New AIConfig

# [vscode] Prompt User for Name When Creating New AIConfig

VS Code api for creating new/untitled files does not allow specifying the extension to use for the file on save. This results in untitled files being prompted to save with either .json or .yaml extension, NOT .aiconfig.json or .aiconfig.yaml. This is a big problem, since using the non-aiconfig extension will result in the file not opening with our custom editor next time it's opened.

As a workaround, we can default the untilted filename to untitle.aiconfig.json/yaml and prompt the user for the proper name when creating it (before showing the contents). See #1334 for the alternative, of saving the file immediately as untitled and requiring the user to rename manually if they want afterwards.

https://github.com/lastmile-ai/aiconfig/assets/5060851/57a0978a-e377-4c88-bc9f-48dcfc03c8bb

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1337).
* #1342
* __->__ #1337

### DIFF
--- a/vscode-extension/src/extension.ts
+++ b/vscode-extension/src/extension.ts
@@ -18,6 +18,7 @@ import {
   isSupportedConfigExtension,
   SUPPORTED_FILE_EXTENSIONS,
   setupEnvironmentVariables,
+  validateNewConfigName,
 } from "./util";
 import {
   getPythonPath,
@@ -76,7 +77,7 @@ export async function activate(context: vscode.ExtensionContext) {
   const createAIConfigJSONCommand = vscode.commands.registerCommand(
     COMMANDS.CREATE_NEW_JSON,
     async () => {
-      return await createNewAIConfig(context, aiconfigEditorManager, "json");
+      return await createNewAIConfig(context, "json");
     }
   );
   context.subscriptions.push(createAIConfigJSONCommand);
@@ -84,7 +85,7 @@ export async function activate(context: vscode.ExtensionContext) {
   const createAIConfigYAMLCommand = vscode.commands.registerCommand(
     COMMANDS.CREATE_NEW_YAML,
     async () => {
-      return await createNewAIConfig(context, aiconfigEditorManager, "yaml");
+      return await createNewAIConfig(context, "yaml");
     }
   );
   context.subscriptions.push(createAIConfigYAMLCommand);
@@ -202,19 +203,8 @@ export function deactivate() {
  */
 async function createNewAIConfig(
   context: vscode.ExtensionContext,
-  aiconfigEditorManager: AIConfigEditorManager,
   mode: "json" | "yaml" = "json"
 ) {
-  const workspaceUri = vscode.workspace.workspaceFolders
-    ? vscode.workspace.workspaceFolders[0].uri
-    : null;
-  const untitledUri = workspaceUri
-    ? workspaceUri.with({
-        scheme: "untitled",
-        path: `${workspaceUri.path}/untitled.aiconfig.${mode}`,
-      })
-    : vscode.Uri.parse(`untitled:untitled.aiconfig.${mode}`);
-
   // Specify the initial content here
   const newAIConfigJSON = vscode.Uri.joinPath(
     context.extensionUri,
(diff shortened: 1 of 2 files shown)


cookbooks/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "vscode-aiconfig.pythonInterpreter": "/opt/homebrew/Caskroom/miniconda/base/envs/aiconfig/bin/python"
+}
```
rossdanlm, line 3: Let's not land these for now since they'll cause merge conflicts

vscode-extension/src/extension.ts
```diff
@@ -233,22 +223,64 @@ async function createNewAIConfig(
   const fileContentBuffer = await vscode.workspace.fs.readFile(fileContentPath);
   const initialContent = fileContentBuffer.toString();
 
-  const doc = await vscode.workspace.openTextDocument({
-    content: initialContent,
-    language: mode,
+  const workspaceUri = vscode.workspace.workspaceFolders
+    ? vscode.workspace.workspaceFolders[0].uri
```
rossdanlm, line 227: Since we're just using `workspaceUri.path` later, let's just use `uri.path` here directly. If workspace folder is defined `.uri.fsPath` must exist and can't be undefined: 

Screenshot 2024-02-24 at 17 26 26

vscode-extension/src/extension.ts
```diff
@@ -233,22 +223,64 @@ async function createNewAIConfig(
...
+  const newConfigName = await vscode.window.showInputBox({
+    prompt: "Enter a name for the new AIConfig file",
+    value: firstAvailableUntitledName,
+    validateInput: (input) => {
+      const nameValidationError = validateConfigName(input, mode);
+      if (nameValidationError) {
+        return nameValidationError;
+      }
+
+      return fs.existsSync(input) ? "File already exists" : null;
```
rossdanlm, line 255: Can we just put this inside of the `validateConfigName()` method to keep it cleaner?

vscode-extension/src/extension.ts
```diff
@@ -233,22 +223,64 @@ async function createNewAIConfig(
...
 
-  //const doc = await vscode.workspace.openTextDocument(untitledUri);
-  await vscode.window.showTextDocument(doc, {
-    preview: false,
-    viewColumn: vscode.ViewColumn.One,
+  const newConfigFilePath = workspaceUri
+    ? path.join(workspaceUri.path, newConfigName)
+    : newConfigName;
+  const newConfigUri = vscode.Uri.file(newConfigFilePath).with({
+    scheme: "untitled",
```
rossdanlm, line 263: Why is this needed? Tried looking it up on [VSCode's github](https://github.com/microsoft/vscode-uri) but not super clear what untitled schema does:

Screenshot 2024-02-24 at 17 31 42

vscode-extension/src/util.ts
```diff
@@ -381,6 +381,26 @@ export async function setupEnvironmentVariables(
   }
 }
 
+export function validateConfigName(name: string, mode: "json" | "yaml") {
+  if (name === "") {
+    return "Name is required";
```
rossdanlm, line 386: Nit: "Filename is required"

vscode-extension/src/util.ts
```diff
@@ -381,6 +381,26 @@ export async function setupEnvironmentVariables(
...
+  }
+  if (name.includes(" ")) {
+    return "Name cannot contain spaces";
```
rossdanlm, line 389: Why is this an issue?

vscode-extension/src/util.ts
```diff
@@ -381,6 +381,26 @@ export async function setupEnvironmentVariables(
...
+  }
+  if (mode === "json" && !name.endsWith("aiconfig.json")) {
```
rossdanlm, line 391: nit: ".aiconfig.json" otherwise things like "myaiconfig.json" are valid

vscode-extension/src/util.ts
```diff
@@ -381,6 +381,26 @@ export async function setupEnvironmentVariables(
...
+    return "Name must end with aiconfig.json";
+  }
+  if (
+    mode === "yaml" &&
+    !name.endsWith("aiconfig.yaml") &&
+    !name.endsWith("aiconfig.yml")
```
rossdanlm, line 397: Same about ".aiconfig.y[aml|ml]"

rossdanlm (approved):
lgtm but please address comments before landing

cookbooks/.vscode/settings.json
rholinshead: Oh, whoops - thanks. Didn't see that sneak in

vscode-extension/src/extension.ts
rholinshead: From https://code.visualstudio.com/api/references/document-selector#:~:text=The%20untitled%20scheme%20is%20for,yet%20been%20saved%20to%20disk.

> The untitled scheme is for new files that have not yet been saved to disk.

Which is technically the case until they save it for the first time

vscode-extension/src/util.ts
rholinshead: I'll test it out, but just assumed having spaces in the file name would be problematic as we've seen in other places

rholinshead: Tested it out, made a config with spaces and it ran and saved fine

rholinshead:
Going to land this so we have something better than current UX. Can fix forward with #1334 if that's the desired ux

PR #1340, opened 2024-02-25
Add completion events to walkthrough steps

rossdanlm:
Add completion events to walkthrough steps

Ok, upon further investigation, I found out that the `completionEvents` doesn't actually do anything anymore: https://github.com/microsoft/vscode/issues/176074#issuecomment-1676995414. However, this no-op is being reverted in next VS Code update in https://github.com/microsoft/vscode/issues/176074#issuecomment-1912698405, so it's still good to included this here!!! We should then advertise to people to pls update their VS Code version!!!

You can read more about completion events in https://code.visualstudio.com/api/references/contribution-points#Completion-events

~~Now we shouldn't auto-mark steps as completed unless command has been run~~

I also added "CMD/CTRL" to the readme files and renamed the readme files to be non-capitalized. I also made some content changes in the walkthrough mardown because it was annoying me, but feel it's better and more precise.

## Test Plan

This actually has no functional changes at all becuase of https://github.com/microsoft/vscode/issues/176074#issuecomment-1676995414 until users update the latest VS Code version (which I tried to do but still doesn't work for me :/ )

https://github.com/lastmile-ai/aiconfig/assets/151060367/d9a74035-1951-4cf7-adbc-4c4b2bcad6c1

Trying on Jupyter notebooks walkthrough also doesn't work for me

https://github.com/lastmile-ai/aiconfig/assets/151060367/39231e92-fe53-41bd-ac4a-cf94725975f5

## Appendix

The timeline for this is pretty wild:
1. Someone doesn't like having checkboxes so removed the functionality in https://github.com/microsoft/vscode/issues/166747 to automatically check once a step is viewed. The PR for addressing this was done in Jan 31, 2023: https://github.com/microsoft/vscode/pull/172941
2. People complained about this, so https://github.com/microsoft/vscode/issues/176074 was filed on March 3, 2023
3. Change for reverting back to original completionEvent feature was done in January 24, 2024: https://github.com/microsoft/vscode/pull/203573

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1340).
* #1346
* #1345
* #1343
* #1341
* __->__ #1340
* #1329

### DIFF
--- a/vscode-extension/package.json
+++ b/vscode-extension/package.json
@@ -119,34 +119,42 @@
           {
             "id": "initializeExtension",
             "title": "Initialize Extension",
-            "description": "Get the extension set up for first use.\n[Initialize Extension](command:vscode-aiconfig.init)",
+            "description": "Ensure you have everything needed before using the extension for the first time\n[Initialize Extension](command:vscode-aiconfig.init)",
             "media": {
               "markdown": "media/initializeExtension.md"
-            }
+            },
+            "completionEvents": ["onCommand:vscode-aiconfig.init"]
           },
           {
             "id": "setEnvironment",
             "title": "Set API Keys",
-            "description": "Set up a .env file with the API keys needed to access the models you want to use.\n[Set API Keys](command:vscode-aiconfig.setApiKeys)",
+            "description": "Define your model API keys in an ``.env`` file\n[Set API Keys](command:vscode-aiconfig.setApiKeys)",
             "media": {
               "markdown": "media/apiKeys.md"
-            }
+            },
+            "completionEvents": ["onCommand:vscode-aiconfig.setApiKeys"]
           },
           {
             "id": "createAIConfig",
             "title": "Create your first AIConfig",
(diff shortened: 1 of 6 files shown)


vscode-extension/package.json
```diff
@@ -119,34 +119,42 @@
...
               "markdown": "media/initializeExtension.md"
-            }
+            },
+            "completionEvents": ["onCommand:vscode-aiconfig.init"]
           },
           {
             "id": "setEnvironment",
             "title": "Set API Keys",
-            "description": "Set up a .env file with the API keys needed to access the models you want to use.\n[Set API Keys](command:vscode-aiconfig.setApiKeys)",
+            "description": "Define your model API keys in an ``.env`` file\n[Set API Keys](command:vscode-aiconfig.setApiKeys)",
```
rholinshead, line 131: nit, an -> a

rholinshead (approved):

vscode-extension/package.json
rossdanlm: I'm gonna keep as "an" because I feel most people would say "env" instead of "dot env" but yea I know what you mean, can fix later if we want

PR #1341, opened 2024-02-25
Cache versions of extension into global settings

rossdanlm:
Cache versions of extension into global settings

This is kind of a hack. I'm not sure why the extension walkthrough isn't opening when it first installs, but whatever we need to do this to ensure that walkthrough runs, because it's not running when users first download the extension (could be perhaps that we've already installed it so global settings is aware of it, idk but whatever)

Adding a global config for `vscode-aiconfig.version`, and also changed the activate settings to activate upon vscode startup. This is fine because we moved the initialize step outside of here so there's no negative consequences of activating it.

Will implement the actual functionality next PRs

## Test Plan
Ok I don't know why I can't test directly bvy running the activation command, but instead I just hooked this up to the 'Set API Keys" command and ran it lol

https://github.com/lastmile-ai/aiconfig/assets/151060367/449e599a-4ed7-457f-9543-7e7895337ebe

### DIFF
--- a/vscode-extension/src/utilities/versionUpdateUtils.ts
+++ b/vscode-extension/src/utilities/versionUpdateUtils.ts
@@ -1,0 +1,44 @@
+/**
+ * Utils file for handling actions to perform whenever user
+ * installs or updates the extension.
+ * VS Code automatically checks for version updates
+ * (https://code.visualstudio.com/docs/editor/extension-marketplace#_extension-autoupdate)
+ * but I haven't been able to find a way to use this to trigger any
+ * actions. Also even though the Walkthrough documentation says
+ * that this gets triggered on first install, I haven't been able
+ * to trigger this either, so I am calling this manually
+ * https://code.visualstudio.com/api/references/contribution-points#contributes.walkthroughs
+ */
+
+import * as vscode from "vscode";
+
+import { EXTENSION_NAME } from "../util";
+import { VERSION_KEY_NAME } from "../constants";
+
+export async function performVersionInstallAndUpdateActionsIfNeeded(
+  context: vscode.ExtensionContext
+) {
+  const extension = vscode.extensions.getExtension(context.extension.id);
+  const currExtensionVersion = extension?.packageJSON.version;
+  const config = vscode.workspace.getConfiguration(EXTENSION_NAME);
+  const lastActivatedVersion = config.get<string>(VERSION_KEY_NAME);
+  if (currExtensionVersion !== lastActivatedVersion) {
+    await config.update(
+      VERSION_KEY_NAME,
+      currExtensionVersion,
+      vscode.ConfigurationTarget.Global
+    );
+  }
+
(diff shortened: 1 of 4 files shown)


rholinshead:
> also changed the activate settings to activate upon vscode startup. This is fine because we moved the initialize step outside of here so there's no negative consequences of activating it.

I'm not entirely sure this is true because our extension requires python extension and our activation uses it, so it does add the negative consequence of initializing the python extension (which may be resource intensive)

vscode-extension/src/utilities/versionUpdateUtils.ts
```diff
@@ -0,0 +1,42 @@
+/**
+ * Utils file for handling actions to perform whenever user
+ * installs or updates the extension.
+ * VS Code automatically checks for version updates
+ * (https://code.visualstudio.com/docs/editor/extension-marketplace#_extension-autoupdate)
+ * but I haven't been able to find a way to use this to trigger any
+ * actions. Also even though the Walkthrough documentation says
```
rholinshead, line 7: Could look into vscode.extensions.onDidChange in the future but not sure if it fires on update or just add/remove/enable/disable

vscode-extension/src/utilities/versionUpdateUtils.ts
```diff
@@ -0,0 +1,42 @@
...
+) {
+  const extension = vscode.extensions.getExtension(context.extension.id);
+  const currExtensionVersion = extension?.packageJSON.version;
+  const config = vscode.workspace.getConfiguration(EXTENSION_NAME);
+  const lastActivatedVersion = config.get<string>(VERSION_KEY_NAME);
+  await config.update(
+    VERSION_KEY_NAME,
+    currExtensionVersion,
+    vscode.ConfigurationTarget.Global
+  );
```
rholinshead, line 29: I think this should only be done if it's not already set / doesn't match. Otherwise it'll unnecessarily write the same version each time

rholinshead (approved):
Accepting to unblock, but please move the config version update to only happen when it's needed

vscode-extension/src/utilities/versionUpdateUtils.ts
rossdanlm: Yea sure I can do that

rossdanlm:
Changed activation events from `onStartupFinished` to `onView:extension` (based on https://code.visualstudio.com/api/references/activation-events#onView and https://github.com/search?q=%2FonView%5C%3A*extension*%2F++path%3A**%2Fpackage.json&type=code&p=5), and `"workspaceContains:**.aiconfig.**"`. This will ensure that when users first download the AIConfig Editor extension, they will most likely be in the extension view, and therefore the extension will activate when they first see it. I wasn't sure if the `onStartupFinished` would still trigger for the current VS Code session if it still opened. 

I really wanted to see if there was a manual way to check an event when VS Code extension first installs, but couldn't find anything for that.

rossdanlm:
Also updated to only write global config version if the current version and last active version are different

PR #1342, opened 2024-02-25
[wip] Don't Create Duplicate Text Editor for new Config

rholinshead:
[wip] Don't Create Duplicate Text Editor for new Config

Still looking into this a bit. We can create a new aiconfig doc with aiconfig scheme and that will go through our content provider to get the initial content. However, this now results in a significant bug where the new file isn't marked as dirty (even with changes), so I think there's some disconnection between the text document and editor :s

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1342).
* __->__ #1342
* #1337

### DIFF
--- a/vscode-extension/src/extension.ts
+++ b/vscode-extension/src/extension.ts
@@ -18,6 +18,7 @@ import {
   isSupportedConfigExtension,
   SUPPORTED_FILE_EXTENSIONS,
   setupEnvironmentVariables,
+  validateNewConfigName,
 } from "./util";
 import {
   getPythonPath,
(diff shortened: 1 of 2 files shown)


rossdanlm:
I only investigated for about 15 mins, but it seems that registering this is designed for read-only content only: https://code.visualstudio.com/api/extension-guides/virtual-documents

> The text document content provider API allows you to create readonly documents in Visual Studio Code from arbitrary sources.


It seems if we want to update, we can emit an event for onDidChange and update that, but that seems kind of weird for our use case since our editor is the one initiating the change itself: https://code.visualstudio.com/api/extension-guides/virtual-documents#update-virtual-documents

Perhaps a hack we can do is

1. Create new command
2. Find the .json text file editor window tab open with that name
3. Force close it
4. When we force close it, force the error not to show up (not sure how we'd do that)

I feel this wouldn't be ideal but could work

rholinshead:
Closing in favour of https://github.com/lastmile-ai/aiconfig/pull/1351

PR #1345, opened 2024-02-25
Change file titles in media folder

rossdanlm:
Change file titles in media folder

This is really annoying, changing the capitalized letter of a filename or some reason is not being detected by Git or Sapling, so changing the entire media file to `media_change`, just to change it back to `media` next PR

## Test PLan
This obviously breaks it but next PR when I "revert" it back with new un-capitalized filenames it will work

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1345).
* #1346
* __->__ #1345


rholinshead:
You should be able to do this in a single PR. Just `sl mv` to the intermediate name, and then `sl mv` to desired name

rholinshead (approved):

rossdanlm:
> You should be able to do this in a single PR. Just `sl mv` to the intermediate name, and then `sl mv` to desired name

Oh ok, will do next time

PR #1351, opened 2024-02-26
Proper Fix for Ideal New/Untitled Flow

rholinshead:
# [rfc] Proper (?) Fix for Ideal New/Untitled Flow

This implements a proper(?) fix for creating new/untitled aiconfig files:
- treats the file as regular Untitled (but with proper `.aiconfig.json/yaml` extension), prompting for filename only on first save, no longer needing to prompt immediately on create (undoing #1337)
- now, only one tab is opened and it opens the custom editor

The main change/fix is using a `WorkspaceEdit` to hold the initial content changes instead of opening/showing a new text editor with the content. I found that solution from https://github.com/microsoft/vscode/issues/93441

However, now the webview seems to reload the file after saving it for the first time, which makes it go back to readonly and start up the server :( --> it seems to dispose the untitled webview (killing the server) and refresh with a new server...

https://github.com/lastmile-ai/aiconfig/assets/5060851/8a471b5e-f6d3-4860-a780-949ca928394a

Not sure if this is overall better because of the server thing. Need to investigate some more. Fwiw, current scenario on main with 2 tabs is probably a worse UX because:
- obviously, 2 tabs is confusing
- closing text editor tab seems to break custom editor (can't make edits in custom editor after)? Also, closing text editor tab has the same notification about unable to update server
- on top of ^, need to name the file beforehand, as opposed to on save

### DIFF
--- a/vscode-extension/src/extension.ts
+++ b/vscode-extension/src/extension.ts
@@ -7,7 +7,6 @@ import {
   PythonExtension,
 } from "@vscode/python-extension";
 import { ufetch } from "ufetch";
-import fs from "fs";
 import path from "path";
 import { AIConfigEditorProvider } from "./aiConfigEditor";
 import {
@@ -26,7 +25,6 @@ import {
   isSupportedConfigExtension,
   SUPPORTED_FILE_EXTENSIONS,
   setupEnvironmentVariables,
-  validateNewConfigName,
   getConfigurationTarget,
 } from "./util";
 import {
@@ -228,7 +226,6 @@ async function createNewAIConfig(
   );
 
   const fileContentPath = mode === "json" ? newAIConfigJSON : newAIConfigYAML;
-
   const fileContentBuffer = await vscode.workspace.fs.readFile(fileContentPath);
   const initialContent = fileContentBuffer.toString();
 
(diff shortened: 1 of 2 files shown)


saqadri (approved):
Wow amazing find! I love this. I think the untitled TextDocument gets disposed and a new one with the actual path is opened, which is why the old webview is getting disposed. 
IMO this isn't too bad. My main concern is the "Failed to start aiconfig server" error, which may be too aggressive and definitely confusing for this state.

My proposal:
1. We ship this
2. We tune down the "failed to start" message. I'm confused why it's showing up in this case but we can investigate that separately?

rholinshead:
> Wow amazing find! I love this. I think the untitled TextDocument gets disposed and a new one with the actual path is opened, which is why the old webview is getting disposed. IMO this isn't too bad. My main concern is the "Failed to start aiconfig server" error, which may be too aggressive and definitely confusing for this state.
> 
> My proposal:
> 
> 1. We ship this
> 2. We tune down the "failed to start" message. I'm confused why it's showing up in this case but we can investigate that separately?

I'm going to look into the "Failed to update" message some more this morning. I think saving results in updating the text document while the untitled tab is being disposed, so probably some race condition there. I honestly don't know if we even want that notification to show in any case

vscode-extension/src/extension.ts
```diff
@@ -228,54 +226,29 @@ async function createNewAIConfig(
...
-    }
-  }
-
-  const newConfigName = await vscode.window.showInputBox({
-    prompt: "Enter a name for the new AIConfig file",
-    value: firstAvailableUntitledName,
-    validateInput: (input) => validateNewConfigName(input, mode),
-  });
-
+  const untitledFileName = `untitled.aiconfig.${mode}`;
```
rossdanlm, line 236: I like having the `firstAvailableUntitledName` like what we had before, but get now that we're not actually saving it yet, so maybe there's a way to check for both existing filenames that exist, and also files that are currently open. Not a blocker

PR #1358, opened 2024-02-26
[editor] Fix Readonly Rendering of Undefined Model Settings

rholinshead:
# [editor] Fix Readonly Rendering of Undefined Model Settings

Currently when rendering the editor in readonly, opening the model settings for a prompt which has no metadata and which has no prompt schema, results in a js/render error:
Screenshot 2024-02-26 at 2 54 14 PM
Screenshot 2024-02-26 at 2 54 32 PM

This is actually the problem Zak surfaced for the vscode doc of a blank screen when opening the settings for an aiconfig in vscode.

The problem is that Prism expects the contents to be a string, which obviously isn't the case if there is no metadata (it's undefined). To fix, just don't render anything in this specific scenario.

## Testing:
- Make sure rendering in and out of readonly works:
Screenshot 2024-02-26 at 2 52 00 PM
Screenshot 2024-02-26 at 2 52 20 PM

### DIFF
--- a/python/src/aiconfig/editor/client/src/components/JSONRenderer.tsx
+++ b/python/src/aiconfig/editor/client/src/components/JSONRenderer.tsx
@@ -16,12 +16,27 @@ export default memo(function JSONRenderer({
   schema,
 }: Props) {
   const { readOnly } = useContext(AIConfigContext);
-  // Prism is a read only renderer.
-  return !onChange || readOnly ? (
-    <Prism language="json" styles={{ code: { textWrap: "pretty" } }}>
-      {JSON.stringify(content, null, 2)}
-    </Prism>
-  ) : (
-    <JSONEditor content={content} onChangeContent={onChange} schema={schema} />
-  );
+  let renderer = null;
+
+  if (!onChange || readOnly) {
+    if (content != null) {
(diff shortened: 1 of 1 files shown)


python/src/aiconfig/editor/client/src/components/JSONRenderer.tsx
```diff
@@ -16,12 +16,27 @@ export default memo(function JSONRenderer({
...
-    </Prism>
-  ) : (
-    <JSONEditor content={content} onChangeContent={onChange} schema={schema} />
-  );
+  let renderer = null;
+
+  if (!onChange || readOnly) {
+    if (content != null) {
+      // Prism is a read only renderer, but requires a string input. Just render
+      // null if the content is empty in readonly
```
rossdanlm, line 24: thanks for the comment, took me a while to understand the the only difference in this PR was L22 for the null-check

rossdanlm (approved):

python/src/aiconfig/editor/client/src/components/JSONRenderer.tsx
rholinshead: Ya, just figured it was better to pull into if/else rather than another ternary :)

PR #1375, opened 2024-02-27
[docs] fixed duplication on extension readme

vanshb03:
Small fix on readme that removed duplication for the video

## Before Change:
image

## After Change: 
image


rholinshead (approved):
Awesome, thanks!

PR #1383, opened 2024-02-29
[server] 3/n Endpoint for set_env

Ankush-lastmile:
[server] 3/n Endpoint for set_env

Summary:

Stack contains changes that updates overall environment setup flow for vscode extension.

This diff implements the server side endpoint. Will test on top with vscode

Test Plan:
```
payload={'env_file_path': {}}
Status Code: 400
Response JSON: {'message': 'Invalid request, specified .env file path is not a string.'}

payload={'env_file_path': '/hi'}
Status Code: 400
Response JSON: {'message': 'Specified .env file path does not exist.'}

payload={'env_file_path': '.env'}
Status Code: 200
Response JSON: {'message': 'Updated .env file path.'}
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1383).
* #1384
* __->__ #1383
* #1382

### DIFF
--- a/python/src/aiconfig/editor/server/server.py
+++ b/python/src/aiconfig/editor/server/server.py
@@ -8,6 +8,7 @@ import threading
 import time
 import uuid
 import webbrowser
+import os
 from typing import Any, Dict, Literal, Type, Union
 
 import lastmile_utils.lib.core.api as core_utils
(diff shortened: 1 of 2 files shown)


python/src/aiconfig/editor/server/server.py
```diff
@@ -829,3 +829,25 @@ def set_aiconfigrc() -> FlaskResponse:
...
+
+    This method does NOT load the .env file into the server state.
+    It only updates the path to the .env file. See api/run
+
+    """
+    request_json = request.get_json()
+    state = get_server_state(app)
+
+    # Validate
+    request_env_path = request_json.get("env_file_path")
```
rholinshead, line 846: Can we do basic validation of the path here as well? It would seem really weird to successfully set an invalid path and only learn that it's not valid when I go to run a prompt later

Ankush-lastmile: yea we could but vscode already validates that. Can add anyway for more robustness if you think we should add it

rholinshead: Just for the sake of correctness, I think we should. Maybe just:
- path exists / is accessible
- filename format is correct

Ankush-lastmile: basic validation sounds good to me, added and updated with testplan

Ankush-lastmile:
- added basic validation of .env 
- updated testplan

python/src/aiconfig/editor/server/server.py
```diff
@@ -829,3 +830,61 @@ def set_aiconfigrc() -> FlaskResponse:
...
+                },
+                400,
+            )
+        )
+    
+    if filename_format_correct is False:
+        return FlaskResponse(
+            (
+                {
+                    "message": "Specified .env file path is not a .env file.",
```
rholinshead, line 882: nit, `Specified env file` instead of `.env` to reduce confusion

rholinshead: also nit, should do this check before checking if path exists

rholinshead (approved):

python/src/aiconfig/editor/server/server.py
Ankush-lastmile: - updated ordering
- modified comment

PR #1384, opened 2024-02-29
[server] 4/n Update Use .env at the run step

Ankush-lastmile:
[server] 4/n Update Use .env at the run step

Summary:

Test Plan:

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1384).
* __->__ #1384
* #1383
* #1382

### DIFF
--- a/python/src/aiconfig/editor/server/server.py
+++ b/python/src/aiconfig/editor/server/server.py
@@ -8,6 +8,7 @@ import threading
 import time
 import uuid
 import webbrowser
+import os
 from typing import Any, Dict, Literal, Type, Union
 
 import lastmile_utils.lib.core.api as core_utils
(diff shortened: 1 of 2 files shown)


python/src/aiconfig/editor/server/server.py
```diff
@@ -334,7 +334,8 @@ def run() -> FlaskResponse:
 
     # Allow user to modify their environment keys without reloading the server.
     # Execution time of `0.001s` is arbitrary, but should be small enough to not be noticeable.
-    dotenv.load_dotenv()
+    # Should override be set to True?
+    dotenv.load_dotenv(state.env_path)
```
Ankush-lastmile, line 338: RFC, Thanks!

rholinshead: Good question. I say yes, we should override if env_path is set.

If we don't override, then calling `set_env_file_path` will effectively be a no-op if the file contains variables that already exist (e.g. from a previous call). This would be confusing in a scenario where we call `set_env_file_path` with one file and then call it again with another. From the server standpoint, I would say:
- if no state.env_path, then don't override (keep existing behaviour)
- if state.env_path is set, there has been an explicit request to set env path and would be ideal for the variables in that env file to take precedence

But, from the vscode perspective, is there any scenario where `set_env_file_path` will be called with different paths for the same server? Is the server persisted when moving a file to a different dir, for example? If there is such a case I think we *must* override if the path is set, otherwise we would not be using the closest/expected env variables. I guess the fact we have a 'set API keys' command means this is the case.

More generally, the only way I could make an argument to not override is:
- if it can't be changed for an existing server
- user is familiar with load_dotenv so they know existing environment (e.g. set in shell) takes precedence over .env file

IMO this argument isn't very strong and as a user if I explicitly do an action to "Set API Keys" I would 100% want the key I just set to be used instead of whatever may already be set for the same variable

Ankush-lastmile: Thanks for the analysis. I'm in agreement. In the meantime I've updated to set override to true

Ankush-lastmile:
- set override to true

python/src/aiconfig/editor/server/server.py
```diff
@@ -335,7 +335,8 @@ def run() -> FlaskResponse:
 
     # Allow user to modify their environment keys without reloading the server.
     # Execution time of `0.001s` is arbitrary, but should be small enough to not be noticeable.
-    dotenv.load_dotenv()
+    # Should override be set to True?
```
rholinshead, line 338: Maybe update comment to state something like,
"Override if env_file_path is specified so that provided .env values take precedence over existing values"

python/src/aiconfig/editor/server/server.py
```diff
@@ -335,7 +335,8 @@ def run() -> FlaskResponse:
...
+    dotenv.load_dotenv(state.env_file_path, override= True)
```
rholinshead, line 339: Let's do `override=(state.env_file_path is not None)` (or whatever the clean python way is). We should not override if no path is set since that changes behaviour

rholinshead (approved):
Accepting to unblock, but please make suggested changes

python/src/aiconfig/editor/server/server.py
Ankush-lastmile: thanks, +1. You're pretty good at python 😉

PR #1385, opened 2024-02-29
[ez] Specify content type in Share request based on JSON vs. YAML document

rossdanlm:
[ez] Specify content type in Share request based on JSON vs. YAML document

TSIA, Sarmad aded change to render it correctly in the lastmile repo (https://github.com/lastmile-ai/lastmile/pull/1380), but still should be changing the content type now:

## Test Plan

It actually still works before, not sure if content-type is a blocker. But just good to specify I guess

https://github.com/lastmile-ai/aiconfig/assets/151060367/4f7e6c8f-7198-45a9-b92c-43560ff5ba0d

### DIFF
--- a/vscode-extension/src/extension.ts
+++ b/vscode-extension/src/extension.ts
@@ -26,6 +26,7 @@ import {
   SUPPORTED_FILE_EXTENSIONS,
   setupEnvironmentVariables,
   getConfigurationTarget,
+  getModeFromDocument,
 } from "./util";
 import {
   initialize,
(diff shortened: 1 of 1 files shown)


vscode-extension/src/extension.ts
```diff
@@ -495,17 +496,13 @@ async function shareAIConfig(
...
   const uploadUrl = "https://s3.amazonaws.com/lastmileai.aiconfig.public/";
   const randomPath = Math.round(Math.random() * 10000);
   const uploadKey: string = `aiconfigs/${getTodayDateString()}/${randomPath}/${sanitizedFileName}`;
-
-  // TODO: Will also need to check for yaml files and change the contentType accordingly
-  const contentType = "application/json";
+  const contentType =
+    getModeFromDocument(activeEditor.document) === "json"
+      ? "application/json"
+      : "application/yaml";
```
rholinshead, line 505:
```suggestion
  const contentType = `application/${getModeFromDocument(activeEditor.document)}`;
```
nit

rholinshead (approved):
Just a nit

vscode-extension/src/extension.ts
rossdanlm: I didn't want to do this in case someday we update `getModeFromDocument()` to also return `"yml"`

PR #1390, opened 2024-02-29
[vscode] 7/n listeners and .env

Ankush-lastmile:
[vscode] 7/n listeners and .env

Summary:

When the .env file path is set, vscode will invoke api/set_env on each AIConfig server.

Defines a listener to a newly introduced setting for the env_file_path.

This ensures that users do not need to restart vscode to pickup env changes.

Test Plan:

1. set env file path
2. changes picked up
3. changes picked up on restart
4. Changes picked up when manually removing vscode setting and re-setting env file path and api key

https://github.com/lastmile-ai/aiconfig/assets/141073967/07ac1d43-28ef-439f-98e7-94a4e73e11c9

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1390).
* #1398
* __->__ #1390
* #1387
* #1389

### DIFF
--- a/vscode-extension/package.json
+++ b/vscode-extension/package.json
@@ -128,6 +128,11 @@
           "type": "string",
           "default": "",
           "description": "Version of the AIConfig Editor extension last time it was activated. We use this value to check if the extension has been updated to prompt users to reload VS Code"
+        },
+        "vscode-aiconfig.env_file_path": {
+          "type": "string",
+          "default": "",
+          "description": "File path to your .env file containing your API keys"
         }
       }
     },
--- a/vscode-extension/src/editor_server/editorServer.ts
+++ b/vscode-extension/src/editor_server/editorServer.ts
@@ -3,6 +3,7 @@ import { EXTENSION_NAME } from "../util";
 import { getPythonPath } from "../utilities/pythonSetupUtils";
 import { ChildProcessWithoutNullStreams, spawn } from "child_process";
 import serverPortManager from "./serverPortManager";
+import { ENV_FILE_PATH } from "../constants";
 
 export enum EditorServerState {
   Starting = "Starting",
@@ -73,6 +74,8 @@ export class EditorServer {
       const modelRegistryPathArgs = modelRegistryPath
         ? ["--parsers-module-path", modelRegistryPath]
         : [];
+      const filePath = config.get(ENV_FILE_PATH) as string;
+      const envFilePathArgs = filePath ? ["--env-file-path", filePath] : [];
 
       this.port = await serverPortManager.getPort();
 
@@ -89,6 +92,7 @@ export class EditorServer {
           "--server-port",
           this.port.toString(),
           ...modelRegistryPathArgs,
+          ...envFilePathArgs,
         ],
         {
           cwd: this.cwd,
--- a/vscode-extension/src/extension.ts
+++ b/vscode-extension/src/extension.ts
@@ -27,6 +27,7 @@ import {
   setupEnvironmentVariables,
   getConfigurationTarget,
   getModeFromDocument,
+  updateServerEnv
 } from "./util";
 import {
   initialize,
@@ -34,6 +35,7 @@ import {
   savePythonInterpreterToCache,
 } from "./utilities/pythonSetupUtils";
 import { performVersionInstallAndUpdateActionsIfNeeded } from "./utilities/versionUpdateUtils";
+import { ENV_FILE_PATH } from "./constants";
 
 // This method is called when your extension is activated
 // Your extension is activated the very first time a command is executed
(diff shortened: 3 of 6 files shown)


vscode-extension/src/util.ts
```diff
@@ -98,6 +102,15 @@ export async function updateServerState(
...
 
+export async function updateServerEnv(serverUrl: string, filePath: string) {
+  return await ufetch.post(
+    EDITOR_SERVER_ROUTE_TABLE.SET_ENV_FILE_PATH(serverUrl),
+    {
+      [ENV_FILE_PATH]: filePath,
+    }
+  );
+}
+
```
Ankush-lastmile, line 113: for anyone unaware, 
`[ENV_FILE_PATH]` is syntax equivalent to resolving the value of `ENV_FILE_PATH` as the key for the js object.

vscode-extension/package.json
```diff
@@ -128,6 +128,11 @@
           "type": "string",
           "default": "",
           "description": "Version of the AIConfig Editor extension last time it was activated. We use this value to check if the extension has been updated to prompt users to reload VS Code"
+        },
+        "vscode-aiconfig.env_file_path": {
+          "type": "string",
+          "default": "",
+          "description": "env_file_path"
```
rholinshead, line 135: Description should be clearer on what this is for

vscode-extension/src/util.ts
```diff
@@ -383,6 +394,12 @@ export async function setupEnvironmentVariables(
       "Please define your environment variables."
     );
   }
+
+  // Update Server Env FLow
```
rholinshead, line 398: Do we technically still need `lowestCommonWorkspacePath` validation above now, or can .env path be anywhere?

vscode-extension/src/util.ts
```diff
@@ -383,6 +394,12 @@ export async function setupEnvironmentVariables(
...
+  // Set the .env file path in the settings
+  // vscode Extension has a listener for changes defined at activation.
+  const config = vscode.workspace.getConfiguration(EXTENSION_NAME);
```
rholinshead, line 401: Can you give some context on why we need to persist this? This would result in this persisted .env file being used for a given config instead of any .env file colocated in the folder with the config, right?

Ankush-lastmile: >This would result in this persisted .env file being used for a given config instead of any .env file colocated in the folder with the config, right?

Correct. With the two following nuances
1.  Setting the env path would be stored in either the workspace or global workspace vscode settings depending on where you are.
2. A potential .env file colocated with the config would work IF the setting from 1 is not set, since `load_dotenv` will pull the the closest env file.

The only other alternative I see is to _not_ store `env_file_path` as an extension setting, and ONLY update the currently open editor file. However, I'm not sure how I feel about that vs. this way to persist and use the env_file_path for all aiconfigs in workspace.

vscode-extension/src/util.ts
Ankush-lastmile: Not needed after this diff, will remove

vscode-extension/package.json
Ankush-lastmile: Updated to "File path to your .env file containing your API keys"

vscode-extension/src/editor_server/editorServer.ts
```diff
@@ -73,6 +74,8 @@ export class EditorServer {
       const modelRegistryPathArgs = modelRegistryPath
         ? ["--parsers-module-path", modelRegistryPath]
         : [];
+      const filePath = config.get(ENV_FILE_PATH) as string;
```
rholinshead, line 77: oh, one last thing - should this be `as string | undefined`?

vscode-extension/src/extension.ts
```diff
@@ -208,6 +210,25 @@ export async function activate(context: vscode.ExtensionContext) {
       }
     )
   );
+
+  // Handle changes to the .env path
+  context.subscriptions.push(vscode.workspace.onDidChangeConfiguration(async (event) => {
+    if (event.affectsConfiguration("vscode-aiconfig")) {
+      // Get new env
+      const envPath = vscode.workspace.getConfiguration("vscode-aiconfig").get(ENV_FILE_PATH) as string;
```
rholinshead, line 218: Same here, unless deleting it makes its value empty string instead of undefined

vscode-extension/src/extension.ts
```diff
@@ -208,6 +210,25 @@ export async function activate(context: vscode.ExtensionContext) {
       }
     )
   );
+
+  // Handle changes to the .env path
+  context.subscriptions.push(vscode.workspace.onDidChangeConfiguration(async (event) => {
```
rholinshead, line 215: no need for async

vscode-extension/src/extension.ts
```diff
@@ -208,6 +210,25 @@ export async function activate(context: vscode.ExtensionContext) {
...
+    if (event.affectsConfiguration("vscode-aiconfig")) {
+      // Get new env
+      const envPath = vscode.workspace.getConfiguration("vscode-aiconfig").get(ENV_FILE_PATH) as string;
+      console.log(`New .env path set: ${envPath}`);
+      // set env on all AIConfig Servers
+      const editors: Array<AIConfigEditorState> = Array.from(
+        aiconfigEditorManager.getRegisteredEditors()
+      );
+      if (editors.length > 0) {
+        editors.forEach(async (editor) => {
```
rholinshead, line 225: no need for async if not awaiting

rholinshead (approved):
Just a few small things

vscode-extension/src/editor_server/editorServer.ts
Ankush-lastmile: same as below

Ankush-lastmile:
- added an await statement when invoking updateServerEnv
- modified event handler to execute only if the extension setting for env_file_path is modified

PR #1395, opened 2024-03-04
[ez] vscode extension - fix diff editor

saqadri:
[ez] vscode extension - fix diff editor

The bug is a simple spacebar between`aiconfig.json,<SPACE>aiconfig.yaml`. This affects the regex used to apply the editor association.

Tested that now opening both aiconfig.yaml and aiconfig.json files in diff editor open them in regular text editor.

For good measure, also added support for *.aiconfig files, and tested that too.

### DIFF
--- a/vscode-extension/package.json
+++ b/vscode-extension/package.json
@@ -128,7 +128,7 @@
     },
     "configurationDefaults": {
       "workbench.editorAssociations": {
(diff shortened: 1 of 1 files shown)


rholinshead (approved):
facepalm


Thanks!

PR #1396, opened 2024-03-04
Implement HuggingFaceVisualQuestionAnsweringRemoteInferencePromptSchema

rholinshead:
Implement HuggingFaceVisualQuestionAnsweringRemoteInferencePromptSchema

# Implement HuggingFaceVisualQuestionAnsweringRemoteInferencePromptSchema

For https://github.com/lastmile-ai/aiconfig/issues/1392

This will add the prompt schema so that visual question answering prompts have the nice UI for input and settings

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1396).
* #1397
* __->__ #1396

PR #1397, opened 2024-03-04
Render Attachment before Data When Both Exist in Input

rholinshead:
# Render Attachment before Data When Both Exist in Input

See comment in https://github.com/lastmile-ai/aiconfig/pull/1396#pullrequestreview-1915432291

When we have a Prompt Input with both attachment and data, most likely the attachment will be the main focus (e.g. Visual Question Answering) so we can show it first in the prompt input:

Screenshot 2024-03-04 at 5 08 17 PM
Screenshot 2024-03-04 at 5 08 44 PM

Was initially thinking we could order it based on the properties order in the schema, but just doing this for now since property order in schema may not be guaranteed when we define it in python and then serialize it to to client (will just be passing JSON object in response)

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1397).
* __->__ #1397
* #1396


Ankush-lastmile (approved):
lgtm, we can iterate on top as needed

PR #1399, opened 2024-03-05
{WIP} RFC 2: Schema update to distinguish model & model parser

Ankush-lastmile:
{WIP} RFC 2: Schema update to distinguish model & model parser

Summary:

Test Plan:


python/src/aiconfig/schema.py
```diff
@@ -136,9 +136,11 @@ class Error(BaseModel):
 
 
 class ModelMetadata(BaseModel):
-    # The ID of the model to use.
+    # The ID of the inference provider aka Model Parser to use.
```
rholinshead, line 139: How would these parser be registered? Would it be `registerModelParser(new Parser(model))` for some list of models, instead of `registerModelParser(new Parser(), [model list])`? This is one thing we should address with this update

Ankush-lastmile: for this method which would essentially unlink the model from model parser, it would be

`registerModelParser(new Parser())`

whereas the other notation would like this:

`registerModelParser(new Parser(), [modelParserAlias list])`

Ankush-lastmile: >This is one thing we should address with this update

+1 from my side, I think the problem boils down to how do we id these and how does that setup look like

Ankush-lastmile:
closing for. Stay tuned for v2 😊

PR #1416, opened 2024-03-07
Implement OpenAIVisionParser

rholinshead:
Implement OpenAIVisionParser

# Implement OpenAIVisionParser

Summary: Using the default openai parser as the basis (and extending it), implement a parser for text-and-image openai models (e.g. gpt-4-v). The key differences are:
- vision models don't support function calling or tools, so exclude from the completion params
- serialize/deserialize need to be a bit different to construct the user messages' content with text and image urls

One thing to note is the default max tokens for GPT-4V is quite small. Bumping it to 100 in #1419

Another thing I found while testing is that the model gets confused sometimes when you refer to the attached image as an image. For example, I had a prompt as "What is this image of" and it responded with "I can't provide information on any images until you upload one". Then I reran a handful of times and it finally gave expected result. Changing the prompt to "What is this?" seems to provide much better results.

Test Plan:

https://github.com/lastmile-ai/aiconfig/assets/5060851/47e0ae90-fab7-4d4f-98e9-163cd992aeed

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1416).
* #1419
* #1418
* #1417
* __->__ #1416
* #1415

### DIFF
--- a/python/src/aiconfig/default_parsers/openai_vision.py
+++ b/python/src/aiconfig/default_parsers/openai_vision.py
@@ -1,0 +1,440 @@
+import copy
+import re
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
+from aiconfig.util.params import (
+    resolve_prompt_string,
+    resolve_system_prompt,
+)
+
+from aiconfig.callback import CallbackEvent
+
+from .openai import DefaultOpenAIParser, build_output_data
+
+from openai.types.chat import ChatCompletionMessage
+
+from aiconfig.schema import (
+    Attachment,
+    AttachmentDataWithStringValue,
+    ExecuteResult,
+    OutputDataWithValue,
+    Prompt,
+    PromptInput,
+    PromptMetadata,
+)
+
+if TYPE_CHECKING:
+    from aiconfig.Config import AIConfigRuntime
+
+IMAGE_FORMAT_PATTERN = r"data:image/(\w+);base64"
+
+
+class OpenAIVisionParser(DefaultOpenAIParser):
+    async def serialize(
+        self,
+        prompt_name: str,
+        data: Dict[str, Any],
+        ai_config: "AIConfigRuntime",
+        parameters: Optional[Dict[str, Any]],
+        **kwargs,
+    ) -> List[Prompt]:
+        """
+        Defines how prompts and model inference settings get serialized in the .aiconfig.
+
+        Similar to serialize of DefaultOpenAIParser, however currently, GPT-4 Turbo
+        with vision does not support the message.name parameter, functions/tools,
+        response_format parameter
+
+        Args:
+            prompt (str): The prompt to be serialized.
+            inference_settings (dict): Model-specific inference settings to be serialized.
+
+        Returns:
+            str: Serialized representation of the prompt and inference settings.
+        """
+        event = CallbackEvent(
+            "on_serialize_start",
+            __name__,
+            {
+                "prompt_name": prompt_name,
+                "data": data,
+                "parameters": parameters,
+                "kwargs": kwargs,
+            },
+        )
+        await ai_config.callback_manager.run_callbacks(event)
+        prompts: list[Prompt] = []
+
+        # Combine conversation data with any extra keyword args
+        conversation_data = {**data}
+
+        if not "messages" in conversation_data:
+            raise ValueError(
+                "Data must have `messages` array to match openai api spec"
+            )
+
+        # Find first system prompt. Every prompt in the config will bet set to use this system prompt.
+        system_prompt = None
+        for message in conversation_data["messages"]:
+            if "role" in message and message["role"] == "system":
+                # Note: system prompt for openai represented will be an object with content and role attributes {content: str, role: str}
+                system_prompt = message
+                break
+
+        # Get the global settings for the model
+        model_name = (
+            conversation_data["model"]
+            if "model" in conversation_data
+            else self.id()
+        )
+
+        model_metadata = ai_config.get_model_metadata(
+            conversation_data, model_name
+        )
+        # Remove messages array from model metadata. Handled separately
+        if model_metadata.settings:
+            model_metadata.settings.pop("messages", None)
+
+        # Add in system prompt
+        if system_prompt:
+            if not model_metadata.settings:
+                model_metadata.settings = {}
+            model_metadata.settings["system_prompt"] = system_prompt
+
+        i = 0
+        while i < len(conversation_data["messages"]):
+            message = conversation_data["messages"][i]
+            role = message["role"]
+            if role == "user":
+                # Serialize User message as a prompt and save the assistant response as an output
+                assistant_response: Union[ChatCompletionMessage, None] = None
+                if i + 1 < len(conversation_data["messages"]):
+                    next_message = conversation_data["messages"][i + 1]
+                    if next_message["role"] == "assistant":
+                        assistant_response = next_message
+                        i += 1
+                new_prompt_name = f"{prompt_name}_{len(prompts) + 1}"
+
+                content = message["content"]
+                input = {}
+                # Message can contain text and multiple images
+                # See https://platform.openai.com/docs/guides/vision/multiple-image-inputs
+                for val in content:
+                    if val["type"] == "text":
+                        input["data"] = val["text"]
+                    if val["type"] == "image_url":
+                        if input.get("attachments") is None:
+                            input["attachments"] = []
+                        image_url: str = val["image_url"]["url"]
+                        attachment_kind = (
+                            "base64"
+                            if image_url.startswith("data:")
+                            else "file_uri"
+                        )
+                        attachment_data = AttachmentDataWithStringValue(
+                            kind=attachment_kind, value=image_url
+                        )
+
+                        mime_type = "image/*"
+                        if attachment_kind == "base64":
+                            type_match = re.search(
+                                IMAGE_FORMAT_PATTERN, image_url
+                            )
+                            if type_match:
+                                mime_type = f"image/{type_match.group(1)}"
+
+                        input["attachments"].append(
+                            Attachment(
+                                data=attachment_data, mime_type=mime_type
+                            )
+                        )
+
+                # openai sdk currently only supports images with user messages
+                assistant_output = []
+                if assistant_response is not None:
+                    output_data = build_output_data(assistant_response)
+                    metadata = {
+                        "raw_response": assistant_response,
+                        role: "assistant",
+                    }
+                    assistant_output = [
+                        ExecuteResult(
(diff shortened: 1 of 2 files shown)


python/src/aiconfig/default_parsers/openai_vision.py
```diff
@@ -0,0 +1,427 @@
...
+                        parameters=parameters,
+                        remember_chat_context=True,
+                    ),
+                    outputs=assistant_output,
+                )
+                prompts.append(prompt)
+            elif i == 0 and role == "assistant":
+                # If the first message is an assistant message,
+                # build a prompt with an empty input,
+                # and the assistant response as the output
```
rossdanlm, line 171: This is so annoying lol

python/src/aiconfig/default_parsers/openai_vision.py
```diff
@@ -0,0 +1,427 @@
...
+
+                # Pull assistant response
+                assistant_output = build_output_data(
+                    conversation_data["messages"][i]
+                )
+                prompt = Prompt(
+                    name=f"{prompt_name}_{len(prompts) + 1}",
```
rossdanlm, line 178: Why `len(prompts) + 1` instead of `i`?

python/src/aiconfig/default_parsers/openai_vision.py
```diff
@@ -0,0 +1,427 @@
...
+                        parameters=parameters,
+                        remember_chat_context=True,
+                    ),
+                    outputs=[
+                        ExecuteResult(
+                            output_type="execute_result",
+                            execution_count=None,
+                            data=assistant_output,
+                            metadata={},
+                        )
```
rossdanlm, line 191: Let's create issue to followup in the edge-case where we receive multiple assistant outputs as the first message, not just a single chatbot text. Basically convert this to iterate over first n-assistant messages and update `i` accordingly

python/src/aiconfig/default_parsers/openai_vision.py
```diff
@@ -0,0 +1,427 @@
...
+            role = message["role"]
+            if role == "user":
+                # Serialize User message as a prompt and save the assistant response as an output
+                assistant_response: Union[ChatCompletionMessage, None] = None
+                if i + 1 < len(conversation_data["messages"]):
+                    next_message = conversation_data["messages"][i + 1]
+                    if next_message["role"] == "assistant":
+                        assistant_response = next_message
+                        i += 1
+                new_prompt_name = f"{prompt_name}_{len(prompts) + 1}"
```
rossdanlm, line 112: Same as comment below, why not use `i` since we can now have prompts with the same name

python/src/aiconfig/default_parsers/openai_vision.py
```diff
@@ -0,0 +1,427 @@
...
+
+                content = message["content"]
+                input = {}
+                # Message can contain text and multiple images
```
rossdanlm, line 116: nit: can you link the API or code pointer. Not a blocker

python/src/aiconfig/default_parsers/openai_vision.py
```diff
@@ -0,0 +1,427 @@
...
+                    if val["type"] == "text":
+                        input["data"] = val["text"]
+                    if val["type"] == "image_url":
+                        if input.get("attachments") is None:
+                            input["attachments"] = []
+                        image_url: str = val["image_url"]["url"]
+                        attachment_kind = (
+                            "base64"
+                            if image_url.startswith("data:")
+                            else "file_uri"
```
rossdanlm, line 127: Going to trust you lol

python/src/aiconfig/default_parsers/openai_vision.py
```diff
@@ -0,0 +1,427 @@
...
+                        )
+                        attachment_data = AttachmentDataWithStringValue(
+                            kind=attachment_kind, value=image_url
+                        )
+                        input["attachments"].append(
+                            Attachment(
+                                data=attachment_data, mime_type="image/*"
```
rossdanlm, line 134: So there's no way of detecting mime type from openAI? If it's a file_uri can't we read the extension?

python/src/aiconfig/default_parsers/openai_vision.py
```diff
@@ -0,0 +1,427 @@
...
+                            data=assistant_output,
+                            metadata={},
+                        )
+                    ],
+                )
+                prompts.append(prompt)
+            i += 1
+
+        if prompts:
+            prompts[len(prompts) - 1].name = prompt_name
```
rossdanlm, line 198: Do we still need this if we do the refactoring part above for prompt name to use `i`?

python/src/aiconfig/default_parsers/openai_vision.py
```diff
@@ -0,0 +1,427 @@
...
+
+    completion_data = {}
+    for key in supported_keys:
+        if key in model_settings:
+            completion_data[key] = model_settings[key]
+
+    # Explicitly set the model to use if not already specified
+    if completion_data.get("model") is None:
+        model_name = aiconfig.get_model_name(prompt)
+        completion_data["model"] = model_name
```
rossdanlm, line 342: cc @Ankush-lastmile please I can't take it anymore!!!! Think about the children!

python/src/aiconfig/default_parsers/openai_vision.py
rholinshead: file_uri is basically any non-base64 uri -- in practice it's usually a url to some online blob, which won't necessarily have any extension. I *think* we could technically use a regex to look for extension in base64 case, if it's in format `data:image/x` so I can try that

python/src/aiconfig/default_parsers/openai_vision.py
rholinshead: This is the exact same logic from existing openai parser, minus the function and tools message handling (since that's not supported for vision).

using len(prompts) won't have the same name since prompts is added to with user message here. I don't know if `i` would work as well since `i` is incremented for each message (not just user message) so you could end up with prompt_name and then prompt_name_3 if there are assistant messages between user messages

python/src/aiconfig/default_parsers/openai_vision.py
rholinshead: Answered above. But basically len(prompts) accounts for user message count (one prompt per user message) whereas i would include assistant message (output) count

python/src/aiconfig/default_parsers/openai_vision.py
rholinshead: I think it's needed regardless. If my understanding is correct it's just assuming that the user is serializing a full message history and wants the latest prompt to be named exactly prompt_name instead of some suffix for uniqueness

python/src/aiconfig/default_parsers/openai_vision.py
```diff
@@ -0,0 +1,427 @@
...
+                completion_params["messages"].append(
+                    {"content": resolved_system_prompt, "role": "system"}
+                )
+
+            # Default to always use chat context
+            if not hasattr(prompt.metadata, "remember_chat_context") or (
+                hasattr(prompt.metadata, "remember_chat_context")
+                and prompt.metadata.remember_chat_context != False
+            ):
+                # handle chat history. check previous prompts for the same model. if same model, add prompt and its output to completion data if it has a completed output
```
rossdanlm, line 258: Thanks for comment, this is pretty tricky dicky

python/src/aiconfig/default_parsers/openai_vision.py
```diff
@@ -0,0 +1,427 @@
...
+                        if prompt_content[j].get("text"):
+                            prompt_content[j]["text"] = resolve_prompt_string(
+                                prompt,
+                                params,
+                                aiconfig,
+                                prompt_content[j]["text"],
+                            )
+
+        # Add in the latest prompt
+        add_prompt_as_message(
```
rossdanlm, line 301: Ok yea, this use case I see is going to make the input adapters pretty difficult when we start allowing previous chat history to included non-text inputs to remember. It'll have to be like a nested check for each prompt --> model --> model parser to allow for input and output verification and mapping to the message completion params

python/src/aiconfig/default_parsers/openai_vision.py
```diff
@@ -0,0 +1,427 @@
...
+
+    # Explicitly set the model to use if not already specified
+    if completion_data.get("model") is None:
+        model_name = aiconfig.get_model_name(prompt)
+        completion_data["model"] = model_name
+
+    return completion_data
+
+
+def add_prompt_as_message(
```
rossdanlm, line 347: Is this a new method you made or is it in the default parsers for openAI in other model parsers? If the second one, can we centralized to a utils? not a blocker, can be done end of diff stack, or create a bootcamp task for it

rossdanlm (approved):
What a complex mode parser

python/src/aiconfig/default_parsers/openai_vision.py
rossdanlm: Ah I see, thanks this makes sense

python/src/aiconfig/default_parsers/openai_vision.py
rossdanlm: Not a blocker, if it takes more than 10 mins we can create a followup issue

python/src/aiconfig/default_parsers/openai_vision.py
rholinshead: The same concept/function name/call exists in default openai parser but it is mainly focused on function call / tools. I had to re-implement here to handle images specifically, and exclude function call / tools (not supported)

python/src/aiconfig/default_parsers/openai_vision.py
rholinshead: We should probably pull chat serialization into the config api instead (but that's a larger refactor). So, in run_inference if remember_chat_history is true we do something like
```
history_completion_params = config.deserialize_chat_history(prompt)
```
And that implementation would iteratively call the parser .deserialize on each one.

And then in run_inference we instead have an array of completion params that work for the model, and the parser logic can piece together the messages from each?

i.e. instead of coupling parser with prompt management logic, we have the parser just handle an array of completion params and it can figure out how to construct the message from that array of params (would just be a map over the completion params and append all messages to an array)

rossdanlm: Hmmm, yea, I like having the abstraction between how each model parser deserializes it's own prompt, and then another model parser that connects them together if needed

rholinshead: Instead of array of completion params, can be a conversation context, array of conversation messages (openai spec)

PR #1417, opened 2024-03-07
Add "gpt-4-vision-preview" as default parser

rholinshead:
Add "gpt-4-vision-preview" as default parser

# Add "gpt-4-vision-preview" as default parser

Summary: Just adding this model/parser as a default alongside the other openai defaults

Test Plan: Can use the parser in the editor

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1417).
* #1419
* #1418
* __->__ #1417
* #1416
* #1415

### DIFF
--- a/python/src/aiconfig/Config.py
+++ b/python/src/aiconfig/Config.py
@@ -12,6 +12,7 @@ from .default_parsers.anyscale_endpoint import (
 from .default_parsers.claude import ClaudeBedrockModelParser
 from .default_parsers.dalle import DalleImageGenerationParser
 from .default_parsers.openai import DefaultOpenAIParser
+from .default_parsers.openai_vision import OpenAIVisionParser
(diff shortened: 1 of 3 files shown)


python/src/aiconfig/Config.py
```diff
@@ -41,6 +42,10 @@
 ]
 for model in gpt_models_main:
     ModelParserRegistry.register_model_parser(DefaultOpenAIParser(model))
+
+ModelParserRegistry.register_model_parser(
+    OpenAIVisionParser("gpt-4-vision-preview")
+)
```
rossdanlm, line 48: cc @tanya-rai this would be the 3rd position from drop-down. If you want to change it, we can easily change in a future PR

rossdanlm (approved):

PR #1418, opened 2024-03-07
Update OpenAIVisionParserPromptSchema

rholinshead:
Update OpenAIVisionParserPromptSchema

# Update OpenAIVisionParserPromptSchema

Summary: The schema was added a while ago to allow us to model the multimodal schema in the config. Now that we have the parser, some updates are needed (e.g. fix the settings supported since they differ a bit from the other chat models).

Test Plan: Input and settings render correctly
Screenshot 2024-03-07 at 5 07 20 PM

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1418).
* #1419
* __->__ #1418
* #1417
* #1416
* #1415


python/src/aiconfig/editor/client/src/shared/prompt_schemas/OpenAIVisionParserPromptSchema.ts
```diff
@@ -0,0 +1,124 @@
...
+        },
+      },
+    },
+  },
+  model_settings: {
+    type: "object",
+    properties: {
+      model: {
+        type: "string",
+      },
```
rossdanlm, line 31: Should we also add this as required field? (I know it doesn't do anything but having UI would still be helpful)

python/src/aiconfig/editor/client/src/shared/prompt_schemas/OpenAIVisionParserPromptSchema.ts
```diff
@@ -0,0 +1,124 @@
...
+      system_prompt: {
+        type: "string",
+      },
```
rossdanlm, line 34: For future, wondering if there's a better way of refactoring this so that you can compose different schemas together. It seems like this is used in a bunch of them (if not all?) and having to repeat seems annoying

rossdanlm (approved):

python/src/aiconfig/editor/client/src/shared/prompt_schemas/OpenAIVisionParserPromptSchema.ts
rholinshead: IMO it would be weird to have as required but then actually not be required (because you can leave it empty and it works with the registered model name)

python/src/aiconfig/editor/client/src/shared/prompt_schemas/OpenAIVisionParserPromptSchema.ts
rholinshead: Yes, definitely. When we move to python side we can have reusable constants to construct the schemas from

PR #1421, opened 2024-03-08
[docs] Remove caution text from Gradio Notebooks Getting Started for incompatible SDK version

rossdanlm:
[docs] Remove caution text from Gradio Notebooks Getting Started for incompatible SDK version

This is now resolved with https://github.com/lastmile-ai/gradio-notebook/pull/199 so we don't need this restriction anymore!

## Before

Screenshot 2024-03-08 at 15 34 54


rholinshead:
Should we still mention this would be needed if the space is using gradio-notebook 0.0.15 or below?

rossdanlm:
> Should we still mention this would be needed if the space is using gradio-notebook 0.0.15 or below?

Nah, I feel like that's nitty gritty. Most people if they create a new space, they will be copying from existing one and that will also copy the README.md too. If you are create a new space, you will be using the latest and greatest `gradio-notebook` package anyways

rholinshead (approved):

PR #1445, opened 2024-03-12
1/n remove create new json

Ankush-lastmile:
1/n remove create new json

Will be adding a create empty command. This would result in 4 commands which is a too many. Since yaml is our intended target, we can get rid of the create new json.

## Testplan:
.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1445).
* #1448
* #1447
* __->__ #1445

### DIFF
--- a/vscode-extension/package.json
+++ b/vscode-extension/package.json
@@ -207,8 +207,7 @@
               "markdown": "media/createAIConfig.md"
             },
             "completionEvents": [
-              "onCommand:vscode-aiconfig.createAIConfigYAML",
-              "onCommand:vscode-aiconfig.createAIConfigJSON"
+              "onCommand:vscode-aiconfig.createAIConfigYAML"
             ]
           },
           {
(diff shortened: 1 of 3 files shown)


saqadri:
Why are we removing it? There may be valid reasons someone prefers json instead of yaml. What's wrong with keeping the command?

Ankush-lastmile:
>Why are we removing it? There may be valid reasons someone prefers json instead of yaml. What's wrong with keeping the command?

Oops sorry this was supposed to be in draft mode. Currently working on an "create empty" command which will be in json and yaml, but that also means there's 4 options to choose from and that might be too many

saqadri:
> >Why are we removing it? There may be valid reasons someone prefers json instead of yaml. What's wrong with keeping the command?
> 
> 
> 
> Oops sorry this was supposed to be in draft mode. Currently working on an "create empty" command which will be in json and yaml, but that also means there's 4 options to choose from and that might be too many

Ah! Understood. Yeah one thing we can do is have a single Create command, but it walks through some steps (like json, yaml), and (empty, from template). Just a thought. It might get annoying to type that each time though

Ankush-lastmile:
closing for now, will revisit

PR #1447, opened 2024-03-13
[vscode] 1/n rename create new yaml -> create example yaml

Ankush-lastmile:
[vscode] 1/n rename create new yaml -> create example yaml

VSCode extension needs to have a "create new empty" command. This diff just renames the current create new commands to create example.

rfc: Maybe this should be create template instead of "example" to be more explicit

## testplan

https://github.com/lastmile-ai/aiconfig/assets/141073967/0f79234c-f62c-4df5-995c-9261e9013538

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1447).
* #1448
* __->__ #1447

### DIFF
--- a/vscode-extension/package.json
+++ b/vscode-extension/package.json
@@ -56,13 +56,13 @@
         "category": "AIConfig"
       },
       {
-        "command": "vscode-aiconfig.createAIConfigJSON",
-        "title": "Create New AIConfig (json)",
+        "command": "vscode-aiconfig.createExampleAIConfigYAML",
+        "title": "Create AIConfig From Example  (yaml)",
         "category": "AIConfig"
       },
       {
-        "command": "vscode-aiconfig.createAIConfigYAML",
-        "title": "Create New AIConfig (yaml)",
+        "command": "vscode-aiconfig.createExampleAIConfigJSON",
+        "title": "Create AIConfig From Example (json)",
         "category": "AIConfig"
       },
       {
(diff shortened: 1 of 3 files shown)


vscode-extension/package.json
```diff
@@ -56,13 +56,13 @@
...
-        "title": "Create New AIConfig (json)",
+        "command": "vscode-aiconfig.createExampleAIConfigJSON",
+        "title": "Create New Example AIConfig (json)",
         "category": "AIConfig"
       },
       {
-        "command": "vscode-aiconfig.createAIConfigYAML",
-        "title": "Create New AIConfig (yaml)",
+        "command": "vscode-aiconfig.createExampleAIConfigYAML",
+        "title": "Create New Example AIConfig (yaml)",
```
rossdanlm, line 65: The main user-facing change is here.

cc @tanya-rai , @saqadri what do you think/prefer? 

Next PR Ankush is implementing a "real" empty new AIConfig

rholinshead: nit, having "Create New E..." as text for both could lead to accidentally picking one when wanting the other. Maybe this could be 'Create AIConfig from Example" and next command can be "Create New AIConfig"?

rholinshead (approved):

vscode-extension/package.json
```diff
@@ -56,13 +56,13 @@
         "category": "AIConfig"
       },
       {
-        "command": "vscode-aiconfig.createAIConfigJSON",
-        "title": "Create New AIConfig (json)",
+        "command": "vscode-aiconfig.createExampleAIConfigJSON",
```
rholinshead, line 59: Should the order be swapped to have YAML first?

rholinshead (approved):

Ankush-lastmile:
- swapped ordering of create example commands to show yaml then json
- renamed command `Create New Example` -> `Create Aiconfig From Example`

vscode-extension/package.json
Ankush-lastmile: I like this. Updated

Ankush-lastmile:
- Fixed walkthrough commands: https://github.com/lastmile-ai/aiconfig/pull/1448#discussion_r1525354591

PR #1448, opened 2024-03-13
[vscode] 2/n Create Empty commands

Ankush-lastmile:
[vscode] 2/n Create Empty commands

This diff adds "create new" aiconfig commands to the vscode extension. Instead of forcing the user to deal with a haiku, they now have the option to start from scratch.

## Testplan

https://github.com/lastmile-ai/aiconfig/assets/141073967/4b314eb0-95cf-4e94-89c5-511ba7580a9f

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1448).
* __->__ #1448
* #1447


vscode-extension/src/extension.ts
```diff
@@ -289,6 +298,25 @@ async function createNewAIConfig(
...
 
+function relativeStaticAIConfigFilePath(
+  mode: "json" | "yaml" = "json",
+  empty: boolean = true
+): string {
+  if (mode === "json") {
+    if (empty) {
+      return "empty.aiconfig.json";
+    } else {
+      return "untitled.aiconfig.json";
```
rholinshead, line 309: nit, should we rename from `untitled` to `example`?

vscode-extension/package.json
```diff
@@ -65,6 +65,16 @@
         "title": "Create New Example AIConfig (yaml)",
         "category": "AIConfig"
       },
+      {
+        "command": "vscode-aiconfig.createEmptyAIConfigJSON",
+        "title": "Create New Empty AIConfig (json)",
```
rholinshead, line 70: nit, from previous PR - can this just be 'New' without 'Empty'?

rholinshead: Same comment on ordering YAML first here as well

vscode-extension/src/extension.ts
```diff
@@ -241,22 +257,15 @@ export function deactivate() {
  */
 async function createNewAIConfig(
   context: vscode.ExtensionContext,
-  mode: "json" | "yaml" = "json"
+  mode: "json" | "yaml" = "json",
```
rholinshead, line 260: Should we default to yaml?

vscode-extension/src/extension.ts
```diff
@@ -289,6 +298,25 @@ async function createNewAIConfig(
   }
 }
 
+function relativeStaticAIConfigFilePath(
+  mode: "json" | "yaml" = "json",
```
rholinshead, line 302: same here r.e. yaml as the default

vscode-extension/src/extension.ts
```diff
@@ -289,6 +298,25 @@ async function createNewAIConfig(
...
+  empty: boolean = true
+): string {
+  if (mode === "json") {
```
rholinshead, line 305:
```
const templateConfigName = empty ? "empty" : "example";
return `${templateConfigName}.aiconfig.${mode}`
```
Can be cleaned up quite a bit

rholinshead (approved):
Accepting to unblock, just minor comments

vscode-extension/package.json
```diff
@@ -202,13 +212,13 @@
           {
             "id": "createAIConfig",
             "title": "Create your first AIConfig",
-            "description": "Turn your VS Code into a generative AI studio\n**Note:** Please complete the ``Initialize Extension`` step before opening an AIConfig file!\n[Create .aiconfig.yaml](command:vscode-aiconfig.createAIConfigYAML)\nIf you prefer JSON:\n[Create .aiconfig.json](command:vscode-aiconfig.createAIConfigJSON)",
+            "description": "Turn your VS Code into a generative AI studio\n**Note:** Please complete the ``Initialize Extension`` step before opening an AIConfig file!\n[Create .aiconfig.yaml](command:vscode-aiconfig.createExampleAIConfigJSON)\nIf you prefer JSON:\n[Create .aiconfig.json](command:vscode-aiconfig.createAIConfigJSON)",
```
rholinshead, line 215:
```suggestion
            "description": "Turn your VS Code into a generative AI studio\n**Note:** Please complete the ``Initialize Extension`` step before opening an AIConfig file!\n[Create .aiconfig.yaml](command:vscode-aiconfig.createExampleAIConfigYAML)\nIf you prefer JSON:\n[Create .aiconfig.json](command:vscode-aiconfig.createExamleAIConfigJSON)",
```

rholinshead: Can we surface the empty commands here as well?

rholinshead (changes requested):
Ah just noticed the walkthrough so requesting changes so it's seen

vscode-extension/package.json
Ankush-lastmile: just confirming, you want all 4 commands in the walkthrough? I initially thought that would be too many

rholinshead: Maybe just fix these ones and then we can add the new ones later if we get feedback about it. They should be discoverable in the command palette when doing the example commands here

vscode-extension/src/extension.ts
rholinshead: Ah, ok, makes sense

Ankush-lastmile:
Since we're going in the direction of yaml as a default:
- - Modified Defaults to be yaml
- - swapped ordering of commands to have yaml first

- renamed "empty" command, now just `Create New AIConfig (..)`

> Ah just noticed the walkthrough so requesting changes so it's seen
Addressed and fixed  in 1/n: https://github.com/lastmile-ai/aiconfig/pull/1447#issuecomment-1999951840. Thanks for catching.

vscode-extension/package.json
```diff
@@ -246,6 +256,12 @@
         },
         {
           "command": "vscode-aiconfig.createExampleAIConfigYAML"
+        },
+        {
+          "command": "vscode-aiconfig.createEmptyAIConfigYAML"
```
rholinshead, line 261: One of these should be JSON.

Also, if order of these matters we should do Empty before Example and YAML before JSON

rholinshead (approved):
Accepting to unblock but one of the commands needs fixed to JSON and order should be swapped if it reflects order in the UI

vscode-extension/package.json
Ankush-lastmile: ordering matters above in the section of
```
"contributes": {
    "commands": [
```

With that in mind I don't know what this section does. In the testplan of this diff the 4 create commands work as intended.

Ankush-lastmile: but I'l add it in

Ankush-lastmile: synced offline with @rossdanlm , this is for the new file flow within vscode

PR #1453, opened 2024-03-13
[ez] gh action remove uneeded step

Ankush-lastmile:
[ez] gh action remove uneeded step

This github action does not need to store builds. It should be fine to build and publish within the same action


rossdanlm (approved):
I trust you manager!

PR #1454, opened 2024-03-13
combine jobs, build and publish in the same step.

Ankush-lastmile:
combine jobs, build and publish in the same step.

also add a secrets web hook

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1454).
* __->__ #1454
* #1453

### DIFF
--- a/.github/workflows/publish_python_sdk.yml
+++ b/.github/workflows/publish_python_sdk.yml
@@ -1,15 +1,17 @@
 # Build and publish the python-sdk to pypi. This workflow does NOT update the version number. Open a pr to update the version number and merge it to master before running this workflow.
 # This workflow is triggered manually. see https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow
-name: Publish Python SDK to PyPI
+name: Build and Publish Python SDK to PyPI
 
 # Controls when the workflow will run
 on:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish_python_sdk.yml
```diff
@@ -23,17 +25,18 @@ jobs:
...
           packages-dir: python/dist/
-
+      - name: Send slack message
+        id: slack
+        uses: slackapi/slack-github-action@v1.25.0
+        with:
+          # This data can be any valid JSON from a previous step in the GitHub Action
+          payload: |
+            {
+              "text": "Headsup - The python-sdk has been version bumped and published to pypi",
```
rossdanlm, line 39: Followup PR, can we also include the version # extracted from pyproject? If not that's ok. Also let's link to the pypi: https://pypi.org/project/python-aiconfig/

rossdanlm (approved):

.github/workflows/publish_python_sdk.yml
Ankush-lastmile: sounds good, will do on top sometime

PR #1458, opened 2024-03-14
[ci/cd] Python Formatting Check

Ankush-lastmile:
[ci/cd] Python Formatting Check

blocks on merging 

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1458).
* __->__ #1458
* #1459

PR #1459, opened 2024-03-14
[ez] Run Formatting on Python Files

Ankush-lastmile:
[ez] Run Formatting on Python Files

`black --exclude '/\.sl\b' . --line-length=79`

A lot of the python files checked into the AIConfig repo have not been formatted. This command formats all python files the same way vscode would do it in the repo.

I noticed differences between black version 23.11 and 24 (latest). I had version 23.11 on my local install. This also causes different behaviours when running autoformat through vscode instead of the command line, I'm fairly certain vscode extension uses the latest version of the package. I upgraded my black formatter local install to 24 (latest).

Note: black is python package.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/lastmile-ai/aiconfig/pull/1459).
* #1458
* __->__ #1459

### DIFF
--- a/cookbooks/Claude-And-Azure/aiconfig_model_registry.py
+++ b/cookbooks/Claude-And-Azure/aiconfig_model_registry.py
@@ -7,7 +7,6 @@
     See example below.
 """
 
-
 from aiconfig import AzureOpenAIParser, ModelParserRegistry
 import dotenv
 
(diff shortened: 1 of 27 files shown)


rossdanlm (approved):
nice pass!

Issue #1451, opened 2024-03-13
[editor] Add button to clear single prompt output


rossdanlm:
Today we have a button to clear all outputs, but it would be great to have something that adds to a single one only.

https://github.com/lastmile-ai/aiconfig/assets/151060367/241b1f38-0295-4550-9a4f-5804f487ab2d

Goals:
1. Get familiar with the AIConfig SDK (you'll learn the schema structure (in schema.py file) as well as SDK API (Config.py + schema.py)
2. Some familiarity with our editor and how it attaches to server and calls into our SDK API to execute operations

## Testing

First must need  to pip install to local repo
You will need to make sure you're on Python 3.10 or higher
`cd aiconfig/python && pip3 install -e .`
`cd aiconfig/extensions/HuggingFace/python && pip3 install -e .`

The only way to test this is through local editor because that will have your local changes connected!

Go to `aiconfig/python/src/aiconfig/editor/client` and run this command (you need to re-run this every time you make changes in the `aiconfig/python/src/aiconfig/editor/client` directory!):
```
rm -rf node_modules && yarn && yarn build
```

Then go to `aiconfig` dir and run this command:
```
aiconfig_path=./cookbooks/Gradio/huggingface.aiconfig.json
parsers_path=./cookbooks/Gradio/aiconfig_model_registry.py
aiconfig edit --aiconfig-path=$aiconfig_path --server-port=8080 --server-mode=debug_servers --parsers-module-path=$parsers_path
```

rholinshead:
> you need to re-run this every time you make changes in the aiconfig/python/src/aiconfig/editor/client directory!

Note that hot reloading will pick up local changes; you only need to run this to build the js for the --server-mode=prod variant (i.e. must build before publishing aiconfig package)

PR #1465, opened 2024-03-17
Feat: Add in ability to delete a single output by prompt name in editor

justinkchen:
https://github.com/lastmile-ai/aiconfig/issues/1451
Screenshot 2024-03-17 at 4 37 15 PM

Adds the ability to delete a single output in the config editor


python/src/aiconfig/editor/client/src/components/prompt/PromptContainer.tsx
```diff
@@ -190,7 +203,21 @@ export default memo(function PromptContainer({
 
           {prompt.outputs && prompt.outputs.length > 0 && (
             <>
+            <Flex justify="space-between" direction="column">
               <PromptOutputBar />
+              {!readOnly && prompt.outputs?.length && deleteOutput ? (
```
saqadri, line 208: Instead of adding the button here, can we add it to the prompt menu?

https://github.com/lastmile-ai/aiconfig/blob/main/python/src/aiconfig/editor/client/src/components/prompt/PromptMenuButton.tsx

I imagine we will have other capabilities like Copy/Duplicate prompt which will also be added in the future, so it will prevent too many buttons everywhere

saqadri (commented):
Looks great overall!! Nice first diff 😃 -- just one suggestion about the button moved to https://github.com/lastmile-ai/aiconfig/blob/main/python/src/aiconfig/editor/client/src/components/prompt/PromptMenuButton.tsx

python/src/aiconfig/editor/client/src/components/prompt/PromptMenuButton.tsx
```diff
@@ -34,10 +49,19 @@ export default memo(function PromptMenuButton({
...
           color="red"
-          onClick={() => onDeletePrompt(promptId)}
+          onClick={deletePrompt}
         >
           Delete Prompt
         </Menu.Item>
+        {prompt.outputs?.length ? (
+          <Menu.Item
+            icon={<IconEraser size={16} />}
+            color="red"
```
saqadri, line 59: nit: clear output may not need to be red, but unsure.

saqadri (approved):
Ship it! 🚀🚀🚀

python/src/aiconfig/editor/client/src/components/prompt/PromptMenuButton.tsx
saqadri: Also Delete Prompt should probably remain the last option in the menu.

python/src/aiconfig/editor/client/src/components/prompt/PromptsContainer.tsx
```diff
@@ -66,8 +68,10 @@ export default memo(function PromptsContainer(props: Props) {
             <Flex mt="md">
               {!readOnly && (
                 <PromptMenuButton
+                  prompt={prompt}
                   promptId={prompt._ui.id}
```
rossdanlm, line 72: nit: we probably don't need to pass in `promptId` anymore if passing in prompt

python/src/aiconfig/editor/client/src/reducers/aiconfigReducer.ts
```diff
@@ -150,6 +150,25 @@ export default function aiconfigReducer(
...
+    case "DELETE_OUTPUT": {
+      const prompts = state.prompts.map((prompt) => {
+        if (prompt._ui.id === action.id) {
+          return {
+            ...prompt,
+            outputs: undefined,
+          };
+        } else {
+          return prompt;
+        }
```
rossdanlm, line 163: We have a utils reducer method called `reduceReplacePrompt()`. Would prefer to use that to keep things cleaner and more abstracted

python/src/aiconfig/editor/client/src/components/prompt/PromptMenuButton.tsx
```diff
@@ -10,13 +11,27 @@ const useStyles = createStyles(() => ({
...
 
+  const deleteOutput = useCallback(
+    async () => await onDeleteOutput(promptId),
+    [promptId, onDeleteOutput]
+  );
+
+  const deletePrompt = useCallback(
+    async () => await onDeletePrompt(promptId),
+    [promptId, onDeletePrompt]
+  );
```
rossdanlm, line 33: This is a front-end thing that Ryan taught me. Since `PromptMenuButton` is already wrapped with a `memo` component, it's actually more overhead for us to wrap these sub-functions in a `useCallback()` hook. Here `useCallback()` checks for changes to any dependencies, but in our case that's simply changes to the props, which is already memoized in the `PromptMenuButton` component itself, so in this case it adds overhead. Would be better to either:

1. Keep inlined
2. Keep separate but don't wrap with `useCallback()`

python/src/aiconfig/editor/client/src/components/prompt/PromptMenuButton.tsx
```diff
@@ -31,10 +46,19 @@ export default memo(function PromptMenuButton({
       </Menu.Target>
 
       <Menu.Dropdown>
+        {prompt.outputs?.length ? (
+          <Menu.Item
+            icon={<IconEraser size={16} />}
+            color="white"
+            onClick={deleteOutput}
+          >
```
rossdanlm, line 54: cc @rholinshead for whether this would work with light mode theming

python/src/aiconfig/editor/server/server.py
```diff
@@ -811,9 +812,32 @@ def _op(
 
     signature: dict[str, Type[Any]] = {}
     return run_aiconfig_operation_with_request_json(
-        aiconfig, request_json, f"method_", _op, signature
+        aiconfig, request_json, f"method_{method_name}", _op, signature
```
rossdanlm, line 815: nice thanks!

rossdanlm: Actually upon further inspection, I think this should just be `method_name`, not `f"method_{method_name}"`: https://github.com/lastmile-ai/aiconfig/blob/f0f052231feee5f15843e2d88a70f7b406527dd5/python/src/aiconfig/editor/server/server_utils.py#L501-L503

python/src/aiconfig/editor/server/server.py
```diff
@@ -811,9 +812,32 @@ def _op(
...
+    if aiconfig is None:
+        LOGGER.info("No AIConfig in memory, can't run clear outputs.")
+        return HttpResponseWithAIConfig(
+            message="No AIConfig in memory, can't run clear outputs.",
+            code=400,
+            aiconfig=None,
+        ).to_flask_format()
+
+    signature: dict[str, Type[Any]] = {"prompt_name": str}
+    operation = make_op_run_method(method_name)
```
rossdanlm, line 837: Oh cool, TIL that we already have existing API for this: https://github.com/lastmile-ai/aiconfig/blob/f0f052231feee5f15843e2d88a70f7b406527dd5/python/src/aiconfig/schema.py#L1005. Thought it was for deleting entire outputs only. We can probly clean this up and delete the line `existing_outputs = prompt.outputs`. cc @Ankush-lastmile do you remember why this line might have been needed?

rossdanlm (approved):
Besides the `useCallbacks` and potential light mode issue, there are no functional blockers I see on my end. I'm happy to unblock and land for now and we can fix lingering nits in a followup! Let me know whether you prefer to fix these small things in this PR (will wait for you) or do a followup (will merge this for now)

python/src/aiconfig/editor/client/src/components/prompt/PromptContainer.tsx
```diff
@@ -190,7 +190,9 @@ export default memo(function PromptContainer({
 
           {prompt.outputs && prompt.outputs.length > 0 && (
             <>
+            <Flex justify="space-between" direction="column">
```
rholinshead, line 193: What does this look like / is this needed now that the button is moved to the menu?

python/src/aiconfig/editor/client/src/components/prompt/PromptMenuButton.tsx
```diff
@@ -10,13 +11,27 @@ const useStyles = createStyles(() => ({
 
 export default memo(function PromptMenuButton({
   promptId,
+  prompt,
+  onDeleteOutput,
   onDeletePrompt,
 }: {
   promptId: string;
+  prompt: ClientPrompt;
```
rholinshead, line 19: Passing `prompt` through just to access the output length will mean this component is re-rendered any time any part of the prompt changes (e.g. if typing in the input). That will result in the React memo constantly re-calculating the diff to check if it should re-render, and also re-rendering unnecessarily. Instead, it would be better to have a prop `showDeleteOutput` which is set to true if the prompt.outputs.length > 0 in the parent component

rossdanlm:
Btw for failing blockers (on autoformatting), you can use either of these commands to reformat and get unblocked (I think, not 100% sure):

1. `black --exclude '/\.sl\b' . --line-length=79`
2. `fd --glob '*.py'  python/src/aiconfig/editor | xargs python -m 'scripts.lint' --mode=fix --files`

python/src/aiconfig/editor/client/src/components/prompt/PromptMenuButton.tsx
rholinshead: Oh just noticed as well but we should just construct the `onDeletePrompt` and `onDeleteOutput` callbacks in the parent, since the promptId is known there. Then `promptId` wouldn't even need to be passed, just the callbacks. Similarly that would make PromptMenuButton memoization actually work as intended since the inline function currently prevents memoization from working (new function prop each time). So in `PromptsContainer` we could have
```
const promptId = prompt._ui.id;

const onDeletePrompt = props.onDeletePrompt;
const onDeleteOutput = props.onDeleteOutput;

const deletePrompt = useCallback(() => onDeletePrompt(promptId), [onDeletePrompt, promptId]);
const deleteOutput = useCallback(() => onDeleteOutput(promptId), [onDeleteOutput, promptId]);
```
And they can be passed through the props to PromptMenuButton

python/src/aiconfig/editor/client/src/components/prompt/PromptMenuButton.tsx
rholinshead: Good catch. I think we can remove color here and it will have the correct color we want. To test this, you can set `mode="gradio"` and `themeMode="dark"` and `themeMode="light"` props for AIConfigEditor in the `LocalEditor` component

python/src/aiconfig/editor/client/src/components/prompt/PromptsContainer.tsx
```diff
@@ -34,6 +35,7 @@ type Props = {
   ) => Promise<void>;
   prompts: ClientPrompt[];
   runningPromptId?: string;
+  readOnly?: boolean;
```
rholinshead, line 38: This can be removed as unused, since it's already obtained from context.

python/src/aiconfig/editor/client/src/components/prompt/PromptsContainer.tsx
rholinshead: Made a comment above -- we shouldn't pass prompt and we can actually just pass callbacks (and a new `showDeleteOutput` or `canDeleteOutput` prop)

python/src/aiconfig/editor/client/src/reducers/aiconfigReducer.ts
rholinshead: +1, it will handle selecting the correct prompt by id

rholinshead (changes requested):
Overall this looks great. Just requesting changes for some small but important changes in the comments

python/src/aiconfig/editor/client/src/components/prompt/PromptContainer.tsx
```diff
@@ -158,54 +175,65 @@ export default memo(function PromptContainer({
...
-          defaultConfigModelName={defaultConfigModelName}
-          prompt={prompt}
-          promptSchema={promptSchema}
-          onUpdateModelSettings={updateModelSettings}
-          onUpdateParameters={updateParameters}
-          onUpdatePromptMetadata={updatePromptMetadata}
+    <>
+      {!readOnly && (
+        <PromptMenuButton
+          showDeleteOutput={!!prompt.outputs?.length}
```
rossdanlm, line 181: nit: feel `prompt.outputs?.length > 0` is easier to read and I don't really like JS implicitly "falsey-ness" but don't really care

python/src/aiconfig/editor/client/src/components/prompt/PromptsContainer.tsx
```diff
@@ -58,25 +58,23 @@ export default memo(function PromptsContainer(props: Props) {
...
+
         return (
           <Stack key={prompt._ui.id}>
             <Flex mt="md">
-              {!readOnly && (
-                <PromptMenuButton
-                  promptId={prompt._ui.id}
-                  onDeletePrompt={() => props.onDeletePrompt(prompt._ui.id)}
-                />
-              )}
```
rossdanlm, line 72: omg this had me confused so hard wondering wtf was going on, but this is the `PromptsContainer`, not `PromptContainer` :S

rossdanlm: I like this change, I feel it makes sense to put with the individual prompt component anyways

python/src/aiconfig/editor/client/src/components/prompt/PromptsContainer.tsx
```diff
@@ -58,25 +58,23 @@ export default memo(function PromptsContainer(props: Props) {
...
               <PromptContainer
                 prompt={prompt}
                 getModels={props.getModels}
                 onChangePromptInput={props.onChangePromptInput}
                 onChangePromptName={props.onChangePromptName}
                 cancel={props.cancelRunPrompt}
                 onRunPrompt={props.onRunPrompt}
+                onDeletePrompt={props.onDeletePrompt}
+                onDeleteOutput={props.onDeleteOutput}
```
rossdanlm, line 77: nice

python/src/aiconfig/editor/client/src/shared/types.ts
```diff
@@ -66,6 +66,7 @@ export function aiConfigToClientConfig(aiconfig: AIConfig): ClientAIConfig {
 export type LogEvent =
   | "ADD_PROMPT"
   | "CLEAR_OUTPUTS"
+  | "DELETE_OUTPUT"
```
rossdanlm, line 69: nit: Would like to use wording "CLEAR_OUTPUT" to match with our UI, but again not a blocker or a big deal

rossdanlm (approved):
Nice, I think this is great! Will let @rholinshead do a final pass before merging!

python/src/aiconfig/editor/client/src/components/prompt/PromptContainer.tsx
justinkchen: it actually complains if you use this syntax unelss you do a `(prompt.outputs?.length || 0) > 0` it seems, so thats why i kept it with this way

python/src/aiconfig/editor/client/src/shared/types.ts
justinkchen: i think maybe as a bigger refactor of the api? i made it clear output at first but then found that all the apis reference delete output so wanted to stay consistent there.

rholinshead (approved):
Very nice - thanks!

rholinshead:
Ah, looks like the formatting lint is still failing. I'll override to merge to unblock and then I can do a quick PR with the formatting. I'd recommend setting up black formatter extension in vs code (if you use it) so that it formats automatically on save instead of needing the command

PR #1469, opened 2024-03-18
Update aiconfig-editor Package and Attach delete_model request/callback

rholinshead:
# Update aiconfig-editor Package and Attach delete_model request/callback

Summary:
Update the aiconfig-editor package to include UI for global model settings. As part of this, add the code to connect the delete_model request. Note that the delete_model call doesn't work because aiconfig python package needs to be updates so that the server has the endpoint.

Test Plan:
Added global settings, system prompt "respond in capital letters" for gpt-3.5-turbo and it works as expected

### DIFF
--- a/vscode-extension/editor/src/VSCodeEditor.tsx
+++ b/vscode-extension/editor/src/VSCodeEditor.tsx
@@ -6,6 +6,8 @@ import {
   RunPromptStreamErrorEvent,
   LogEvent,
   LogEventData,
+  ThemeMode,
+  AIConfigEditorNotification,
 } from "@lastmileai/aiconfig-editor";
 import { Flex, Loader, createStyles } from "@mantine/core";
 import {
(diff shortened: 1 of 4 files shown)


vscode-extension/editor/src/VSCodeEditor.tsx
```diff
@@ -233,6 +231,15 @@ export default function VSCodeEditor() {
     [aiConfigServerUrl, vscode]
   );
 
+  const deleteModelSettings = useCallback(
```
rossdanlm, line 234: nit: Why is this called `deleteModelSettings` instead of `deleteModel`?

rossdanlm (approved):
Do we also need to run `yarn && yarn build` from editor/client?

vscode-extension/editor/src/VSCodeEditor.tsx
rholinshead: I called the callback deleteModelSettings because that's a much clearer user-facing name. IMO `delete_model` is not clear in aiconfig sdk right now since I actually would have interpreted it as deleting the model parser if I didn't look through the code

PR #1475, opened 2024-03-22
Update to throw IndexError if model name not registered

justinkchen:
I wasn't really sure where else this change needed to be done. I added this error message in this registry code since it will be called by the same run_batch call that checks for prompts too


saqadri (approved):
Thanks! Change looks reasonable to me -- @Ankush-lastmile to confirm

Ankush-lastmile:
> Thanks! Change looks reasonable to me -- @Ankush-lastmile to confirm

Looks good to me, thanks for adding this @justinkchen

PR #1490, opened 2024-05-09
Create AI NIDE

ReymondVillaSenor11:


saqadri:
@ReymondVillaSenor11 did you forget to commit some changes here? I see only the new directory being created but no contents for the cookbook. We would be happy to take additional cookbook contributions :)

